PR #323, opened 2014-12-29
Configurable file_descriptors limit for warden container

shashidharatd:
file_descriptors for a warden container is made configurable through
deployment manifest.
Currently the default is 16384 which is read from the database table
and is not configurable. To configure, cf operator needs to add the
following config variable under cc
cc:
  default_app_file_descriptors: 4096

[#82011156]

### DIFF
--- a/bosh-templates/cloud_controller_clock.yml.erb
+++ b/bosh-templates/cloud_controller_clock.yml.erb
@@ -78,6 +78,8 @@ default_app_memory: <%= p("cc.default_app_memory") %>
 default_app_disk_in_mb: <%= p("cc.default_app_disk_in_mb") %>
 maximum_app_disk_in_mb: <%= p("cc.maximum_app_disk_in_mb") %>
 
+container_file_descriptor_limit: <%= p("cc.container_file_descriptor_limit") %>
+
 request_timeout_in_seconds: <%= p("request_timeout_in_seconds") %>
 
 cc_partition: <%= p("cc.cc_partition") %>
--- a/app/models/runtime/app.rb
+++ b/app/models/runtime/app.rb
@@ -138,6 +138,10 @@ module VCAP::CloudController
       self.memory ||= Config.config[:default_app_memory]
       self.disk_quota ||= Config.config[:default_app_disk_in_mb]
 
+      if Config.config[:container_file_descriptor_limit]
+        self.file_descriptors ||= Config.config[:container_file_descriptor_limit]
+      end
+
       set_new_version if version_needs_to_be_updated?
 
       AppStopEvent.create_from_app(self) if generate_stop_event?
--- a/bosh-templates/cloud_controller_api.yml.erb
+++ b/bosh-templates/cloud_controller_api.yml.erb
@@ -74,6 +74,8 @@ default_app_memory: <%= p("cc.default_app_memory") %>
 default_app_disk_in_mb: <%= p("cc.default_app_disk_in_mb") %>
 maximum_app_disk_in_mb: <%= p("cc.maximum_app_disk_in_mb") %>
 
+container_file_descriptor_limit: <%= p("cc.container_file_descriptor_limit") %>
+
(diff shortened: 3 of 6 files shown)


bosh-templates/cloud_controller_clock.yml.erb
```diff
@@ -76,6 +76,9 @@ billing_event_writing_enabled: <%= p("cc.billing_event_writing_enabled") %>
 
 default_app_memory: <%= p("cc.default_app_memory") %>
 default_app_disk_in_mb: <%= p("cc.default_app_disk_in_mb") %>
+<% if_p("cc.default_app_file_descriptors") do |file_descriptors|%>
```
crhino: Hi @shashidharatd,

You should not need to guard against this property not being set. Cloud Controller should be robust to this configuration value being null. You should follow the same pattern as the properties above it, e.g. `default_app_memory`. Can you also make this change for `bosh-templates/cloud_controller_worker.yml.erb` and `bosh-templates/cloud_controller_api.yml.erb` as well.

Thanks,

@crhino and @jfmyers9

shashidharatd:
Hi @crhino,
I did fix the review comment. But now there are merge conflicts. Shall i resolve these and raise a fresh PR?

Thanks
Shashi

shashidharatd:
Hi @crhino,
  I have rebased with master and have taken care of the review comment. Plz check.

Thanks
@shashidharatd

jpalermo:
Hi @shashidharatd 

We've discussed this change and have a few modifications.

We don't want the ability to change this on a per-app basis. So it should not be settable per app, and should not be returned in the response when fetching an app.

There should be just a single global manifest defined value for now. We would prefer "cc.container_file_descriptors" since the role of app is changing, and the value is no longer a default but the value for all apps.

It would be great if you could make these changes.

shashidharatd:
Hi @jpalermo, @dieucao,
  I have done the changes. hope it is in sync with the changes you expected.

Regarding the current default file descriptor limit per container, it is 16384. Checking `ulimit -n` will not give the correct value. It is only the default limit in any Linux bash shell as we are executing `ulimit -n` in the shell we setup for debugging the container. To exactly know the file descriptor limit for the app, we should check the `/proc/<pid>/limits`, where pid is process id of the app process in container. below is one such output.
```
vcap@18d2mqidul2:~$ ps -elf
F S UID        PID  PPID  C PRI  NI ADDR SZ WCHAN  STIME TTY          TIME CMD
4 S root         1     0  0  60 -20 -   281 poll_s 13:49 ?        00:00:00 wshd: 18d2mqidul2
4 S vcap        31     1  0  60 -20 - 22764 futex_ 13:49 ?        00:00:00 ruby /home/vcap/app/vendor/bundle/bin/rackup -p 61001
4 S vcap        37     1  0  60 -20 -  4871 wait   14:06 pts/0    00:00:00 /bin/bash
0 R vcap        53    37  0  60 -20 -  3833 -      14:10 pts/0    00:00:00 ps -elf

vcap@18d2mqidul2:~$ cat /proc/31/limits
Limit                     Soft Limit           Hard Limit           Units
Max cpu time              unlimited            unlimited            seconds
Max file size             unlimited            unlimited            bytes
Max data size             unlimited            unlimited            bytes
Max stack size            8388608              8388608              bytes
Max core file size        0                    0                    bytes
Max resident set          unlimited            unlimited            bytes
Max processes             512                  512                  processes
Max open files            16384                16384                files
Max locked memory         65536                65536                bytes
Max address space         unlimited            unlimited            bytes
Max file locks            unlimited            unlimited            locks
Max pending signals       1024                 1024                 signals
Max msgqueue size         819200               819200               bytes
Max nice priority         0                    0
Max realtime priority     0                    0
Max realtime timeout      unlimited            unlimited            us
```
we should also add the variable `cc.container_file_descriptor_limit` to cloud_controller jobs in cf-release.

PR #331, opened 2015-02-13
Allow db:rollback to roll back multiple migrations

Kaixiang:
story here: https://www.pivotaltracker.com/story/show/87150028


zrob:
merged

PR #339, opened 2015-03-20
Add "create-service-key" to cloud controller, based on story #87057732

wsxiaozhang:
Add the rest API controller and DB model for "create service key", this patch includes the following part:

1. 1 new rest controller for "create-service-key"
2. DB migration script for adding a new table to record service keys
3. 1 new DB model for service key
4. Changes the v2 service broker client to support create service key

This patch is submitted to implement story #87057732 in Service Key API
More test cases will be added in other commits

Signed-off-by: Tom Xing <xingzhou@cn.ibm.com>

### DIFF
--- a/app/access/service_key_access.rb
+++ b/app/access/service_key_access.rb
@@ -1,0 +1,13 @@
+module VCAP::CloudController
+  class ServiceKeyAccess < BaseAccess
+    def create?(service_key, params=nil)
+      return true if admin_user?
+      return false if service_key.in_suspended_org?
+      service_key.service_instance.space.developers.include?(context.user)
+    end
+
+    def delete?(service_key)
+      create?(service_key)
+    end
+  end
+end
--- a/app/controllers/services/lifecycle/service_key_manager.rb
+++ b/app/controllers/services/lifecycle/service_key_manager.rb
@@ -1,0 +1,62 @@
+module VCAP::CloudController
+  class ServiceKeyManager
+    class ServiceInstanceNotFound < StandardError; end
+    class ServiceInstanceNotBindable < StandardError; end
+
+    def initialize(services_event_repository, access_validator, logger)
+      @services_event_repository = services_event_repository
+      @access_validator = access_validator
+      @logger = logger
+    end
+
+    def create_service_key(request_attrs)
+      service_instance = ServiceInstance.first(guid: request_attrs['service_instance_guid'])
+      raise ServiceInstanceNotFound unless service_instance
+      raise ServiceInstanceNotBindable unless service_instance.bindable?
+
+      service_key = ServiceKey.new(request_attrs)
+      @access_validator.validate_access(:create, service_key)
+      raise Sequel::ValidationFailed.new(service_key) unless service_key.valid?
+
+      lock_service_instance_by_blocking(service_instance) do
+        attributes_to_update = service_key.client.bind(service_key)
+        begin
+          service_key.set_all(attributes_to_update)
+          service_key.save
+        rescue
+          safe_unbind_instance(service_key)
+          raise
+        end
+      end
+
+      service_key
+    end
+
+    private
+
+    def safe_unbind_instance(service_key)
+      service_key.client.unbind(service_key)
+    rescue => e
+      @logger.error "Unable to unbind #{service_key}: #{e}"
+    end
+
+    def lock_service_instance_by_blocking(service_instance, &block)
+      return block.call unless service_instance.managed_instance?
+
+      original_attributes = service_instance.last_operation.try(:to_hash)
+      begin
+        service_instance.lock_by_failing_other_operations('update') do
+          block.call
+        end
+      ensure
+        if original_attributes
+          service_instance.last_operation.set_all(original_attributes)
+          service_instance.last_operation.save
+        else
+          service_instance.service_instance_operation.destroy
+          service_instance.save
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/services/service_keys_controller.rb
+++ b/app/controllers/services/service_keys_controller.rb
@@ -1,0 +1,42 @@
+require 'services/api'
+
+module VCAP::CloudController
+  class ServiceKeysController < RestController::ModelController
+    define_attributes do
+      to_one :service_instance
+      attribute :name, String
+    end
+
+    post path, :create
+    def create
+      @request_attrs = self.class::CreateMessage.decode(body).extract(stringify_keys: true)
+      logger.debug 'cc.create', model: self.class.model_class_name, attributes: request_attrs
+      raise InvalidRequest unless request_attrs
+      service_key_manager = ServiceKeyManager.new(@services_event_repository, self, logger)
+      service_key = service_key_manager.create_service_key(@request_attrs)
+      [HTTP::CREATED,
+       { 'Location' => "#{self.class.path}/#{service_key.guid}" },
+       object_renderer.render_json(self.class, service_key, @opts)
+      ]
+    rescue ServiceKeyManager::ServiceInstanceNotFound
+      raise VCAP::Errors::ApiError.new_from_details('ServiceInstanceNotFound', @request_attrs['service_instance_guid'])
+    rescue ServiceKeyManager::ServiceInstanceNotBindable
+      raise VCAP::Errors::ApiError.new_from_details('UnbindableService')
+    end
+
+    private
+
+    def self.translate_validation_exception(e, attributes)
+      unique_errors = e.errors.on([:name, :service_instance_id])
+      if unique_errors && unique_errors.include?(:unique)
+        Errors::ApiError.new_from_details('ServiceKeyTaken', "#{attributes['name']} #{attributes['service_instance_guid']}")
+      elsif e.errors.on(:service_instance) && e.errors.on(:service_instance).include?(:presence)
+        Errors::ApiError.new_from_details('ServiceInstanceNotFound', attributes['service_instance_guid'])
+      else
+        Errors::ApiError.new_from_details('ServiceKeyInvalid', e.errors.full_messages)
+      end
+    end
+
+    define_messages
+  end
+end
--- a/app/models/services/service_key.rb
+++ b/app/models/services/service_key.rb
@@ -1,0 +1,71 @@
+module VCAP::CloudController
+  class ServiceKey < Sequel::Model
+    class InvalidAppAndServiceRelation < StandardError; end
+
+    many_to_one :service_instance
+
+    export_attributes :name, :service_instance_guid, :credentials
+
+    import_attributes :name, :service_instance_guid, :credentials
+
+    delegate :client, :service, :service_plan, to: :service_instance
+
+    plugin :after_initialize
+
+    encrypt :credentials, salt: :salt
+
+    def to_hash(opts={})
+      if !VCAP::CloudController::SecurityContext.admin? && !service_instance.space.developers.include?(VCAP::CloudController::SecurityContext.current_user)
+        opts.merge!({ redact: ['credentials'] })
+      end
+      super(opts)
+    end
+
+    def in_suspended_org?
+      space.in_suspended_org?
+    end
+
+    def space
+      service_instance.space
+    end
+
+    def validate
+      validates_presence :name
+      validates_presence :service_instance
+      validates_unique [:name, :service_instance_id]
+    end
+
+    def credentials_with_serialization=(val)
+      self.credentials_without_serialization = MultiJson.dump(val)
+    end
+    alias_method_chain :credentials=, 'serialization'
+
+    def credentials_with_serialization
+      string = credentials_without_serialization
+      return if string.blank?
+      MultiJson.load string
+    end
+    alias_method_chain :credentials, 'serialization'
+
+    def self.user_visibility_filter(user)
+      { service_instance: ServiceInstance.user_visible(user) }
+    end
+
+    def after_initialize
+      super
+      self.guid ||= SecureRandom.uuid
+    end
+
+    def logger
+      @logger ||= Steno.logger('cc.models.service_key')
+    end
+
+    private
+
+    def safe_unbind
+      client.unbind(self)
+    rescue => unbind_e
+      logger.error "Unable to unbind #{self}: #{unbind_e}"
+    end
+  end
+end
--- a/db/migrations/20150316184259_create_service_key_table.rb
+++ b/db/migrations/20150316184259_create_service_key_table.rb
@@ -1,0 +1,13 @@
+Sequel.migration do
+  change do
+    create_table(:service_keys) do
+      VCAP::Migration.common(self, :sk)
+      String :name, null: false
+      String :salt
+      String :credentials, null: false, size: 2048
+      Integer :service_instance_id, null: false
+      foreign_key [:service_instance_id], :service_instances, name: :fk_svc_key_svc_instance_id
+      index [:name, :service_instance_id], unique: true, name: :svc_key_name_instance_id_index
+    end
+  end
+end
--- a/lib/services/service_brokers/v2/client.rb
+++ b/lib/services/service_brokers/v2/client.rb
@@ -92,11 +92,15 @@ module VCAP::Services::ServiceBrokers::V2
 
     def bind(binding)
       path = service_binding_resource_path(binding)
-      response = @http_client.put(path, {
-        service_id:  binding.service.broker_provided_id,
-        plan_id:     binding.service_plan.broker_provided_id,
-        app_guid:    binding.app_guid
-      })
+      attr = {
+          service_id:  binding.service.broker_provided_id,
+          plan_id:     binding.service_plan.broker_provided_id
+      }
+      if binding.respond_to? 'app_guid'
+        attr[:app_guid] = binding.app_guid
+      end
+
+      response = @http_client.put(path, attr)
       parsed_response = @response_parser.parse(:put, path, response)
 
       attributes = {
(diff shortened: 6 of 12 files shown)


app/models/services/service_key.rb
```diff
@@ -0,0 +1,92 @@
...
+      rescue => e
+        safe_unbind
+        raise e
+      end
+    end
+
+    def after_initialize
+      super
+      self.guid ||= SecureRandom.uuid
+      puts self.guid
```
dsabeti: It looks like this was left behind from debugging. Can we delete this?

app/controllers/services/lifecycle/service_key_manager.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def initialize(services_event_repository, access_validator, logger)
+      @services_event_repository = services_event_repository
+      @access_validator = access_validator
+      @logger = logger
+    end
+
+    def create_service_key(request_attrs)
+      service_instance = ServiceInstance.first(guid: request_attrs['service_instance_guid'])
+      raise ServiceInstanceNotFound unless service_instance
```
dsabeti: This case needs a test

app/controllers/services/lifecycle/service_key_manager.rb
```diff
@@ -0,0 +1,59 @@
...
+      raise ServiceInstanceNotBindable unless service_instance.bindable?
```
dsabeti: This case also needs a test

app/controllers/services/lifecycle/service_key_manager.rb
```diff
@@ -0,0 +1,59 @@
...
+    def lock_service_instance_by_blocking(service_instance, &block)
+      return block.call unless service_instance.managed_instance?
+
+      original_attributes = service_instance.last_operation.try(:to_hash)
+      begin
+        service_instance.lock_by_failing_other_operations('update') do
+          block.call
+        end
+      ensure
+        if original_attributes
```
dsabeti: I'm glad you guys followed the pattern that we established in the ServiceBindingManager. This logic changed a bit since you last looked at it though. Please take a look at the newest version of that file and incorporate the same logic. In this case, we want to make sure that we get rid of `service_instance_last_operation` when `original_attributes` is nil (meaning that the original `service_instance.last_operation` was nil. Ideally, the operation will not change the state of the instace).

app/access/service_key_access.rb
```diff
@@ -0,0 +1,13 @@
+module VCAP::CloudController
+  class ServiceKeyAccess < BaseAccess
```
dsabeti: Please add a file with tests for this class.

app/controllers/services/service_keys_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+module VCAP::CloudController
+  class ServiceKeysController < RestController::ModelController
+    define_attributes do
+      to_one :service_instance
+      attribute :name, String
+    end
+
+    post path, :create
+    def create
+      @request_attrs = self.class::CreateMessage.decode(body).extract(stringify_keys: true)
```
dsabeti: This method is also missing many tests. A good rule of thumb is that commenting out any line of code should result in a test failure. Mostly, I'm thinking that all of the errors that are getting raised need to have associated tests.

maximilien:
Thanks @dsabeti for the quick comments. We will try to address them all today and resubmit.

Best,

max

app/controllers/services/lifecycle/service_key_manager.rb
maximilien: Hi, all, looking at the code more carefully and noticed we don't have any `ServiceBindingManager` specs. Is there a reason that this class does not have its own unit tests?

app/controllers/services/lifecycle/service_key_manager.rb
zhang-hua: "orphan mitigation" is planned to be covered when we implement story [#89967302], so we will submit the test cases in another patch later, thanks

xingzhou:
@dsabeti , thanks for the comments and we have resbumitted the PR according to your suggestions. For "orphan mitigation" and "delete keys" and "details" relevant code and test cases, we will add them in relevant stories later.

xingzhou:
Current PR depends on another PR in errors project: https://github.com/cloudfoundry/errors/pull/18, this will lead to the failure of Travis CI build.

app/controllers/services/lifecycle/service_key_manager.rb
dsabeti: The `ServiceBindingManager` is covered by the ServiceBindingController specs. We're talking about backfilling unit tests for the object. In the meantime, you guys can follow our lead and add the tests for the manager class to the controller specs.

spec/unit/controllers/services/service_keys_controller_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+
+          it 'does not send a bind request to broker' do
+            expect(a_request(:put, bind_url_regex(service_instance: instance))).to_not have_been_made
+          end
+        end
+
+        context 'when the service instance is invalid' do
+          context 'because service_instance_guid is invalid' do
+            before do
+              service.save
```
dsabeti: I think this before block is unnecessary.

app/controllers/services/lifecycle/service_key_manager.rb
maximilien: Sure.

xingzhou:
@dsabeti , have removed the useless lines in service_keys_controller_spec, please take a look at the latest patch

dsabeti:
We've merged in the PR for vendor/errors, so you can now update this PR to include the correct submodule reference.

spec/unit/controllers/services/service_keys_controller_spec.rb
```diff
@@ -0,0 +1,227 @@
...
+              post '/v2/service_keys', req, json_headers(headers_for(developer))
+
+              hash_body = JSON.parse(last_response.body)
+              expect(hash_body['error_code']).to eq('CF-ServiceInstanceNotFound')
+              expect(last_response.status).to eq(404)
+            end
+          end
+
+          context 'because the service instance is destroyed after controller validation and before binding save' do
+            let(:service_instance_guid) { 'THISISWRONG' }
```
dsabeti: This test is exactly the same as the test above it. Let's remove the dup.

app/controllers/services/lifecycle/service_key_manager.rb
dsabeti: Ok, I suppose that's ok for now. 

For future reference, please don't add code if you aren't going to write a corresponding test. For example, I was originally planning to ask for removal of the following line:
`safe_unbind_instance(service_key)`
because it has no tests (because they are part of the orphan mitigation suite that you're planning to add later). We didn't immediately understand why that line was there, and assumed it was part of this story. It's good that you're thinking ahead, but it is easier for us to follow which logic pertains to the current story if it's kept as minimal as possible. Also, we'll have to remember that this particular line of code is untested and double-check test coverage when the orphan mitigation story is submitted. This separation of tests and code will create more work for us in the future.

Rather than ask you to remove code and then re-add it later, we can keep it so long as the subsequent story about orphan mitigation includes backfilled tests. I've added a task in the story.

spec/unit/controllers/services/service_keys_controller_spec.rb
```diff
@@ -0,0 +1,227 @@
...
+      context 'for managed instances' do
+        let(:instance) { ManagedServiceInstance.make }
+        let(:space) { instance.space }
+        let(:service) { instance.service }
+        let(:developer) { make_developer_for_space(space) }
+        let(:name) { 'fake-service-key' }
+        let(:service_instance_guid) { instance.guid }
+        let(:req) {
+          {
+              name: :name,
```
dsabeti: I think instead of `:name`, you want `name`. This means that you will uses the name created by the `let(:name) ...` above.

spec/unit/controllers/services/service_keys_controller_spec.rb
```diff
@@ -0,0 +1,227 @@
...
+          expect(service_key.credentials).to eq(CREDENTIALS)
+        end
+
+        context 'when attempting to create service key for an unbindable service' do
+          before do
+            service.bindable = false
+            service.save
+
+            req = {
+                name: :name,
```
dsabeti: Same as above. Change from `:name` to `name`

spec/unit/controllers/services/service_keys_controller_spec.rb
```diff
@@ -0,0 +1,227 @@
...
+          end
+
+          describe 'creating key errors' do
+            subject(:make_request) do
+              post '/v2/service_keys', req, json_headers(headers_for(developer))
+            end
+
+            context 'when attempting to create key and service key already exists' do
+              before do
+                ServiceKey.make(name: :name, service_instance: instance)
```
dsabeti: change from `:name` to `name`

app/models/services/service_key.rb
```diff
@@ -0,0 +1,95 @@
...
+  class ServiceKey < Sequel::Model
+    class InvalidAppAndServiceRelation < StandardError; end
+
+    many_to_one :service_instance
+
+    export_attributes :name, :service_instance_guid, :credentials, :syslog_drain_url
+
+    import_attributes :name, :service_instance_guid, :credentials, :syslog_drain_url
+
+    alias_attribute :broker_provided_id, :gateway_name
```
dsabeti: This has no tests. Luckily, this is a relic from the days of V1 service gateways and we don't need it anymore. I'm adding a chore to investigate removing this for service_bindings and service_instances.

app/models/services/service_key.rb
```diff
@@ -0,0 +1,95 @@
...
+    def in_suspended_org?
+      space.in_suspended_org?
+    end
+
+    def space
+      service_instance.space
+    end
+
+    def validate
+      validates_presence :name
```
dsabeti: Removing this validation does not cause a test to fail. I saw that there was clearly a test that is *supposed* to make it fail, so maybe you could investigate what's going wrong.

xingzhou: yes, should be `validate_presence` not `validate_db_presence` in the test case, have modified and resubmitted the PR

app/models/services/service_key.rb
```diff
@@ -0,0 +1,93 @@
+module VCAP::CloudController
+  class ServiceKey < Sequel::Model
+    class InvalidAppAndServiceRelation < StandardError; end
+
+    many_to_one :service_instance
+
+    export_attributes :name, :service_instance_guid, :credentials, :syslog_drain_url
```
dsabeti: `syslog_drain_url` is not applicable to service_keys (it only matters for bindings with apps). Please remove all fields and logic regarding syslog_drain_url

app/models/services/service_key.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    def space
+      service_instance.space
+    end
+
+    def validate
+      validates_presence :name
+      validates_presence :service_instance
+      validates_unique [:name, :service_instance_id]
+      validate_logging_service_binding if service_instance.respond_to?(:service_plan)
```
dsabeti: Remove this validation. See above regarding syslog_drain_url

app/models/services/service_key.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    def validate_logging_service_binding
```
dsabeti: remove this method. It will not be used.

db/migrations/20150316184259_create_service_key_table.rb
```diff
@@ -0,0 +1,14 @@
+Sequel.migration do
+  change do
+    create_table(:service_keys) do
+      VCAP::Migration.common(self, :sk)
+      String :name, null: false
+      String :salt
+      String :syslog_drain_url
```
dsabeti: No need to have a syslog_drain_url

spec/support/fakes/blueprints.rb
```diff
@@ -194,6 +194,13 @@ module VCAP::CloudController
     syslog_drain_url  { nil }
   end
 
+  ServiceKey.blueprint do
+    credentials       { Sham.service_credentials }
+    service_instance  { ManagedServiceInstance.make }
+    name               { Sham.name }
+    syslog_drain_url  { nil }
```
dsabeti: No syslog_drain_url

spec/unit/models/services/service_key_spec.rb
```diff
@@ -0,0 +1,218 @@
...
+    describe 'Validations' do
+      it { is_expected.to validate_presence :service_instance }
+      it { is_expected.to validate_presence :name }
+      it { is_expected.to validate_db_presence :service_instance_id }
+      it { is_expected.to validate_db_presence :credentials }
+      it { is_expected.to validate_uniqueness [:name, :service_instance_id] }
+    end
+
+    describe 'Serialization' do
+      it { is_expected.to export_attributes :name, :service_instance_guid, :credentials, :syslog_drain_url }
```
dsabeti: No syslog_drain_url

spec/unit/models/services/service_key_spec.rb
```diff
@@ -0,0 +1,218 @@
...
+            service_instance: service_instance,
+            credentials: value_to_encrypt
+        )
+      end
+
+      let(:encrypted_attr) { :credentials }
+      let(:attr_salt) { :salt }
+    end
+
+    describe 'logging service bindings' do
```
dsabeti: We can remove this whole section

spec/unit/models/services/service_key_spec.rb
```diff
@@ -0,0 +1,218 @@
...
+      it { is_expected.to validate_db_presence :credentials }
+      it { is_expected.to validate_uniqueness [:name, :service_instance_id] }
+    end
+
+    describe 'Serialization' do
+      it { is_expected.to export_attributes :name, :service_instance_guid, :credentials, :syslog_drain_url }
+      it { is_expected.to import_attributes :name, :service_instance_guid, :credentials, :syslog_drain_url }
+    end
+
+    describe '#insert_service_key_data' do
```
dsabeti: This test is really testing `#new` method. Let's change this string to `'#new'`

app/models/services/service_key.rb
```diff
@@ -0,0 +1,93 @@
...
+    alias_method_chain :credentials=, 'serialization'
+
+    def credentials_with_serialization
+      string = credentials_without_serialization
+      return if string.blank?
+      MultiJson.load string
+    end
+    alias_method_chain :credentials, 'serialization'
+
+    def create!
```
dsabeti: Nobody calls this method, and it's logic is duplicated in the ServiceKeyManager. We can probably just delete this method.

app/models/services/service_key.rb
```diff
@@ -0,0 +1,93 @@
...
+    def self.user_visibility_filter(user)
+      { service_instance: ServiceInstance.user_visible(user) }
+    end
+
+    def after_initialize
+      super
+      self.guid ||= SecureRandom.uuid
+    end
+
+    def before_destroy
```
dsabeti: We're trying to move away from `before_destroy` hooks. Take a look at the service_binding_delete.rb to see how we're working around them. Please remove this method and build a corresponding object for service keys.

spec/unit/models/services/service_key_spec.rb
```diff
@@ -0,0 +1,218 @@
...
+      it 'has a guid when constructed' do
+        service_key = described_class.new
+        expect(service_key.guid).to be
+      end
+    end
+
+    describe '#destroy' do
```
dsabeti: Please remove these tests as a result of removing the before_destroy hook

lib/services/service_brokers/v2/client.rb
```diff
@@ -92,11 +92,15 @@ def fetch_service_instance_state(instance)
...
-      response = @http_client.put(path, {
-        service_id:  binding.service.broker_provided_id,
-        plan_id:     binding.service_plan.broker_provided_id,
-        app_guid:    binding.app_guid
-      })
+      attr = {
+          service_id:  binding.service.broker_provided_id,
+          plan_id:     binding.service_plan.broker_provided_id
+      }
+      if binding.respond_to? 'app_guid'
```
dsabeti: There's a missing test case here. If you comment out the `if` and `end`, but leave the code that adds the app_guid to `attr`, you'll find that no tests fail. That means there is no test that drives out the desired change here: if the "binding" object does not respond to `#app_guid`, we want to exclude app_guid from the request.

dsabeti:
Hi @xingzhou . We've made a complete pass through the code, and this should be ready to merge once you address these last few concerns. Sorry about the endless back-and-forth. 

Two things to note:
- Through reviewing your code, we noticed several issues that we will also be changing for service bindings. When looking to the service bindings code for guidance, please point out any confusing/incorrect code/tests, and feel free to change your own implementation for service keys from what we currently have. It's an interesting side-effect of copying old, crufty code that we get to re-evaluate and see all of its problems. We definitely want to continue seeing this feedback, but it will make things easier if they don't make it in to new code for Service Keys. So please, do let us know if you see anything strange in the Service Binding code so we can refactor / clean up appropriately.
- In a few places, this commit includes code that is outside the scope of the story. Please try to include the minimum change to implement the current feature. It's easier for us to review the pull request, to keep track of corresponding tests, and keeping changing patterns up-to-date. (For an example of the last case, we're referring to the `before_destroy hook` vs `ServiceBindingDelete` class) This will help us review your PR's more quickly.

lib/services/service_brokers/v2/client.rb
xingzhou: yes, we missed to test this piece of code in client spec, add one more test case in the latest patch

xingzhou:
@dsabeti, thanks for the review, it's the fast way for us to pick up the CC code by these back-and-forth review process, we will keep these in mind in the following development.

PR #347, opened 2015-03-25
346 - have DropletsController inherit from a richer parent class

ericpromislow:
Update the class.  Running the tests in spec/unit/controllers/v3/droplets_controller_spec.rb
is sufficient to test this change, so no new test is necessary to test breakage. (Testing third-party
code that monkeypatches DropletsController is another matter.)

### DIFF
--- a/app/controllers/v3/droplets_controller.rb
+++ b/app/controllers/v3/droplets_controller.rb
@@ -4,7 +4,7 @@ require 'queries/droplet_delete_fetcher'
 require 'actions/droplet_delete'
 
 module VCAP::CloudController
-  class DropletsController < RestController::BaseController
+  class DropletsController < RestController::ModelController
     def self.dependencies
       [:droplets_handler, :droplet_presenter]
     end


DanLavine:
Hi @ericpromislow,

As part of our v3 refactor we are actively moving away from the model controller and its established patterns. We are trying to be more explicit with our database access and have the flexibility in our design to support future features. 

We are curious as to why you would like this change? As it stands the droplets controller would not use any of the functionality provided by the model controller, therefore we are are unclear as to the motivation behind this change.

Thanks,

@jfmyers9 && @DanLavine, CF Runtime Team

ericpromislow:
Stackato has for a long time reopened DropletsController as a subclass of ModelController and then called define_attributes at the class level to add a few more attributes.  Before v201 this class wasn't explicitly defined, so there was no problem. I see all the classes in controllers/v3 are subclassed off Base, while the legacy classes are mostly off Model, with a few off Base.

DanLavine:
Hi @ericpromislow,

It is very important to our team to not use model controller for v3. It came up on many of our retrospectives as a complex controller to inherit from.

One solution to this problem could be to merge these changes into your `DropletsController` and change the controller to inherit from `ModelController` on your fork.

The Runtime Team
@danlavine && @sergueif

ericpromislow:
Hi @DanLavine and thanks for the reply.

Could you let me know what some of the issues that came up were?

sergueif:
Well, it's meta-programmed, so it performs validation/authorization/queries/serialization in a particular way, irrespective of the model in question extra tersely but rigidly. When either aspect of the request (validation/authorization/queries/serialization) was needed to be tweaked slightly, you can either copy-paste controller action implementation or attempt to flush-out the hooks exposed on the model like user_visibility_filter.

Inheriting from a smaller BaseController in v3, we enjoy being able to compose endpoints out of separate ruby objects and leave room for some query optimization or more involved authorization business rules.

The Runtime Team,
@sergueif && @DanLavine

PR #348, opened 2015-03-26
Enable update of docker images

hsiliev:
[#[86969440](https://www.pivotaltracker.com/story/show/86969440)]

CC will update the `docker_image` column in `apps` table if Diego changed the image name during staging in order to cache it.  

### DIFF
--- a/lib/cloud_controller/diego/docker/staging_completion_handler.rb
+++ b/lib/cloud_controller/diego/docker/staging_completion_handler.rb
@@ -12,8 +12,9 @@ module VCAP::CloudController
         def self.success_parser
           @staging_response_schema ||= Membrane::SchemaParser.parse do
             {
-              execution_metadata: String,
-              detected_start_command: Hash,
+                :execution_metadata => String,
+                :detected_start_command => Hash,
+                optional(:lifecycle_data) => Hash
             }
           end
         end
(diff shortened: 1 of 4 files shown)


ematpl:
Hi, @hsiliev,

We're concerned that this clobbers the `docker_image` value that the user specified when they pushed their docker image app. Would that make a subsequent restage a no-op, because the staging process would pull the cached image from the internal registry and then push that back to the registry? It might make more sense to add a separate field on App to store the staging completion data, and then to prefer its value when running if set. 

We have a few other remarks that we'll make as line comments.

Thanks,
Eric && @Amit-PivotalLabs

spec/unit/lib/cloud_controller/diego/docker/staging_completion_handler_spec.rb
```diff
@@ -43,6 +43,24 @@ module Docker
             expect(runner).to have_received(:start)
           end
 
+          context 'when it receives docker image' do
+            let(:docker_image_name) { '10.244.2.6:8080/generated_id:latest' }
+            let(:payload) do
+              {
+                  execution_metadata: '"{\"cmd\":[\"start\"]}"',
+                  detected_start_command: { web: 'start' },
+                  lifecycle_data: { docker_image: docker_image_name }
```
ematpl: Maybe add a test for the case when `lifecycle_data: {}`? Also, what should the behavior be when `docker_image` is `""`?

lib/cloud_controller/diego/docker/staging_completion_handler.rb
```diff
@@ -12,8 +12,9 @@ def initialize(runners)
         def self.success_parser
           @staging_response_schema ||= Membrane::SchemaParser.parse do
             {
-              execution_metadata: String,
-              detected_start_command: Hash,
+                :execution_metadata => String,
+                :detected_start_command => Hash,
+                optional(:lifecycle_data) => Hash
```
ematpl: Let's leave the indentation at 2 spaces.

PR #350, opened 2015-03-31
Enable docker image caching

hsiliev:
[#86969440]

CC will update the docker_image column in apps table if Diego changed the image name during staging in order to cache it.

Signed-off-by: Georgi Sabev georgethebeatle@gmail.com

### DIFF
--- a/db/migrations/20150327080540_add_cached_docker_image_to_droplets.rb
+++ b/db/migrations/20150327080540_add_cached_docker_image_to_droplets.rb
@@ -1,0 +1,5 @@
+Sequel.migration do
+  change do
+    add_column :droplets, :cached_docker_image, String
+  end
+end
--- a/db/migrations/20150327080540_add_cached_docker_image_to_droplets.rb
+++ b/db/migrations/20150327080540_add_cached_docker_image_to_droplets.rb
@@ -1,5 +1,5 @@
 Sequel.migration do
   change do
-    add_column :droplets, :cached_docker_image, String
+    add_column :droplets, :cached_docker_image, String, text: true
   end
 end
--- a/lib/cloud_controller/diego/docker/protocol.rb
+++ b/lib/cloud_controller/diego/docker/protocol.rb
@@ -40,6 +40,8 @@ module VCAP::CloudController
         end
 
         def desire_app_message(app, default_health_check_timeout)
+          cached_docker_image = app.current_droplet.cached_docker_image if app.current_droplet
+
           {
             'process_guid' => ProcessGuid.from_app(app),
             'memory_mb' => app.memory,
@@ -52,7 +54,7 @@ module VCAP::CloudController
             'num_instances' => app.desired_instances,
             'routes' => app.uris,
             'log_guid' => app.guid,
-            'docker_image' => app.docker_image,
+            'docker_image' => cached_docker_image || app.docker_image,
             'health_check_type' => app.health_check_type,
             'health_check_timeout_in_seconds' => app.health_check_timeout || default_health_check_timeout,
             'egress_rules' => @common_protocol.running_egress_rules(app),
--- a/lib/cloud_controller/diego/docker/staging_completion_handler.rb
+++ b/lib/cloud_controller/diego/docker/staging_completion_handler.rb
@@ -12,8 +12,9 @@ module VCAP::CloudController
         def self.success_parser
           @staging_response_schema ||= Membrane::SchemaParser.parse do
             {
-              execution_metadata: String,
-              detected_start_command: Hash,
+              :execution_metadata => String,
+              :detected_start_command => Hash,
+              optional(:lifecycle_data) => Hash
             }
           end
         end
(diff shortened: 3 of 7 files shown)


lib/cloud_controller/diego/docker/staging_completion_handler.rb
```diff
@@ -12,8 +12,9 @@ def initialize(runners)
         def self.success_parser
           @staging_response_schema ||= Membrane::SchemaParser.parse do
             {
-              execution_metadata: String,
-              detected_start_command: Hash,
+              :execution_metadata => String,
```
sykesm: This change doesn't appear to be necessary.  The code that was here before the PR uses the new hash syntax. Is there a reason this was done or can you put it back to how it was?

db/migrations/20150327080540_add_cached_docker_image_to_droplets.rb
```diff
@@ -0,0 +1,5 @@
+Sequel.migration do
+  change do
+    add_column :droplets, :cached_docker_image, String
```
sykesm: If the cached_docker_image can ever exceed ~250 characters, it's better to make this a text field by adding `, text: true` after `String`.

lib/cloud_controller/diego/docker/staging_completion_handler.rb
hsiliev: We re-use the lifecycle_metadata to return the cached/generated docker image. This field however is optional (users can opt-in for caching). Another reason for having it optional is that until we merge the changes in Diego the response won't contain the cached image anyway.

The new syntax is fine until we have `optional(:symbol)`. Seems this requires the old syntax. We saw the same pattern in all places where `optional` is used. We can use the new hash syntax for the first 2 fields and the old for the optional `lifecycle_data`, but this does not improve the block in my opinion.

lib/cloud_controller/diego/docker/protocol.rb
```diff
@@ -40,6 +40,8 @@ def desire_app_request(app, default_health_check_timeout)
         end
 
         def desire_app_message(app, default_health_check_timeout)
+          cached_docker_image = app.current_droplet.cached_docker_image if app.current_droplet
```
sykesm: I was able to comment out the `if app.current_droplet` from the code and no tests failed. Looks like you need a test there.

spec/unit/lib/cloud_controller/diego/docker/staging_completion_handler_spec.rb
```diff
@@ -43,6 +43,96 @@ module Docker
...
+
+              it 'does not update the cached image' do
+                handler.staging_complete(staging_guid, payload)
+
+                expect(droplet.cached_docker_image).to be_nil
+              end
+            end
+          end
+
+          RSpec.shared_examples 'a handler with no lifecycle data' do
```
sykesm: You don't need the `RSpec` here. `shared_examples` will work on its own.

lib/cloud_controller/diego/docker/staging_completion_handler.rb
```diff
@@ -36,6 +37,12 @@ def save_staging_result(app, payload)
               end
             end
 
+            if payload.key?(:lifecycle_data) && payload[:lifecycle_data][:docker_image].present?
+              droplet.update_cached_docker_image(payload[:lifecycle_data][:docker_image])
+            else
+              droplet.update_cached_docker_image(nil)
+            end
+
```
sykesm: Maybe consider something like this to avoid duplicating the walk of the keys?

```ruby
            cached_image = payload[:lifecycle_data] && payload[:lifecycle_data][:docker_image]
            droplet.update_cached_docker_image(cached_image) if cached_image.present?
```

sykesm:
Please see the comments. Overall it looks pretty good.

Two issues that need to be resolved:
1. The missing test for the dangling if in protocol.rb
2. A decision on whether or not we need the column that can hold the cached image to be text. (May not be necessary.)

Thanks.
@sykesm & @atulkc

lib/cloud_controller/diego/docker/staging_completion_handler.rb
hsiliev: The idea here was to remove the `cached_docker_image` if the user opted-out of the docker caching. This should allow re-staging with another docker image ref/url and starting new instances of the app.

sykesm:
manually merged as fd7b435ae735b95f18ca93d9d97b1512b1d3e9d6

spec/unit/lib/cloud_controller/diego/docker/staging_completion_handler_spec.rb
```diff
@@ -43,6 +43,96 @@ module Docker
...
+
+              it 'does not update the cached image' do
+                handler.staging_complete(staging_guid, payload)
+
+                expect(droplet.cached_docker_image).to be_nil
+              end
+            end
+          end
+
+          shared_examples 'a handler with no lifecycle data' do
```
maximilien: I'd question why creating a shared example when it seems (from this PR) being used only twice...  

I say this because, shared example have caused a lot of "pain" in services team that the policy has been to avoid them. I'd say especially in a situation when the shared example is not shared so many times...  

Finally, if you still decide to use shared example, might be good to aggregate in one place.

Anyhow, my .02

PR #355, opened 2015-04-08
Added order-by query parameter

shashidharatd:
Now user can get sorted response based on order set provided.
Order set is comma-delimited.
This patch is submitted to implement CLI story
https://www.pivotaltracker.com/n/projects/892938/stories/91314240
https://www.pivotaltracker.com/n/projects/892938/stories/62876392
to get org/space list alphabetically sorted.

### DIFF
--- a/app/controllers/base/model_controller.rb
+++ b/app/controllers/base/model_controller.rb
@@ -7,6 +7,32 @@ module VCAP::CloudController::RestController
 
     attr_reader :object_renderer, :collection_renderer
 
+    def initialize(config, logger, env, params, body, sinatra=nil, dependencies={})
+      validate_parameters(params)
+      super(config, logger, env, params, body, sinatra, dependencies)
+    end
+
+    def validate_parameters(params)
+      valid_params = {}
+      [
+        ['order-by',               String],
+
+      ].each do |key, klass|
+        val = params[key]
+        valid_params[key.underscore.to_sym] = Object.send(klass.name, val) if val
+      end
+
+      if valid_params[:order_by]
+        valid_params[:order_by].split(',').each do |col|
+          unless self.class.query_parameters.to_a.include?(col)
+            raise VCAP::Errors::ApiError.new_from_details(
+                'BadQueryParameter',
+                "invalid request parameter '#{col}' in order_by")
+          end
+        end
+      end
+    end
+
     def inject_dependencies(dependencies)
       super
       @object_renderer = dependencies.fetch(:object_renderer)
--- a/spec/unit/lib/rest_controller/paginated_collection_renderer_spec.rb
+++ b/spec/unit/lib/rest_controller/paginated_collection_renderer_spec.rb
@@ -293,6 +293,44 @@ module VCAP::CloudController::RestController
         end
       end
 
+      context 'when order-by' do
+        before do
+          VCAP::CloudController::TestModel.make
+          VCAP::CloudController::TestModel.make
+          VCAP::CloudController::TestModel.make
+        end
+
+        let(:opts) do
+          {
+            results_per_page: 1,
+            order_by: order_by,
+            page: 2
+          }
+        end
+
+        context 'is specified' do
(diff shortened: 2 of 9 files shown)


spec/unit/lib/rest_controller/paginated_collection_renderer_spec.rb
```diff
@@ -293,6 +293,44 @@ module VCAP::CloudController::RestController
         end
       end
 
+      context 'when order-by' do
```
maximilien: What happens if one of the `order-by` values is not valid, e.g., `bogus`? Should you just pass it or should we attempt to fail fast soon? 

Also, small mistakes like: `order-by=id, ,name` will result in `['id', ' ', 'name']`. But maybe this is not something that can be prevented.

Anyhow, just curious. My .02

shashidharatd:
Thanks for review @maximilien. We will put some checks on user data validation.

utako:
Hi @shashidharatd,

If we were to introduce an 'order by' query parameter, we would feel more comfortable only sorting by a valid database column (ie name) instead of sorting by arbitrary parameters. We'd like to avoid handling unforeseen use cases.

Thanks,
@DanLavine and @utako, CF Runtime Team

shashidharatd:
Hi @utako,
  We agree with the comments and would change the code to handle invalid data(database column) in the query request.
We would like to know which is the better choice, when handling invalid database column.
Option-1: Send error response to the user query, if any column is invalid.
Option-2: Ignore the invalid column(s) and continue ordering based on valid columns.

Thanks
@shashidharatd and @AmitRoushan

crhino:
Hi @shashidharatd,

Thanks for implementing the `BadQueryParameter` error case. We also think that a user should not be able to order-by any column on in the dataset, as it may have adverse effects on database performance if the column is not indexed. We would prefer if the `order-by` attribute only accepted columns that we have specified as `queryable_attributes` on controllers. You can access these values on a controller by calling `self.class.query_parameters`. [You can view an example of that here](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/controllers/base/model_controller.rb#L248). If you could make it so that we only accept these values for the order-by column, that would be great.

Thanks,

@jfmyers9 and @crhino, CF Runtime Team

shashidharatd:
Hi @crhino,
  Thank you for the comments. We have updated the code so that order-by attribute accepts only query-able columns as parameters. please check the code if modifications are inline with the expectations.

Thanks
@AmitRoushan and @shashidharatd

app/controllers/base/model_controller.rb
```diff
@@ -7,6 +7,32 @@ class ModelController < BaseController
...
+        valid_params[:order_by].split(',').each do |col|
+          unless self.class.query_parameters.to_a.include?(col)
+            raise VCAP::Errors::ApiError.new_from_details(
+                'BadQueryParameter',
+                "invalid request parameter '#{col}' in order_by")
+          end
+        end
+      end
+    end
+
```
wendorf: This validation should live outside the controller. `OrderApplicator` looks like the correct home. Look at [query.rb](https://github.com/cloudfoundry/cloud_controller_ng/blob/26b6cdc812800f62c8166444ca4feff79cc38316/lib/vcap/rest_api/query.rb#L57) for an example of injecting the controller's `query_parameters` into another class to do the validation.

This will also reduce the complexity of the validation (`valid_params` shouldn't be necessary to build), since you will only need to worry about the `order-by` parameter.

wendorf:
Since the PR is not a bit stale, please rebase on top of the latest master branch.

Otherwise, your PR looks good. If you rebase and address the model_controller notes, we should be able to merge it.

wendorf:
Hi @shashidharatd ,

This PR has been open for a while. Are you planning on addressing the concerns above or should we close this PR?

Thanks,
@wendorf & @jberkhahn , Cloud Foundry CAPI team

SocalNick:
Closing in favor of feature: https://www.pivotaltracker.com/story/show/116232501

PR #356, opened 2015-04-09
Add audit event for create-service-key

xingzhou:
story number: [#87481308]

### DIFF
--- a/app/controllers/services/service_keys_controller.rb
+++ b/app/controllers/services/service_keys_controller.rb
@@ -10,6 +10,15 @@ module VCAP::CloudController
     get path,      :enumerate
     query_parameters :name, :service_instance_guid
 
+    def self.dependencies
+      [:services_event_repository]
+    end
+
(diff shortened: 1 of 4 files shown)


xingzhou:
cc: @maximilien , @shalako 

We noticed that for other controllers in CC, there are no test cases for the dependencies related method, as they are public method for controllers, we also created some new test cases to cover the test of these methods, we are thinking it's better to supplement the test cases for other controllers, please check if the test cases we added are proper and if needed, we can help on adding test cases for other controllers.

maximilien:
Cool, since the general policy is to have tests for public methods, good to add them.

Unless others feel otherwise, might make sense to create a chore to do the same for other controllers. 

/cc @dsabeti, @jeffh, and @rmasand

jeffh:
Looks good :+1:. We probably won't backfill tests related to dependencies, because they're implicitly tested when the CC is able to start up. But the tests you've written are fine as-is and can be left in with the rest of the commit.

PR #359, opened 2015-04-13
add API doc for list and delete service key

xingzhou:
implement story [#90469846] and [#90471234]

### DIFF
--- a/spec/api/api_version_spec.rb
+++ b/spec/api/api_version_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 require 'digest/sha1'
 
 describe 'Stable API warning system', api_version_check: true do
-  API_FOLDER_CHECKSUM = '58773324449147857429086f9ca4a380817bcd96'
+  API_FOLDER_CHECKSUM = '1168d221e57996aae58b9ded56e60062d003e5be'
 
   it 'double-checks the version' do
     expect(VCAP::CloudController::Constants::API_VERSION).to eq('2.25.0')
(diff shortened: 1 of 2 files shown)


spec/api/api_version_spec.rb
```diff
@@ -2,7 +2,7 @@
 require 'digest/sha1'
 
 describe 'Stable API warning system', api_version_check: true do
-  API_FOLDER_CHECKSUM = '58773324449147857429086f9ca4a380817bcd96'
+  API_FOLDER_CHECKSUM = '1168d221e57996aae58b9ded56e60062d003e5be'
```
maximilien: Generally, if I recall correctly, the team likes having thee API checksum as a separate commit...  Maybe it's OK for this one.

spec/api/documentation/service_keys_api_spec.rb
```diff
@@ -5,13 +5,18 @@
...
   before do
     create_response_body = { 'credentials' => service_instance.credentials }
     stub_bind(service_instance, body: create_response_body.to_json)
+    stub_unbind(service_key)
   end
 
+  standard_model_list :service_key, VCAP::CloudController::ServiceKeysController
+  standard_model_delete :service_key, async: false
+
   post '/v2/service_keys' do
```
maximilien: How come there is no `delete` method?

spec/api/api_version_spec.rb
xingzhou: the checksum will affect the unit test result, so if we submit it in separate PR, it will lead the failure of one test case

spec/api/documentation/service_keys_api_spec.rb
xingzhou: the standard_model_delete method will test the delete function of the model, so this will cover the delete API doc

PR #361, opened 2015-04-14
Remove space_id from events

fraenkel:
- Use space_guid, space_id is redundant and expensive

[#92032548]


dieucao:
@fraenkel I think this was addressed a while back, correct?  Okay to close this one?

fraenkel:
Yes. I know you have another story in the icebox to actually remove the space_id.

PR #364, opened 2015-04-16
Fix some RuboCop violations

xingzhou:
Fix some RuboCop violations


zaksoup:
@xingzhou Looks like we fixed this internally at the same time as you submitted this. Thanks for catching it, though!

PR #365, opened 2015-04-16
add job_name to registration with collector

holgero:
will have an effect only if vcap-common gets a new release
that uses the job_name

Signed-off-by: Marco Voelz <marco.voelz@sap.com>


jpalermo:
We've decided not to merge these pull requests in. They do provide a piece of data the is currently missing, but the number of changes that will be required to the jobs and their templates seems to great considering this functionality is already planned for inclusion in loggregator/metron/firehose components.

PR #368, opened 2015-04-22
Propagate actual error message upon staging failure

georgethebeatle:
Give the user some more context upon staging failure

### DIFF
--- a/lib/cloud_controller/diego/stager_client.rb
+++ b/lib/cloud_controller/diego/stager_client.rb
@@ -78,8 +78,8 @@ module VCAP::CloudController
       end
 
       def error_message(response)
-        JSON.parse(response.body)['error']['message'] || response.code
-      rescue JSON::ParserError, NoMethodError
+        JSON.parse(response.body).fetch('error', {})['message'] || response.code
+      rescue JSON::ParserError
         response.code
       end
     end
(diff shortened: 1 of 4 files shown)


dieucao:
@georgethebeatle Thanks for your contribution.  Could you make your membership of cloudfoundry-incubator public? Thanks.

georgethebeatle:
Should be public now

lib/cloud_controller/diego/stager_client.rb
```diff
@@ -76,6 +76,12 @@ def http_client
       def logger
         @logger ||= Steno.logger('cc.stager.client')
       end
+
+      def error_message(response)
+        JSON.parse(response.body)['error']['message'] || response.code
+      rescue JSON::ParserError, NoMethodError
```
matt-royal: Are you rescuing NoMethodError to handle the case when the request doesn't have an "error" key? This will hide exceptions that occur when other objects are nil unexpectedly. It would be better to deal with the problem directly. Here is one way you could do that...

```ruby
JSON.parse(response.body).fetch('error', {})['message'] || response.code
```

matt-royal:
Could you please address the comment above? Once that is addressed, we believe we can merge this.

georgethebeatle:
It turned out that this proposal is merged by the Diego team with another commit. Closing.

Issue #371, opened 2015-04-27
Cannot reset the command if set on the initial create/push of an application


fraenkel:
Easiest way to reproduce it.

1. cf push myapp -c abc --no-start
2. cf push myapp -c null --no-start

cf curl /v2/apps/GUID will show that the command is still abc.

The problem is during the app creation, console touches metadata which populates the metadata with the command. The command is then stored in the metadata as part of the app creation.

Any further attempt to reset/remove the command will pick it up from the metadata.

dieucao:
This was addressed by #373

PR #373, opened 2015-04-28
Commands are not metadata

fraenkel:
Allow commands to be reset to nothing
Merge command into metadata when necessary but don't save it as metadata

[#93386810]


fraenkel:
Fixes Issue #371

PR #374, opened 2015-04-30
Respond correct error message if service broker does not support binding...

xingzhou:
... without app_guid

implement of story [#87485072]


lib/services/service_brokers/v2/response_parser.rb
```diff
@@ -28,7 +28,7 @@ def parse_provision_or_bind(path, response)
           when 409
             FailingValidator.new(Errors::ServiceBrokerConflict)
           when 422
-            FailWhenValidator.new('error', ['AsyncRequired'], Errors::AsyncRequired,
+            FailWhenValidator.new('error', { 'AsyncRequired' => Errors::AsyncRequired, 'RequiresApp' => Errors::AppRequired },
```
dsabeti: There's a very subtle mistake here. The RequiresApp error is only appropriate for the bind request, but this implementation would also allow provision requests to pass this error. If you write two tests, one for provision and one for bind, both of which pass the RequiresApp error, both tests will show the client raising AppRequired error, even though in the case of provision, the RequiresApp error should be ignored.

Look just a few lines above this at the IfElsePathMatcher. This allows us to give a condition and two validators; one runs if the condition is true, and the other validator runs if the condition is false. We can use this to make sure that we have two different FailWhenValdiators, one for provision and one for bind.

xingzhou:
Updated the PR according to David's comment

dsabeti:
Hi @xingzhou 

There's actually still a bug in the code, and it can be driven out with a just a few tests. Take a look at response_parser_spec.rb. This file contains a full list of test cases for the response parser in a very condensed format. If you look at https://github.com/cloudfoundry/cloud_controller_ng/blob/master/spec/unit/lib/services/service_brokers/v2/response_parser_spec.rb#L206, you can see that we have a test for AsyncRequired errors raised during provision. We could write a similar test for RequiresApp:
```
test_case(:bind, 422, { error: 'RequiresApp' }.to_json,                      error: Errors::AppRequired)
```

Also, we actually need to backfill a test for AsyncRequired during bind requests:
```
test_case(:bind, 422, { error: 'AsyncRequired' }.to_json,                    error: Errors::AsyncRequired)
```

Prior to your change, this backfilled test would pass. But with the change, it fails. This is because we need to handle AsyncRequired for both provision and bind, but the new change currently only handles AsyncRequired for provision, and RequiresApp for bind.

In the first iteration of this commit, you actually included a change to FailWhenValidator that allowed many kinds of errors to be handled. Maybe that would work here?

Ultimately, we need both of the tests above to pass.

xingzhou:
@dsabeti, sure, will backfill these two tests and make the changes to FailWhenValidator, thanks

dsabeti:
This last commit looks great! The pipelines are all blocked right now, but I'll merge this as soon as they go green again.

PR #375, opened 2015-04-30
Implement the orphan_mitigator for creating service_key

xingzhou:
story [#89967302]

### DIFF
--- a/lib/services/service_brokers/v2/client.rb
+++ b/lib/services/service_brokers/v2/client.rb
@@ -79,15 +79,35 @@ module VCAP::Services::ServiceBrokers::V2
       result.merge(parsed_response.symbolize_keys)
     end
 
+    def create_service_key(key, arbitrary_parameters: {})
+      path = service_binding_resource_path(key)
+      attr = {
+          service_id:  key.service.broker_provided_id,
+          plan_id:     key.service_plan.broker_provided_id
+      }
+
+      attr[:parameters] = arbitrary_parameters if arbitrary_parameters.present?
+
+      response = @http_client.put(path, attr)
+      parsed_response = @response_parser.parse_provision_or_bind(path, response)
+
+      attributes = {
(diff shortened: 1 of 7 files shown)


lib/services/service_brokers/v2/client.rb
```diff
@@ -79,15 +79,35 @@ def fetch_service_instance_state(instance)
       result.merge(parsed_response.symbolize_keys)
     end
 
+    def create_service_key(key, arbitrary_parameters: {})
```
dsabeti: Instead of adding a method like this, can we just add a little more logic to the `bind` method? There is so much duplication between `bind` and `create_service_key`, I think it's better to have one method, and it choose which orphan mitigation technique to use. Something like this maybe:
```
    ...
    rescue Errors::ServiceBrokerApiTimeout, Errors::ServiceBrokerBadResponse => e
      if binding.respond_to? :app_guid
        @orphan_mitigator.cleanup_failed_bind(@attrs, binding)
      else
        @orphan_mitigator.cleanup_failed_key(@attrs, binding)
      end
      raise e
    end
```

xingzhou: @dsabeti, yes, it's duplicate of these two methods, one thing I'm thinking is create-service-key is just reusing the name of "binding", is it possible that in the future, broker will support two kinds of endpoint which will support create-service-key and service-binding separately? this may make the code more clear. if you think it's better to reuse the "bind" method, I'll change it and resubmit this one, thanks

dsabeti: Hi Tom,

It's a good point. I don't think we'll ever want to have both endpoints (one for create-service-key and another for service-binding), but I also think that your implementation might be more clear than the one I suggested. Let's stick with two methods for now, and if the duplicate code becomes a problem we can merge them later.

I'll merge this as soon as the pipeline goes green.

PR #377, opened 2015-05-08
Add private docker image credentials to app

hsiliev:
Adds docker_login_server, docker_auth_token and docker_email used to authenticate in Docker Hub to fetch metadata and pull private images.

[#86969554]


hsiliev:
The code in this PR won't work correctly since in-memory data populated by the controller during app create/update:
- is not available to other CC instances/servers/VMs, unless stored in DB
- breaks the current flow `cf push <app> --no-start`, then `cf start <app>`

PR #378, opened 2015-05-08
current usage against the org quota

AmitRoushan:
Provide current status of routes, service instances and memory usage of organization quota.
[#91309412](https://www.pivotaltracker.com/story/show/91309412)

If this implementation looks good then I would like to do PR for space quota [#91309436](https://www.pivotaltracker.com/story/show/91309436) also.

### DIFF
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -217,6 +217,10 @@ module CloudController
       create_paginated_collection_renderer(collection_transformer: UsernamesAndRolesPopulator.new(username_lookup_uaa_client))
     end
 
+    def quota_usage_populating_renderer
+      create_object_renderer(transformer: QuotaUsagePopulator.new)
+    end
+
     def username_lookup_uaa_client
       client_id = @config[:cloud_controller_username_lookup_client_name]
       secret = @config[:cloud_controller_username_lookup_client_secret]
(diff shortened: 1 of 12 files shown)


lib/cloud_controller/dependency_locator.rb
```diff
@@ -201,6 +201,10 @@ def paginated_collection_renderer
       create_paginated_collection_renderer
     end
 
+    def object_collection_renderer
```
jpalermo: Has this method been added for a reason? We're only seeing it used in tests.

goehmen: Love this.  Can't wait to expose it in the CLI.  I was going to try to
accomplish this via a plugin but happy to just hit CC endpoints!

Greg Oehmen
Cloud Foundry Product Manager
415.205.6596

AmitRoushan:
Hi @jpalermo 
Thanks for review.
We have added object_collection_renderer to test create_object_collection_renderer function only.

Thanks 
@AmitRoushan

jpalermo:
I'm not sure create_object_collection_renderer makes sense. A collection renderer is responsible for taking a collection of things and rendering them all, here you only have a single thing.

Similarly the `ObjectRenderer` should not have a `collection_transformer`, but just a `transformer`since it only deals with single objects and not collections.

AmitRoushan:
Hi @jpalermo 
Thanks again for reviewing and providing your valuable comments.
As per your suggestion, I have removed 'collection' from possible places.
Please check if modifications are inline with the expectations.

Thanks
@AmitRoushan

lib/cloud_controller/dependency_locator.rb
```diff
@@ -201,6 +201,10 @@ def paginated_collection_renderer
       create_paginated_collection_renderer
     end
 
+    def created_object_renderer
```
jpalermo: This method is never used by quota_usage_populating_renderer, or any other production code. It should probably be removed.

Tests for the values passed to ObjectRenderer.new, should just be moved to the test for quota_usage_populating_renderer.

Since the implementation of quota_usage_populating_renderer could change, having a test on the created_object_renderer doesn't provide much use.

AmitRoushan:
Hi @jpalermo 
As per your suggestion, I have removed 'created_object_renderer' and added testcases for create_object_renderer in quota_usage_populating_renderer context.
I am not sure about why CI is not triggered.
Thanks again for review comments.

Thanks
@AmitRoushan

PR #381, opened 2015-05-19
Add private docker image credentials to app

hsiliev:
[#86969554]

This pull request enables storing encrypted credentials for the Docker Hub in ccdb. Please have a look and tell us what you think?

We are concerned about encrypting several things with the same salt. We now have 5 columns (server, user, password, email and salt). We can either:
* combine all of these into single json
* add salt columns for all of the credentials
* any other ideas are welcome

### DIFF
--- a/app/controllers/base/model_controller.rb
+++ b/app/controllers/base/model_controller.rb
@@ -5,6 +5,8 @@ module VCAP::CloudController::RestController
   class ModelController < BaseController
     include Routes
 
+    CENSORED_MESSAGE = 'PRIVATE DATA HIDDEN'.freeze
+
     attr_reader :object_renderer, :collection_renderer
 
     def inject_dependencies(dependencies)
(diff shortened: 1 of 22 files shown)


luan:
Hi @hsiliev,

I'm not particularly concerned with the encrypting many things with the same salt issue, they're all in the same row anyway. I think I like the idea of having them in a single JSON. I expected docker to have a better way of expressing login info (such as a URI like user:pass@server) but it seems like you need user AND email there which rules that option out.

What is interesting is that we have to do all of this in the `App`, and I really see no better way of doing this for V2. We should consider this when migrating docker "Packages"/"Droplets" to V3 so we can clean up some of the App code. I don't know if we want to be tackling any of that right know though.
(/cc @zrob)

Thanks,
@luan, CF Diego Team

georgethebeatle:
We moved all docker credentials to a single JSON attribute/column. We also added validation logic. Please take a look.

PR #383, opened 2015-05-22
Current usage against space quota

AmitRoushan:
Provide current status of routes, service instances and memory usage for space against associated space quota.
[#91309436]

As of now we are returning error when space is not associated with any quota.
So one idea we can think of is to populate space_usage in 'QuotaDefinition' model when space is not associated with any space quota.
We need feedback from Runtime Team on this.

### DIFF
--- a/spec/unit/controllers/runtime/spaces_controller_spec.rb
+++ b/spec/unit/controllers/runtime/spaces_controller_spec.rb
@@ -194,6 +194,22 @@ module VCAP::CloudController
       end
     end
 
+    describe 'GET /v2/spaces/:guid/quota_usage' do
+      context 'for an space that does not exist' do
+        it 'returns a 404' do
+          get '/v2/spaces/foobar/quota_usage', {}, admin_headers
+          expect(last_response.status).to eq(404)
+        end
+      end
+      context 'when the user does not have permissions to read' do
+        let(:user) { User.make }
+        it 'returns a 403' do
+          get "/v2/spaces/#{space_one.guid}/quota_usage", {}, headers_for(user)
+          expect(last_response.status).to eq(403)
+        end
+      end
+    end
+
     describe 'GET /v2/spaces/:guid/service_instances' do
       let(:space) { Space.make }
       let(:developer) { make_developer_for_space(space) }
--- a/spec/unit/models/runtime/space_quota_definition_spec.rb
+++ b/spec/unit/models/runtime/space_quota_definition_spec.rb
@@ -60,5 +60,15 @@ module VCAP::CloudController
         expect { space_quota_definition.destroy }.to change { space.reload.space_quota_definition }.from(space_quota_definition).to(nil)
       end
     end
+
+    describe '#to_hash' do
+      it 'does not include space usage when space_usage has not been set' do
+        expect(space_quota_definition.to_hash).to_not include('space_usage')
+      end
+      it 'includes space usage when space_usage has been set' do
+        space_quota_definition.space_usage = 'someusage'
+        expect(space_quota_definition.to_hash).to include('space_usage')
+      end
+    end
   end
 end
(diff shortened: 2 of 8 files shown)


DanLavine:
Hi @AmitRoushan,

This seems like a fine implementation. The one thing I would be cautious about is eager loading all of the dependencies onto the space in the quota_usage_populator. Have you considered just selecting the count directly? 

Thanks!
@DanLavine && @MarcSchunk, The CF Runtime Team

AmitRoushan:
Hi @DanLavine && @MarcSchunk 

Thanks for reviewing and providing comments. I have removed eager loading while populating space quota usage. Please check if modifications are inline with the expectations.

Thanks
@AmitRoushan && @ShwethaKumbla

zaksoup:
Hey @AmitRoushan + @ShwethaKumbla 

This looks good from a technical standpoint. Since it's adding a new API endpoint we'd like to talk it over with our PM before merging.

Thanks for your contributions!
@zaksoup + @wendorf CF Runtime Team

zaksoup:
We'll be merging this in today. Thanks again for your contributions!

spec/unit/controllers/runtime/spaces_controller_spec.rb
```diff
@@ -194,6 +194,22 @@ module VCAP::CloudController
...
+          expect(last_response.status).to eq(404)
+        end
+      end
+      context 'when the user does not have permissions to read' do
+        let(:user) { User.make }
+        it 'returns a 403' do
+          get "/v2/spaces/#{space_one.guid}/quota_usage", {}, headers_for(user)
+          expect(last_response.status).to eq(403)
+        end
+      end
```
maximilien: what about the "posive" context when user has permission and quota exists, we should have a context and test(s) to return it.

Also, add new lines between contexts to keep things easier to read and consistent.

spec/unit/models/runtime/space_quota_definition_spec.rb
```diff
@@ -60,5 +60,15 @@ module VCAP::CloudController
         expect { space_quota_definition.destroy }.to change { space.reload.space_quota_definition }.from(space_quota_definition).to(nil)
       end
     end
+
+    describe '#to_hash' do
+      it 'does not include space usage when space_usage has not been set' do
+        expect(space_quota_definition.to_hash).to_not include('space_usage')
+      end
+      it 'includes space usage when space_usage has been set' do
```
maximilien: new line, same as above...

spec/unit/models/runtime/space_quota_definition_spec.rb
```diff
@@ -60,5 +60,15 @@ module VCAP::CloudController
...
+        space_quota_definition.space_usage = 'someusage'
+        expect(space_quota_definition.to_hash).to include('space_usage')
+      end
+    end
```
maximilien: Are we only testing "space_usage"?

utako:
Hi @AmitRoushan,

After discussing this and the previous pull request #378 we have decided we'd like to make some changes to these endpoints and their responses. In light of this discussion, we've decided to revert the merges for these pull requests, pending these changes:

1. We'd like to see the space and org usages separated from logic around quotas. After reviewing the implementation of the usage populating renderer we would like you to get usages directly from the org and space, without first getting the quota. This would take care of our second request, which is that we would like to return a usage when there's no quota definition applied (we are currently returning a 400 error).
1. In the response to the usages endpoints we'd like the name of each key to specify units. In this case this will affect only one key: `memory` to `memory_in_mb`.
1. Given the above change we believe the endpoints should be renamed to `/organizations/:guid/usage` and `/organizations/:guid/spaces/:guid/usage`

Thank you for your continued contributions!
@utako + @zaksoup, CF Runtime Team

AmitRoushan:
Hi @utako && @zaksoup ,

Apologies for the delay in responding. I was busy with some version work.
Thanks for providing detailed info and expectation from both of PRs.
I will certainly take up this task and try to finish it within this week.

Thanks 
@AmitRoushan

PR #394, opened 2015-06-12
Encryptor fips

ScarletTanager:
I updated lib/cloud_controller/encryptor.rb so that it can be used when OpenSSL::fips_mode = true is set.  This required moving off of the (deprecated) keyivgen method and following the better practice of using a random IV for every message, which IV is then returned as part of the encrypted string for use in decryption.  I updated the spec file so that it enables fips_mode, but I tagged this so that you can avoid enabling fips mode with rspec -t ~fips. 

RSPEC: 

[sandy@oc1718273258 cloud_controller_ng]$ bundle exec rspec spec/unit/lib/cloud_controller/encryptor_spec.rb 
.......................

Finished in 3.38 seconds (files took 42.58 seconds to load)
23 examples, 0 failures

RSPEC (fips disabled): 

[sandy@oc1718273258 cloud_controller_ng]$ bundle exec rspec -t ~fips spec/unit/lib/cloud_controller/encryptor_spec.rb 
Run options: exclude {:fips=>true}
......................

Finished in 3.01 seconds (files took 1 minute 6.34 seconds to load)
22 examples, 0 failures

RUBOCOP:

[sandy@oc1718273258 cloud_controller_ng]$ bundle exec rubocop
warning: parser/current is loading parser/ruby21, which recognizes
warning: 2.1.5-compliant syntax, but you are running 2.1.6.
Inspecting 1228 files
[...]
1228 files inspected, no offenses detected

### DIFF
--- a/lib/cloud_controller/encryptor.rb
+++ b/lib/cloud_controller/encryptor.rb
@@ -1,5 +1,5 @@
 require 'securerandom'
-require 'openssl/cipher'
+require 'openssl'
 require 'base64'
 
 module VCAP::CloudController::Encryptor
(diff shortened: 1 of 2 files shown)


ScarletTanager:
bundle exec rake spec on my system runs successfully w/o errors.  I will work on a modification to encryptor_spec.rb to let it run with a non-FIPS OpenSSL.

ScarletTanager:
I've decided to close this for now - I need to rethink how to introduce this, and I think it will involve retaining the current (IMHO, less-than-ideal) encryption while enabling FIPS mode via a configuration flag or similar (with it disabled by default).

jpalermo:
Hi @ScarletTanager,

There is a desire to have the ability to rotate the encryption keys used with CloudController. No work has been scheduled, but that work might overlap nicely with this effort. There is some context on the cf-dev mailing list:
http://lists.cloudfoundry.org/pipermail/cf-dev/2015-June/000416.html

PR #399, opened 2015-06-25
fix API behavior for SpaceManager and SpaceAuditor to receive 403

zhang-hua:
this PR is to fix the bug of both SpaceManager and SpaceAuditor should
receive 403 when list keys for a service instance, not 0 results.

[#87481016]

Signed-off-by: Hua Zhang <zhuadl@cn.ibm.com>

### DIFF
--- a/app/access/service_key_access.rb
+++ b/app/access/service_key_access.rb
@@ -14,5 +14,11 @@ module VCAP::CloudController
       return true if admin_user?
       service_key.service_instance.space.developers.include?(context.user)
     end
+
+    def index?(_, params=nil)
+      return true if admin_user?
+      return true unless params && params.key?(:related_obj)
+      params[:related_obj].space.developers.include?(context.user)
+    end
   end
 end
--- a/spec/unit/controllers/services/service_instances_controller_spec.rb
+++ b/spec/unit/controllers/services/service_instances_controller_spec.rb
@@ -2319,14 +2319,14 @@ module VCAP::CloudController
         context 'when the user is not of developer role' do
           it 'return an empty service key list if the user is of space manager role' do
             get "/v2/service_instances/#{instance_a.guid}/service_keys", {}, headers_for(manager)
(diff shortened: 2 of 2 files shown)


app/access/service_key_access.rb
```diff
@@ -14,5 +14,11 @@ def read?(service_key)
       return true if admin_user?
       service_key.service_instance.space.developers.include?(context.user)
     end
+
+    def index?(_, params=nil)
+      return true if admin_user?
+      return true unless params && params.key?(:related_obj)
```
rmasand: Could you explain more about this line? It looks like we'll continue to the next line in the case of developer/manager/auditor role. There isn't any test coverage for the case where we return true, so we don't understand what kind of user it would apply to. Did you mean to put false here?

zhang-hua: @rmasand : the basic model controller is using validate_access for associated model of service instance to check if the current user has access to the endpoint (/v2/service_instances/:guid/service_keys) in enumerate_related method at [117](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/controllers/base/model_controller.rb#L117). The service instance object will be passed to `params` of `index?` method through line [129](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/controllers/base/model_controller.rb#L129). But the index method will also apply the existing service key endpoint  (/v2/service_keys). So for this case, it will be validate with true value and leave it to be checked by user_visibility at line [134](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/controllers/base/model_controller.rb#L134). This is expected API behavior to return 0 results. Does it make sense? Thanks for your comments.

rmasand:
Thanks for your response, the organization is more clear now. I see the test coverage in `service_keys_controller_spec.rb` for the enumerate endpoint, and the rest is covered in `service_key_access_spec.rb`. Looks good! Will merge when the pipeline is ready for new commits.

PR #401, opened 2015-07-06
Increase open and read timeouts for TPS client

hsiliev:
[#97390140]


jfmyers9:
We don't believe that the failures in [#97390140] will be fixed with these increased timeouts, and we feel uncomfortable increasing these timeouts as they seem to be working for all other requests. Let's continue the conversation in [#97390140]

PR #416, opened 2015-08-24
Service binding validation should assert syslog_drain_url is empty

mheath:
Service binding validation should assert syslog_drain_url is empty rather than whether it's present or not.

Many service brokers (noteably the ones using https://github.com/cloudfoundry-community/spring-boot-cf-service-broker will error out at service binding time because they return a null value for syslog_drain_url.

[#101927804]

### DIFF
--- a/lib/services/service_brokers/v2/response_parser.rb
+++ b/lib/services/service_brokers/v2/response_parser.rb
@@ -213,7 +213,7 @@ module VCAP::Services
           def validate(method:, uri:, code:, response:)
             service = VCAP::CloudController::Service.first(guid: @service_guid)
             parsed_response = MultiJson.load(response.body)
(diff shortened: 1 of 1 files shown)


jeffpak93:
Hey @mheath ,

This change looks pretty reasonable! If you could just add a test to be written in addition to the commit here, we'd be happy to accept this PR. Also, we were thinking that `!parsed_response['syslog_drain_url'].nil?` could be simplified to  `parsed_response['syslog_drain_url']`.

rmasand:
Wrote some tests locally and merged the original changes. 

Thanks,
@rmasand and @utako 
CF CAPI Team

mheath:
That's funny. I was just about to to get back to writing a test for this
change. Thanks Raina!

PR #418, opened 2015-08-25
Added new events table index on timestamp and id.

andrew-edgar:
removed timestamp only index as now it is not required.

Cloud controller now added "id" to the order by in /v2/events we need to update the index so that queries on large tables do not do a seq scan and offline disk sort.  Now it will use the index to get the sorted results.

### DIFF
--- a/db/migrations/20150825075900_add_events_index.rb
+++ b/db/migrations/20150825075900_add_events_index.rb
@@ -1,0 +1,11 @@
+Sequel.migration do
+  up do
+    add_index :events, [:timestamp, :id]
+    drop_index :events, :timestamp
+  end
+
+  down do
+    add_index :events, :timestamp
+    drop_index :events, [:timestamp, :id]
+  end
+end


wendorf:
Hi @andrew-edgar ,

Thank you for the PR -- it does look like this would speed up the events controller. We would need the migration to do a truncation as well if we did this due to the size of the events table, so we're running it by our product manager @dieucao . You can track that conversation on the corresponding Pivotal Tracker story [here](https://www.pivotaltracker.com/story/show/101985370).

zrob:
Hey @andrew-edgar 

We'd like to pull this in, but with a truncate for the events table in the up part of the migration.  Would you like to update the PR?  If not we can just make the change on our end.

Thanks,
@zrob & @jberkhahn

andrew-edgar:
Hi @zrob & @jberkhahn,

I'm fine with you making the change on your end.  I'm not sure what the command is to do a truncation but I would be interested when you do it and I'll take a look :).

Thanks,

Andrew

Issue #419, opened 2015-08-27
Need a way to configure how many stagers to select across


fraenkel:
In larger environments, selecting one from the top 5 isn't always ideal. In the case of freshly rolled DEAs (although this is a more serious issue) or imbalanced DEAs, the selection will tend to select the same 5 DEAs for all future staging requests until they have more running applications.

During this window which can be days, users experience frequent staging failures because they are all competing for the disk and network.

Ideally we would want to be smarter, however, in the interim, it might prove better to at least spread the pain to more DEAs. Whether its a "fixed" number or a percentage of known DEAs doesn't really matter. The goal is to improve the end user perception.

dieucao:
The PR has been merged and should be available in the next cf-release.

PR #421, opened 2015-08-28
Select from a larger pool of stagers if needed

fraenkel:
As the number of potential DEA stagers grow, allow the pool to grow as well.


fraenkel:
One way to solve issue #419

zrob:
@fraenkel this seems pretty solid.   We were thinking we go with minimum of 5, make the percentage configurable with a default of the 1% you have now.   

We'll pull it in if you can add some tests and the configurability change.

@dieucao do you have a preference for the manifest property for the percentage?

dieucao:
How about `placement_top_stager_percentage` ? It's 4 words though...

fraenkel:
Just updated the code. I don't have a fast internet so I cannot test the BOSH changes in cf-release yet.

wendorf:
Hi @fraenkel ,

Looks good! We just merged it and will bump cf-release to have the change.

Thanks,
@wendorf and @jberkhahn , Cloud Foundry CAPI team

PR #422, opened 2015-08-28
Routing api v2 info

msmykowski:
Added routing_api endpoint to the v2/info endpoint

-routing_api is optional

### DIFF
--- a/app/controllers/runtime/info_controller.rb
+++ b/app/controllers/runtime/info_controller.rb
@@ -19,6 +19,10 @@ module VCAP::CloudController
         app_ssh_host_key_fingerprint: @config[:info][:app_ssh_host_key_fingerprint]
       }
 
+      if @config[:routing_api] && @config[:routing_api][:url]
+        info[:routing_endpoint] = @config[:routing_api][:url]
+      end
+
       if @config[:loggregator] && @config[:loggregator][:url]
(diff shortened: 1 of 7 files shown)


zrob:
hey, I'm trying to deploy this to bosh-lite and I'm getting the following error:
`Error 100: Error filling in template `cloud_controller_api.yml.erb' for `api_z1/0' (line 190: Can't find property `["routing_api.enabled"]')`

Where did you intend for this value to come from?

dieucao:
After discussion with @shalako, we think it would be better to take advantage of context path routes and make it available at:

api.system.domain/routing

So potentially this would be:
api.system.domain/routing/v1/routes

Closing this PR.

shalako:
@dieu 

Whether the route is `routing-api.system-domain.com` or `api.system-domain.com/routing`, having the url discoverable on /v2/info still seems valuable. This allows us to change the route later but not break clients. 

Do you plan to remove the urls for logging and authentication from /v2/info also?

dieucao:
@shalako Good point.  Sounds reasonable to make the url discoverable still in /v2/info for that reason.

@zrob had a comment above about the original PR.  Could you have your team review that and resubmit a PR?

shalako:
Yes, we'll resubmit with the new url.

PR #423, opened 2015-09-03
Enable nginx log format and destination config

sykesm:
Enable operators to provide custom access log formats and destinations.

This needs to be merged along with https://github.com/cloudfoundry/cf-release/pull/779.

### DIFF
--- a/bosh-templates/nginx.conf.erb
+++ b/bosh-templates/nginx.conf.erb
@@ -1,7 +1,7 @@
 # deployment cloudcontroller nginx.conf
 user  vcap vcap;
 
-error_log /var/vcap/sys/log/nginx_cc/nginx.error.log;
+error_log <%= p("cc.nginx_error_log_destination") %> <%= p("cc.nginx_error_log_level") %>;
 pid       /var/vcap/sys/run/nginx_cc/nginx.pid;
 
 events {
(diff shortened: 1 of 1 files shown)


wendorf:
Hi @sykesm,

This looks great! Merged.


Thanks,
@wendorf & @jberkhahn , CF CAPI team

bosh-templates/nginx.conf.erb
```diff
@@ -1,7 +1,7 @@
 # deployment cloudcontroller nginx.conf
 user  vcap vcap;
 
-error_log /var/vcap/sys/log/nginx_cc/nginx.error.log;
+error_log <%= p("cc.nginx_error_log_destination") %> <%= p("cc.nginx_error_log_level") %>;
```
dieucao: @sykesm What was the behavior previously with log level if none was specified?  I noticed the default you added on the cf-release PR is `error` https://github.com/cloudfoundry/cf-release/pull/779/files#diff-517e6c86965cf7638732689e362ecb1fR460  so just wanted to know if this is more or less or the same as before.

sykesm: Same: http://nginx.org/en/docs/ngx_core_module.html#error_log

PR #428, opened 2015-09-10
384 extraneous log stmts

ericpromislow:


zrob:
Hey ericpromislow, 

Thanks for contributing!

It does look like a lot of the logging in the model_controller and the various enumerate_related could be merged into the log line [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/controllers/base/front_controller.rb#L33), if the line was changed to include the route that was requested.  Would that change help with your issues?

We do find value in the blobstore stuff, in fact we've used those just this week to help debug blobstore issues in some of our dev environments.

We would also like to keep this focused on just the logging issues and not include the rspec or rake changes.

Thanks!

ericpromislow:
Thanks, zrob. Excuse the .rspec file, but "--color" is misspelled in it, so I'll file a separate bug and pull it out of this pr

I'll reinstate the blobstore statements, but could we get them downgraded to logger.debug?
When I push apps with large files there can be a lot of output to the CC log.

ericpromislow:
Updated the PR as requested.

rmasand:
Hi @ericpromislow,

Thanks for moving the unrelated work to different PR's.

We would like to find a solution where we can retain these logs in some form, while reducing some of the pain you're experiencing from them. I see that many of the logs you deleted are already being emitted by "logger.debug"; do you need to keep the logging level on debug for some other reasons? It would be helpful for us to understand what you're doing to filter the logs currently, so we can choose a solution more relevant to that filtering technique.

Also, @zrob suggested another solution of merging those logs into [this](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/controllers/base/front_controller.rb#L33) line, if consolidation would help. This would reduce the number of lines in the logs, but the resulting lines would probably have additional information (other than the current "Vcap-Request-Id" and "User"). Does this sound like something that would help you?

Let us know what you think.

Thanks!
Raina and @jeffpak93 
CF CAPI Team

ericpromislow:
We use logger.debug to report anomalous conditions that could point to an eventual bug.
When a user reports an error, we ask them to run Stackato with logging turned on,
and snapshot the log files when the bug is reproduced. 

The problem is that when a user is running, say a 12-node cluster with 3 controllers, the
logs as they are will very quickly fill up with "parse params" statements and the like, and
on a busy system the lines of interest will be scrolled out of log.3 within 5 minutes. 

Most of the statements I identified looked like they would have been of interest when the
feature was initially implemented, and can now be dropped.

ericpromislow:
@zrob's sample statement looks more useful, but I don't think it's applicable to this issue.  I simply want to remove logs that no longer of use.

zrob:
Hey @ericpromislow,

We would still prefer to see the route information moved into the log lines I mentioned above before removing the lines in the controllers.  We generally find that information very useful when debugging.

We would also like to keep the param parsing debug info as we have recently found that useful.

If you're having trouble with logs rotating on the physical machines, you could consider attaching a syslog drain where those logs can be aggregated and searched.  That's the setup we generally recommend for production environments.

Thanks,
@zrob

dieucao:
@ericpromislow Just checking in on this PR. If you could make the changes that @zrob suggests above, we can pull this in.  Let us know if you have any questions/concerns.

rmasand:
This issue has been inactive for 3 weeks. Going to go ahead and close if we don't see a response within another week.

ericpromislow:
Sorry, I've been working on a different code base the last month, and didn't have time to spend on this. @zrob's suggestion sounds fine: push as much info on a request and its status result as possible into those two lines, and ensure that the logs don't span multiple lines so they're easy to filter out, and I can drop my request here. I assume the params are available in the front controller.

utako:
Hi @ericpromislow,

Checking in on the status of this PR again. Have you been able to make @zrob's suggested changes? 

Thanks,
@utako and @karafelix, CF CAPI Team

dieucao:
@ericpromislow The next cf-release I believe will contain the agreed upon changes.
See the tracker story for example log lines.

PR #431, opened 2015-09-11
430 use U.S. spelling of 'rspec', drop --profile

ericpromislow:


zrob:
we agree that --profile is mostly white noise to have on by default.

the british spelling of colour is fully supported by rspec, however we feel we should change it mostly just to spite @julz

PR #436, opened 2015-09-23
Allow modifying ssh-enabled apps always

krishicks:
In the case that an app is pushed with ssh enabled, and later ssh is
disabled either globally or for the space that the app was pushed to,
allow restarting, scaling, and pushing the same app. Only when flipping
ssh off and on at the app level should an error occur in this scenario
(when trying to switch it back on).

Additionally, the error message now properly reflects at which level ssh
was disabled (global, space).

Some tests in apps_controller_spec were deleted as they were testing at
the same granularity as the tests in enable_ssh_policy_spec.

Signed-off-by: Andrew Edgar <aedgar@ca.ibm.com>


rmasand:
Looks good, merged the commits locally.

PR #438, opened 2015-09-23
Use Ruby 2.2.3 for CloudController

MarcSchunk:
Update Ruby to 2.2.3 in CloudController and bump dependencies where necessary (mysql2 gem is update to 0.3.20 - older versions do not support Ruby 2.2.3)

CC unit tests and CATs have been run with both dbs: postgress and mysql. 

Attention: This change needs to be included in cf-release together with the [PR against cf-release](https://github.com/cloudfoundry/cf-release/pull/793) which provides the ruby-2.2.3 package and updates jobs/packages to ruby 2.2.3 

See Tracker Story for reference [#101225348](https://www.pivotaltracker.com/story/show/101225348)


wendorf:
@MarcSchunk The PR was failing on Travis due to api_spec.rb being broken. I believe the branch just needed to be rebased on top of master, so I did that and force-pushed. If Travis goes green, this should be fine to merge in.

Gerg:
Merged. Thanks!

PR #441, opened 2015-09-25
Cache buildpack information

fraenkel:
To prevent a flurry of blobstore activity when constructing buildpack urls for
each DEA, just cache the info and rebuild when it changes.

### DIFF
--- a/app/controllers/internal/listbuildpacks_controller.rb
+++ b/app/controllers/internal/listbuildpacks_controller.rb
@@ -1,0 +1,21 @@
+module VCAP::CloudController
+  class ListBuildpacksController < RestController::BaseController
+    # Endpoint does its own (non-standard) auth
+    allow_unauthenticated_access
+
+    def initialize(*)
+      super
+      auth = Rack::Auth::Basic::Request.new(env)
+      unless auth.provided? && auth.basic? && auth.credentials == InternalApi.credentials
+        raise Errors::ApiError.new_from_details('NotAuthenticated')
+      end
+    end
+
+    get '/internal/buildpacks', :list
+    def list
+      dependency_locator = CloudController::DependencyLocator.instance
+      buildpacks_presenter = dependency_locator.buildpacks_presenter
+      [HTTP::OK, MultiJson.dump(buildpacks_presenter.to_staging_message_array)]
+    end
+  end
+end
--- a/lib/cloud_controller/admin_buildpacks_presenter.rb
+++ b/lib/cloud_controller/admin_buildpacks_presenter.rb
@@ -1,20 +1,37 @@
 module VCAP::CloudController
   class AdminBuildpacksPresenter
-    def initialize(blobstore_url_generator)
+    def initialize(blobstore_url_generator, buildpack_blobstore)
       @blobstore_url_generator = blobstore_url_generator
+      @buildpack_blobstore = buildpack_blobstore
+      @blobcache = {}
     end
 
     def to_staging_message_array
-      Buildpack.list_admin_buildpacks.
-        select(&:enabled).
-        collect { |buildpack| admin_buildpack_entry(buildpack) }.
-        select { |entry| entry[:url] }
+      buildpacks = Buildpack.list_admin_buildpacks.
+        select(&:enabled)
+
+      new_cache = {}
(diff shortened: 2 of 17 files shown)


fraenkel:
This is a potential solution for issue #441

rmasand:
Hey @fraenkel,

Your solution looks good. We reviewed it alongside cloudfoundry/dea_ng#174 and cloudfoundry/cf-release#802. We're wondering whether the caching is 100% necessary at this point, when it seems like the use of HTTP rather than NATS should make a significant improvement in performance. We'd prefer to avoid reasoning about caching until we determine it's still necessary. Would you be okay limiting the scope of this PR?

There's also a bigger question about the migration path.  If DEAs roll before CCs, then they will try to get buildpack URIs via the CC HTTP endpoint, which won't exist.  If CCs roll before DEAs, we'll be fine.  In more complicated mixed scenarios, it can also be problematic, but perhaps we can assume that those mixed scenarios don't happen.  Would it make sense to support both behaviours for a release or so, to avoid running into these issues until everything has migrated?

Thanks,
Raina and @Amit-PivotalLabs

fraenkel:
The caching is absolutely necessary. The math I show is per CC. When we roll a CC, we have 400 DEAs appear which cause the storm (400x15=6000). By memoizing we fix that little issue.
Flipping from a NATs message to a HTTP fixes when a DEA bounces, you don't have 20x15 or 300 messages to the blobstore. So it's less of an issue but still not good when you are rolling DEAs.

I thought about the migration and decided it isn't something we should be overly concerned. Yes, on any given roll, a DEA might not get the "update" on start. It is a best effort today anyhow and will be rectified on the first stage. Regardless of whether we have old or new, you always have the case that a DEA can come up without any CC and won't retrieve the buildpacks.
It won't matter how many releases you wait someone will be skipping and will hit this so you might as well prove that it shouldn't matter rather than having your environments "work" properly and everyone else experience the skip.

Amit-PivotalLabs:
Maybe I'm missing something.  If CCs provide buildpack info only when a DEA
starts up and makes an HTTP request, when would the scenario you're
describing happen?  New CC rolls, wakes up, what will trigger it to make
400x15 blobstore requests if it's no longer reacting to dea.advertise
messages with a blobstore fetch?

fraenkel:
A CC provides the buildpack info to any DEA it has never seen before. If the CC rolls, all DEAs look new or if a DEA rolls, all CCs see the DEA as new.

rmasand:
That seems to be the case before your changes, but it looks like DEA should now be initiating all requests for buildpack info over HTTP. CCs shouldn't be doing anything special when they roll, unless they immediately receive a direct request from a DEA. Shouldn't this change https://github.com/cloudfoundry/cloud_controller_ng/pull/441/files#diff-05cd812b72cb744abd42252bcde6bf56L19 eliminate the behavior you're talking about? 

I can see that caching information from the blobstore would be helpful, but not convinced that it would significantly improve things after CCs have been changed to wait patiently for HTTP requests. 400 DEAs' requests spread over 30 CCs doesn't seem like a huge issue (~12 requests per CC, spread out over the time it takes to roll all the DEAs) even if they have to repeatedly talk to the blobstore. 

If our understanding of the world is wrong, could you link us to the lines you're concerned about?

fraenkel:
Yes by switching to just the HTTP, we end up with 400x15 across the CCs given we have 15 buildpacks but that will probably grow as you start versioning things. Anyhow, there are 2 issues at play. 1) Consuming all the available threads across the CCs. If I were to roll X DEAs at any time, each DEA would consume all the available threads. The router doesn't know a busy CC from a non-busy CC but what does happen is monit may freak out if it doesn't get its health heartbeat through. At the same time, you could have stagings which will also make the same requests although the is part of the "START" transition so we cannot control how that works other than it too will steal 15 threads. And then you have the combination of all requests to your blobstore across all CCs.

I don't want to have to think about whether we are hammering a blobstore or not for these buildpacks on every STAGING or DEA roll. It all comes back to how scalable and responsive your blobstore is. They all have limits.

fraenkel:
#446 breaks out the endpoint and removes the buildpack message.

rmasand:
Looks like that has been merged, thanks for splitting it out. Going to close this one for now.

PR #442, opened 2015-09-28
Introducing idempotency to service instance deletion.

jrozansk:
Deleting already deleted service instance does not return error.

Fixes following scenario:
While very vast space is deleted, one can delete service instance manually/independently in meantime.
This will end with error when cloud controller tries to remove instance again.
As service instance is missing why to return errors from delete_service_instance method?

==== Addditional notes ====
This situation occurred on CF instance I use on daily basis and is fully reproducible.
I got following error:
Deletion of space X failed because one or more resources within could not be deleted.nntRecord not found
Response from cloud controller HTTP Bad Gateway (502)

If this problem will be solved in other way in the future, please let me know.
There can be more elegant way to fix this problem, also. I am open to suggestions.

### DIFF
--- a/app/actions/services/service_instance_delete.rb
+++ b/app/actions/services/service_instance_delete.rb
@@ -1,5 +1,6 @@
 require 'actions/services/service_binding_delete'
 require 'actions/services/locks/deleter_lock'
+require 'sequel'
 
 module VCAP::CloudController
   class ServiceInstanceDelete
@@ -33,6 +34,10 @@ module VCAP::CloudController
 
     def delete_service_instance(service_instance)
       errors = []
+
+      if !service_instance.exists? # for idempotency purposes
+        return []
+      end
 
       begin
         lock = DeleterLock.new(service_instance)
--- a/spec/unit/actions/services/service_instance_delete_spec.rb
+++ b/spec/unit/actions/services/service_instance_delete_spec.rb
@@ -318,6 +318,17 @@ module VCAP::CloudController
           expect(errors[0].message).to eq 'BOOM'
         end
       end
+
+      context 'when deleting already deleted service instance' do
+        it 'does not throw errors as element is missing anyway' do
+          expect(ServiceInstance.count).to eq 2
+          service_instance_delete.delete([service_instance_1])
+          errors = service_instance_delete.delete([service_instance_1])
+          expect(ServiceInstance.count).to eq 1
+
+          expect(errors.count).to eq(0)
+        end
+      end
     end
   end
 end


dieucao:
@jrozansk Have you signed the CLA? Or you part of a company that's covered by the CLA as mentioned above by dreddbot?

jrozansk:
@dieucao How to check whether my employer (Intel Corp) signed Corporate CLA? Do you hold list of such organizations?

Gerg:
@jrozansk You need to make your membership in your parent company's CF github org public, so we can verify that you are a member. See instructions [here](https://help.github.com/articles/publicizing-or-hiding-organization-membership/).

Gerg:
This seems reasonable.

Instead of catching the Sequel error, you should be able to check `service_instance.exists?` before attempting the delete.

Also, there were some Rubocop code style violations with your pull request (see Travis build).

dieucao:
@jrozansk It appears that Intel has an org https://github.com/IntelCFContrib that @amhuber and @jmprice are members of.
Our CLA checking tool appears to have Aaron and John as individual contributors and not cleared by membership of that Org.  I would suggest contacting Aaron and John about how you want to manage contributions, whether that github org should be added to the dreddbot tool so that Intel can independently manage membership, or if they would prefer for you to sign the individual CLA, and emailing contributors@cloudfoundry.org with the preference.

amhuber:
Hmm, we definitely wanted to switch to the group but I don’t recall where that process left off.  How do I request updating our CLA to associate the group instead of Jon and I individually?  Once that is done I can add Jakub.

Aaron

jrozansk:
Update: I applied changes suggested by @Gerg. Thanks for pointing that out.
I fixed problems listed by rubycop, also. Now travis is green.

Well, I made my org association public in profile (intel-data and trustedanalytics).
Anyway I filled and sent Individual CLA to make things happen sooner.

dieucao:
@amhuber I think you can email contributors@cloudfoundry.org to get the CLA checking stuff sorted out.
@chipchilders is that right?

chipchilders:
We just got the ICLA this morning... @dieucao you can proceed with the merge whenever it's technically appropriate for the project team.

@jrozansk - Thanks for the ICLA. We do not have a CCLA on file for any Intel GitHub organizations that I can see. Feel free to have Intel fill out the CCLA and send it in with a specific GitHub organization that you would like to use to identify your approved contributors!

amhuber:
OK, I'll work on getting the CCLA signed and associated with the IntelCFContrib group.  Thanks!

PR #452, opened 2015-10-14
blobstore dir list: Don't include explicit attribs

mtekel:
When you include explicit 'limit' attribute, you make OpenStack SWIFT
work, but you break compatibility with other storage providers, like
Google Cloud Storage. Revert this again. Fog should support remapping
max_keys into limit specifically for SWIFT. Fog issue risen here:
https://github.com/fog/fog/issues/3723

This should be merged when the Fog issue is fixed. Until then a workaround would be to add a vendor specific code (if provider is openstack then include limit).

### DIFF
--- a/lib/cloud_controller/blobstore/directory.rb
+++ b/lib/cloud_controller/blobstore/directory.rb
@@ -11,7 +11,7 @@ module CloudController
       end
 
       def get
-        @connection.directories.get(@key, 'limit' => 1, max_keys: 1)
+        @connection.directories.get(@key, max_keys: 1)
       end
     end
   end
--- a/spec/unit/lib/cloud_controller/blobstore/directory_spec.rb
+++ b/spec/unit/lib/cloud_controller/blobstore/directory_spec.rb
@@ -30,7 +30,7 @@ module CloudController
 
       describe '#get' do
         it 'retrieves the directory' do
-          expect(directories).to receive(:get).with(directory_key, 'limit' => 1, max_keys: 1).and_return(fog_directory)
+          expect(directories).to receive(:get).with(directory_key, max_keys: 1).and_return(fog_directory)
           expect(directory.get).to eq(fog_directory)
         end
       end


simonleung8:
Sorry, our tracker bot did not pick this up for some reason. We have created a story manually: https://www.pivotaltracker.com/story/show/113314479

-@simonleung8 and @Gerg

dieucao:
@mtekel Can you confirm if this is still an issue for you?

mtekel:
Because of this we could not use GCS storage for cloud controller. This issue is the only thing preventing it. The implementation is not nice, as it is adding that limit attribute to all provider's calls, so potentially it's also breaking other providers. This looks like a workaround just for OpenStack, so I think there should be some logic that applies it only in that case. From the git history I see this has been added and reverted many times. I guess it's time for a proper fix...

jeffpak:
Hi @mtekel,

We agree that this is slightly odd, but we do not want to fix it in cloud controller until we can use a fog provided key that works across all storage providers.  We'll continue to track the referenced fog issue and  make updates to cloud controller after fog gives us a way forward.

Thanks,
@jeffpak and @zrob

SocalNick:
@jeffpak @mtekel 

The [fog issue](https://github.com/fog/fog/issues/3723) has been closed in favor of an [issue in fog-openstack](https://github.com/fog/fog-openstack/issues/51). We don't use fog-openstack, I'm not sure how fog works when it natively supports OpenStack but there is also an OpenStack plugin.

mtekel:
Looks like they are just moving openstack from the core into separate repo (as openstack is no longer present in https://github.com/fog/fog/tree/master/lib/fog). A move bit similar how cf-release is being split into components...

This is openstack (swift) specific issue and once it's fixed there, you no longer need to be applying the workaround you have currently, which breaks other providers (Google storage).

SocalNick:
@mtekel we got a new PR that might fix this, would you mind taking a look and potentially close this? https://github.com/cloudfoundry/cloud_controller_ng/pull/645

mtekel:
Yes it looks like that PR might fix it. Unfortunately I currently don't have access to GCS storage to test. But because that PR is solving the same problem, closing this issue.

Issue #449, opened 2015-10-13
Wrong return type for delete operation


cgfrost:
See: http://apidocs.cloudfoundry.org/214/users/remove_audited_organization_from_the_user.html

A delete operation should not return `201 created`.

dieucao:
@cgfrost 
Thanks for the report.
It's apparently extremely hard to fix this on the v2 end points because of some of the meta-programming magic in cloud controller.
We'll address this when we move the users resource over to /v3.
I'll mark this with the wontfix label though and leave it open

fraenkel:
@dieucao It's actually quite easy to fix. I thought the original reason was to prevent breakage even though I believe most folks would rather it be fixed.

dieucao:
@fraenkel I'll follow up with @zrob . I could have sworn it was because it was hard to change.  I get how it could be considered a backwards breaking change, but it seems silly.

cgfrost:
Possibly related to https://github.com/cloudfoundry/cloud_controller_ng/issues/450

There may be others, we will continue to raise issues as we find them. We're working through the API endpoint one by one to implement a new CF Java Client.

dieucao:
This was addressed as part of the above linked PR from @fraenkel .
http://apidocs.cloudfoundry.org/225/users/remove_audited_organization_from_the_user.html

Closing this issue.

PR #453, opened 2015-10-15
Return 204 on successful delete

fraenkel:
Fixes issue #449

### DIFF
--- a/app/controllers/base/model_controller.rb
+++ b/app/controllers/base/model_controller.rb
@@ -204,6 +204,7 @@ module VCAP::CloudController::RestController
 
       after_update(obj)
 
+      return [HTTP::NO_CONTENT] if verb == 'remove'
       [HTTP::CREATED, object_renderer.render_json(self.class, obj, @opts)]
     end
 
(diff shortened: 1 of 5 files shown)


dieucao:
@fraenkel I'll post a question to cf-dev about how people feel about making this type of change.
Does this PR affect the async services end points that currently return 202 on deletes? Or is that stuff special cased.

fraenkel:
@dieucao The 201 fiasco only affects related routes. A normal delete always returns a 202 or 204 depending on if it queues a job.

dieucao:
@fraenkel Everyone I've talked to so far that has implemented a client thinks this would be okay to change. 
I'll have the team review.
Thanks!

jberkhahn:
@fraenkel There was a change made in spec/support/legacy_api_dsl.rb that should affect the docs. However when we generate the docs it still shows the expected response status as 201. Fix that up and this looks good.

fraenkel:
I don't know what you are referring to. I rebased the commit and re-pushed. Everything is showing a 204 without any code change.
Please identify the exact command that still shows a 201.

rmasand:
Not seeing the docs issue anywhere, maybe they needed to be regenerated? Also the travis failure from before looked to be a flake. Going to go ahead and merge. Thanks @fraenkel !

dieucao:
This will be available in the next cf-release.

Issue #440, opened 2015-09-24
100GB of orphan packages & droplets in blobstore


rkoster:
In one of our client's environments we noticed an unexpected increase in blobstore size. Further investigation revealed their where a lot of orphan packages laying around (I used [this script](https://gist.github.com/rkoster/9a95ee65eb79988dfaae) to find the orphan packages). I also noticed a lot of failed jobs in the worker queue.

I have not yet found the root cause of the issue but was able to reproduce the behaviour on bosh-lite by running the following bash snippet:
```bash
for i in {1..10}; do
    echo "PUSH: $i"
    dd if=/dev/random of=big_file count=102400 bs=1024
    cf push foo &
    sleep 10
    cf delete-space -f bar && cf create-space bar && cf target -s bar
    kill $!
done
```

The sleep statement in the above script is something which needs to be tweaked according to your setup. 

I was able to see packages build up (while there where no apps deployed) by running:
```bash
bosh ssh api_z1/0
sudo su -
watch ls -lah /var/vcap/store/10.244.0.34.xip.io-cc-packages/??/??/
```

I'm aware that the above is not a typical use-case, but it demonstrates packages can become orphaned. Maybe it would make sense to have a scheduled orphan package/droplets cleaner job.

Gerg:
Using your ruby script, we didn't see any orphans on our test environments. We also tried to reproduce this on bosh-lite using the provided bash snippet. Even after playing around with the length of the sleep, we did not create any orphans. 

Have you made any progress on finding the underlying cause? Any additional information or reproduction steps you can provide will help us characterize this issue.

rkoster:
On further investigation I found multiple failed delayed_jobs for `VCAP::CloudController::Jobs::Runtime::AppBitsPacker`. The passed `app_guid`s matched the files a found in `/var/vcap/store/10.244.0.34.xip.io-cc-packages/`.

Below the full stacks trace from `last_error`, in short the application was `nil`. Which happend because it was deleted before the task completed. After a while the tasks get cleaned up but the blobs stay in the blobstore.

Full stack trace:
```
undefined method `allow_ssh' for nil:NilClass
/var/vcap/data/packages/cloud_controller_ng/9b84bcc79ce079128bb3a06c360d465eb83ad0d1.1-d489916692c5925a0ea8111f9fe0d2dcb066b7f7/cloud_controller_ng/app/models/runtime/constraints/enable_ssh_policy.rb:10:in `validate'
[...]

PR #461, opened 2015-11-18
Orphaned packages and droplets cleanup jobs

rkoster:
This PR addresses the orphaned packages issue raised [here](https://github.com/cloudfoundry/cloud_controller_ng/issues/440).

The cleanup of orphaned droplets will be submitted as a separated PR.


rkoster:
Since submitting orphaned droplets cleanup as a separate PR would result in merge conflicts I have added it to this PR.

dieucao:
Thanks @rkoster for the PR.  We'll try to get a quick sanity check code review some time next week and then I'll prioritize it for deeper review after that.

utako:
Hi @rkoster,

As of [this story](https://www.pivotaltracker.com/story/show/109604284), we'll be making some changes to our blobstore client so there will be a few changes to make with regards to how this delayed job interacts with files.

Our main concern is how we load the blobstore files into memory and interact with them for each type of file system (NFS, s3, WebDAV, etc). For instance, the S3 fog client is smart enough to not load all blobstore files into memory, but this won't be the case for WebDAV. In WebDAV's case, we'll have to figure out a way to load the files in an enumerable object or something equally conservative with memory use. In addition to this concern, we are currently working on a client wrapper interface in which we'll be removing the "#files" method as part of the aforementioned story, so we'll have to incorporate those changes into the implementation of droplet/package orphan mitigation. 

Feel free to comment here or ping us if you have any ideas about how to elegantly solve the challenges mentioned above. We'll be happy to answer any questions you might have about the changes we're making to the blobstore client, especially on completion of story #109604284.

Thanks for your patience,
@utako, CF CAPI Team

rkoster:
I would say the best place for solving the above would be the client wrapper. It could  a #files method which would returns an enumerable object?

simonleung8:
Hi @rkoster,

We're going to keep this PR open and revisit it when we finish the [story](https://www.pivotaltracker.com/story/show/109604284) to rewrite the blobstore clients.

Thanks,
@simonleung8 & @youreddy 
CF CAPI Team

dieucao:
As discussed with @rkoster I'm going to close this PR because the scope of work remaining is large and he's currently focused on other things.

@smoser-ibm FYI, this could be a useful thing to consider for the bits service/[Flintstone team](https://www.pivotaltracker.com/n/projects/1406862) at some point down the road.  It's a valid pain point that at times there are blobs that are orphaned and no good way to clean them up from the blobstore.

Issue #468, opened 2015-12-03
create-service-key fails when credentials section is long


karol-brejna-i:
It turns out (for us) that credentials field of service_key table is to small.

According to initial schema definition (
https://github.com/cloudfoundry/cloud_controller_ng/blob/master/db/migrations/20150316184259_create_service_key_table.rb) it is:

```String :credentials, null: false, size: 2048```

Doing a quick and dirty patch on the db - `alter table service_keys alter credentials type text` - solved the issue.

Please, consider changing the field type/size.

karol-brejna-i:
Probably already solved by https://github.com/cloudfoundry/cloud_controller_ng/blob/965dbc4bdf65df89f382329aef39f86a916b3f05/db/migrations/20150713133551_enlarge_service_keys_credentials.rb

dieucao:
Closing this issue then.

PR #470, opened 2015-12-03
Change type of credential field to text in service_key table.

karol-brejna-i:
Solves https://github.com/cloudfoundry/cloud_controller_ng/issues/468


karol-brejna-i:
Probably already solved by https://github.com/cloudfoundry/cloud_controller_ng/blob/965dbc4bdf65df89f382329aef39f86a916b3f05/db/migrations/20150713133551_enlarge_service_keys_credentials.rb

PR #504, opened 2016-01-18
Upgrade rubocop to 0.29

Gerg:
- Provides additional cops and options for existing cops
- Fix easy formatting errors
- Disable (new) struct inheritance cop for now

PR #506, opened 2016-01-19
Upgrade rubocop to 0.30.1

Gerg:
**Don't look at this until #504 has been addressed**

- Disable ClosingParenthesisIndentation because it conflicts with our project style in situations like:
    ```rb
    method(:argument1,
      hash1: :arg1,
      hash2: :arg2,
      hash3: :arg3,
    )
    ```
It would like to do this:
    ```rb
    method(:argument1,
      hash1: :arg1,
      hash2: :arg2,
      hash3: :arg3,
          )
    ```
This is because it is expecting a new line between the method name and the first argument.

- Performance/Sample is broken in this version. It tries to apply in situations where you are iterating over the collection.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,9 +5,6 @@ AllCops:
   Exclude:
     - vendor/**/*
 
-RegexpLiteral:
-  MaxSlashes: 1
-
 UnusedMethodArgument:
   Enabled: false
 
@@ -25,7 +22,7 @@ LineLength:
 
 MethodLength:
   Max: 60
- 
+
 Metrics/AbcSize:
   Max: 30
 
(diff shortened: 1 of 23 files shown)


wendorf:
Hi @Gerg ,

We just merged in #504 , so can you please rebase this on top of master?

Additionally, https://github.com/cloudfoundry/cloud_controller_ng/pull/506/files#diff-3e986c24321412db3ad103e17eeaf6eeL215 looks like it should be a separate PR since it is functionally different. Can you split that out?

Thanks,
@wendorf, Cloud Foundry CAPI team

Gerg:
@wendorf I have rebased and addressed your concern

PR #510, opened 2016-01-21
Upgrade rubocop to 0.31 

Gerg:
**Don't look at this until #506 has been addressed**

- Disable rubocop for migrations. Modifying old migrations could be
  dangerous and style isn't as important for code that should
  (hopefully) never need to change.
- Disable module length check. It would be great to have reasonable
  modules, but this is a much larger undertaking than can be
  accomblished in this PR.
- Disable trailing underscore variable check (a, b, _ = foo()). The
  suggested style is less clear (a, b, = foo())

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,9 +5,6 @@ AllCops:
   Exclude:
     - vendor/**/*
 
-RegexpLiteral:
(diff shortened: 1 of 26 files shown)


simonleung8:
Looks good. Merged

PR #522, opened 2016-01-25
Set TMPDIR for worker

rkoster:
While investigating an app upload issue we found the local worker is using `/tmp`.
In our case the `/tmp` dir was world writable which caused us to run into [this](https://blog.diacode.com/fixing-temporary-dir-problems-with-ruby-2). 

After consulting with @cppforlife we found that using `/tmp` is considered a bad practise for bosh-releases.


simonleung8:
Looks good. merged

PR #523, opened 2016-01-26
Don't rely on permission on  /tmp

syslxg:
Depending on the permissions for `/tmp`, calling the Dir.tmpdir method can [sometimes raise an error](https://blog.diacode.com/fixing-temporary-dir-problems-with-ruby-2). And root_path does not have to be exist at all.


dieucao:
I believe this was addressed with https://www.pivotaltracker.com/story/show/112387205 and this https://github.com/cloudfoundry/cloud_controller_ng/pull/522 which was merged

PR #548, opened 2016-02-28
Upgrade to Rubocop 0.37.2

wendorf:
Rubocop 0.36.0 was skipped because it has some bugs that prevent using it for our project.

- Style/IndentArray changed its default behavior, so set EnforcedStyle to `consistent` to maintain our current behavior
- Style/Alias changed its default behavior, so set EnforcedStyle to `prefer_alias_method` to maintain our current behavior
- Disabled the new Performance/Casecmp and Style/ZeroLengthPredicate cops because they make the code harder to read.


Gerg:
:tada:

PR #575, opened 2016-04-12
Limit app starts to 5 at a time

DanLavine:
- Reduce the keep alive to 1s
- Only perform 5 starts at a time

[#115771421]

Signed-off-by: Michael Fraenkel <fraenkel@us.ibm.com>

### DIFF
--- a/lib/cloud_controller/dea/client.rb
+++ b/lib/cloud_controller/dea/client.rb
@@ -22,7 +22,7 @@ module VCAP::CloudController
             client.connect_timeout = 5
             client.receive_timeout = 5
             client.send_timeout = 5
-            client.keep_alive_timeout = 5
+            client.keep_alive_timeout = 1
 
             ssl = client.ssl_config
             ssl.verify_mode = OpenSSL::SSL::VERIFY_PEER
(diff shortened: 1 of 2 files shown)


youreddy:
Hello @DanLavine, 

Thank you for the pull request. We created a branch to test the PR and it passed all our unit and acceptance tests. We wanted to be able to manually test the start instances being limited to 5 on our bosh-lite for good measure. Can you please advise the best way to do so?

We would also appreciate it if you could adjust the added test to be more readable. 

Best,
@youreddy & @karafelix 
CF CAPI Team

ronakbanka:
@DanLavine Just curious what is the use case for this restriction on number of app instances which can be started at a time by CC ??

ScarletTanager:
@ronakbanka - to be clear, this is not a limit on the number of apps the user can start at a time - it is a limit on the number of start requests (CC->DEA) which can be sent at a time.  This should be more or less transparent to end users.

ronakbanka:
@ScarletTanager yeah that's exactly what my question was why restrict this on CC->DEA??

DanLavine:
@youreddy You would be able to test this by starting a large number of instances for a single app and looking for DEA log line 'http.request.received'. These should be in batches of 5. You should be able to correlate these with log lines in CC (probably easiest to just add something around sending each batch of requests)

@ronakbanka We would like to batch the request to the DEA as it provides less network traffic bursts.

utako:
I've merged this [here](https://github.com/cloudfoundry/cloud_controller_ng/commit/08bda793cbfa6054b9fc2fca887cbbf1302f69a9). 

Closing this, feel free to reopen and comment.

@utako, CF CAPI Team

PR #579, opened 2016-04-19
Should not fail if target to be deleted is already gone

MarcSchunk:
BuildpacksCacheController#delete removes all entries of the buildpack_cache. With
a DavClient this will create an issue as soon as an app is deleted, because the
delete_all_in_path method (called when the buildpack_cache for the app is removed)
will raise after not finding the app guid in the first place.

This is not an issue with FogClient, which just silently returns. We believe this
is the correct behavior because the target state (app_guid and children are gone)
has already been reached.

BuildpacksCacheController#delete removes all entries of the buildpack_cache. With
fog_client.delete_all_in_path and dav_client.delete_all_in_path behave
differently:

* fog_client will not attempt to delete anything if the given path does not exist.
It will not fail or raise an exeption.
* dav_client will raise an exception when to path to be deleted does not exist.

Note: This is only an issue when using the CC v3 API.

This commit brings the behavior of DavClient in line with FogClient when attempting
to delete a resource that already is gone.

thx,
@suhlig and @MarcSchunk

### DIFF
--- a/lib/cloud_controller/blobstore/webdav/dav_client.rb
+++ b/lib/cloud_controller/blobstore/webdav/dav_client.rb
@@ -158,7 +158,9 @@ module CloudController
         response = with_error_handling { @client.delete(url, header: @headers) }
         return if response.status == 204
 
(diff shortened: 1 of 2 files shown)


wendorf:
Looks great! Thanks!

PR #583, opened 2016-04-27
CC stages over HTTPS

ScarletTanager:
This is a change to allow the CC to send staging requests to DEAs over https instead of NATS.  The behavior is configurable and can be determined on a per-DEA basis (meaning we support a mixed deployment of some DEAs supporting NATS, some support HTTPS).

* [ YES] I have viewed signed and have submitted the Contributor License Agreement

* [YES ] I have made this pull request to the `master` branch

* [ YES] I have run all the unit tests using `bundle exec rake`

* [ YES] I have run CF Acceptance Tests on bosh lite

[#112002909]

Thanks,

@ScarletTanager & @jberkhahn CF RuntimeOG team

### DIFF
--- a/lib/cloud_controller/dea/app_stager_task.rb
+++ b/lib/cloud_controller/dea/app_stager_task.rb
@@ -21,11 +21,9 @@ module VCAP::CloudController
       end
 
       def stage(&completion_callback)
-        @stager_id = @dea_pool.find_stager(@app.stack.name, staging_task_memory_mb, staging_task_disk_mb)
-        raise CloudController::Errors::ApiError.new_from_details('StagingError', 'no available stagers') unless @stager_id
-
-        subject = "staging.#{@stager_id}.start"
-        @multi_message_bus_request = MultiResponseMessageBusRequest.new(@message_bus, subject)
+        stager = @dea_pool.find_stager(@app.stack.name, staging_task_memory_mb, staging_task_disk_mb)
+        raise CloudController::Errors::ApiError.new_from_details('StagingError', 'no available stagers') unless stager
+        @stager_id = stager.dea_id
 
         # Save the current staging task
         @app.update(package_state: 'PENDING', staging_task_id: task_id)
@@ -39,6 +37,13 @@ module VCAP::CloudController
 
         logger.info('staging.begin', app_guid: @app.guid)
         staging_msg = staging_request
+
+        if stager.url && Client.enabled?
+          return stage_with_http(stager.url, staging_msg)
+        end
+
+        subject = "staging.#{@stager_id}.start"
+        @multi_message_bus_request = MultiResponseMessageBusRequest.new(@message_bus, subject)
 
         staging_result = EM.schedule_sync do |promise|
           # First response is blocking stage_app.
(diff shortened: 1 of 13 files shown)


lib/cloud_controller/dea/app_stager_task.rb
```diff
@@ -40,6 +38,13 @@ def stage(&completion_callback)
         logger.info('staging.begin', app_guid: @app.guid)
         staging_msg = staging_request
 
+        if stager.url && Client.enabled?
+          return stage_with_http(stager.url, staging_msg)
+        end
+
+        subject = "staging.#{@stager_id}.start"
+        @multi_message_bus_request = MultiResponseMessageBusRequest.new(@message_bus, subject)
```
fraenkel: Why isn't this a separate method for stage_with_nats?

lib/cloud_controller/dea/app_stager_task.rb
```diff
@@ -80,8 +103,8 @@ def handle_first_response(response, error, promise)
 
       def handle_second_response(response, error)
         @multi_message_bus_request.ignore_subsequent_responses
-        ensure_staging_is_current!
-        check_staging_error!(response, error)
+        ensure_staging_is_current!(false)
```
fraenkel: this looks very similar to handle_http_response....

lib/cloud_controller/dea/app_stager_task.rb
```diff
@@ -142,15 +165,19 @@ def ensure_staging_is_current!
           raise CloudController::Errors::ApiError.new_from_details('StagingError', "failed to stage application: can't retrieve staging status")
         end
 
-        if @app.staging_task_id != task_id
-          raise CloudController::Errors::ApiError.new_from_details('StagingError', 'failed to stage application: another staging request was initiated')
-        end
+        check_task_id unless use_http
```
fraenkel: why isn't this done for both?

lib/cloud_controller/dea/client.rb
```diff
@@ -37,6 +37,21 @@ def configure(config, message_bus, dea_pool, blobstore_url_generator)
           end
         end
 
+        def enabled?
```
fraenkel: is what enabled?

lib/cloud_controller/dea/staging_message.rb
```diff
@@ -20,6 +20,7 @@ def staging_request(app, task_id)
           start_message:                start_app_message(app),
           admin_buildpacks:             admin_buildpacks,
           egress_network_rules:         staging_egress_rules,
+          accepts_http:                 @config[:dea_client] ? true : false
```
fraenkel: this is wrong.

dea_client: 'foobar' is true?

lib/cloud_controller/dea/app_stager_task.rb
ScarletTanager: Addressed.

lib/cloud_controller/dea/app_stager_task.rb
ScarletTanager: Because if we're using http, we check the task id when we receive the message (in the controller), and if it doesn't match, we fail early (discard the message).  We know it makes the paths slightly different for NATS and http, but we didn't think it made sense to get this far on the http path before discarding a no-longer-current staging response.

lib/cloud_controller/dea/app_stager_task.rb
ScarletTanager: It is similar but slightly different - there is no multi_message_bus_request in the http flow, and we pass in a boolean to ensure_staging_is_current to avoid checking the task_id (in the http flow we have to create a new app_stager_task, whereas the one for NATS is the same throughout the request lifecycle).  An alternative would have been to make check_task_id know whether we're using http or not and pick whether to use the current task's task_id or that from the message.  Would you prefer that over discarding a no-longer-current response in the controller upon receipt?

fraenkel: checking the task_id is a nit and costs us nothing other than more code to make it not happen.
Again, not sure why I care if it is nats or http. we should always verify the task_id. Don't assume the path above you.

One can start an application, have staging start, stop the app, start the app, and the previous stage may not complete so you still need to verify that you are the current task_id even with NATs.

lib/cloud_controller/dea/app_stager_task.rb
fraenkel: See above. I think you have made this more complicated when its the same regardless of the path.

lib/cloud_controller/dea/staging_message.rb
ScarletTanager: Fixing.

lib/cloud_controller/dea/app_stager_task.rb
ScarletTanager: Discussed in Slack, @fraenkel and I are on the same page regarding checking of task_id now.  Only change required is to move @app.refresh earlier in the NATS path.

lib/cloud_controller/dea/app_stager_task.rb
ScarletTanager: See above, discussed in Slack with Michael.

simonleung8:
Thanks,

This has been merged. I put a commit on top to satisfy rubocop and github couldn't recognize the merge because of the commit.

PR #590, opened 2016-05-03
Make staging response controller respond correctly

ScarletTanager:
Currently the CC returns a 400 (incorrectly) when the DEA sends a staging response including an error over http.  This fixes that (should be a 200) and fixes any other errors not resulting from a bad request to be 500s.

* [Yes ] I have viewed signed and have submitted the Contributor License Agreement

* [ Yes] I have made this pull request to the `master` branch

* [ Yes] I have run all the unit tests using `bundle exec rake`

* [ Yes] I have run CF Acceptance Tests on bosh lite

- return 200 when DEA response contains error
- return 500 (not 400) for all CC errors

[#112002909]

Signed-off-by: Dan Lavine <dlavine@us.ibm.com>


simonleung8:
Thanks, LGTM. Merged

PR #591, opened 2016-05-05
Fallback to NATs if staging over http isn't there

fraenkel:

When staging over http isn't supported on the DEA, fallback to NATs

Rolling upgrade when CC supports staging and the DEAs don't yet.

[#118970125]


rizwanreza:
Looks like this pull request was merged already, but not closed.

PR #592, opened 2016-05-05
Add last operation type to broker API request

craigfurman:
We have added a query parameter, `operation` to the service broker API `last_operation` request. The parameter can take one of 3 values: `create`, `delete`, or `update`.

This allows service broker developers to avoid duplicating service instance lifecycle state between Cloud Controller and the broker.

We were unsure of how to structure the broker API acceptance tests: the test suites for the 2.7 and 2.8 API compatibility needed to change even though our API change is non-breaking to reflect new stubbing for webmock. If we had added a 2.9/2.10 test file, it would contain an exact copy-paste of that block in 2.8/2.7. We are open to suggestions on how to proceed here.

We have not submitted a PR to [the broker API docs](https://github.com/cloudfoundry/docs-services) as the change is so small that it seemed more pragmatic to wait until this is merged, and then we/you can make that PR?

* [ ] I have not run CF Acceptance Tests on bosh lite. We weren't able to bump `cloud_controller_ng` inside `capi-release` inside `cf-release` without significant work.

Cheers!

cc @avade @utako

### DIFF
--- a/lib/services/service_brokers/v2/client.rb
+++ b/lib/services/service_brokers/v2/client.rb
@@ -237,7 +237,7 @@ module VCAP::Services::ServiceBrokers::V2
     end
 
     def service_instance_last_operation_path(instance)
-      "#{service_instance_resource_path(instance)}/last_operation"
+      "#{service_instance_resource_path(instance)}/last_operation?operation=#{instance.last_operation.type}"
     end
 
     def service_binding_resource_path(binding_guid, service_instance_guid)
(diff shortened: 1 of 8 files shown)


shalako:
@SocalNick see convo occurring in https://www.pivotaltracker.com/n/projects/1443654/stories/117839149

avade:
As per our discussion, we will not be proceeding with this PR. Drafting a proposal for an alternative solution.

PR #604, opened 2016-05-19
Handle socket error

DanLavine:

If Consul or HM9K cannot be connected to we will capture the socket error. This allows users to then still type cf app 'name' and have it report back with a ? rather than an internal server error.


adowns01:
Hi @DanLavine, 

Looks good! I merged this in, but the shas changed; authors are still the same though. 

Thanks, 
@adowns01

PR #610, opened 2016-05-31
Emit error when consul is down.

jberkhahn:
Small change to actually emit the error when consul is down. Should make it easier to diagnose what's going wrong if/when this happens.

Thanks,
@jberkhahn & @swetharepakula 
Runtime_OG


sax:
@jberkhahn thank you!

PR #614, opened 2016-06-08
Config handles minimum_candidate_stagers set to < 1

ScarletTanager:
If minimum_candidate_stagers is set to a value less than 1 in a manifest, reset it to the default value.

[#119362989]

Signed-off-by: Swetha Repakula <srepaku@us.ibm.com>

### DIFF
--- a/lib/cloud_controller/config.rb
+++ b/lib/cloud_controller/config.rb
@@ -378,7 +378,7 @@ module VCAP::CloudController
         config[:broker_client_default_async_poll_interval_seconds] ||= 60
         config[:packages][:max_valid_packages_stored] ||= 5
         config[:droplets][:max_staged_droplets_stored] ||= 5
-        config[:minimum_candidate_stagers] ||= 5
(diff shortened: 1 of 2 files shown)


adowns01:
Hello @ScarletTanager, 

Just curious, is there a reason that you need an extra check here for the `minimum_candidate_stagers` and nothing else? 

Before we merge this, can you change the description of this property to include a description of this new behavior? 

Thanks, 
@adowns01 && @rizwanreza

ScarletTanager:
@adowns01 - I'll definitely see about updating the property description.  We only need the extra check for minimum_candidate_stagers because that is the only property that really matters for the math done by the CC in determining which DEAs may be used to stage an application.  The current minimum size of the candidate pool is hard-coded to 5 - so if you have 5 or fewer DEAs in your deployment, it is really just selecting one at random, to some extent.  This addresses a bug identified by a user with a small deployment who noticed that it was still possible for his most loaded DEAs to be targeted for staging.

thausler786:
Thanks for the commit! It has been merged now.

Issue #617, opened 2016-06-13
Removal of db.log_level / cc.db_logging_level


dcarley:
076e68fa9325e7af176e305e238c9008d46a023f and c155c3c3a439d706a573451b392253ab4635f228 hardcoded the `Sequel` logging level that was previously configured by `cc.db_logging_level`.

Should the following be removed now that they're not used?

- `log_level` docstring for `VCAP::CloudController::DB#connect`
  - https://github.com/cloudfoundry/cloud_controller_ng/blob/e7452b01ed6dadde8fd0550c364b0ef79116fa9f/lib/cloud_controller/db.rb#L11
- `db.log_level` config options in `bosh-templates/`
  - https://github.com/cloudfoundry/cloud_controller_ng/search?utf8=%E2%9C%93&q=db_logging_level
- `cc.db_logging_level` property in cloudfoundry/capi-release specs
  - https://github.com/cloudfoundry/capi-release/search?utf8=%E2%9C%93&q=db_logging_level
- `cc.db_logging_level` property in cloudfoundry/cf-release examples
  - https://github.com/cloudfoundry/cf-release/search?utf8=%E2%9C%93&q=db_logging_level

I'd offer a PR, but I'm not sure how to co-ordinate it across that many release repos.

thausler786:
Hi @dcarley,

Thanks for bringing these chores to our attention.  We've decided to revert 076e68fa9325e7af176e305e238c9008d46a023f and c155c3c3a439d706a573451b392253ab4635f228 since we decided that they were unneccessary.  For anyone interested, the story of these commits can be found in issue #572.

Best,

@thausler786, CAPI team member

dcarley:
Reverts were e2dde7131a89f18a81b97d7455fc2c4a0814a099 and 23bc280af26b5f995ee8d4d54519ce71e67191e9. Thanks!

Issue #572, opened 2016-04-08
CC ignores log_level for DB log


poligraph:
I am trying to reduce logs from CF and I have found that CC ignores settings for DB operations

I use CF V230

-1. go to api machine
-2. open /var/vcap/jobs/cloud_controller_ng/config/cloud_controller_ng.yml and change loglevel

```
db: &db
  database: postgres://ccadmin:password@192.168.111.16:5524/ccdb
  max_connections: 25
  pool_timeout: 10
  log_level: fatal
```

 

-3. check logfile /var/vcap/sys/log/cloud_controller_ng/cloud_controller_ng.log
There is a lot of messages with log level fatal
Even if I change to error or info nothing happened and I still can see a lot of messages like this:

> {"timestamp":1460121236.5210814,"message":"(0.000533s) SELECT * FROM \"delayed_jobs\" WHERE ((((\"run_at\" <= '2016-04-08 13:13:56.519899+0000') AND (\"locked_at\" IS NULL)) OR (\"locked_by\" = 'cc_api_worker.api_z2.0.1')) AND (\"failed_at\" IS NULL) AND (\"queue\" IN ('cc-api_z2-0'))) ORDER BY \"priority\" ASC, \"run_at\" ASC LIMIT 5","log_level":"fatal","source":"cc.background","data":{},"thread_id":47458667846040,"fiber_id":47458708540280,"process_id":24480,"file":"/var/vcap/packages/cloud_controller_ng/cloud_controller_ng/vendor/bundle/ruby/2.2.0/gems/sequel-4.29.0/lib/sequel/database/logging.rb","lineno":70,"method":"block in log_each"}
> 
[...]

PR #618, opened 2016-06-13
Reduce cc.logging_level default from debug2 to info

dcarley:
CC should ship with a default value that is safe and unsurprising for
production deployments, where you want to:

- be able to see errors and warnings without scrolling through lots of debug
  information that may not be relevant
- reduce the throughput and volume of logs that are going into centralised
  stores like Logsearch
- be confident that sensitive information like credentials aren't being
  captured in logs under normal circumstances

cloudfoundry/cf-release may also need updating after this is merged because
the value is repeated in `templates/` and `spec/fixtures/`.

I haven't modified `db_logging_level` because I think `logging_level` takes
precedence and there are some outstanding issues described in:

- cloudfoundry/cloud_controller_ng#572
- cloudfoundry/cloud_controller_ng#617

---

### DIFF
--- a/bosh/jobs/cloud_controller_clock/spec
+++ b/bosh/jobs/cloud_controller_clock/spec
@@ -156,7 +156,7 @@ properties:
     description: "Log level for cc database operations"
 
   cc.logging_level:
-    default: "debug2"
+    default: "info"
     description: "Log level for cc"
   cc.logging_max_retries:
     default: 1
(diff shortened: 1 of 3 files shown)


SocalNick:
@utako @zrob can you comment on why the default log level for CC is `debug2` and not something less verbose like `info`?

thausler786:
We checked this out today and it looks good to us.  We've also referenced this pull request in issue #572.

PR #619, opened 2016-06-14
Service Broker API 2.9 compatibility specs

craigfurman:
As discussed in [tracker](https://www.pivotaltracker.com/story/show/121395851), we've added some service broker API compatibility tests for when a broker returns `operation` data from provision, update, and delete endpoints.

cc @avade


thausler786:
Looks good, merged.

@thausler && @simonleung8 

CAPI team

PR #626, opened 2016-06-21
Add internal controller to accept http start requests from hm9k

swetharepakula:
Hello,

This is the companion controller for the stop one. It's pretty much exactly the same. We've deployed it, ran CATS, tested it manually, seems fine. Story is [here.](https://www.pivotaltracker.com/story/show/112003111)

Let us know if you need anything else.

Thanks,
@jberkhahn & @swetharepakula
Runtime_OG


zrob:
Hey @swetharepakula & @jberkhahn 

This duplicated code is making codeclimate sad.  Could you look at refactoring it so this isn't just a copy and paste?  It seems like almost the entire thing is the same.  Maybe start and stop can just be the same controller with two different paths.

Thanks,
@zrob & @utako

fraenkel:
Combine them into a hm9000_controller

swetharepakula:
We combined them into a single controller, but code climate is still complaining. We don't have access to code climate for the CC, so we can't see what's wrong, though.

thausler786:
@swetharepakula a lot of the complaints from CodeClimate are because it sees this as an entirely new controller, which it is comparing with the other internal controllers. So the duplication of the `initialize` and `read_body` methods is the biggest problem. That does not seem fair to ask you to fix as a part of this pull request, however.

The one bit that does seem actionable is this duplication:

```
      def start(app_guid)
        raise CloudController::Errors::ApiError.new_from_details('NotFound') unless App.find(guid: app_guid)
        logger.info('dea.hm9000.start', app_guid: app_guid)
        SubSystem.hm9000_respondent.process_hm9000_start(read_body)
```

and

```
      def stop(app_guid)
        raise CloudController::Errors::ApiError.new_from_details('NotFound') unless App.find(guid: app_guid)
        logger.info('dea.hm9000.stop', app_guid: app_guid)
        SubSystem.hm9000_respondent.process_hm9000_stop(read_body)
```

This does not seem too bad, however, so we are going to merge this.

@thausler && @sax

PR #636, opened 2016-07-06
make services queryable by id provided by the service broker

tinygrasshopper:
We need to fetch the service from the CC api, using the identifier provided in the service catalog, so adding unique_id to be a query-able parameter. 

Thoughts?

### DIFF
--- a/app/controllers/services/services_controller.rb
+++ b/app/controllers/services/services_controller.rb
@@ -6,7 +6,7 @@ module VCAP::CloudController
       to_many :service_plans
     end
 
-    query_parameters :active, :label, :provider, :service_broker_guid
+    query_parameters :active, :label, :provider, :service_broker_guid, :unique_id
 
     def self.dependencies
       [:services_event_repository]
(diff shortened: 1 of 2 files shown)


adowns01:
Hi @tinygrasshopper, 

You can currently query based on `label` and `provider` which are [guaranteed to be unique](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/db/migrations/20130131184954_new_initial_schema.rb#L93).  Is there a specific reason why that won't work for you and `unique_id` is better?

Thanks, 
@adowns01, CAPI Team Member

tinygrasshopper:
Hi @adowns01 

We are looking at the CC API from the perspective of the service broker author. The service author publishes a unique identifier to the cloud controller via the service catalog. The cloud controller then assigns its own unique identifier to the service, which is used as a 'primary' key in the CC API. As a service broker author I don't have ability to every query on the unique id that I provided to the cloud controller. 
Currently our workaround for this is to fetch all the services(and navigating through all the pages) and detecting the service that we want by matching the unique id. With this change we will need to make fewer calls to the CC API. 
We would also like to make a similar change to the service_plans api.

Best,
@craigfurman and @tinygrasshopper
Services Enablement Team, London

adowns01:
Thanks for the longer explanation @craigfurman and @tinygrasshopper. 👍

PR #643, opened 2016-07-15
Include causing error when raising StatsUnavailable

keymon:
The stats_controller will catch any InstancesUnavailable and raise an
error without any information.

This can make more complicated troubleshoot issues on CF.

Note: Currently this PR simply appends the underlying error to the message, which will result in something like: `Stats unavailable: Stats server temporarily unavailable., Error: Could not connect to the remote server`. I did not want to change it in case some user relies on this exact message.

It might be worth it to change the message, at least to remote the last dot.

### DIFF
--- a/app/controllers/runtime/stats_controller.rb
+++ b/app/controllers/runtime/stats_controller.rb
@@ -17,8 +17,8 @@ module VCAP::CloudController
 
       begin
         [HTTP::OK, MultiJson.dump(instances_reporters.stats_for_app(app))]
-      rescue CloudController::Errors::InstancesUnavailable
-        raise ApiError.new_from_details('StatsUnavailable', 'Stats server temporarily unavailable.')
+      rescue CloudController::Errors::InstancesUnavailable => e
+        raise ApiError.new_from_details('StatsUnavailable', ['Stats server temporarily unavailable.', e.to_s])
       rescue StandardError => e
         raise ApiError.new_from_details('StatsError', e)
       end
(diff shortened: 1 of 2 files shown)


keymon:
Fixed typo that made tests fail.

thausler786:
Hi keymon,

The pull request doesn't break anything, but  we are concerned about leaking credentials.  What kind of errors and logging are you getting when you run this code? Can you give us steps to reproduce your error so we can test this too?

@thausler786 && @adowns01 
CAPI team members

keymon:
For instance, we were having issues with CC callling  `tps_listener`, which was caused by a timeout calling it. `tps_listener` was not responding because doppler was not responding on time (due some network issue).

In consequence I was getting a "Stats unavailable: Stats server temporarily unavailable" which gave us no info at all. We had to read the code and consider any kind of problem when communicating. It was a tricky one,  not easy to troubleshoot at all and took a while.

You can simulate that with any problems when running the underlying code: Network refused, DNS resolution error, bad permissions, etc... For instance, make `tps_listener` hang when calling the `/stats` endpoint: CC it will silently timeout after 10s 

BTW, we wrote a blogpost about that case: https://gdstechnology.blog.gov.uk/2016/07/20/how-our-government-paas-team-solved-an-aws-packet-mystery/

rizwanreza:
Looking at other instances where this error is raised, it's safe to merge these changes. Thanks.

PR #653, opened 2016-07-25
Add option to install buildpacks when not bootstrapping

aemengo:

A manifest property allowing buildpacks to be installed.

We would like buildpacks to be installed on our VM but spec.bootstrap is false.

Signed-off-by: Mark DeLillo <mdelillo@pivotal.io>

### DIFF
--- a/bosh/jobs/cloud_controller_ng/spec
+++ b/bosh/jobs/cloud_controller_ng/spec
@@ -387,6 +387,9 @@ properties:
     description: "Key pair name for signed download URIs"
     default: ""
 
+  cc.run_buildpacks_install:
+    description: "Indicate whether to install buildpacks."
+    default: false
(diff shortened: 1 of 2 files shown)


thausler786:
Hey @aemengo!

Could you give us a bit more context on what led to this pull request?  Do you want buildpacks to be installed on non-bootstrap vms?  Are you running on an old bosh that does not have `spec.bootstrap`? Or is there another reason? 

Thanks,

@thausler786 && @rizwanreza 
CAPI team members

aemengo:
Actually, I'm sorry for leaving this here. The PCF Dev team uses a flavor of bosh that didn't have a `spec.bootstrap` property. But we've since seen the need to modify this [https://github.com/pcfdev-forks/bosh-provisioner](bosh) to have the bootstrap property.

We no longer have a use case for this pull request.

PR #658, opened 2016-08-04
Pass encryption headers on copy.

jmcarp:
This patch updates `FogClient.cp_file_between_keys` to both copy and save files with the specified encryption options. This handles the edge case where encryption is mandatory for an S3 bucket--to ensure that all of our files are encrypted, we configure our buckets to reject any PUTs without encryption set. With this patch, S3 objects are encrypted at all times, instead of copying an object unencrypted and then re-encrypting it afterwards.


thausler786:
Hi @jmcarp,

Looks good, thanks for the pull request!

@adowns01 && @thausler786
CAPI team members

PR #676, opened 2016-09-08
Add validation for volume mount data in service broker bind responses.

julian-hj:

This change adds new validations during service binding to make sure that any volume_mounts returned by the service broker are formatted correctly to be consumed by NSYNC/bbs later on.

Prior to this change, the bind operation would succeed even when the service broker returned a bad volume mount.  The application would then fail to start with json deserialization errors buried somewhere in the NSYNC logs.  We think it is particularly important to provide better feedback for this feature since the expected format from service brokers has changed from API 2.9 to API 2.10 and any volume service brokers out there are still experimental.  After this change, the bind operation will fail with directly with a message telling the user to contact the service provider.

- Validate all required fields
- Validate that 'device' is a json object
- Validate that 'device.mount_config' is a json object if it is included
- Update test payloads in all tests to pass cleanly through new
validations
- add plumbing for v2 and v3 controllers to emit a meaningful message to
the end user in case of bad broker behavior.

In Tracker: [#126647149](https://www.pivotaltracker.com/story/show/126647149)

CATS says yes

### DIFF
--- a/app/actions/service_binding_create.rb
+++ b/app/actions/service_binding_create.rb
@@ -8,6 +8,7 @@ module VCAP::CloudController
     class ServiceBrokerInvalidSyslogDrainUrl < StandardError; end
     class InvalidServiceBinding < StandardError; end
     class VolumeMountServiceDisabled < StandardError; end
+    class InvalidVolumeMount < StandardError; end
 
     include VCAP::CloudController::LockCheck
 
@@ -39,6 +40,7 @@ module VCAP::CloudController
       rescue => e
         logger.error "Failed to save state of create for service binding #{service_binding.guid} with exception: #{e}"
         mitigate_orphan(service_binding)
+        raise InvalidVolumeMount if e.instance_of?(ServiceBindingModel::InvalidVolumeMount)
         raise e
       end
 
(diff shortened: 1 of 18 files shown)


zrob:
hey @julian-hj 

we've made significant changes to service bindings now that we have merged v2 and v3 data storage which have made this pr out of date.

additionally we noticed that there is already volume mount validation [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/services/service_brokers/v2/response_parser.rb#L209).  we would prefer to keep validations of broker responses in one place and not inside the model.

i'm not sure the best way to proceed with this one, would you like to submit a new pr? or possibly a cross-team pair could be set up.

/cc @SocalNick @tedsuo 

thanks,
@zrob & @rizwanreza

julian-hj:
Thanks Zach. Probably we should cross team pair if you guys can accommodate that. We could take another crack at the PR on our side, but since we lack context on the v2/v3 merge it would probably be a bit of a crap shoot.

julian-hj:
This pull request is no longer current.

We moved the validation code into the response parser as Zach suggested, and those changes are on a branch waiting for the pipeline to go green.  See the commit here: https://github.com/cloudfoundry/cloud_controller_ng/commit/75383c8c6c3ef269c36bd8c32b3100a02d23813d

PR #685, opened 2016-09-23
Isolation segments <-> Organizations

DanLavine:
Created a "shared" Isolation Segment that gets seeded in the database.

Organizations have a list of "allowed Isolation Segments"
- Only Admins can manage the allowed list
- Space's Isolation Segments must belong to an Organization's allowed list
- Org managers can manage Space's Isolation Segment

Organizations can set a 'default' from the allowed list.

A number of endpoints to manage Isolation Segments:
- post '/isolation_segments/:guid/relationships/organizations'
- delete '/isolation_segments/:guid/relationships/organizations'
- get '/isolation_segments/:guid/relationships/organizations'
- get '/isolation_segments/:guid/relationships/spaces'
- put '/v2/organizations/:guid' -d '{"default_isolation_segment_guid":"guid"}'
- get /v2/organizations/:guid/isolation_segments


spec/api/documentation/organizations_api_spec.rb
```diff
@@ -90,6 +91,48 @@
...
+              'next'          =>  nil,
+              'previous'      =>  nil
+            },
+            'resources' => [
+              {
+                'name'        => isolation_segment_model.name,
+                'guid'        => isolation_segment_model.guid,
+                'created_at'  => iso8601,
+                'updated_at'  => nil,
+                'links'       => {
```
Gerg: Related to my other comment: the pagination and resource links are v3-style here.

app/controllers/runtime/organizations_controller.rb
```diff
@@ -214,6 +228,20 @@ def share_domain(guid, domain_guid)
...
     end
 
+    get '/v2/organizations/:guid/isolation_segments', :list_isolation_segments
+    def list_isolation_segments(guid)
+      org = find_guid_and_validate_access(:read, guid)
+
+      dataset = IsolationSegmentModel.dataset.where(organizations: org)
+      message = IsolationSegmentsListMessage.from_params(@params)
+
+      [HTTP::OK, MultiJson.dump(Presenters::V3::PaginatedListPresenter.new(
```
ScarletTanager: My understanding from @zrob was that using the v3 presenter was okay, since these are v3-only objects - for example, GET /v3/isolation_segments will use this same presenter, so we'd like to keep the same look/feel.  There is no v2 isolation segment API.  Zach, did we misunderstand the outcome of our last conversation on this?

spec/api/documentation/organizations_api_spec.rb
ScarletTanager: Yep, please see my response above.  Need @zrob or @utako to weigh in with they remember us having agreed on.

app/controllers/runtime/organizations_controller.rb
Gerg: I've talked some with @SocalNick about this.

By exposing this resource on the v2 API, it is not longer a v3-only object. Using `V3:PaginatedListPresenter` means that you will get v3 pagination even though you are hitting a URL that begins with "v2." If I'm building an API client, this means that I can no longer key off of the api version number to determine what pagination scheme will be returned.

Given that, the two options are:
1) Don't expose isolation segments through the v2 api
2) Build v2 presenters for isolation segments

Since these are still experimental endpoints, this change doesn't have to block merging this PR, but we need to change it before going GA.

Gerg:
Since we have been so slow reviewing/merging this, there were a bunch of merge issues. I rebased and pushed to the [isolation_segments branch](https://github.com/cloudfoundry/cloud_controller_ng/tree/isolation_segments)

app/controllers/runtime/organizations_controller.rb
ScarletTanager: @Gerg Thanks for checking with Nick on this.  If you can let me know when you've finished merging, I'll see about a follow-on PR to implement a v2 presenter for Isolation Segments.

-Sandy

app/presenters/v3/isolation_segment_presenter.rb
```diff
@@ -10,7 +10,7 @@ def to_hash
             name:       isolation_segment.name,
             created_at: isolation_segment.created_at,
             updated_at: isolation_segment.updated_at,
-            links:      build_links
+            links: build_links,
```
Gerg: Whitespace

app/presenters/v3/relationship_presenter.rb
```diff
@@ -0,0 +1,30 @@
...
+          }
+        end
+
+        private
+
+        def build_relations
+          data = []
+
+          @relationships.each do |relationship|
+            data << { name: relationship.name, guid: relationship.guid, link: "/v2/#{@relation_url}/#{relationship.guid}" }
```
Gerg: @SocalNick I'm guessing we want this link to be a full path now?

bosh/jobs/cloud_controller_ng/spec
```diff
@@ -225,6 +225,14 @@ properties:
     default: 180
     description: "Maximum health check timeout (in seconds) that can be set for the app"
 
+  cc.shared_isolation_segment_name:
+    default: "shared"
+    description: |
+      Name of the shared isolation segment created at startup.
+      This field can be updated, but subject to the following caveat:
+         Using the name of an existing IS other than the default one will cause a deployment to fail.
+         To recover, redeploy using the old (current) name of the previous Isolation Segment.
```
Gerg: I'm confused about what this is saying. Does this mean "redeploy using the previous shared_isolation_segment_name"?

config/routes.rb
```diff
@@ -66,6 +66,12 @@
   get '/isolation_segments/:guid', to: 'isolation_segments#show'
   delete '/isolation_segments/:guid', to: 'isolation_segments#destroy'
   put '/isolation_segments/:guid', to: 'isolation_segments#update'
+  post '/isolation_segments/:guid/relationships/organizations', to: 'isolation_segments#assign_allowed_organizations'
+  delete '/isolation_segments/:guid/relationships/organizations', to: 'isolation_segments#unassign_allowed_organizations'
+
+  get '/isolation_segments/:guid/relationships/organizations', to: 'isolation_segments#relationships_orgs'
+  # To build these relationships. OrgManagers/Admins set the isolation segment on Spaces
```
Gerg: I don't think this comment is needed. This type of information should be in specs/ API docs instead.

db/migrations/20160922213611_organization_isolation_segments.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    alter_table :organizations do
+      add_column :default_isolation_segment_guid, String, collate_opts
+
+      add_foreign_key [:guid, :default_isolation_segment_guid], :organizations_isolation_segments, :name => 'organizations_isolation_segments_pk'
+    end
+  end
+
+  down do
+    alter_table :organizations do
```
Gerg: Do you also need to remove the `default_isolation_segment_guid` column? Or does Sequel magically do that when dropping the foreign key?

spec/api/documentation/organizations_api_spec.rb
```diff
@@ -19,6 +19,7 @@
       field :status, 'Status of the organization'
       field :quota_definition_guid, 'The guid of quota to associate with this organization', example_values: ['org-quota-def-guid']
       field :billing_enabled, 'If billing is enabled for this organization', deprecated: true
+      field :default_isolation_segment_guid, 'The guid for the default isolation segment', experimental: true
```
Gerg: We are no longer automatically updating the v2 api docs via these tests. In order for this change to be reflected in the docs, you will have to manually edit them.

spec/request/isolation_segments_spec.rb
```diff
@@ -90,121 +207,71 @@
...
-        VCAP::CloudController::IsolationSegmentModel.make(name: 'segment6')
-      ]
-    }
-
-    it 'returns a paginated list of the isolation segments' do
-      get '/v3/isolation_segments?per_page=2', nil, user_header
+    let(:org1) { VCAP::CloudController::Organization.make }
+    let(:org2) { VCAP::CloudController::Organization.make }
 
+    it 'returns the seeded isolation segment' do
```
Gerg: This seems more appropriate as a controller spec

spec/request/isolation_segments_spec.rb
```diff
@@ -214,14 +281,119 @@
...
+        }
+
+        parsed_response = MultiJson.load(last_response.body)
+
+        expect(last_response.status).to eq(200)
+        expect(parsed_response['resources'].map { |r| r['name'] }).to eq([models[3].name, models[5].name])
+        expect(parsed_response['pagination']).to eq(expected_pagination)
+      end
+
+      context 'when the user is not an admin' do
```
Gerg: This would also be more appropriate as a controller spec. Generally, we want the request specs to only cover super happy path cases since they are relatively slow.

Correct me if I'm wrong @utako

spec/spec_helper.rb
```diff
@@ -12,6 +12,8 @@
 require 'timecop'
 require 'awesome_print'
 
+require 'pry'
```
Gerg: Unrelated to PR.

app/controllers/runtime/spaces_controller.rb
```diff
@@ -212,11 +211,14 @@ def delete_isolation_segment(guid)
 
     def before_update(obj)
       if request_attrs['isolation_segment_guid']
-        raise CloudController::Errors::ApiError.new_from_details('NotAuthorized') unless SecurityContext.admin?
+        check_isolation_segment_access!(obj)
 
-        isolation_segment = IsolationSegmentModel[guid: request_attrs['isolation_segment_guid']]
-        raise CloudController::Errors::ApiError.new_from_details('InvalidRelation',
-                "Could not find Isolation Segment with guid: #{request_attrs['isolation_segment_guid']}") unless isolation_segment
+        isolation_segment_guids = obj.organization.isolation_segment_models.map(&:guid)
```
ScarletTanager: Not sure what you mean - organization.isolation_segment_models is the accessor for ISes from orgs...?  The space needs to navigate to the ISes to which the owning org is entitled, I don't really see a shorter path to get there?

app/controllers/v3/isolation_segments_controller.rb
```diff
@@ -86,15 +96,81 @@ def update
...
+
+  def organizations_lookup
+    isolation_segment_model = IsolationSegmentModel.where(guid: params[:guid]).first
+    resource_not_found!(:isolation_segment) unless isolation_segment_model
+
+    message = IsolationSegmentAssignOrgMessage.create_from_http_request(params[:body])
+    unprocessable!(message.errors.full_messages) unless message.valid?
+
+    organizations = Organization.where(guid: message.guids).all
+    resources_not_found!("Organization guids: #{message.guids - organizations.map(&:guid)} cannot be found") unless organizations.length == message.guids.length
```
ScarletTanager: I'll double-check in the AM and either point you to them or add them.

app/controllers/v3/application_controller.rb
```diff
@@ -16,6 +16,14 @@ def unprocessable!(message)
   def unauthorized!
     raise CloudController::Errors::ApiError.new_from_details('NotAuthorized')
   end
+
+  def method_not_allowed!(method, obj)
```
ScarletTanager: lIRC we had at least one legit case for a 405 - but I'll look at each of the cases you pointed out and be sure we're staying on the reservation.

app/models/runtime/organization.rb
```diff
@@ -5,6 +5,19 @@ class Organization < Sequel::Model
...
+      primary_key: :guid,
+      key: :default_isolation_segment_guid
+
+    many_to_many :isolation_segment_models,
+      left_key: :organization_guid,
+      left_primary_key: :guid,
+      right_key: :isolation_segment_guid,
+      right_primary_key: :guid,
+      join_table: :organizations_isolation_segments,
+      before_add: proc { raise CloudController::Errors::ApiError.new_from_details('OperationNotImplemented', 'add_isolation_segment', 'Organization') },
```
ScarletTanager: Hrm...not sure why we went with a 501, because what you say seems entirely correct.

app/models/runtime/organization.rb
```diff
@@ -166,8 +185,37 @@ def billing_enabled?
       billing_enabled
     end
 
+    def default_isolation_segment(isolation_segment)
```
ScarletTanager: Reasonable, will fix.

app/models/runtime/organization.rb
```diff
@@ -166,8 +185,37 @@ def billing_enabled?
...
+      if isolation_segment_models.length == 1
+        self.lock!
+        self.update(isolation_segment_model: isolation_segment)
+        self.save
+      end
+    end
+
+    def unset_default_isolation_segment(isolation_segment)
+      raise CloudController::Errors::ApiError.new_from_details('AssociationNotEmpty', 'Space', 'Isolation Segment') unless
+      !isolation_segment_associated_with_space?(isolation_segment)
```
ScarletTanager: Yuck.  I'll fix that.

db/migrations/20160922213611_organization_isolation_segments.rb
ScarletTanager: I'll check.

spec/spec_helper.rb
ScarletTanager: Whoops, this leaked in.  I'll remove it.  As a side note, is there any good reason *not* to include this in the spec_helper?

app/controllers/v3/isolation_segments_controller.rb
```diff
@@ -86,15 +96,81 @@ def update
...
+
+    isolation_segment_model.db.transaction do
+      isolation_segment_model.lock!
+      orgs.each do |org|
+        isolation_segment_model.remove_organization(org)
+      end
     end
-    dataset
+
+    head :no_content
```
Gerg: Looks like JSON API agrees with you: http://jsonapi.org/format/#crud-updating-relationship-responses

app/models/runtime/organization.rb
DanLavine: 501 I think should be correct here. Right now, there are no endpoints to add an isolation segment to an organization. Or to word it another way, you should not be able to do `organization.add_isolation_segment(isolation_segment.guid)`. This is a safeguard for developers, as it does not use all of the logic when adding an isolation segment (setting the default, etc).

app/models/v3/persistence/isolation_segment_model.rb
```diff
@@ -1,20 +1,32 @@
...
       primary_key: :guid
 
+    many_to_many :organizations,
+      left_key: :isolation_segment_guid,
+      left_primary_key: :guid,
+      right_key: :organization_guid,
+      right_primary_key: :guid,
+      join_table: :organizations_isolation_segments,
+      after_add: proc { |isolation_segment, org| org.default_isolation_segment(isolation_segment) },
+      before_remove: proc { |isolation_segment, org| org.unset_default_isolation_segment(isolation_segment) }
```
DanLavine: To me, thats seems like a very odd place to keep the logic. If you house all of the logic in a v3 action, you could write tests where you do `isolation_segment.add_organization(org)`, but then the default isolation segment is not set. Then you could be writing code for behavior which should not happen.

app/controllers/runtime/spaces_controller.rb
Gerg: Sorry about the confusion. This was a note to myself that I forgot to edit.

I'm talking about `obj.organization.isolation_segment_models.map(&:guid)` vs something like `obj.organization.get_isolation_segment_guids`. That way you define a public interface for organizations instead of reaching into their internal state. We do the former a lot in CC anyway, so this isn't blocking the PR.

app/controllers/v3/isolation_segments_controller.rb
ScarletTanager: https://github.com/cfibmers/cloud_controller_ng/blob/isolation_segments/spec/unit/controllers/v3/isolation_segments_controller_spec.rb#L269-L301  and https://github.com/cfibmers/cloud_controller_ng/blob/isolation_segments/spec/unit/controllers/v3/isolation_segments_controller_spec.rb#L419-L444 hit the resources_not_found! logic.

https://github.com/cfibmers/cloud_controller_ng/blob/isolation_segments/spec/unit/controllers/v3/isolation_segments_controller_spec.rb#L419-L444 (for example) hits the unprocessable! path.

Gerg: Sorry, I was specifically talking about testing the error message logic: `"Organization guids: #{message.guids - organizations.map(&:guid)} cannot be found"`

config/routes.rb
ScarletTanager: I think this was meant as an editorial/reminder for ourselves.  I'll simply delete it.

app/models/v3/persistence/isolation_segment_model.rb
Gerg: In v3 we are trying to avoid putting business logic in our models as much as possible. The model hooks distribute logic from single user actions across multiple different classes and can result in v2's signature scary action at a distance.

In v3 we allow our models be more flexible and use actions and test factories to enforce business rules.

What are your feelings on this @zrob?

ScarletTanager: I think what Dan says is the critical part - we want to avoid enabling a scenario whereby one could entitle an org for a segment (segment.add_organization(org)) without setting that segment as the org default - this is IMHO the cleanest way to get there - you remove the ability to shoot yourself in the foot.  While I understand the philosphical/design point, I think it's throwing out the baby with the bathwater not to leverage some of what Sequel gives you when appropriate.

app/controllers/v3/isolation_segments_controller.rb
```diff
@@ -56,6 +60,9 @@ def destroy
     isolation_segment_model = IsolationSegmentModel.where(guid: params[:guid]).first
     resource_not_found!(:isolation_segment) unless isolation_segment_model
 
+    method_not_allowed!('DELETE', "the #{isolation_segment_model.name} Isolation Segment") if
```
ScarletTanager: I think a 405 is exactly the correct response here - from the spec: "The method specified in the Request-Line is not allowed for the resource identified by the Request-URI."  A 405 doesn't mean you can't DELETE any IS, just this one IS.

app/controllers/v3/isolation_segments_controller.rb
```diff
@@ -67,12 +74,15 @@ def destroy
   def update
     unauthorized! unless roles.admin?
 
-    message = IsolationSegmentCreateMessage.create_from_http_request(params[:body])
-    unprocessable!(message.errors.full_messages) unless message.valid?
-
     isolation_segment_model = IsolationSegmentModel.where(guid: params[:guid]).first
     resource_not_found!(:isolation_segment) unless isolation_segment_model
 
+    method_not_allowed!('PUT', "the #{isolation_segment_model.name} Isolation Segment") if
```
ScarletTanager: Same 405 response as above. :)

app/controllers/runtime/spaces_controller.rb
ScarletTanager: Okay, thanks for the clarification.

app/controllers/v3/application_controller.rb
ScarletTanager: I think both cases where we return 405s are legit, so I think this should stay.

app/messages/isolation_segment_assign_org_message.rb
```diff
@@ -0,0 +1,34 @@
...
+    def self.create_from_http_request(body)
+      IsolationSegmentAssignOrgMessage.new(body.symbolize_keys)
+    end
+
+    validates_with NoAdditionalKeysValidator
+    validates :data, presence: true, array: true, allow_nil: false, allow_blank: false
+    validates_each :data do |record, attr, values|
+      if values.is_a? Array
+        values.each do |value|
+          guid = value[:guid] || value['guid']
```
ScarletTanager: Fixed.

app/controllers/v3/isolation_segments_controller.rb
```diff
@@ -25,11 +29,11 @@ def create
...
-    resource_not_found!(:isolation_segment) unless isolation_segment_model
-
-    unauthorized! unless roles.admin? || isolation_segment_model.spaces.any? do |space|
-      can_read?(space.guid, space.organization.guid)
-    end
+    resource_not_found!(:isolation_segment) unless isolation_segment_model && (
+      roles.admin? ||
+      isolation_segment_model.spaces.any? { |space| can_read?(space.guid, space.organization.guid) } ||
+      isolation_segment_model.organizations.any? { |org| can_read_from_org?(org.guid) }
+    )
```
ScarletTanager: Fixed.

spec/api/documentation/organizations_api_spec.rb
zrob: @Gerg @ScarletTanager I remember that we considered the body the same, but we never expected to see links or v3 pagination on the v2 api

app/models/v3/persistence/isolation_segment_model.rb
zrob: We'd like to stick to the same patterns across our v3 code base so our team can have the same expectations for how all associations work.

app/controllers/v3/isolation_segments_controller.rb
ScarletTanager: Fixed.

app/controllers/v3/isolation_segments_controller.rb
SocalNick: Does 409 seem more appropriate? https://httpstatuses.com/409

app/messages/isolation_segment_assign_org_message.rb
```diff
@@ -0,0 +1,34 @@
...
+    attr_accessor(*ALLOWED_KEYS)
+
+    def self.create_from_http_request(body)
+      IsolationSegmentAssignOrgMessage.new(body.symbolize_keys)
+    end
+
+    validates_with NoAdditionalKeysValidator
+    validates :data, presence: true, array: true, allow_nil: false, allow_blank: false
+    validates_each :data do |record, attr, values|
+      if values.is_a? Array
```
ScarletTanager: This doesn't work as well, because it's a validates_each, not a simple validates - if we move this totally to a custom validator, that validator seems to get invoked before the `validates :data` one, which means we have to reimplement the type checking (for arrays), etc., so it's not really an improvement.  The only alternative I can suggest, if you're simply after simplification, is to move the body to a private method (e.g. `validates_each ... do |...| private_method(args) end`

>

app/models/runtime/organization.rb
ScarletTanager: Fixed.

app/models/runtime/organization.rb
ScarletTanager: Fixed.

spec/unit/controllers/runtime/spaces_controller_spec.rb
```diff
@@ -1348,6 +1499,22 @@ def decoded_guids
           end
         end
 
+        context 'as an org manager' do
+          before do
+            organization.add_manager(user)
+          end
+
+          # it 'fails with a 403' do
```
ScarletTanager: Fixed, also later in the same file.

spec/unit/controllers/runtime/organizations_controller_spec.rb
```diff
@@ -158,15 +319,30 @@ module VCAP::CloudController
...
 
+          it 'does not add creator as an org manager' do
             post '/v2/organizations', MultiJson.dump({ name: 'my-org-name' })
 
             expect(last_response.status).to eq(201)
             org = Organization.find(name: 'my-org-name')
             expect(org.managers.count).to eq(0)
           end
+
+          it 'does not allow creation with a default isolation segment set' do
```
ScarletTanager: Fixed.  Spec description was incorrect, code is correct.

spec/spec_helper.rb
ScarletTanager: Fixed.

spec/unit/models/runtime/v3/persistence/isolation_segment_model_spec.rb
```diff
@@ -2,81 +2,210 @@
...
+                expect {
+                  isolation_segment_model_2.remove_organization(org)
+                }.to raise_error(CloudController::Errors::ApiError)
+              end
+            end
+          end
+        end
+
+        context 'when setting the default isolation segment' do
+          it 'must be in the allowed list' do
```
ScarletTanager: Ugh.  This behavior is partially handled (where it should be) in the organization model spec, but there are empty tests there as well.   Fixing.

ScarletTanager: Fixed.

spec/unit/models/runtime/v3/persistence/isolation_segment_model_spec.rb
```diff
@@ -2,81 +2,210 @@
...
+          end
+
+          it 'can be updated' do
```
ScarletTanager: Fixed.

app/controllers/v3/isolation_segments_controller.rb
ScarletTanager: @SocalNick I don't think so - a 409 implies that you could tweak the request - but same method, same resource - and have it succeed.  That isn't the case here, there is literally no way to delete (or update) this particular segment.

ScarletTanager:
I fixed most of the issues, but I think the following remain:

1. Moving the IS model hooks into v3 action(s)
2. Adding a v2 presenter for ISes (does not block PR, but needs to happen before GA)
3. Moving one negative spec from IS request spec into controller spec
4. Need to come to some agreement on response codes - for the most part, I think the 405s are correct.  The one that still sticks is the 501 case, which may need to become something else (405 or 409, possibly?).  Need a verdict on that.  Does this block the PR?

Gerg:
Regarding all the discussion around HTTP status codes. One of our objectives for v3 was to avoid the status code pedantry that I've been gleefully engaging in. 

Zach reminded me that in our v3 style guide we have a small list of response codes: https://github.com/cloudfoundry/cc-api-v3-style-guide#response-codes. The idea is to use the response codes somewhat generically, but have descriptive error messages, since those are what most people care about anyway.

Given that, it seems like 422 is appropriate for the 501/405 cases. Thoughts?

ScarletTanager:
@Gerg - I don't think so - I actually think that a 405 is correct for the two cases where I'm defending it (there is an entity identifiable by the URI path for which the verb used will not be accepted ever).  It's not because there is some semantic error (I realize that the scope of what could qualify for a 422 is pretty broad, which is one of the reasons to be careful about using it).  In the case where we had a 501, but you argued for a 405, I think you are probably correct - I think that should be a 405.  Since I'm X-team pairing with CAPI today, we can probably work this out in real time.

ScarletTanager:
@gerg @SocalNick any news on when we might get this merged?  It's blocking a bunch of stuff for us.

Gerg:
Our pipeline went green Friday EOD, so we should be able to push today.

ScarletTanager:
@gerg @SocalNick @zrob - it appears that all the commits from this PR are now in master.  Can we close this?

Issue #672, opened 2016-08-30
Last operation API invoked with stale broker credentials


smoser-ibm:
Thanks for submitting an issue to `cloud_controller_ng`. We are always trying to improve! To help us, please fill out the following template.
## Issue

Looking at the latest cloud_controller_ng code, the client's credentials are retrieved at instantiation of the async_job and never retrieved again:  https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/actions/services/service_instance_create.rb#L42  I see the same behavior for service_instance_update and service_instance_delete.  Async job code found here https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/jobs/services/service_instance_state_fetch.rb  Credentials have to be retrieved prior to each call to the service broker.
## Context

see steps to reproduce
## Steps to Reproduce

We have test automation that verifies asynchronous provisions verified the following flow:
1. Issue an asynchronous provision (ie, the broker PROVISION API returned a 202)
2. Wait for the last operation to be invoked 1 time
3. Timed out after 10 minutes
4. At this point we deemed the provision failed (it was still in a "in progress" state)
5. Other testing updated the broker credentials and invoked "cloud-cli update-service-broker" with new credentials
6. Some time later the last operation for the instanced created in 1 was invoked on our broker
7. Broker returns a 401
## Expected result

Scenario should work
## Current result

Scenario doesnt work
## Possible Fix

[not obligatory, but suggest fixes or reasons for the bug]
## `name of issue` screenshot

[if relevant, include a screenshot]

michaelxupiv:
Hello @smoser-ibm,

We tried reproducing this issue, but were unable to do so. We're a bit unsure what your instructions mean. What we tried was:

1. Have a service broker with an async plan.
2. Start creating a service instance with that async service plan.
3. Update the service broker with a new username/password before the above request completes.

However, for us the service instance finished creating successfully. Is there some step we're missing?

Best,
@jberkhahn @michaelxupiv

michaelxupiv:
The linked commit should prevent this from happening. We're going to close this issue, please feel free to reopen if you continue to experience this.

Thanks,
@jberkhahn & @michaelxupiv
CF CAPI Team

PR #693, opened 2016-10-04
Always use the broker client

fraenkel:

Always get the broker client from the Broker.

If the broker credentials or url are updated, any outstanding async create will fail with a 401 or worse.

* [ x] I have viewed signed and have submitted the Contributor License Agreement

* [x ] I have run all the unit tests using `bundle exec rake`

* [x ] I have run CF Acceptance Tests on bosh lite

- Get the latest client in case the broker attributes have been updated

[#129486569]

### DIFF
--- a/app/jobs/services/service_instance_state_fetch.rb
+++ b/app/jobs/services/service_instance_state_fetch.rb
@@ -18,7 +18,7 @@ module VCAP::CloudController
         def perform
           service_instance = ManagedServiceInstance.first(guid: service_instance_guid)
           return if service_instance.nil?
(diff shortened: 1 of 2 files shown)


fraenkel:
Fixes  #672 or https://www.pivotaltracker.com/story/show/129486569

fraenkel:
I would suggest someone fix all the jobs.

Gerg:
We pushed a fix for this with slightly different tests here: https://github.com/cloudfoundry/cloud_controller_ng/commit/b9c9d65b5aa76a0c463796b93fd2d3b4da7dea14

PR #699, opened 2016-10-19
Make service plans queryable by id provided by the service broker

Samze:
Similar to #636 we would like to be able to query service plans by the unique identifier provided in the service catalog.

Sam & @ablease
PCF Services Enablement


Samze:
@utako Is there an issue with Travis builds at the moment? We don't know why this failed.

Samze:
My Github org membership should now be sufficient for the CLA checks.

PR #657, opened 2016-08-02
Introduce nginx configuration for rate limiting

suhlig:
Thanks for contributing to the cloud_controller_ng. To speed up the process of reviewing your pull request please provide us with:

* A short explanation of the proposed change:
  
  This change introduces rate limiting to API calls to the Cloud Controller by exposing the [limit_req](http://nginx.org/en/docs/http/ngx_http_limit_req_module.html ) nginx configuration setting. The rate limiting proposed is applied per user, using nginx' `$http_authorization` variable.

* An explanation of the use cases your change solves

  It will prevent overloading the cloud controller with too many requests. It protects legit requests by applying the limit per-user.

* [x] I have viewed signed and have submitted the Contributor License Agreement

* [x] I have made this pull request to the `master` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [x] I have run CF Acceptance Tests on bosh lite

Please note that the change was implemented by @petergtz and @jordanj.

@smoser-ibm & @suhlig from the Flintstone team

PR #703, opened 2016-10-21
Fix for rate limiting. nginx sublocations do not inherit all parameters from upper locations

joachimjordan:

See pull request #657. I worked on this together with @suhlig and @petergtz. We noticed that when rate limiting is activated, this breaks the nginx config. Reason is that nginx does not document for each parameter if it is inherited by sub locations. The proxy_pass parameter is not inherited and thus must be duplicated for sublocations.

This fix adds the duplication for the proxy_pass parameter to the sublocations. Do you also require a test to ensure that this is working correctly? To test this, a CF installation is neccessary with properties for the rate limiting added to the deployment manifest. 

FYI, @smoser-ibm. @petergtz, @suhlig, do you want to comment?


joachimjordan:
Hi @MarcSchunk, can you also please take a look?

joachimjordan:
Will try again later

PR #707, opened 2016-10-28
Proposal of Private Stacks (Minimal Viable Feature) 

t-seyama:
- A short explanation of the proposed change:  
  We implement Minimal Viable Feature of "Private Stacks"  which is "Stacks" for limited users.  
  This Pull Request is not for merge but for discussion. Please comment for implementation to this  and for feature to Design Document( https://docs.google.com/document/d/1_540iCJlMWQw7q4-JZHKYY_Pk3fmhcJk-vurAM4DcZQ/edit?usp=sharing ) if you have any.  
- An explanation of the use cases your change solves  
  It is possible to provide a special Stack which require additional supports, to use proprietary resource and so on.  
- [ x] I have viewed signed and have submitted the Contributor License Agreement
- [ x] I have made this pull request to the `master` branch
- [ x] I have run all the unit tests using `bundle exec rake`
- [ x] I have run CF Acceptance Tests on bosh lite

### DIFF
--- a/app/controllers/runtime/apps_controller.rb
+++ b/app/controllers/runtime/apps_controller.rb
@@ -185,7 +185,7 @@ module VCAP::CloudController
     end
 
     def before_create
-      space = VCAP::CloudController::Space[guid: request_attrs['space_guid']]
+      space = find_guid(request_attrs['space_guid'], Space)
       verify_enable_ssh(space)
       verify_private_stack(space)
     end
@@ -277,8 +277,7 @@ module VCAP::CloudController
       @request_attrs = self.class::CreateMessage.decode(body).extract(stringify_keys: true)
       logger.debug 'cc.create', model: self.class.model_class_name, attributes: redact_attributes(:create, request_attrs)
 
-      space = find_guid(request_attrs['space_guid'], Space)
-      verify_enable_ssh(space)
+      before_create
 
       creator = V2::AppCreate.new(access_validator: self)
       process = creator.create(request_attrs)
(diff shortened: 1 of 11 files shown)


dieucao:
@t-seyama I believe this feature request may be satisfied by support for Isolation Segments - compute only.
https://docs.google.com/document/d/1JPjKxo0JT5rqKZV1XtpydL6ehlyKIk7ipNWMCz2xyKE/edit#

t-seyama:
Hi @dieucao,
 
Thanks for your comment.
Sorry for the delay.

As we are referring in our Design Document [1],  we agree that Isolation Segments seems to be able to address similar use cases.

However, we think there are two different points between Private Stacks and Isolation Segments.

First, Private Stacks allows different stacks to coexist on the same Cell. For example, cflinuxfs2 and centos can be delivered on in one Cell, while Private Stacks can provide centos stack for the limited users. This will contriubute reducing both resource cost and management cost because Isolation Segments requires different cells for each different stack.

Second, Private Stacks can control visibility of stacks. On the other hand, Isolation Segments can provide special stack for only limited number of users, but existence of the stack is known by all user. We think there are cases that users (expecially using stacks that contains proprietary resources, referred in [1]) want to hide even existence of their stacks from public.

We think Private Stacks and Isolation Segments provide solutions of different levels for the similar use cases, and can be used in combination.

[1] https://docs.google.com/document/d/1_540iCJlMWQw7q4-JZHKYY_Pk3fmhcJk-vurAM4DcZQ/edit#heading=h.v0o3dngekgnu

Regards,

SocalNick:
@t-seyama we haven't received any further comment on your cf-dev post. I'm going to close this for now, we can revisit if we receive further evidence.

PR #716, opened 2016-11-09
Send Isolation Segments as part of the staging and running requests

DanLavine:
This PR enables CC to use Isolation Segments. It can be merged as is, but also requires the PRs to nsync, stager, and runtime_schema to take effect.

Signed-off-by: Sandy Cash <scarlet.tanager@gmail.com>

### DIFF
--- a/app/models/v3/persistence/isolation_segment_model.rb
+++ b/app/models/v3/persistence/isolation_segment_model.rb
@@ -27,5 +27,9 @@ module VCAP::CloudController
       raise CloudController::Errors::ApiError.new_from_details('AssociationNotEmpty', 'Organization', 'Isolation Segment') unless organizations.empty?
       super
     end
+
+    def self.shared_segment
+      @shared_segment ||= IsolationSegmentModel.first(guid: SHARED_ISOLATION_SEGMENT_GUID)
+    end
   end
 end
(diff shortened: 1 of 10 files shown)


app/models/v3/persistence/isolation_segment_model.rb
```diff
@@ -27,5 +27,9 @@ def before_destroy
       raise CloudController::Errors::ApiError.new_from_details('AssociationNotEmpty', 'Organization', 'Isolation Segment') unless organizations.empty?
       super
     end
+
+    def self.shared_segment
+      @shared_segment ||= IsolationSegmentModel.first(guid: SHARED_ISOLATION_SEGMENT_GUID)
```
SocalNick: @ScarletTanager @DanLavine how do we guarantee the first Isolation Segment model is intended to be used as the shared one?

ScarletTanager: I'm not sure what you mean.  If you are referring to the use of `IsolationSegmentModel.first()`, that is simply the current recommend way to find a single row in Sequel.  It's essentially the same semantics as `.find()` for returning a single row.  There will only ever be a single row with that guid, of course.

SocalNick: Couldn't the admin delete the seeded `shared` isolation segment and create another one that is not intended to be the shared IS?

ScarletTanager: Nope.  We do not allow deletion of that segment.

SocalNick: Got it. Thanks!

PR #723, opened 2016-11-16
Isolation segments cleanup

ScarletTanager:
- Move state logic to sequel validations
- refactor to accommodate recent changes to master
- cleanup of specs
- add IsolationSegmentDelete action
- add is_shared_segment? check for individual objects rather than re-query DB
- sort by organization id when doing POST bulk updates

Story reference https://www.pivotaltracker.com/story/show/133869287

### DIFF
--- a/app/controllers/runtime/spaces_controller.rb
+++ b/app/controllers/runtime/spaces_controller.rb
@@ -199,44 +199,28 @@ module VCAP::CloudController
     delete '/v2/spaces/:guid/isolation_segment', :delete_isolation_segment
     def delete_isolation_segment(guid)
       space = find_guid(guid)
-      check_isolation_segment_access!(space)
-
-      check_space_is_empty!(space, 'Removing')
+      check_org_update_access!(space)
 
       space.db.transaction do
         space.lock!
-        space.update(isolation_segment_guid: nil)
-        space.save
+        space.update(isolation_segment_model: nil)
       end
 
       [HTTP::OK, object_renderer.render_json(self.class, space, @opts)]
     end
 
-    def before_update(obj)
+    def before_update(space)
       if request_attrs['isolation_segment_guid']
-        check_isolation_segment_access!(obj)
+        check_org_update_access!(space)
 
-        check_space_is_empty!(obj, 'Adding')
-
-        isolation_segment_guids = obj.organization.isolation_segment_models.map(&:guid)
-        unless isolation_segment_guids.include?(request_attrs['isolation_segment_guid'])
-          raise CloudController::Errors::ApiError.new_from_details('UnableToPerform',
-                                                                   'Adding the Isolation Segment to the Space',
-                                                                   "Only Isolation Segments in the Organization's allowed list can be used.")
-        end
+        raise CloudController::Errors::ApiError.new_from_details('ResourceNotFound', 'Isolation Segment not found') if
+          IsolationSegmentModel.where(guid: request_attrs['isolation_segment_guid']).empty?
       end
 
-      super(obj)
+      super(space)
     end
 
     private
-
-    def check_space_is_empty!(space, action)
-      raise CloudController::Errors::ApiError.new_from_details(
-        'UnableToPerform',
-        "#{action} the Isolation Segment to the Space",
-        'Cannot change the Isolation Segment for a Space containing Apps') unless space.app_models.empty?
-    end
 
     def after_create(space)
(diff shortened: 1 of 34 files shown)


app/controllers/runtime/spaces_controller.rb
```diff
@@ -199,45 +199,29 @@ def delete(guid)
...
-        check_space_is_empty!(obj, 'Adding')
+        check_org_update_access!(space)
 
-        isolation_segment_guids = obj.organization.isolation_segment_models.map(&:guid)
-        unless isolation_segment_guids.include?(request_attrs['isolation_segment_guid'])
-          raise CloudController::Errors::ApiError.new_from_details('UnableToPerform',
-                                                                   'Adding the Isolation Segment to the Space',
-                                                                   "Only Isolation Segments in the Organization's allowed list can be used.")
-        end
+        raise CloudController::Errors::ApiError.new_from_details('ResourceNotFound', 'Isolation Segment not found') if
```
luan: This is a little confusing to read. It may be the go developer talking but I don't really like the if at the end of the line with no conditions. If you were to merge the condition in the same line it'd be a really long line. So I'd prefer if this was just a regular if around the `raise`.

spec/unit/controllers/runtime/spaces_controller_spec.rb
```diff
@@ -1269,120 +1270,23 @@ def decoded_guids
       let(:isolation_segment_model) { IsolationSegmentModel.make }
       let(:organization) { Organization.make }
       let(:space) { Space.make(organization: organization) }
+      let(:assigner) { IsolationSegmentAssign.new }
 
       context 'associating an isolation_segment' do
-        context 'when the space contains no apps' do
```
ScarletTanager: @luan Yes, @DanLavine and I moved these into the model spec for space, since the check is now in a validation in that class.

spec/unit/actions/isolation_segment_assign_spec.rb
```diff
@@ -8,14 +8,22 @@ module VCAP::CloudController
     let(:org2) { Organization.make }
 
     it 'sorts the organizations passed in for assignment' do
-      org.update(guid: 'b')
-      org2.update(guid: 'a')
+      rval = Random.new.rand(1..2)
```
ScarletTanager: Fixed.

spec/unit/actions/isolation_segment_unassign_spec.rb
```diff
@@ -11,14 +11,22 @@ module VCAP::CloudController
     let(:org2) { Organization.make }
 
     it 'sorts the organizations passed in for unassignment' do
-      org.update(guid: 'b')
-      org2.update(guid: 'a')
+      rval = Random.new.rand(1..2)
 
-      org.reload
-      org2.reload
+      if rval == 1
```
ScarletTanager: Fixed.

ScarletTanager:
@luan - Pushed a commit with the requested changes.  Thanks!

@ScarletTanager & @DanLavine

spec/unit/controllers/runtime/spaces_controller_spec.rb
thausler786: @ScarletTanager We were able to find the tests for when spaces contains apps, but we were unable to find the tests for spaces that had no apps in them, particularly testing associating the space with the isolation segment as different roles.  Could you point us to where those tests landed?

spec/unit/actions/isolation_segment_assign_spec.rb
thausler786: Looks good!

spec/unit/actions/isolation_segment_unassign_spec.rb
thausler786: Looks good!

spec/unit/controllers/runtime/spaces_controller_spec.rb
DanLavine: @tharrisca What is the difference in the space having apps or not when assigning an Isolation Segment? If the user is unable to perform the action, then they will get a 403 regardless. Those tests are still in this `spaces_controller_spec.rb` file.

ScarletTanager: @SocalNick I'm not sure who's in the queue to pick this up, but hopefully we've answered the question here?

SocalNick: Community pair changes every day. Maybe ping us in Slack to see if you can get some realtime confirmation during working hours today.

PR #721, opened 2016-11-16
Add bits-service signing credentials to cc config

petergtz:
Thanks for contributing to cloud_controller_ng. To speed up the process of reviewing your pull request please provide us with:

* A short explanation of the proposed change:
    See PR title.

* An explanation of the use cases your change solves
    We are preparing a change in the bits-service that makes requires CC to use basic auth when requesting signed URLs from bits-service. In preparation,  we'd like to add the necessary configuration to CC.

* [x] I have viewed signed and have submitted the Contributor License Agreement

* [x] I have made this pull request to the `master` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [x] I have run CF Acceptance Tests on bosh lite

Signed-off-by: Peter Goetz <peter.gtz@gmail.com>

/cc @smoser-ibm @suhlig

PR #728, opened 2016-11-21
Add bits-service signing credentials to cc worker/clock config

petergtz:
As [proposed]( https://github.com/cloudfoundry/cloud_controller_ng/pull/721#issuecomment-261123213)  by @zrob in #721 

    See PR title.

    We are preparing a change in the bits-service that makes requires CC to use basic auth when requesting signed URLs from bits-service. In preparation,  we'd like to add the necessary configuration to CC.

Signed-off-by: Peter Goetz <peter.gtz@gmail.com>

/cc @smoser-ibm @suhlig

### DIFF
--- a/bosh/jobs/cloud_controller_clock/spec
+++ b/bosh/jobs/cloud_controller_clock/spec
@@ -548,3 +548,9 @@ properties:
   cc.bits_service.private_endpoint:
     description: "Private url for the bits-service service"
     default: ""
+  cc.bits_service.username:
+    description: "Username for the bits-service"
+    default: ""
+  cc.bits_service.password:
+    description: "Password for the bits-service"
+    default: ""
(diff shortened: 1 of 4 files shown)


aashah:
Hey @petergtz,

We'd like for the clock's set of properties to remain minimal. We do not believe these bit_service properties should affect the cloud_controller_clock job. Could we ask for you to remove that portion of the PR? Everything looks good otherwise.

Thanks,
CAPI Community Pair - @aashah & @michaelxupiv

petergtz:
@aashah @michaelxupiv 
The cloud_controller_clock job has a dependency on the blobstore_client, which is the bits_service_client when the bits_service is configured to be enabled. I think it has these dependencies, because it has  jobs regularly delete bits from the blobstore as cleanup tasks. I therefore request to merge this PR as is.

We had to add these properties explicitly to make the cloud_controller_clock job run correctly.

Btw, the fact that the cloud_controller_clock job swallows such configuration errors (like omitting required properties) silently makes it extremely difficult to discover and troubleshoot. Please see #732 for more context.

Thanks.

aashah:
Ah, thanks for that. It was difficult to follow dependencies for some of the configuration options.

We'll go ahead and merge as is then 👍 

Thanks,
CAPI Community Pair - @aashah & @michaelxupiv

Issue #739, opened 2016-12-02
Wrong trevis settings (default has changed)


tyyko:
## Issue

Previous default virtualization environment for travis was the 'SUDO-ENABLED' (the required one to run the tests) but now is the 'CONTAINER-BASED' and test will not work on forks.

Source: 
First row of the table on
https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

## Steps to Reproduce

Fork the repository and enable travis on the fork.

## Expected result

Travis should work out of the box and tests that runs on the original branch should work also on the forked one.

## Current result

Test fails because are killed by Travis due to 'Tests running in parallel using too many processes or threads (e.g. using the parallel_test gem)'

Source: 
https://docs.travis-ci.com/user/common-build-problems/#My-build-script-is-killed-without-any-error

## Possible Fix

Explicitly specify the virtualization environment on the .travis.yml file. Pull request will follow.

thausler786:
Closing this in favor of #740.

PR #740, opened 2016-12-02
Explicitly set the correct virtualization environment

tyyko:
Refers to #739

Default has been changed, forks are now use the wrong environment and test fail.
Source: https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

Not necessary to run CATs, the only change is about travis configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@
 language: ruby
 bundler_args: --deployment --without development
 cache: bundler
+sudo: required
 
 # Travis cannot find the binary for Ruby 2.3, so we specifically choose 2.3.1
 # instead of using the .ruby-version file


thausler786:
Hi @tyyko,

We noticed that other pull requests and forks have been green on Travis. So we're having a hard time understanding exactly what the problem this issue is solving. We'd prefer to not override the travis default unless we really have to.
Could you point us to a failed build that this would fix so that we can better understand the problem?

Also, we'll close the other issue in favor of this PR so that we can have only one place for this discussion to happen.

Thanks,
@thausler786 and @luan, CF CAPI Team

tyyko:
Hello @thausler786 and @luan ,

Thanks for your reply and sorry for the issue, I was used to open an issue before a pull request as a standard process.

I can point you to the build of my other pull request, where on your[1] trevis works (because the trevis job has been created before 2015 = with default SUDO-ENABLED virtualization environment as described) and mine[2] fails because is ran in a container (CONTAINER-BASED virt. env, please refer to build log 'Worker information' because I enabled trevis just a couple of days ago and the default is the container now)

The reason my builds are failing in my opinion are:
- the machine where the container is running have a lot of cores, so parallelization of testing start a lot of parallel jobs instead of 2 like in your build
- container have smaller limits and some tests gets killed

you can check the difference on each build log on the Worker information details, like comparing the startup time (copule of milliseconds VS couple of seconds) as described https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments  or by just looking for 'docker' as part of the hostname

[1] https://travis-ci.org/cloudfoundry/cloud_controller_ng/builds/180734520
[2] https://travis-ci.org/tyyko/cloud_controller_ng/builds/180352692

If you need any further details don't hesitate to ask

Best,
Michael

aashah:
Hey @tyyko,

Thanks for the detailed response. Everything looks good then, merged!

CAPI Community Pair - @aashah & @michaelxupiv

Issue #735, opened 2016-11-30
Default settings for enable/disable ssh


tyyko:
## Issue

When ssh access is enabled by the operator, it is enabled by default for all apps and spaces and an action from space owner or developers is required to locally disable it.

It would make sense to allow the configuration of the default setting and not having it hardcoded to true.

## Context

according to [the official documentation](https://docs.cloudfoundry.org/devguide/deploy-apps/app-ssh-overview.html#ssh-access-control-hierarchy) there are 3 levels of configuration to enable/disable ssh access to apps running on Diego:
 
- **Entire deployment:** I guess this translates to the configuration property [cc.allow_app_ssh_access](https://github.com/cloudfoundry/cloud_controller_ng/blob/654e39c6d22f99e75ba0833236c5e6a3f12ad967/bosh/jobs/cloud_controller_ng/spec#L522)
- **Space:** cf allow-space-ssh/disallow-space-ssh
- **Application:** cf enable-ssh/disable-ssh
 
It seems that, with the current implementation, there's no way to disable ssh
by default while configuring `cc.allow_app_ssh_access` as true.

It is since the default value of `allow_ssh` for a space is true [1], and
the value of `enable_ssh` for an app becomes true if it is nil
and`cc.allow_app_ssh_access`and space.allow_ssh` is true [2].
[1] https://github.com/cloudfoundry/cloud_controller_ng/blob/f2b4118fb23665d8f021ef2f82de9952c563cddd/app/controllers/runtime/spaces_controller.rb#L12
[2] https://github.com/cloudfoundry/cloud_controller_ng/blob/f2b4118fb23665d8f021ef2f82de9952c563cddd/app/models/runtime/app.rb#L274

## Steps to Reproduce

Set `cc.allow_app_ssh_access` to true

## Expected result

I would like to have another configuration parameter, like `default_app_ssh_access` to choose if the ssh service will be enabled or disabled by default when the operator allows ssh with `cc.allow_app_ssh_access`

## Current result

Currently ssh is enabled on all spaces and applications as soon as the operator sets `cc.allow_app_ssh_access` to true.

## Possible Fix

We are working on a pull request to add the described new configuration parameter on the line already [linked](https://github.com/cloudfoundry/cloud_controller_ng/blob/f2b4118fb23665d8f021ef2f82de9952c563cddd/app/models/runtime/app.rb#L274)

aashah:
Hey @tyyko,

[...]

PR #741, opened 2016-12-02
Default value configurable for app ssh access on new applications

tyyko:
Refers to: #735

When ssh access is allowed by the operator, it is enabled by default for all apps and spaces and an action from space owner or developers is required to locally disable it.

It would make sense to allow the configuration of the default setting and not having it hardcoded to true.

On productive deployments, especially when migrating from DEA to to Diego, it would make sense to allow the operator to deploy Diego with the ssh service as 'opt-in' feature where developers can activate the feature but by default we don't open any channel that can lead to security concerns.

I still have to run CATs, will do as soon as possible.

### DIFF
--- a/app/models/runtime/app.rb
+++ b/app/models/runtime/app.rb
@@ -271,7 +271,7 @@ module VCAP::CloudController
     end
 
     def before_save
-      self.enable_ssh = Config.config[:allow_app_ssh_access] && space.allow_ssh if enable_ssh.nil?
+      self.enable_ssh = Config.config[:allow_app_ssh_access] && space.allow_ssh && Config.config[:default_app_ssh_access] if enable_ssh.nil?
       set_new_version if version_needs_to_be_updated?
       super
     end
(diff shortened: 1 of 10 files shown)


tyyko:
I confirm CATS behave exactly the same with upstream version and with my fix applied

aashah:
Hey @tyyko,

Everything looks good. We have one small request, we have been trying to prevent copy/pasta'ing properties over to the clock and worker templates as many of them are not relevant for those jobs. In this case, we believe that this property is not relevant for the clock. Could you please update the PR to remove the property being sent into the clock job's template (see `bosh/jobs/cloud_controller_clock/templates/cloud_controller_clock.yml.erb`)

Thanks,
CAPI Community Pair - @michaelxupiv & @aashah

tyyko:
Hello @michaelxupiv & @aashah,

Thanks for your feedback!
I just removed the changes to the clock job.

Best,
Michael

tyyko:
Hello, are there any other issues about the PR? 

Thanks,
Michael

simonleung8:
@tyyko Sorry for the delay. We had issues with our CI and it was blocking us from merging this PR.

It is merged now. Thanks

PR #761, opened 2017-01-11
Improve DEA queries when determining app health

ScarletTanager:
* improve HM9k bulk/app requests by collecting all information in three
  queries for package state.
* improve lookup for processes in the running state.
* improve lookup for processes in the starting or running state.
* The previous 2 bullet points rely on the fact that a process's
  current droplet is set to nil as part of the staging process and only
  gets set when staging a package is successful.

Signed-off-by: Dan Lavine <dlavine@us.ibm.com>

### DIFF
--- a/lib/cloud_controller/backends/runners.rb
+++ b/lib/cloud_controller/backends/runners.rb
@@ -73,31 +73,93 @@ module VCAP::CloudController
       query.all
     end
 
-    def dea_apps_hm9k(batch_size, last_id)
-      query = App.select_all(App.table_name).
-              runnable.
-              dea.
-              where("#{App.table_name}.id > ?", last_id).
-              order("#{App.table_name}__id".to_sym).
-              limit(batch_size).
-              eager(:latest_droplet, :latest_package, current_droplet: :package)
+    def dea_apps_hm9k
+      # query 1
+      # get all process information where the process is STARTED and running on the DEA
+      process_query = App.db["Select p.id, p.app_guid, p.instances, p.version, apps.droplet_guid from processes p
+                      inner join apps ON (apps.guid = p.app_guid AND p.state ='STARTED' AND p.diego IS FALSE)"]
+      processes = process_query.all
 
-      apps = query.all.reject { |a| a.package_state == 'FAILED' }
+      # query 2
+      # get all necessary droplet information. This includes:
+      #    where the droplet's associated process is running on the DEA and the process is STARTED
+      #    Finding only the latest droplet associated with the process
+      droplets_query = App.db["select d.id, d.guid, d.app_guid, d.created_at, d.package_guid, d.state from droplets d
+                              join processes p ON (d.app_guid = p.app_guid AND p.state ='STARTED' AND p.diego IS FALSE)
+                              inner join (select app_guid, max(created_at) as _max from droplets group by app_guid) as x
+                              ON d.app_guid = x.app_guid and d.created_at=x._max"]
+      latest_droplets = latest(droplets_query.all)
 
-      app_hashes = apps.map do |app|
-        {
-          'id'            => app.guid,
-          'instances'     => app.instances,
-          'state'         => app.state,
-          'memory'        => app.memory,
-          'version'       => app.version,
-          'updated_at'    => app.updated_at || app.created_at,
-          'package_state' => app.package_state,
-        }
+      # query 3
+      # get all necessary package information. This includes:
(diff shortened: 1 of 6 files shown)


spec/unit/lib/cloud_controller/backends/runners_spec.rb
```diff
@@ -344,10 +324,199 @@ module VCAP::CloudController
         staging_pending_app = dea_app1
         PackageModel.make(app: dea_app1.app, state: PackageModel::PENDING_STATE)
 
-        batch, _ = runners.dea_apps_hm9k(100, 0)
+        batch, _ = runners.dea_apps_hm9k
 
-        guids = batch.map { |entry| entry['id'] }
-        expect(guids).to include(staging_pending_app.guid)
+        found = false
+        batch.each do |item|
```
Gerg: This can be rewritten as:
```rb
expect(batch).to include(hash_including({
  'id' => staging_pending_app.guid,
  'package_state' => 'PENDING'
}))
```

lib/cloud_controller/backends/runners.rb
```diff
@@ -73,31 +73,93 @@ def dea_apps(batch_size, last_id)
...
-        {
-          'id'            => app.guid,
-          'instances'     => app.instances,
-          'state'         => app.state,
-          'memory'        => app.memory,
-          'version'       => app.version,
-          'updated_at'    => app.updated_at || app.created_at,
-          'package_state' => app.package_state,
-        }
+    def dea_apps_hm9k
```
Gerg: This method is pretty big and does a lot. It might be clearer to at least break out the queries into their own methods. You could then unit test the queries, so you don't need the comments describing them.

lib/cloud_controller/backends/runners.rb
```diff
@@ -73,31 +73,93 @@ def dea_apps(batch_size, last_id)
...
+        })
+      end
+
+      [process_list, largest_id]
+    end
+
+    def latest(items)
+      current = {}
+
+      items.each do |item|
```
Gerg: You can also use https://ruby-doc.org/core-2.1.0/Enumerable.html#method-i-reduce instead of `each`

lib/cloud_controller/backends/runners.rb
DanLavine: It feels weird making these queries a public function just for testing purposes. Besides, what we really care about is the interaction between all 3 of them and determining the package state for the response to send back.

Gerg: I agree that they feel weird as public methods on `Runners`. Block comments like these can indicate that there is complexity worth testing. You could either extract them into a helper class and unit test them, or just make them private methods.

lib/cloud_controller/legacy_api/legacy_bulk.rb
```diff
@@ -70,7 +70,8 @@ def bulk_apps
       bulk_token = MultiJson.load(params.fetch('bulk_token'))
       last_id = Integer(bulk_token['id'] || 0)
 
-      if last_id > 0
+      max_process_id = App.db["Select max(id) from processes p WHERE p.state ='STARTED' AND p.diego IS FALSE"].first
```
fraenkel: What happens when a new app is added, the max changes.
Why not just send back a sentinel instead like a negative number or the actual MAX value of id.
Sequel can do this for you too via App.max(:id).where(state: 'STARTED').where(diego: FALSE)

lib/cloud_controller/dea/instances_reporter.rb
```diff
@@ -13,16 +13,23 @@ def all_instances_for_app(app)
...
         health_manager_client.healthy_instances(app)
       end
 
-      def number_of_starting_and_running_instances_for_processes(apps)
-        apps_without_running_instances = apps.select { |app| !app.started? || app.staging_failed? || app.pending? }
-        apps_without_running_instances.inject(
-          healthy_instances_bulk(apps - apps_without_running_instances)
-        ) do |result, app|
-          result.update(app.guid => 0)
+      def number_of_starting_and_running_instances_for_processes(processes)
```
DanLavine: I would rather leave this all as one commit since wraps up all of the DEA query logic into one commit. If it was based off of endpoints, then there would be number of commit.

jsievers:
thanks for the fix! will this make it into CF v251?

SocalNick:
@jsievers CF-251 has already been released, we are trying to get this into a fast-follow CF-252

PR #764, opened 2017-01-25
Add VCAP_APP_HOST environment variable to restore behavior as for DEAs

ScarletTanager:
This is the CC-bridge merge implementation of [this pr](https://github.com/cloudfoundry/nsync/pull/17) for nsync.

### DIFF
--- a/lib/cloud_controller/diego/buildpack/desired_lrp_builder.rb
+++ b/lib/cloud_controller/diego/buildpack/desired_lrp_builder.rb
@@ -58,6 +58,7 @@ module VCAP::CloudController
           [
             ::Diego::Bbs::Models::EnvironmentVariable.new(name: 'PORT', value: ports.first.to_s),
             ::Diego::Bbs::Models::EnvironmentVariable.new(name: 'VCAP_APP_PORT', value: ports.first.to_s),
+            ::Diego::Bbs::Models::EnvironmentVariable.new(name: 'VCAP_APP_HOST', value: '0.0.0.0'),
           ]
         end
 
--- a/spec/unit/lib/cloud_controller/diego/buildpack/desired_lrp_builder_spec.rb
+++ b/spec/unit/lib/cloud_controller/diego/buildpack/desired_lrp_builder_spec.rb
@@ -136,6 +136,7 @@ module VCAP::CloudController
             expect(builder.port_environment_variables).to match_array([
               ::Diego::Bbs::Models::EnvironmentVariable.new(name: 'PORT', value: '11'),
               ::Diego::Bbs::Models::EnvironmentVariable.new(name: 'VCAP_APP_PORT', value: '11'),
+              ::Diego::Bbs::Models::EnvironmentVariable.new(name: 'VCAP_APP_HOST', value: '0.0.0.0'),
             ])
           end
         end


lib/cloud_controller/diego/app_recipe_builder.rb
```diff
@@ -176,6 +176,12 @@ def generate_ssh_action(user, environment_variables)
       def generate_environment_variables(lrp_builder)
         environment_variables = lrp_builder.port_environment_variables
 
+        if process.app.lifecycle_type == BuildpackLifecycleDataModel::LIFECYCLE_TYPE
```
thausler786: We have been trying to not introduce lifecycle specific code into `AppRecipeBuilder`. If you wanted to add a method `#foo` to the`Buildpack:: DesiredLrpBuilder` one, we ask that there be a matching one on the `Docker::DesiredLrpBuilder` to avoid having branch statements on lifecycle type.

lib/cloud_controller/diego/app_recipe_builder.rb
```diff
@@ -176,6 +176,12 @@ def generate_ssh_action(user, environment_variables)
...
+          lrp_builder.host_environment_variables.each do |var|
```
thausler786: Ideally, `#host_environment_variables` gets collapsed into `port_environment_variables`

lib/cloud_controller/diego/buildpack/desired_lrp_builder.rb
```diff
@@ -61,6 +61,10 @@ def port_environment_variables
           ]
         end
 
+        def host_environment_variables
```
thausler786: As mentioned above, this should be collapsed into `#port_environment_variables`

spec/unit/lib/cloud_controller/diego/app_recipe_builder_spec.rb
```diff
@@ -216,7 +221,8 @@ module Diego
               global_environment_variables: env_vars,
               privileged?:                  false,
               ports:                        lrp_builder_ports,
-              port_environment_variables:   expected_action_environment_variables,
+              port_environment_variables:   expected_action_port_variables,
```
thausler786: Since we stub out the `Buildpack::DesiredLrpBuilder` with an instance double here, we should not have to touch the `app_recipe_builder_spec` since our changeset is isolated within the `Buildpack::DesiredLrpBuilder`. This statement only holds true if the two methods are collapsed as mentioned above.

ScarletTanager: @thausler786 @aashah Single comment to respond to all of yours :).  Makes sense, I'll modify as requested.

ScarletTanager:
@thausler786 have you had a chance to look at my updated commit yet?

spec/unit/lib/cloud_controller/diego/app_recipe_builder_spec.rb
aashah: Could we revert all changes to `app_recipe_builder_spec`, this change should not have impacted the `AppRecipeBuilder` in anyway? I'll watch this PR and merge it when that happens, so you don't have to wait for the next community pair to pick this up.

ScarletTanager: @aashah Sure thing.

ScarletTanager:
@aashah Replaced the commit with one with all changes to `app_recipe_builder_spec.rb` reverted.

PR #765, opened 2017-01-27
Clarify when IS entitlement cannot be revoked

SocalNick:

V3 Docs Fix

### DIFF
--- a/docs/v3/source/includes/experimental_resources/isolation_segments/_unassign.md.erb
+++ b/docs/v3/source/includes/experimental_resources/isolation_segments/_unassign.md.erb
@@ -26,4 +26,4 @@ Example Response
 HTTP/1.1 204 No Content
 ```
 
-This endpoint revokes the entitlement for the specified organization to the isolation segment.  Only Admins can revoke isolation segment entitlements.  If the isolation segment is assigned to a space within an organization, the entitlement cannot be revoked.  If the isolation segment is the organization's default, and the organization has more than one isolation segment, the entitlement cannot be revoked.
+This endpoint revokes the entitlement for the specified organization to the isolation segment.  Only Admins can revoke isolation segment entitlements.  If the isolation segment is assigned to a space within an organization, the entitlement cannot be revoked.  If the isolation segment is the organization's default, the entitlement cannot be revoked.


SocalNick:
@ScarletTanager we were talking to docs and this clause didn't make sense for us. Can you let us know if the above change is accurate?

CAPI - don't merge until @ScarletTanager gives this the OK.

ScarletTanager:
@SocalNick I agree that my original wording is a bit...wordy and confusing.  Having said that, it is also accurate.  Two examples:

1) Org X is entitled to ISes A, B.  A is set as the org default for X.  Revocation of entitlement X->A will fail.

2) Org X is entitled to IS A _only_.  A is set as the org default for X.  Revocation of entitlement X-> will succeed.

I'm totally good with rewording, but I think we need to include this information?

SocalNick:
That's a super confusing constraint. Any chance you might want to simplify by making it such that you can't revoke entitlement if the IS is set as the default?

SocalNick:
cc @zrob

ScarletTanager:
Given that the default is not (IIRC) nullable, there is currently no way to remove entirely the default IS from an org.  So the constraint is something of a (admittedly) kludgy way to provide this mechanism.  But I agree, it's definitely not super awesome.  I frankly like your idea better - it's at least much more straightforward.  But if we go that route, then we should have a backlog item to look at allowing an org manager to unset the org default.

aashah:
Hello, 

/cc @zrob 

We've been having difficulty following this conversation. We've confirmed that Sandy's description of the behavior is how things currently work, i.e. you can only revoke entitlement to the default iso seg if it's the only one the org is entitled to.

It is unclear why there is different behavior with varying a number of entitlements to the org. What's the problem with revoking the default entitlement if there are other entitlements? I can start with a new org, add an entitlement and not set a default to arrive at the same state.

- CAPI Community Pair @aashah & @jberkhahn

aashah:
Hey @SocalNick,

Thanks for reaching out about our docs. From looking at our backlog, we are currently reworking the way entitlement revocation works. We do agree with the docs being confusing, and hope to fix them alongside the current prioritized work.

You can view the isolation segments epic here: https://www.pivotaltracker.com/epic/show/2977797

CAPI Community Pair, @aashah & @ScarletTanager

PR #769, opened 2017-01-31
Use the internal routing api endpoint

alexjh:
Use the cluster's internal routing API endpoint rather than making a
hairpin through the router.

### DIFF
--- a/bosh/jobs/cloud_controller_ng/spec
+++ b/bosh/jobs/cloud_controller_ng/spec
@@ -730,6 +730,13 @@ properties:
     description: "Whether to expose the routing_endpoint listed at /v2/info. Enable this after deploying the Routing API"
     default: false
 
+  routing_api.uri:
+    description: "URL where the routing API can be reached internally"
+    default: http://routing-api.service.cf.internal
+  routing_api.port:
+    description: "Port on which Routing API is running."
+    default: 3000
+
   cc.reserved_private_domains:
     description: "File location of a list of reserved private domains (for file format, see https://publicsuffix.org/)"
     default: ~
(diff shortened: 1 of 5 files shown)


zrob:
@shalako It seems reasonable for the CC to use the internal endpoint when talking to routing api.  Does the routing api present https internally also?  If so we would prefer to move that.

thanks
@zrob && @ljfranklin

shalako:
@zrob Roger that; CC will now use internal hostname for Routing API. HTTPS is not supported but I'll record your request.

shashwathi:
@shalako : Yes, that's correct.

@zrob @ljfranklin 
Eventually we(Cloudfoundry) would like to move away from consul and this change is not going in that direction. I would vote for CC to talk to routing api through Gorouter(current state) and not use internal endpoint. When we decide to move away from consul DNS, we have less components affected by this change.
Another advantage is Gorouter can track [response metrics](https://docs.cloudfoundry.org/loggregator/all_metrics.html#gorouter) for all internal components and in long run this helps us track the health of platform.

@shalako correct me if I am wrong.

Regards
Shash

shalako:
@shashwathi mentions some potential benefits of requests going through the router, but I don't have product requirements for it at this time.

I believe that in order to eliminate Consul, another solution for internal system communication will have to take it's place; I've heard mention of BOSH-provided internal DNS. Not sure whether this solution will support our 'one-active-instance' requirement that comes from trying to provide event consistency backed by a relational db.

aashah:
@alexjh 

Thanks for opening up this discussion and PR. The change does not seem suitable at the present time as per the Routing team's comments. We'd like to close this PR until we have a better approach.

CAPI Community Pair, @aashah @ScarletTanager 

/cc @zrob

PR #778, opened 2017-02-14
Add user role events at space, org creation time

ScarletTanager:
The features to record events for role assignment both missed the cases where the assignments are done as part of the POST (create) and PUT (update) requests on the org/space.  This provides the POST function, PUT will be a separate PR.

### DIFF
--- a/app/controllers/runtime/organizations_controller.rb
+++ b/app/controllers/runtime/organizations_controller.rb
@@ -285,9 +285,26 @@ module VCAP::CloudController
     end
 
     def after_create(organization)
-      return if SecurityContext.admin?
-      organization.add_user(user)
-      organization.add_manager(user)
+      unless SecurityContext.admin?
+        organization.add_user(user)
+        organization.add_manager(user)
+      end
+
+      organization.users.each do |user|
+        @user_event_repository.record_organization_role_add(organization, user, 'user', UserAuditInfo.from_context(SecurityContext), request_attrs)
+      end
+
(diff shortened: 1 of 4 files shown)


app/controllers/runtime/organizations_controller.rb
```diff
@@ -285,9 +285,26 @@ def user_guid_parameter
...
+      unless SecurityContext.admin?
+        organization.add_user(user)
+        organization.add_manager(user)
+      end
+
+      organization.users.each do |user|
+        @user_event_repository.record_organization_role_add(organization, user, 'user', UserAuditInfo.from_context(SecurityContext), request_attrs)
+      end
+
+      organization.auditors.each do |auditor|
```
aashah: Since our code only looks to add them as a "user" and "manager", do we need to: a) Loop through all possible users of type {user/billing manager/audit/manager) b) try to record events for the other types not explicitly added above (billing manager & auditor)

ScarletTanager: @aashah I'm honestly not sure what you mean - the events for auditor and billing manager are added just below this.

aashah: Spoke with @ScarletTanager on Slack, I definitely misunderstood the change. It seems that during org creation, we can also supply params `billing_manager_guids` (not documented). I initially believed we only had to emit events for the two add_user/add_manager calls.

Merging once I fix the api_version spec.

PR #785, opened 2017-02-23
Generate role events on updates to spaces and orgs

ScarletTanager:
- Covers when the roles are enumerated in the request body

These are the last (I think) endpoints whereby user roles can be assigned in spaces/orgs, so making the behavior of them consistent with the others.  Long-term, all of these endpoints should be refactored to use a common code path, but that isn't part of this PR.

### DIFF
--- a/app/controllers/runtime/organizations_controller.rb
+++ b/app/controllers/runtime/organizations_controller.rb
@@ -66,8 +66,6 @@ module VCAP::CloudController
           'Could not find Isolation Segment to set as the default.') unless IsolationSegmentModel.first(guid: request_attrs['default_isolation_segment_guid'])
       end
 
-
-
       super(org)
     end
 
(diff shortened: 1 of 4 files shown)


app/controllers/runtime/organizations_controller.rb
```diff
@@ -59,14 +59,44 @@ def self.translate_validation_exception(e, attributes)
...
           'Could not find Isolation Segment to set as the default.') unless IsolationSegmentModel.first(guid: request_attrs['default_isolation_segment_guid'])
       end
 
-      super(obj)
+
+
+      super(org)
+    end
+
+    def update(guid)
```
ScarletTanager: @ericpromislow Unfortunately, it's not that simple - b/c of the way API dispatch happens, if you cross lines between objects it gets a lot hairier.  The ideal way to do this would be to move all of the role assignment (not just these update methods) into workflow actions, but that's a pretty sizable refactor.

If you look, all this is is the `update()` method copied from the `ModelController`  - which I had to do b/c otherwise it's instantiating a new controller object on each`before_update()` and `after_update()` call - which breaks the logic, b/c you have to have a single controller object to hold the Tx state as it sits right now.  So no, I don't think this can be moved into a library method (nor do I think it should be), b/c I think that ends up just being throwaway code.

app/controllers/runtime/organizations_controller.rb
```diff
@@ -318,5 +348,65 @@ def after_update(organization)
...
+
+          user_guids_added.each do |user_id|
+            user = User.first(guid: user_id) || User.create(guid: user_id)
+            user.username = '' unless user.username
+
+            @user_event_repository.record_organization_role_add(
+                organization,
+                user,
+                role,
+                UserAuditInfo.from_context(SecurityContext),
```
ScarletTanager: Hrm, how'd rubocop miss that on my end?  I'll fix it.

ScarletTanager:
@ericpromislow Pushed a new commit to fix the stuff you asked me to fix, plus a little bit extra as we discussed (other uses of UserAuditInfo).

aashah:
Merged in [51939eb](https://github.com/cloudfoundry/cloud_controller_ng/commit/51939eb)

PR #800, opened 2017-03-29
Add option to overwrite security groups when seeding

MatthiasWinzeler:
In the current implementation of the CC, security groups specified in the configuration are created in the database **only** on the first deploy. On subsequent deploys, any changes in the configuration are not applied to the database.

In our use case, we'd like to have the CC configuration/CF deployment manifest as a single source of truth for some security groups, i.e. the global ones like public_networks, services, dns.

Currently, we specify these security groups on the initial CF deploy.
When we change them after this, we have to update them manually using the CF API (which is ok).
But the security groups specified in the CF deployment manifest are now outdated (which leads to confusion), so we need to update them there as well.

This PR adds an option that, if enabled, applies the security group definition from the CC configuration on every deploy. This allows us do manage the global security groups only using the CC config.

We're aware that not all security groups will be in the CC configuration (i.e. those created for specific spaces only), but we'd still to manage the set of global security groups using the deployment manifest alone. Also, the case that old security groups might be left in the database if removed from the config definition is fine for our use case since we most probably won't delete them ever.

* [ ] I have run CF Acceptance Tests on bosh lite

I was not able to run the CF Acceptance tests successfully on my laptop - even the current master branch had some failures (mostly due to timeouts - the whole suite took almost 3 hours). I did test it intensely on bosh-lite though and I'm pretty confident it doesn't break stuff.

### DIFF
--- a/bosh/jobs/cloud_controller_clock/spec
+++ b/bosh/jobs/cloud_controller_clock/spec
@@ -534,11 +534,11 @@ properties:
       description: "Set of buildpacks to install during deploy"
 
   cc.security_group_definitions:
-    description: "Array of security groups that will be seeded into CloudController."
+    description: "Array of security groups that will be seeded into CloudController. Note: security groups are only seeded on the first deploy, after which they should be managed via the API"
   cc.default_running_security_groups:
-    description: "The default running security groups that will be seeded in CloudController."
+    description: "The default running security groups that will be seeded in CloudController. Note: security groups are only seeded on the first deploy, after which they should be managed via the API"
   cc.default_staging_security_groups:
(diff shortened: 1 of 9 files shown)


zrob:
Hi @MatthiasWinzeler 

Thanks for taking the time to work on this.

I am concerned about introducing a second possible source of truth for this resource. We expect security groups to be managed by the API. If an operator were to enable this config, then only one user would have to make a change via API for things to get out of sync.  It also introduces an additional maintenance point for managing security groups via manifest.  I would prefer we stick with the API always being the source of truth.  In the future please consider opening an issue so we can reach an agreement before putting efforts into a PR.  

Thanks!
-Zach

PR #801, opened 2017-03-29
Add option to overwrite quota definitions when seeding

MatthiasWinzeler:
Same use case as https://github.com/cloudfoundry/cloud_controller_ng/pull/800 but for quotas: We'd like to manage our default quotas using the CF deployment manifest as a single source of truth.

* [ ] I have run CF Acceptance Tests on bosh lite

Unfortunately, I was not able to run them due to timeouts but tested them using bosh-lite.


zrob:
Hi @MatthiasWinzeler 

Closing for the same reason as https://github.com/cloudfoundry/cloud_controller_ng/pull/800

PR #813, opened 2017-04-27
Increase ssl_verify_depth for more complex ca ceritficate validation

andrew-edgar:
We use a more complex CA cert and we are seeing failures on the internal communication between TPS and CC as the client cert cannot be validated with the default value.

All tests pass as this is a very small change and does not effect current tests.

Did not add a new test as that would be more an integration level test to add new more complex certs.  This will allow us to support our deeper cert chain.


jenspinney:
Sounds good. Thanks for the PR!

- @jenspinney and @ericpromislow, CAPI Community Pair

PR #834, opened 2017-06-19
Service broker create instance schemas

Samze:
As a service broker author, I can register a single plan with a create service instance schema [#145580729](https://www.pivotaltracker.com/story/show/145580729)

As a client tooling author, I can retrieve a create service instance schema for a plan [#146453595](https://www.pivotaltracker.com/story/show/146453595)

## Why

The Open Service Broker API is [proposing](https://github.com/openservicebrokerapi/servicebroker/issues/59) allowing brokers to define JSON schema for their configuration parameters. This will allow tooling to validate parameters and UIs to auto generate forms.

Schemas are to be defined as part of the catalog on a plan and support create/update parameters on a service instance and create parameters on a service binding (update does not exist yet).

The updated spec can be found [here](https://github.com/avade/servicebroker/blob/method-attached-schemas/spec.md#schema-object).

Example of what a new catalog with schemas will look like:
```
{
     ....
    "plans": [{
      "name": "fake-plan-1",
      "id": "d3031751-XXXX-XXXX-XXXX-a42377d3320e",
      "description": "Shared fake Server, 5tb persistent disk, 40 max concurrent connections",
      "schemas": {
        "service_instance": {
          "create": {
            "parameters": {
              "$schema": "http://json-schema.org/draft-04/schema#",
              "type": "object",
              "properties": {
                "billing-account": {
                  "description": "Billing account number used to charge use of shared fake server.",
                  "type": "string"
                }
              }
            }
          }
        }
      }
    }]
}
```

## What

This PR just adds basic support for create instance schemas. Schemas are parsed during registration, stored in the service plan model and retrieved on the /v2/service_plan/:guid api endpoint. If a broker does not provide a schema, then we default to an empty schema.

* Add `create_instance_schema` to plan model object
* Add migration to service_plans to include text blob for `create_instance_schema`
* Modify service catalog objects to represent and validate schemas on broker registration
* Propagate schema errors back to users on failed registration
* Expose schemas on /v2/service_plans/:guid
* New presenter for service_plans to format and provide default schemas when nil
* Added new 2.13 service broker api spec

1. Validating that the provided schema is valid JSON Schema v4 draft.
1. Placing limits on the size of schemas that can be stored.
1. Support for update instance / create binding.

## Notes
1. For the proposed 2.13 there can be a maximum of 3 schemas attached to a plan, create/update instance and create binding. We think the best way to represent this in the DB is by adding fields to the plan. This seemed simpler than alternatives, such as adding new DB tables. 
1. During parsing of the catalog we interrogate the expected JSON to pull out the create schema to store. When presented on the API /v2/service_plans/:guid, we reconstruct this JSON. However, by doing this it means we separate the concerns of presentation to storage. This means we could iterate on our presentation separate to the OSBAPI spec if needed.
1. We have added the 2.13 broker api spec which verifies that this functionality works but have not bumped to 2.13. The bump to 2.13 will happen when it has been accepted by the working group. This allows us to experiment with this change.
1. We want to support schemas that are up to 128 of KB's in size and enforce this by the DB column type. You can set the limit on a varchar but not a text blob. Unfortunately MySql has a row limit of 64KB, but this excludes text and blobs. This means we must use the text type and will have to validate size before storing the schema. More info [here](https://www.pivotaltracker.com/story/show/145580729/comments/173039645).
1. We have no updated the *docs* yet as this is still an experimental change. 

Feedback appreciated!

## PR 

Sam & @ablease

cc: @matthewmcnew @st3v


spec/acceptance/broker_api_compatibility/broker_api_v2.13_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    context 'when a broker catalog defines plan schemas' do
+      let(:create_instance_schema) {
+        {
+          '$schema' => 'http://example.com/schema',
+          'type' => 'object'
+        }
+      }
+
+      it 'is responds with the schema for a service plan entry' do
```
ericpromislow: Should be `it 'responds with the....`

ljfranklin:
Merged in [here](https://github.com/cloudfoundry/cloud_controller_ng/commit/a08668046292f1129856a6e29e0ad72c662ac468). Thanks for the PR, very thorough!

PR #839, opened 2017-07-03
Update AddConstraintName Rubocop linter to account for additional Sequel methods

tcdowney:
- The linter now checks additional Sequel methods that can add indices
  or constraints to make sure that we are explicitly setting a name
- Added exclusion for preexisting migrations since we do not want to
  edit any indices in them after they have already been ran in user
  environments

[#148173629]

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -104,6 +104,13 @@ Metrics/ParameterLists:
 Metrics/PerceivedComplexity:
   Enabled: false
 
+Migration/AddConstraintName:
+  Include:
+    - 'db/migrations/**/*'
+  Exclude:
+    # skip old migration files since we do not want to fix them after they are in the wild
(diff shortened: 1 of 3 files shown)


Nino55N:
+1

anniesing:
Looks good to me @tcdowney!

ljfranklin:
@tcdowney would this PR handle cases where the unique flag is given as part of the column definition?

```ruby
create_table(:some_table) do
  column :a1, :string, unique: true
end
# or
alter_table(:some_table) do
  add_column :a1, :string, unique: true
end
```

tcdowney:
Yeah @ljfranklin, you're correct that this change won't handle indices that are added during the `add_column` call -- only ones that are added in the explicit methods.  I've moved the chore back to the Icebox in our backlog since this would be good future work on this.

PR #844, opened 2017-07-08
Use hashdiff to show smaller diffs in hashes in failing tests

ericpromislow:

When a test fails due to a diff in hash-type output in a ccng test, the diff shows up as two large blobs, and we need to use a tool like diffchecker to find the differences. It would be better to see a summary of the diffs in the rspec output.

Run `bake` or `be rspec TEST`. See the diffs clearly in the output.

* [x ] I have viewed signed and have submitted the Contributor License Agreement

* [x ] I have run all the unit tests using `bundle exec rake`

* [na ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats) on bosh lite - this PR changes only specs.

### DIFF
--- a/spec/request/apps_spec.rb
+++ b/spec/request/apps_spec.rb
@@ -180,12 +180,12 @@ RSpec.describe 'Apps' do
       expect(parsed_response).to be_a_response_like(
         {
           'pagination' => {
-            'total_results' => 334,
+            'total_results' => 3,
             'total_pages'   => 2,
             'first'         => { 'href' => "#{link_prefix}/v3/apps?page=1&per_page=2" },
             'last'          => { 'href' => "#{link_prefix}/v3/apps?page=2&per_page=2" },
             'next'          => { 'href' => "#{link_prefix}/v3/apps?page=2&per_page=2" },
-            'previousMoose'      => nil,
+            'previous'      => nil,
           },
           'resources' => [
             {
@@ -195,8 +195,8 @@ RSpec.describe 'Apps' do
               'lifecycle' => {
                 'type' => 'buildpack',
                 'data' => {
-                  'buildpacks' => ['bp-name', 'jerp'],
-                  'stackolago'      => 'stacko-name',
+                  'buildpacks' => ['bp-name'],
+                  'stack'      => 'stack-name',
                 }
               },
               'created_at'              => iso8601,
--- a/spec/request/apps_spec.rb
+++ b/spec/request/apps_spec.rb
@@ -180,12 +180,12 @@ RSpec.describe 'Apps' do
       expect(parsed_response).to be_a_response_like(
         {
           'pagination' => {
-            'total_results' => 3,
+            'total_results' => 334,
             'total_pages'   => 2,
             'first'         => { 'href' => "#{link_prefix}/v3/apps?page=1&per_page=2" },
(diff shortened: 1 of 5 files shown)


spec/request/apps_spec.rb
```diff
@@ -180,12 +180,12 @@
       expect(parsed_response).to be_a_response_like(
         {
           'pagination' => {
-            'total_results' => 3,
+            'total_results' => 334,
```
ljfranklin: @ericpromislow why does the test output need to change in this file?

spec/support/matchers/be_a_response_like.rb
```diff
@@ -42,13 +52,39 @@
...
+            if expected_value.is_a?(Regexp)
+              expected_value = expected_value.inspect
+            end
+            summary << "  - #{truncate(expected_value, 80)}"
+            summary << "  + #{truncate(actual_value, 80)}"
+          end
+        end
+      end
+    rescue => ex
+      $stderr.puts("Error in hashdiff: #{ex} \n #{ex.backtrace[0..5]}")
```
ljfranklin: @ericpromislow at a glance, it seems like the return value from `failure_message` is printed to the screen. Seems like we should be using the return value rather than printing straight to stderr

spec/unit/support/matchers/be_a_response_like_spec.rb
```diff
@@ -17,90 +17,98 @@
     it 'fails when comparing an integer' do
       expect {
         expect({ 'a' => 1 }).to be_a_response_like({ 'a' => /\d+/ })
-      }.to raise_expectation_not_met_with_keys(bad_keys: ['a'])
+      }.to raise_expectation_not_met_with_summary_parts('! a:', '- /\d+/', '+ 1')
```
ljfranklin: @ericpromislow is there a slightly more understandable name for this method (`raise_expectation_not_met_with_summary_parts`)? As someone looking at these assertions for the first time, it's hard to figure out what's actually being tested without looking closely at the method implementation itself.

ljfranklin:
@ericpromislow added some comments to the code, you've also got some rubocop failures in Travis. Otherwise looks good!

spec/request/apps_spec.rb
ericpromislow: Yes, those changes shouldn't have been committed. I was working with them to view the generated diffs, but the testing for them should be done in `spec/support/matchers/be_a_response_like.rb`

spec/support/matchers/be_a_response_like.rb
ericpromislow: @ljfranklin I thought given that the exception should be unlikely it would be better to write it directly to stderr, but given that in testing stderr can be redirected, good point.

spec/unit/support/matchers/be_a_response_like_spec.rb
ericpromislow: @ljfranklin I assumed that anyone looking at the unit/spec would be familiar enough with the underlying `be_a_response_like.rb` implementation to know that a `summary` is something it generates, and a closer look shows that a summary part contains one of `-: <dropped key>`, `+ <added key>`, or `! - <old value> + <new value>`. Any suggestions?

tusing:
Thanks @ericpromislow!

@jenspinney && @tusing

PR #847, opened 2017-07-11
Service Broker Create Service Instance Schema Validation

ablease:
As a Service Broker Author, when I register a plan with a create service instance schema, I get fast feedback when my schema is invalid.
 [#146276815](https://www.pivotaltracker.com/story/show/146276815), [#146276747](https://www.pivotaltracker.com/story/show/146276747).

This PR follows from a previous [PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/834).

Why
---
The Open Service Broker API is proposing allowing brokers to define JSON schema for their configuration parameters. This will allow tooling to validate parameters and UIs to auto generate forms.

What
---
This PR improves support for create instance schemas. Schemas are parsed during registration, and validated against the following criteria:
1. Must conform to JSON schema Draft 04
1. Must contain the following property `"type": "object"`
1. Cannot be larger than 64kb
1. Cannot contain external references
1. Cannot contain file references

This PR address the 3 notes that were not implemented in the previous [PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/834).

1. [Update Service Instance Schema support](https://www.pivotaltracker.com/story/show/146082131). 

Feedback appreciated!

### DIFF
--- a/lib/services/service_brokers/v2/catalog_schemas.rb
+++ b/lib/services/service_brokers/v2/catalog_schemas.rb
@@ -4,9 +4,9 @@ module VCAP::Services::ServiceBrokers::V2
   class CatalogSchemas
     attr_reader :errors, :create_instance
 
-    def initialize(attrs)
+    def initialize(schema)
       @errors = VCAP::Services::ValidationErrors.new
-      validate_and_populate_create_instance(attrs)
+      validate_and_populate_create_instance(schema)
     end
 
     def valid?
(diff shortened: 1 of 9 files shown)


ljfranklin:
Hey @ablease, we've prioritized this story and will take a look soon.

lib/services/service_brokers/v2/catalog_schemas.rb
```diff
@@ -1,10 +1,13 @@
...
 module VCAP::Services::ServiceBrokers::V2
+  MAX_SCHEMA_SIZE = 65_536
   class CatalogSchemas
     attr_reader :errors, :create_instance
 
-    def initialize(attrs)
+    def initialize(schema)
       @errors = VCAP::Services::ValidationErrors.new
-      validate_and_populate_create_instance(attrs)
+      validate_and_populate_create_instance(schema)
```
aashah: (This comment doesn't necessarily reflect changes form this specific PR, but the set of PRs)

We were surprised to see validations happening in the constructor. We are more familiar with the ActiveModel pattern of not explicitly implementing `valid?`, but rather including [`validate`](http://guides.rubyonrails.org/active_record_validations.html) statements.

With this approach, we'd be able to convert the basic validations (data types, presence checks, length) with the validations provided by ActiveModel, and separate those from the content validations.

We're asking for this change to make the code consistent with recent V3 code where validations are important. Examples are in the [`messages/`](https://github.com/cloudfoundry/cloud_controller_ng/tree/master/app/messages) folder

lib/services/service_brokers/v2/catalog_schemas.rb
```diff
@@ -13,26 +16,94 @@ def valid?
...
       end
 
-      @create_instance = attrs
+      create_instance_schema = schemas
+      create_instance_path = path.join('.')
+
+      validate_schema(create_instance_path, create_instance_schema)
+      return unless errors.empty?
+
+      @create_instance = create_instance_schema
```
aashah: (This comment doesn't necessarily reflect changes form this specific PR, but the set of PRs)

We were also surprised to see the object being created conditionally through the validations. We'd prefer that the object's validity rely on `valid?`'s return value, rather than presence of an attribute.

lib/services/service_brokers/v2/catalog_schemas.rb
```diff
@@ -13,26 +16,94 @@ def valid?
...
+    end
+
+    def validate_schema(path, schema)
+      schema_validations.each do |validation|
+        break if errors.present?
+        send(validation, path, schema)
+      end
+    end
+
+    def schema_validations
```
aashah: This method is a prime example of being close to how ActiveModel validations are represented. See [ActiveModel::Validations Custom Methods](http://guides.rubyonrails.org/active_record_validations.html#custom-methods)

Samze:
Thanks for the feedback @aashah & @tcdowney. I agree that the `lib/service` validation code would benefit from being less custom.

Since submitting this we have added additional schema support which we are looking to raise as a PR in the next few days. This means that this branch and our head have diverged quite a lot. 

To save us from having to modify this branch, merging, then fixing all the commits from our newer branch would you be willing to accept this PR and have us do the refactoring you suggest as part of our next PR [Update Service Instance Schemas](https://www.pivotaltracker.com/story/show/149064029)? 

Let us know what you think.

aashah:
Seems good to us! Feel free to close the issue.

- Aakash & @tusing

aashah:
Just kidding, let's just merge this PR right now. An upcoming PR can address some of the concerns above.

PR #852, opened 2017-07-19
Remove pry

Samze:


tusing:
Thanks @Samze!

@tusing && @jenspinney

PR #863, opened 2017-08-01
Update network policy endpoints

chinangela:
[#149753377]

We now have a v1 endpoint for network policy and need to update this information for `cf curl /`

Signed-off-by: David McClure <dmcclure@pivotal.io>


jenspinney:
Looks good, thanks!

- @jenspinney && @aashah

PR #865, opened 2017-08-02
Service broker update instance schemas

lurraca:
## Why
The Open Service Broker API is [proposing](https://github.com/openservicebrokerapi/servicebroker/issues/59) allowing brokers to define JSON schema for their configuration parameters. This will allow tooling to validate parameters and UIs to auto generate forms.

Schemas are to be defined as part of the catalog on a plan and support create/update parameters on a service instance and create parameters on a service binding (update does not exist yet).

The updated spec can be found [here](https://github.com/avade/servicebroker/blob/method-attached-schemas/spec.md#schema-object).

Example of what a new catalog with schemas will look like:
```
{
     ....
    "plans": [{
      "name": "fake-plan-1",
      "id": "d3031751-XXXX-XXXX-XXXX-a42377d3320e",
      "description": "Shared fake Server, 5tb persistent disk, 40 max concurrent connections",
      "schemas": {
        "service_instance": {
          "create": {
            "parameters": {
              "$schema": "http://json-schema.org/draft-04/schema#",
              "type": "object",
              "properties": {
                "billing-account": {
                  "description": "Billing account number used to charge use of shared fake server.",
                  "type": "string"
                }
              }
            }
          },
           "update": {
            "parameters": {
              "$schema": "http://json-schema.org/draft-04/schema#",
              "type": "object",
              "properties": {
                "billing-account": {
                  "description": "Billing account number used to charge use of shared fake server.",
                  "type": "string"
                }
              }
            }
          }
        }
      }
    }]
}
```
## What
This PR just adds basic support for update instance schemas (create instance schemas functionality already merged in #834). Schemas are parsed during registration, stored in the service plan model and retrieved on the /v2/service_plan/:guid api endpoint. If a broker does not provide a schema, then we default to an empty schema.
* Add `update_instance_schema` to plan model object
* Add migration to service_plans to include text blob for `update_instance_schema`
* Add support for Update schema to catalog objects retrieved from broker.
* Change service_plans presenter to support update schemas
* Added update schema tests to the 2.13 service broker api spec
* Schema is now a separate object for simpler validation

1. Support create binding schema

## Notes
1. This PR implements feedback given as part of merging #847 
	1. Schemas now use ActiveModel validation
	1. CatalogSchema no longer conditionally creates objects as part of validations
	1. CatalogSchema still validates structure during initialisation as it is required for setting up schema objects
	1. While Schema uses ActiveModel validations, CatalogSchema is the bridge between this world and the custom validation used previously, this can be refactored with the rest of the catalog_* objects later.
1. This PR is an extension of the #834 and #847
1. We have no updated the *docs* yet as this is still an experimental change.
Feedback appreciated!
## PR
Luis & @Samze


lurraca:
Bump CLA check

tusing:
Thanks for the PR @lurraca!

@tusing && @ericpromislow - CF CAPI

PR #877, opened 2017-08-18
Add Originating Identity header to requests to Service Brokers

Samze:
## Why
The Open Service Broker API is [proposing](https://github.com/openservicebrokerapi/servicebroker/pull/222) that requests to service brokers should contain an Originating User ID as a HTTP header.

The header should have the format:

```
X-Broker-API-Originating-Identity: cloudfoundry <value>
```

where _value_ should be Base64 encoded.

For example, a value of:

```
{
  "user_id": "683ea748-3092-4ff4-b656-39cacc4d5360"
}
```
would appear in the HTTP Header as

```
X-Broker-API-Originating-Identity: cloudfoundry 
eyANCiAgInVzZXJfaWQiOiAiNjgzZWE3NDgtMzA5Mi00ZmY0LWI2NTYtMzljYWNjNGQ1MzYwIiwNCiAgInVzZXJfbmFtZSI6ICJqb2VAZXhhbXBsZS5jb20iDQp9
```

## What
This PR makes a change to the service broker HTTP client, and provides that client with the necessary header. The value of the header set to the user who originated the request to the platform.

For requests where there is no originating user, such as last operation and orphan mitigation, the header is omitted. 

The user's id is contained in a JSON string which is base 64 encoded. 

For all cf CLI commands that can trigger calls to the service broker, we have added the appropriate broker acceptance tests:

-    `cf create-service-broker`
-    `cf create-service`
-    `cf update-service`
-    `cf bind-service`
-    `cf delete-service`
-    `cf unbind-service`
-    `cf create-service-key`
-    `cf delete-service-key`
-    `cf bind-route-service`
-    `cf unbind-route-service`

## Notes
1. Docs will be pulled from the OpenServiceBrokerAPI when we bump the version of the spec. 

## PR

Sam, @ablease @lurraca

### DIFF
--- a/lib/cloud_controller/security_context.rb
+++ b/lib/cloud_controller/security_context.rb
@@ -15,6 +15,10 @@ module VCAP::CloudController
       Thread.current[:vcap_user]
     end
 
+    def self.current_user_guid
+      current_user.guid if current_user
+    end
+
     def self.admin?
       roles.admin?
     end
--- a/lib/services/service_brokers/v2/http_client.rb
+++ b/lib/services/service_brokers/v2/http_client.rb
@@ -126,6 +126,9 @@ module VCAP::Services
 
         client.default_header = default_headers
         opts[:header]['Content-Type'] = content_type if content_type
+
+        user_guid = VCAP::CloudController::SecurityContext.current_user_guid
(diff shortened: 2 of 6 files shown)


ericpromislow:
LGTM

PR #886, opened 2017-08-29
Execute bpm from /var/vcap/jobs/bpm/bin

andrewedstrom:
This is the recommended approach suggested by BOSH as /var/vcap/packages
should only be used internally by the job that provides them. For more
information see: https://www.pivotaltracker.com/story/show/150503041


andrewedstrom:
This commit goes with https://github.com/cloudfoundry/capi-release/pull/52 but it shouldn't be an issue if they are accepted separately.

zrob:
@andrewedstrom can you please make your membership to pivotal public so we can make the CLA bot happy?

thanks!

andrewedstrom:
@zrob Is it working now?

zrob:
@andrewedstrom yep, it's picking it up now, thanks!

PR #887, opened 2017-08-30
Schema Validation Refactors

lurraca:
## Why
The SAPI Team was not happy with the current implementation and abstraction of the Catalog Plan Schema Validation piece of work in Cloud Controller.

## What
Refactor and improve the abstraction around Catalog Plan Schema Validation.

Changed the order that Schema Validations run.

## Notes
This PR doesn't change any important functionality in Cloud Controller, except for the order that Schema Validations are run.

## PR
Luis & @Ablease


lib/services/service_brokers/service_manager.rb
```diff
@@ -58,16 +58,34 @@ def update_or_create_plans(catalog)
           plan.public = false
         end
 
+        create_instance = nil
+        begin
+          create_instance = catalog_plan.schemas.service_instance.create.schema.try(:to_json)
```
ericpromislow: `create_instance = catalog_plan.schemas.service_instance.create.schema.try(:to_json) rescue nil` 

is more idiomatic and then you don't need the other 4 lines

lib/services/service_brokers/service_manager.rb
```diff
@@ -58,16 +58,34 @@ def update_or_create_plans(catalog)
...
+        rescue
+        end
```
ericpromislow: There's a potential for exceptions to be lost here -- in the old code they would have been propagated to the top. Here the exceptions should be logged.

Also, in this block there's no need for the `.try` component.

lib/services/service_brokers/v2/catalog_plan.rb
```diff
@@ -37,10 +46,13 @@ def validate!
...
       validate_bool!(:free, free) if free
       validate_bool!(:bindable, bindable) if bindable
+      validate_hash!(:schemas, @schemas_data) if @schemas_data
     end
 
     def validate_schemas!
-      errors.add_nested(schemas, schemas.errors) unless schemas.valid?
+      if schemas
+        errors.add_nested(schemas, schemas.errors) unless schemas.valid?
+      end
```
ericpromislow: I would simplify this as

    if schemas && !schemas.valid?  
      errors.add_nested(schemas, schemas.errors)
    end

lib/services/service_brokers/v2/parameters_schema.rb
```diff
@@ -0,0 +1,47 @@
...
+      @errors = VCAP::Services::ValidationErrors.new
+      @parameters = parameters
+      @parameters_data = parameters['parameters']
+      @path = path + ['parameters']
+
+      build_schema
+    end
+
+    def build_schema
+      return unless @parameters_data
```
ericpromislow: No need for the first test because `nil.is_a?(Hash)` => `false`

lib/services/service_brokers/v2/service_instance_schema.rb
```diff
@@ -0,0 +1,52 @@
...
+      @update_data = instance['update']
+
+      @path = ['service_instance']
+
+      build_create
+      build_update
+    end
+
+    def build_create
+      return unless @create_data
```
ericpromislow: `return unless @create_data` not needed.

ericpromislow:
Bunch of minor points addressed inline.

lurraca:
We made some changes to the content of this PR. Please let us know what you think.

Luis & @AlbertoImpl

matt-royal:
Hey @lurraca. The changes look good to us. Please resolve the conflicts and then we'll merge this.

Matt & @anniesing

utako:
Hi @matt-royal and @anniesing!

We've resolved all of the conflicts and it should be good to go.

Thanks,
@ablease and @utako

lib/services/service_brokers/service_manager.rb
```diff
@@ -58,16 +59,37 @@ def update_or_create_plans(catalog)
           plan.public = false
         end
 
+        create_instance = nil
+        begin
+          create_instance = catalog_plan.schemas.service_instance.create.schema.to_json
+        rescue => e
+          @logger.error("Error parsing service_instance create schema: #{e.message}")
+        end
+
```
ljfranklin: I'm concerned about these rescues as the intent is unclear. Do you expect one of the fields to be nil and rescue a nil pointer exception? Do you expect a JSON marshal error due to bad input? Is there error actually cosmetic or was there a misunderstanding that `@logger.error` would re-raise?

Could you rewrite this to be more explicit about which failure mode you expect?

ljfranklin:
Hey @ablease & @utako, added one more ask inline about the `rescue => e` blocks. Sorry about the delay in getting this merged, we're working on ironing out our community process to be a little more responsive going forward. Thanks for your patience!

lib/services/service_brokers/service_manager.rb
utako: As far as I can tell, we're just protecting against the absence of create_instance, update_instance, and/or create_binding schema (e.g. if catalog_plan.schemas.service_instance is nil, etc.). We shouldn't actually expect any exceptions aside from one that can be avoided by a conditional and some `try`ing.

tcdowney:
Hey @ablease and @utako, we're taking another look at this PR since it looks like you've added commits since Lyle's comment.  There's quite a few of them though, could you respond to his comment with a commit SHA if one of those addresses it?  That could also be helpful going forward to speed these up. 👍

lib/services/service_brokers/service_manager.rb
tcdowney: Looks like this was updated in this commit: 804050c91ee05d425b4637d85c8d451732df87c6

Samze: We updated this and did some refactoring in 1046fd554920502ef2286d00491358f8e3bfcf14 

We no longer use rescue as a control flow instead we use try. I hope this makes the code a bit clearer.

Samze:
Thanks @tcdowney. We have added more commits and replied to @ljfranklin

PR #888, opened 2017-08-30
Fix infinite loop bug when UAA returns empty key map

ScarletTanager:
- Log correct retry count (was off by 1 before)

[Finishes #150683798]

Signed-off-by: Steven Whitehead <steven.whitehead@ibm.com>

This PR fixes [this bug](https://www.pivotaltracker.com/story/show/150683798).


ljfranklin:
Whoops, nice catch. Thanks for the PR y'all!

ScarletTanager:
@ljfranklin No worries, sorry if our PR was insufficiently metal!

PR #902, opened 2017-09-19
Fix base 64 encoding of HTTP header in test

Samze:
In ruby 2.3.5 (possibly also 2.3.4), new validation rules have been
added when constructing an HTTP request to insure that an HTTP header
doesn't contain a newline. This change fixes the varz test to no longer
include a newline in a HTTP request header.

* [N/A] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats) on bosh lite

Thanks,
Sam and @jenspinney


ljfranklin:
Good catch, we had rebuilt our ruby docker image yesterday but hadn't run units yet. Thanks for the quick fix!

PR #907, opened 2017-09-20
Modify validations on service broker schemas

jenspinney:
## What
This PR modifies two validation rules used when the CC validates the JSON schema provided by service brokers. First, we are removing the requirement that the schema have a `"type":"object"` field. Secondly, we are now requiring that all schemas have a `"$schema"` key.

## Why
Details for these changes can be found in the following tracker stories:

[As a SBA, I can register a JSON schema without the type field](https://www.pivotaltracker.com/story/show/150877576)

[As a SBA, I cannot register a JSON schema without the $schema field](https://www.pivotaltracker.com/story/show/150969868)

These changes reflect the latest requirements set forth in the [Open Service Broker API Spec](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#schema-object).

## PR

Thanks,
Jen and @Samze

### DIFF
--- a/lib/services/service_brokers/v2/schema.rb
+++ b/lib/services/service_brokers/v2/schema.rb
@@ -8,7 +8,9 @@ module VCAP::Services::ServiceBrokers::V2
     MAX_SCHEMA_SIZE = 65_536
 
     validates_length_of :to_json, maximum: MAX_SCHEMA_SIZE, message: 'Must not be larger than 64KB'
-    validate :validate_metaschema_conforms_to_json_draft, :validate_open_service_broker_restrictions
+    validate :validate_metaschema_provided,
+      :validate_metaschema_conforms_to_json_draft,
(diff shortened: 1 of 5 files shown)


lib/services/service_brokers/v2/schema.rb
```diff
@@ -20,7 +22,12 @@ def to_json
 
     private
 
-    def validate_against_metaschema
+    def validate_metaschema_provided
+      return unless errors.blank?
+      add_schema_error_msg('Schema must have $schema key but was not present') unless @schema['$schema']
```
tcdowney: @jenspinney 

Minor nitpick, but I think this error message could be worded a bit more clearly -- maybe something like `Schema missing required $schema key`? 🤔 💭 

Just something to consider if you touch this again, not going to block merging the PR over that.

PR #909, opened 2017-09-20
Perm: Make roles when we make the resource and add role unassignment

iredelmeier:

We centralized when roles get created, so that now they get created when you create the org/space (respectively). This was then removed from creation during role assignments.

We also made CAPI role unassignment trigger a Perm role unassignment.

### DIFF
--- a/app/controllers/runtime/organizations_controller.rb
+++ b/app/controllers/runtime/organizations_controller.rb
@@ -239,6 +239,16 @@ module VCAP::CloudController
       define_method("remove_#{role}_by_user_id") do |guid, user_id|
         response = remove_related(guid, "#{role}s".to_sym, user_id, Organization)
 
+        if config.get(:perm, :enabled)
+          actor = CloudFoundry::Perm::V1::Models::Actor.new(id: user_id, issuer: SecurityContext.token['iss'])
+
+          begin
+            @perm_client.unassign_role(actor, "org-#{role}-#{guid}")
+          rescue GRPC::NotFound
+            # ignored
+          end
+        end
+
         user = User.first(guid: user_id)
         user.username = '' unless user.username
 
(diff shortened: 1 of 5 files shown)


app/controllers/runtime/organizations_controller.rb
```diff
@@ -339,11 +330,44 @@ def add_role(guid, role, user_id, username)
       [HTTP::CREATED, object_renderer.render_json(self.class, org, @opts)]
     end
 
+    def remove_role(guid, role, user_id)
+      response = remove_related(guid, "#{role}s".to_sym, user_id, Organization)
+
+      if config.get(:perm, :enabled)
```
tcdowney: @chendrix @iredelmeier 
I think eventually we'll want the conditional calling of perm tested at the unit level to make sure we're calling perm when we want to with the expected stuff.

Since these attempt to make a network call, tests fail when the perm_client is invoked so it's clear that the feature flag is working.

Does that sound reasonable?

app/controllers/runtime/organizations_controller.rb
```diff
@@ -326,10 +314,13 @@ def add_role(guid, role, user_id, username)
       org = find_guid_and_validate_access(:update, guid)
 
       if config.get(:perm, :enabled)
-        r = @perm_client.create_role("org-#{role}-#{guid}")
         actor = CloudFoundry::Perm::V1::Models::Actor.new(id: user_id, issuer: SecurityContext.token['iss'])
 
-        @perm_client.assign_role(actor, r.id)
+        begin
+          @perm_client.assign_role(actor, "org-#{role}-#{guid}")
+        rescue GRPC::AlreadyExists
```
tcdowney: Could this be done at the client level?

tcdowney: Talked about it in person, this is meant to be a generic client.  They will consider making a CC wrapper class to handle these.

PR #913, opened 2017-09-26
Enable rotation of database encryption keys

ScarletTanager:
Signed-off-by: Tiffany Harris <tiffany.harris@ibm.com>

[#150312752]

See referenced story for workflow details.

### DIFF
--- a/db/migrations/20170830183100_add_key_label_column_to_tables_with_encrypted_columns.rb
+++ b/db/migrations/20170830183100_add_key_label_column_to_tables_with_encrypted_columns.rb
@@ -1,0 +1,32 @@
+Sequel.migration do
+  up do
+    tables_to_alter = %w(
+      env_groups
+      service_brokers
+      service_bindings
+      service_instances
+      service_keys
+      apps
+      buildpack_lifecycle_buildpacks
+      buildpack_lifecycle_data
+      droplets
+      packages
+      tasks
+    )
+    tables_to_alter.each do |table|
+      alter_table(table.to_sym) do
+        add_column :key_label, String
+      end
+    end
+  end
+
+  down do
+    if supports_table_listing?
+      tables.each do |table|
+        alter_table(table) do
+          drop_column :key_label
+        end
+      end
+    end
+  end
+end
--- a/lib/cloud_controller/encryptor.rb
+++ b/lib/cloud_controller/encryptor.rb
@@ -1,35 +1,62 @@
 require 'securerandom'
+require 'openssl'
 require 'openssl/cipher'
+require 'openssl/digest'
+
+# require 'openssl/ossl'
 require 'base64'
 
 module VCAP::CloudController::Encryptor
   class << self
-    ALGORITHM = 'AES-128-CBC'.freeze
+    ALGORITHM = 'AES-256-CBC'.freeze
 
     def generate_salt
-      SecureRandom.hex(4).to_s
+      SecureRandom.hex(8).to_s
     end
 
+    # takes no label, looks up in global config
     def encrypt(input, salt)
       return nil unless input
-      Base64.strict_encode64(run_cipher(make_cipher.encrypt, input, salt))
+
+      label = current_encryption_key_label
+      key = key_to_use(label)
+
+      Base64.strict_encode64(run_cipher(make_cipher.encrypt, input, salt, key))
     end
 
-    def decrypt(encrypted_input, salt)
+    # this takes a label
+    def decrypt(encrypted_input, salt, label=nil)
       return nil unless encrypted_input
-      run_cipher(make_cipher.decrypt, Base64.decode64(encrypted_input), salt)
+
+      key = key_to_use(label)
+
+      run_cipher(make_cipher.decrypt, Base64.decode64(encrypted_input), salt, key)
     end
 
-    attr_accessor :db_encryption_key
+    attr_writer :db_encryption_key
+    attr_writer :database_encryption_keys
+    attr_accessor :current_encryption_key_label
 
     private
+
+    attr_reader :db_encryption_key
+    attr_reader :database_encryption_keys
+
+    def key_to_use(label)
+      if database_encryption_keys.nil? || !database_encryption_keys.key?(label)
+        return db_encryption_key
+      end
+
+      database_encryption_keys[label]
+    end
 
     def make_cipher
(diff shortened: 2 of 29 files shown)


elenasharma:
Error when running 'cf apps'.  We had pushed dora before applying the
PR.  Then we deployed the PR, and ran 'cf apps' and got this error:

```
{
"message":"Request failed: 500: {"error_code"=>"UnknownError"
 "description"=>"An unknown error occurred."
 "code"=>10001
 "test_mode_info"=>{"description"=>"iv must be 16 bytes"
 "error_code"=>"CF-ArgumentError"

 "backtrace"=>[".../cloud_controller_ng/lib/cloud_controller/encryptor.rb:59:in `iv='"
 ".../cloud_controller_ng/lib/cloud_controller/encryptor.rb:59:in `run_cipher'"
 ".../cloud_controller_ng/lib/cloud_controller/encryptor.rb:33:in `decrypt'"
 ".../cloud_controller_ng/lib/cloud_controller/encryptor.rb:107:in `block in set_field_as_encrypted'"
 ".../cloud_controller_ng/lib/cloud_controller/serializer.rb:7:in `block in serializes_via_json'"
 "/var/vcap/packages/cloud_controller_ng/cloud_controller_ng/app/models/runtime/process_model.rb:431:in `environment_json'"
 "/var/vcap/packages/cloud_controller_ng/cloud_controller_ng/app/models/runtime/process_model.rb:525:in `to_hash'"
 ".../cloud_controller_ng/app/controllers/runtime/space_summaries_controller.rb:48:in `block in app_summary'"
 ".../cloud_controller_ng/app/controllers/runtime/space_summaries_controller.rb:40:in `collect'"
 ".../cloud_controller_ng/app/controllers/runtime/space_summaries_controller.rb:40:in `app_summary'"
 ".../cloud_controller_ng/app/controllers/runtime/space_summaries_controller.rb:33:in `space_summary'"
 ".../cloud_controller_ng/app/controllers/runtime/space_summaries_controller.rb:15:in `summary'"
```

lib/cloud_controller/encryptor.rb
```diff
@@ -64,25 +104,42 @@ def encrypt(field_name, options={})
...
 
-          encrypted_value =
-            if value.blank?
-              nil
+          encrypted_value = nil
+
+          if value.blank?
+            encrypted_value = nil
+          else
+            if !VCAP::CloudController::Encryptor.current_encryption_key_label.nil? && send(:key_label) != VCAP::CloudController::Encryptor.current_encryption_key_label
```
tcdowney: @ScarletTanager why do we have to call `send(:key_label)` instead of just calling the method directly?

db/migrations/20170830183100_add_key_label_column_to_tables_with_encrypted_columns.rb
```diff
@@ -0,0 +1,32 @@
...
+      apps
+      buildpack_lifecycle_buildpacks
+      buildpack_lifecycle_data
+      droplets
+      packages
+      tasks
+    )
+    tables_to_alter.each do |table|
+      alter_table(table.to_sym) do
+        add_column :key_label, String
```
tcdowney: We were thinking that the name `key_label` on its own might not be very meaningful to future devs (and current devs not familiar with this work).  Maybe we prefix it with `encryption_`? So the column would be `encryption_key_label` or `encryption_key_used` or something even better of your own choosing 😸 .

spec/support/shared_examples/models/encrypted_attribute.rb
```diff
@@ -65,7 +65,7 @@ def last_row
     end
 
     it 'must have a salt of length 8' do
```
tcdowney: Update the description to 16? 😈

db/migrations/20170830183100_add_key_label_column_to_tables_with_encrypted_columns.rb
ScarletTanager: @tcdowney Yeah, totally reasonable.

spec/support/shared_examples/models/encrypted_attribute.rb
ScarletTanager: Whoops.

db/migrations/20170830183100_add_key_label_column_to_tables_with_encrypted_columns.rb
ScarletTanager: @elenasharma So you had an existing deployment, you applied the PR as a patch, then pushed?  Roughly correct?  Just want to make sure I test the right scenario.

ScarletTanager: @elenasharma @tcdowney I know what's going on.  We bumped the cipher length from 128 to 256, but we didn't test for stored 8 byte salt values (256 requires an IV of 16 bytes), which is what you ran into.  I think that, for the time being, we'll back off to 128, then we can figure out a way to move to AES-256 down the road (which we do want to do).

ScarletTanager: Nope.  Using the correct method for computing the key (what we're doing now, instead of the old way does require a 16 bit iv, unfortunately.  We'll have to think of a way to handle legacy data.

lib/cloud_controller/encryptor.rb
ScarletTanager: Seems like we don't.  I was just being consistent with the way the other property accessors were being called.  Changing it to `self.encryption_key_label` instead.

db/migrations/20170830183100_add_key_label_column_to_tables_with_encrypted_columns.rb
ScarletTanager: Got it fixed.

ScarletTanager:
@tcdowney @elenasharma Commit pushed with requested changes.

ScarletTanager:
@Gerg @tcdowney @elenasharma I am fine with you all merging this without me (if you think the code is okay).  Please don't block on my availability.

matt-royal:
@ScarletTanager and I paired on resolving the merge conflict and I pushed it to master directly. Here's the merge commit for that work: `34e71cb183ac0ce952eddd6dd14692a6b6f9720e`

PR #915, opened 2017-09-26
Add internal-routes to the route info.

tylerschultz:
This adds internal routes to the routing_info. This information is generated for each app that is pushed and will be used to provide platform assigned names for service discovery.

[#151332695]

Signed-off-by: Angela Chin <achin@pivotal.io>


tylerschultz:
I've updated my profile to show I'm in the CF org, so I am CLA compliant.

elenasharma:
Merged!

PR #922, opened 2017-09-29
Delete roles in Perm when deleting by username

iredelmeier:
[Finishes #150978621, #150978620, #150978619]


elenasharma:
Hi @iredelmeier - sending this back to you because when we run our unit tests, we get hundreds of `500` errors instead of successful responses. It seems that there's likely one cause for these failures.

iredelmeier:
@elenasharma sorry about that The Gemfile.lock had a newer version of the client. This is what happens when you have other things going on locally and then push right before retro ;) Should be sorted out now!

PR #930, opened 2017-10-06
Add current_user_guid to service unbind

suhlig:
Fixes [#151431663](https://www.pivotaltracker.com/n/projects/2105761/stories/151431663). This PR is based on a collaboration with the SAPI team (@samze).

### DIFF
--- a/lib/services/service_brokers/v2/http_client.rb
+++ b/lib/services/service_brokers/v2/http_client.rb
@@ -2,6 +2,12 @@ require 'httpclient'
 
 module VCAP::Services
   module ServiceBrokers::V2
+    module IdentityEncoder
+      def self.encode(user_guid)
+        "cloudfoundry #{Base64.strict_encode64({ user_id: user_guid }.to_json)}"
+      end
+    end
+
     class HttpResponse
       attr_reader :code, :message, :body
 
@@ -86,22 +92,22 @@ module VCAP::Services
       end
 
       def get(path)
-        make_request(:get, uri_for(path), nil, nil)
+        make_request(:get, uri_for(path))
       end
 
       def put(path, message)
-        make_request(:put, uri_for(path), message.to_json, 'application/json')
+        make_request(:put, uri_for(path), message.to_json, content_type: 'application/json')
       end
 
       def patch(path, message)
-        make_request(:patch, uri_for(path), message.to_json, 'application/json')
+        make_request(:patch, uri_for(path), message.to_json, content_type: 'application/json')
       end
 
-      def delete(path, message)
+      def delete(path, message, user_guid=nil)
         uri = uri_for(path)
         uri.query = message.to_query
 
-        make_request(:delete, uri, nil, nil)
+        make_request(:delete, uri, nil, user_guid: user_guid)
       end
 
       private
@@ -112,7 +118,7 @@ module VCAP::Services
         URI(url + path)
       end
 
-      def make_request(method, uri, body, content_type)
+      def make_request(method, uri, body=nil, options={})
         client = HTTPClient.new(force_basic_auth: true)
         client.set_auth(uri, auth_username, auth_password)
         client.ssl_config.set_default_paths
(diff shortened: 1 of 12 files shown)


lib/services/service_brokers/v2/http_client.rb
```diff
@@ -86,22 +92,22 @@ def initialize(attrs, logger=nil)
       end
 
       def get(path)
-        make_request(:get, uri_for(path), nil, nil)
+        make_request(:get, uri_for(path))
       end
```
ljfranklin: @suhlig prior to this PR, we were setting the `HEADER_BROKER_API_ORIGINATING_IDENTITY` for every HTTP method (`GET`, `PUT`, `PATCH`, `DELETE`), but this PR only sets the user context for `DELETE`. Was this intentional?

suhlig: I refactored the `make_request` method to default to `body = nil` and take additional options (incl. the former `content_type`); all defaulting to `nil`, too. The reason for this refactoring was either Rubocop or Codeclimate (don't remember which one). I believe the resulting behaviour is the same.

suhlig:
I should have been more clear in the initial description of this PR: For calling with `async=true`, this PR fixes the passing of `HEADER_BROKER_API_ORIGINATING_IDENTITY` only for `DELETE`.

lib/services/service_brokers/v2/http_client.rb
ljfranklin: @suhlig in the previous implementation, `make_request` would try to set the `HEADER_BROKER_API_ORIGINATING_IDENTITY` for `GET`, `PUT`, `PATCH`, and `DELETE` by pulling it off the `VCAP::CloudController::SecurityContext`.

While I like the change to instead pass in the user guid prior to enqueuing the job, I'm concerned that the new code no longer sets the `HEADER_BROKER_API_ORIGINATING_IDENTITY` header for `GET`, `PUT`, or `PATCH`. Seems like the user guid should also be passed to the other HTTP methods, not just `DELETE`.

ljfranklin: @suhlig @Samze, this story is still towards the top of our backlog. We can get it merged once you've addressed the above comment.

suhlig: I _think_ the implementation of `make_request` in this PR behaves the same as before.

*   In master, we always read the `user_guid` from the security context:

    ```ruby
    user_guid = VCAP::CloudController::SecurityContext.current_user_guid
    opts[:header][VCAP::Request::HEADER_BROKER_API_ORIGINATING_IDENTITY] = originating_identity(user_guid) if user_guid
    ```

*   In this PR, we read `user_guid` from `options[:user_guid]`, and if that one is nil, we fall back to the security context as before (see the implementation of the `user_guid` method below):

    ```ruby
    user_guid = user_guid(options)
    opts[:header][VCAP::Request::HEADER_BROKER_API_ORIGINATING_IDENTITY] = IdentityEncoder.encode(user_guid) if user_guid
    ...
    def user_guid(options)
      if options[:user_guid]
        options[:user_guid]
      else
        VCAP::CloudController::SecurityContext.current_user_guid
      end
    end
    ```

Perhaps I am overlooking something else? I am happy to jump on a screen sharing session in your morning.

Samze: @ljfranklin sorry for the delay. The [previous pr](https://github.com/cloudfoundry/cloud_controller_ng/pull/877) added the originating header to outgoing broker requests. It achieved this by reading the user information from the SecurityContext. It purposefully did not send user information when a task was asynchronous, as we considered it to be the "platform" making the request for things like last operation.

However, the async=true flag for unbind service creates an async job which would not have the user information. @suhlig's PR adds this info explicitly on the call to cover the async case.

Deprovision has the same issue which we discussed addressing in a different PR.

ljfranklin:
@suhlig whoops, not sure what I was seeing when I first reviewed the PR. Thanks for the clarification, merging now!

PR #932, opened 2017-10-11
Share a service instance into another space

Samze:
As an app dev (sharer), I can share a service instance into another space [#150801596](https://www.pivotaltracker.com/story/show/150801596)

As an auditor, I see an event when a service instance is shared [#150801603](https://www.pivotaltracker.com/story/show/150801603)

## Why

The services api team have been working on service instance sharing feature as documented in these
[inception notes](https://docs.google.com/document/d/1BYLonYwsMLyjQG0FM8Q-qtoCIpkCz1OUX87odpNKl6Y/edit#heading=h.7tqm9trjzejv). 

This feature will allow space developers to share service instances into other spaces for binding. You can see a broad summary in [these slides](https://docs.google.com/presentation/d/1NMGK5dzW7WEqJ5bUjX6uXPRThb1v-oRthQFHaphpCLE/edit#slide=id.gd3f84e0d5_3_52).

## What

This first PR simply adds the ability for a source space developer to create the sharing relationship into a target space. To perform the share the developer must be a space developer in both source and target space.

From the the [API design doc](https://docs.google.com/document/d/1V5m7L6TPMdR1phRWqzjzCoQIlw44FWZHjKxBsDjUaQQ/edit#):

```
POST /v3/service_instances/:guid/relationships/shared_spaces/

'{ “data”: [ {“guid”: “space-guid”} ] }'
Response
200 Created
{
  "data": [
    {
      "guid": "space-guid"
    }
  ],
  "links": {...}
}
```

Changes:
* New service-instance-sharing feature flag
* New /v3/service_instances/:guid/shared_spaces endpoint
* Migration to support shares
* New audit event
* Updated v3 docs & v2 event docs
* We removed an unused shared context in feature_flags_api_spec.rb

1. Unsharing
1. Visibility of the share
1. Binding to a shared service
1. Name collisions
1. Excluding user-provided-services

## Notes
1. Our new migration uses `on_delete: cascade`. So a share will be deleted when a space/service instance is removed. We noticed that typically [before_destroy](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/models/runtime/route.rb#L175-L183) is used in the code. Is our use of `on_delete: cascade` ok?
1. We created a new generic `to_many_relationship_message.rb` for service sharing. We believe this could be used in other areas of the code, i.e. [isolation_segment_relationship_org_message](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/messages/isolation_segments/isolation_segment_relationship_org_message.rb). To keep the PR small we did not include this refactor.
1. We are extending CATs tests in our [fork](https://github.com/cloudfoundry-incubator/cf-acceptance-tests-sapi/tree/service-instance-sharing).

Feedback appreciated!

## PR 

Sam


Gerg:
I think PR size is something we are going to iterate on going forward. We will review this PR as is, but  I think this could have been decomposed into separate PRs for the feature flag, the core logic, and the audit event.

spec/unit/controllers/v3/service_instance_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+        post :share_service_instance, service_instance_guid: service_instance.guid, body: req_body
+        expect(response.status).to eq 422
+      end
+    end
+
+    describe 'permissions by role' do
+      context 'when the user is a space developer in the source space' do
+        before do
+          set_current_user_as_role(role: 'space_developer', org: source_space.organization, space: source_space, user: user)
+        end
```
ljfranklin: Looks like this context and before are redundant and can be removed.

spec/unit/controllers/v3/service_instance_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+        role_to_expected_http_response = {
+          'admin'               => 200,
+          'space_developer'     => 200,
+          'admin_read_only'     => 403,
+          'global_auditor'      => 403,
+          'space_manager'       => 403,
+          'space_auditor'       => 403,
+          'org_manager'         => 403,
+          'org_auditor'         => 422,
+          'org_billing_manager' => 422,
```
ljfranklin: Why do `org_auditor` and `org_billing_manager` get a 422 when they have a role in the source space but get a 404 in the target space tests below?

spec/unit/controllers/v3/service_instance_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+              end
+            end
+          end
+        end
+      end
+
+      context 'when the user is a space developer in the target space' do
+        before do
+          set_current_user_as_role(role: 'space_developer', org: target_space.organization, space: target_space, user: user)
+        end
```
ljfranklin: Looks like this context and before are redundant and can be removed.

ljfranklin:
Couple small comments / cosmetic changes but otherwise looks good to merge 👍

spec/unit/controllers/v3/service_instance_controller_spec.rb
Gerg: The rule of thumb we follow is: If they don't have access to a resource in the URL, they get a 404. If they don't have access to a resource in the body, they get a 422.

spec/unit/controllers/v3/service_instance_controller_spec.rb
deniseyu: Sharing a service instance requires Space developer access in both a source and target space. The test setups up the user as a space developer in the source and then table tests their access to the target space. The second context does the opposite. 

So we need both these contexts to have a complete set of permutations.

spec/unit/controllers/v3/service_instance_controller_spec.rb
deniseyu: See [comment](https://github.com/cloudfoundry/cloud_controller_ng/pull/932#discussion_r145740706) above

spec/unit/controllers/v3/service_instance_controller_spec.rb
ljfranklin: @deniseyu whoops, yep missed that the line farther down was targeting a different space.

PR #938, opened 2017-10-17
Remove invalid event types from v2 event docs

jenspinney:
The existing v2 docs for events list every event type as a valid value in the response, despite the fact that the documentation is for a querying a specific event type.  This is confusing to doc consumers because they would never actually encounter that variety of event types when issuing the documented query. It also makes it awkward when adding a new event type, since the pattern is to add it to each of these 40 files, even though the new event has nothing to do with the other 40 events.

The "List all Events", "List events associated with an App since January 1, 2014", and "Retrieve a Particular Event" pages were left untouched.

Thanks,
@jenspinney and @Samze

### DIFF
--- a/docs/v2/events/list_app_create_events.html
+++ b/docs/v2/events/list_app_create_events.html
@@ -292,89 +292,12 @@ Cookie: </pre>
                 </td>
                 <td>
                   <ul class="valid_values">
-                      <li>app.crash</li>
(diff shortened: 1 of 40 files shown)


tcdowney:
@zrob does this change seem fine to you?

elenasharma:
Good catch, @jenspinney and @Samze . Thanks! Merging.

PR #944, opened 2017-10-19
Support Multipart messages with CRLF in message parts

bradylove:
Previously, the MultipartParser would split a message part on CRLF. This meant
that if a message part contained a CRLF (which is valid), the parser would
fail to parse correctly.

This commit resolves the issue and supports CRLF in message parts.

What exposed this issue was that Loggregator recently made changes in v100 that applied more tags to envelopes. The protobuf encoding does some checking on the number of tags in an envelopes, does some bit shifting and inserts that data into the envelope. Because of the extra tags `\r` (13) was inserted to the envelope right before a `\n` that is inserted at the beginning of the tags segment of the encoded envelope. This caused the multipart parsing to break in the trafficcontroller multipart parser.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,7 @@ gem 'thin'
 gem 'unf'
 gem 'vcap_common', '~> 4.0.4'
 gem 'yajl-ruby'
+gem 'multipart-parser'
 
 gem 'fog-aws'
 gem 'fog-azure-rm'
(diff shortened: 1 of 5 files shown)


bradylove:
I have fixed the failed checks but do not have permission to push to @enocom's fork. Reopening.

PR #945, opened 2017-10-19
Support Multipart messages with CRLF in message parts

bradylove:
Previously, the MultipartParser would split a message part on CRLF. This meant
that if a message part contained a CRLF (which is valid), the parser would
fail to parse correctly.

This commit resolves the issue and supports CRLF in message parts.

What exposed this issue was that Loggregator recently made changes in v100 that applied more tags to envelopes. The protobuf encoding does some checking on the number of tags in an envelopes, does some bit shifting and inserts that data into the envelope. Because of the extra tags `\r` (13) was inserted to the envelope right before a `\n` that is inserted at the beginning of the tags segment of the encoded envelope. This caused the multipart parsing to break in the trafficcontroller multipart parser.


lib/traffic_controller/client.rb
```diff
@@ -76,31 +77,36 @@ def initialize(body:, boundary:)
...
-        # going to parse part headers
-        delim, _header, body = chunks.next
-        if delim == boundary_delimiter
-          return body
-        end
-        nil
-      rescue StopIteration
+        @chunks ||= parse(@body, @boundary)
+        @chunks.next
+      rescue StopIteration, ParseError
```
elenasharma: @bradylove why is the rescue on `ParseError` needed? Looks like the code treats `truncated multipart message` as a cosmetic error, was this intended?

enocom: We tried in this PR to be as minimally invasive as possible. The `ParseError` attempts to document to a future reader the multipart body was not formatted correctly. If the caller of the `MultipartParser` wanted to handle this error differently, they would at least have a clear place to start.

enocom:
This PR resolves the numerous failures in cf-release CATs revealed by Loggregator v100.

/cc @dsabeti

lisamcho:
Cool thanks!

lib/traffic_controller/client.rb
lisamcho: We'll discuss with team consuming loggregator's gRPC API.

Issue #792, opened 2017-03-15
Service instance permissions should distinguish between read and write


svrc-pivotal:
## Issue (Enhancement request)

Many service instance dashboards are available that provide read-only diagnostic / monitoring information that should be made available to SpaceAuditors.   Per #405 the current permissioning model is a "true/false" where only SpaceDevelopers have permission to service instances {manage: true}.   

## Context

Examples of this enhancement in action: 

*Spring Cloud Services Registry (Eureka)* dashboards contain registered services across a peer replicated mesh.  For debugging/diagnostic purposes,  developers might have SpaceDeveloper Role in their main space but are interoperating with other microservices in other spaces, and should be able to be given SpaceAuditor role in other spaces so they can look at the dashboard of other interoperating peers.

*Spring Cloud Services Circuit Breaker (Hystrix/Turbine)* dashboards are read-only dashboards that should be made available to a broader group (SpaceAuditors) beyond just SpaceDevelopers.
 
*Pivotal SSO* dashboards may declare Oauth resource/scope mappings for microservices that may be browsed by those interoperating with these microservices.

## Steps to Reproduce

1. Create a user with SpaceAuditor role only in a space.
2. Create a service instance with an available dashboard
3. Attempt to browse to that dashboard, it will be rejected

## Expected result

1. Create a user with SpaceAuditor role only in a space.
2. Create a service instance with an available dashboard
3. Attempt to browse to that dashboard, user should be granted read-only rights

## Possible Fix

Introduce a new field to `/v2/service_instances/:guid/permissions` such as `{"read": "true"}` that checks for an auditor role and allows newer service instance dashboards to provide a read-only view of their dashboard.

zrob:
@shalako @avade 

Would we consider this to be part of the broker api spec?

avade:
@zrob No, I think this is a CF permissions issue

@svrc-pivotal when you say "Attempt to browse to that dashboard, it will be rejected", the SSO service is checking against CF's permissions to grant access to the dashboard, correct?

cc. more evidence of permission challenges with the current CF model.

@pivotal-mikegresham 
@matthewmcnew

svrc-pivotal:
@avade I think any service with a dashboard (MySQL, Rabbit, SCS, SSO) is supposed to call `/v2/service_instances/:guid/permissions` to grant access to their dashboard, yeah (this happens after authentication via OAuth).  So you could test this with any service broker that has a dash.

zrob:
[...]

PR #961, opened 2017-10-24
Include shared service instances in /v2/spaces/:guid/service_instances

jenspinney:
As an app dev (receiver), I can see information regarding service instances that have been shared with me. [#152035378](https://www.pivotaltracker.com/story/show/152035378)

## What

This PR changes the `/v2/space/:guid/service_instances` endpoint such that it now includes any services instances that have been shared into the queried space. With this change, the following CLI commands now work with shared services:
* `cf service <shared-service-name>`
* `cf bind-service`
* `cf unbind-service`

Exciting!

### First Important Note
This PR may be a bit more controversial than some of our other recent PRs. We're very interested in your feedback and suggestions if you have ideas for how we could improve our approach. "Weirdness" in summary:
1. Imagine a user is querying `/v2/spaces/:guid/service_instances` for a space where a service instance has been shared into. Imagine also that this user does not have visibility permissions (i.e., space auditor or similar) for the source space of the shared service instance.
1. At the point at which the dataset to retrieve the service instances is actually executed against the database, we use the [secure eager loader](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/cloud_controller/rest_controller/secure_eager_loader.rb). In addition to loading the data for the main model being queried (i.e., service instance), the eager loader also attempts to fill in certain relationship information for the desired resource. In the case of service_instance, the `service_instance.space` relationship is also typically loaded. However, in this scenario, the user does not have visibility access to the service instance's space, so the space information is not loaded and `service_instance.space` will be null for this service instance. The code that does this is [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/cloud_controller/rest_controller/secure_eager_loader.rb#L53). The default visibility filter calls the user_visibility_filter method on the model object, which for this example will return false.
1. The cloud controller code is generally written in a way that assumes `service_instance.space` is never nil. ***We are now breaking this assumption.*** We did an audit of all the places we could find that reference `service_instance.space`, and added nil guarding in every place were we could conceivably run into this scenario. This feels dirty and prone to edge-casey 500s in the future (i.e., if someone adds code later that assumes this value can never be nil and they don't test this narrow scenario). However, loading data that the user doesn't actually have permission to see feels dirty as well and would potentially require some messy surgery of very generic methods. In that case we also have the potential ability to accidentally reveal more information than the user is authorised to see. Despite the problems of letting `service_instance.space` be null, it seems like the better of two evils since this condition will only occur in a very narrow strip of code (i.e., `#enumerate_service_instances` in the v2 space controller) which will probably not be changed dramatically in the future, since it's v2.

### Second Important Note
As part of doing this change, we also noticed the `#read_permissions?` implementation for service instance access seemed to be unnecessarily duplicating the behaviour and intent of `#read?`, but going through a very different codepath. By modifying `#read_permissions?` to use `#read?`, we believe we changed the code to actually be more correct by additionally allowing the global_auditor role to have access to this operation.

The `#read_permissions` access method is used to ultimately tell a service broker whether the active user has sufficient permission to view the service dashboard for a service instance. Based on our understanding of the global_auditor role, it seems to make sense that they would be allowed to view service dashboards. If our understanding is wrong, please let us know :)

## Summary of Changes:
* Backfilled missing unit tests for `service_instance.user_visibility_filter` function.
* Modified `user_visibility_filter` on the `ServiceInstance` model to additionally allow user who have visibility to any space the service instance has been shared into.
* Changed controller method for `/v2/spaces/:guid/service_instance` such that it returns all service instances either created in the given space or shared with the given space.
* Guard against nil values for `service_instance.space` in `service_instance_access.rb` and in `service_instance.rb`.

## PR 

### DIFF
--- a/app/models/services/service_instance.rb
+++ b/app/models/services/service_instance.rb
@@ -138,7 +138,7 @@ module VCAP::CloudController
     alias_method_chain :credentials, 'serialization'
 
     def in_suspended_org?
-      space && space.in_suspended_org?
+      space&.in_suspended_org?
     end
 
     def after_create
--- a/app/models/services/service_instance.rb
+++ b/app/models/services/service_instance.rb
@@ -69,6 +69,9 @@ module VCAP::CloudController
         [:space, user.audited_spaces_dataset],
         [:space, user.managed_spaces_dataset],
         [:shared_spaces, user.spaces_dataset],
+        [:shared_spaces, user.managed_spaces_dataset],
+        [:shared_spaces, user.audited_spaces_dataset],
+        [:shared_spaces, managed_organizations_spaces_dataset(user.managed_organizations_dataset)],
       ])
     end
 
--- a/app/models/services/service_instance.rb
+++ b/app/models/services/service_instance.rb
@@ -68,6 +68,7 @@ module VCAP::CloudController
         [:space, user.spaces_dataset],
         [:space, user.audited_spaces_dataset],
         [:space, user.managed_spaces_dataset],
+        [:shared_spaces, user.spaces_dataset],
       ])
     end
 
--- a/app/models/services/service_instance.rb
+++ b/app/models/services/service_instance.rb
@@ -135,7 +135,7 @@ module VCAP::CloudController
     alias_method_chain :credentials, 'serialization'
 
     def in_suspended_org?
-      space.in_suspended_org?
+      space && space.in_suspended_org?
     end
 
     def after_create
--- a/app/access/service_instance_access.rb
+++ b/app/access/service_instance_access.rb
@@ -1,18 +1,16 @@
 module VCAP::CloudController
   class ServiceInstanceAccess < BaseAccess
     def create?(service_instance, params=nil)
-      return false unless service_instance.space
       return true if admin_user?
       FeatureFlag.raise_unless_enabled!(:service_instance_creation)
       return false if service_instance.in_suspended_org?
-      service_instance.space.has_developer?(context.user) && allowed?(service_instance)
+      service_instance.space && service_instance.space.has_developer?(context.user) && allowed?(service_instance)
     end
 
     def read_for_update?(service_instance, params=nil)
-      return false unless service_instance.space
       return true if admin_user?
       return false if service_instance.in_suspended_org?
-      service_instance.space.has_developer?(context.user)
+      service_instance.space && service_instance.space.has_developer?(context.user)
     end
 
     def update?(service_instance, params=nil)
@@ -20,16 +18,14 @@ module VCAP::CloudController
     end
 
     def delete?(service_instance)
-      return false unless service_instance.space
       return true if admin_user?
       return false if service_instance.in_suspended_org?
-      service_instance.space.has_developer?(context.user)
+      service_instance.space && service_instance.space.has_developer?(context.user)
     end
 
     def manage_permissions?(service_instance)
-      return false unless service_instance.space
       return true if admin_user?
-      service_instance.space.has_developer?(context.user)
+      service_instance.space && service_instance.space.has_developer?(context.user)
     end
 
     def manage_permissions_with_token?(service_instance)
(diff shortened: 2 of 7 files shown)


tcdowney:
@zrob there are some product-facing questions in the description of this SAPI PR. Figured you'd want to weigh-in as well.

app/models/services/service_instance.rb
```diff
@@ -134,7 +138,7 @@ def credentials_with_serialization
     alias_method_chain :credentials, 'serialization'
 
     def in_suspended_org?
-      space.in_suspended_org?
+      space && space.in_suspended_org?
```
tcdowney: @jenspinney @deniseyu  

I'm not sure of all of the places where this is called, but I'm a bit worried that this will allow users to do things with the service instance when the org is suspended that they shouldn't be able to do.

E.g. Org One binds some email sending service to Space A and shares it with Space B in Org Two. If Org One is spamming people with their email sending service and is suspended will this mean that Org Two can keep on making changes to the service instance and using it since `space` will be nil and always return `false`?

jenspinney: Thanks for the feedback, @tcdowney!

This method appears to only be called in [service_instance_access.rb](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/blob/service-instance-sharing/app/access/service_instance_access.rb).

As a general rule, if space is `nil`, the user may only do read-ish things (nothing involving modification, deletion, etc.) on the service instance because they don't have access to the originating space of the service instance. The places where we check whether the org is suspended are modification-type access checks, so we'll fail those anyway because space is nil. In our version of service_instance_access, you'll see that wherever this method is called, we've made changes to also explicitly check if the space is available. 

If we reversed this and returned true for `service_instance.in_suspended_org?` when space is nil, it seems counterintuitive that it would be asserting that the org of the service instance is suspended even though we know nothing about the originating org. I would imagine if `service_instance.in_suspended_org?` returned true, I could return some error message like, "The org you're in has been suspended", which is not true in this case.

Thanks!
Jen and @deniseyu

Gerg:
@tcdowney What were the product questions you wanted Zach to address?

@jenspinney I don't have strong feelings about the nil guarding for `service_instance.space`. The pair that reviews this next can check if there is a better way to encapsulate the the space maybe-not-being-there. 

For the second important note, it looks like https://github.com/cloudfoundry/cloud_controller_ng/issues/792 introduced `#read_permissions?`. It might have been an oversight to not allow global auditors, but worth digging into.

Gerg:
If you want to be extra-ruby, you can also use the new [safe navigation operator](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/). Instead of `foo && foo.bar`, you can now do `foo&.bar`.

jenspinney:
Let's be extra ruby! Thanks for the info @Gerg  :)

Updated the PR with `&.`'s.

And yes, we assumed the change to address #792 mistakenly left out global auditors. (If it was intentional though, we can unwind that change.)

tcdowney:
@Gerg @jenspinney 
> And yes, we assumed the change to address #792 mistakenly left out global auditors. (If it was intentional though, we can unwind that change.)

That's the product question we had for @zrob. Sorry for not making that very clear.

Gerg:
From our discussion with SAPI this morning/evening, it sounds like we want to break the global auditor permissions change into a separate endpoint.

deniseyu:
We have rolled back the global auditors read-permissions. I've also rebased the branch against the latest upstream. Please let us know if you have further feedback!

app/models/services/service_instance.rb
```diff
@@ -68,6 +68,10 @@ def self.user_visibility_filter(user)
         [:space, user.spaces_dataset],
         [:space, user.audited_spaces_dataset],
         [:space, user.managed_spaces_dataset],
+        [:shared_spaces, user.spaces_dataset],
+        [:shared_spaces, user.managed_spaces_dataset],
+        [:shared_spaces, user.audited_spaces_dataset],
+        [:shared_spaces, managed_organizations_spaces_dataset(user.managed_organizations_dataset)],
```
matt-royal: We dug into this method, and realized that it writes very inefficient SQL (involving many redundant subselects). Your PR didn't introduce this problem, it just exacerbated it by adding more subselects. We have a dedicated story to address this problem so this won't prevent us from merging this PR: https://www.pivotaltracker.com/story/show/152736102

jenspinney: Good spot! Thanks for making a separate story for it :)

PR #965, opened 2017-10-25
Implement DB encryption key rotation.

ScarletTanager:
Implement DB encryption key rotation.  Includes reviewer-requested changes and keeps the 128 bit algo (as opposed to moving to 256).


elenasharma:
Thanks, @ScarletTanager !

PR #976, opened 2017-11-08
Add /v2/service_instances/:guid/shared_from

deniseyu:
As an app dev (receiver), I can see sharing information regarding service instances that have been shared with me (shared_from). [#150973038](https://www.pivotaltracker.com/story/show/150973038)

**NOTE**: This PR builds on top of #972, which should be merged first. The actual changes on top of #972 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-service-instance-sharing-v3-service-instances...cloudfoundry-incubator:pr-service-instance-sharing-shared-from-url).

## What

This PR adds `/v2/service_instances/:guid/shared_from`. This is required by the CLI in order to display the originating space and org of a shared service instance when a user would otherwise not have permission to view those names.

Changes:
* All service_instance response bodies have shared_from_url
* /v2/service_instances/:guid/shared_from endpoint returns space name and org name of service
instance, if that service instance has been shared
* If not shared, this endpoint returns 204 No Content

## PR 

### DIFF
--- a/app/access/service_instance_access.rb
+++ b/app/access/service_instance_access.rb
@@ -1,18 +1,16 @@
 module VCAP::CloudController
   class ServiceInstanceAccess < BaseAccess
     def create?(service_instance, params=nil)
-      return false unless service_instance.space
       return true if admin_user?
       FeatureFlag.raise_unless_enabled!(:service_instance_creation)
       return false if service_instance.in_suspended_org?
-      service_instance.space.has_developer?(context.user) && allowed?(service_instance)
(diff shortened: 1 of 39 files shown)


jenspinney:
A note on the codeclimate failure: It's complaining that the shared_from_information method is too "cognitively complex". It could be that we are simply cognition masters, but the function does not appear difficult to understand. We'd suggest dismissing the warning.

PR #978, opened 2017-11-08
Add /v2/service_instances/:guid/shared_to

jenspinney:
As an app dev (sharer), I can see sharing information regarding service instances that I have shared (shared_to). [#151432798](https://www.pivotaltracker.com/story/show/151432798)

**NOTE**: This PR builds on top of #976, which should be merged first. The actual changes on top of #976 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-service-instance-sharing-shared-from-url...cloudfoundry-incubator:pr-service-instance-sharing-shared-to-url).

## What

This PR adds `/v2/service_instances/:guid/shared_to`. This is required by the CLI in order to display the space and org names where a service instance has been shared into. Names are used instead of guids because a dev in the service instance space may not normally have access to read the name of the space/org that the instance has been shared into.

Changes:
* All service_instance response bodies have shared_to_url
* /v2/service_instances/:guid/shared_to endpoint returns a paginated list of space and org names

## PR 

### DIFF
--- a/app/access/service_instance_access.rb
+++ b/app/access/service_instance_access.rb
@@ -1,18 +1,16 @@
 module VCAP::CloudController
   class ServiceInstanceAccess < BaseAccess
     def create?(service_instance, params=nil)
-      return false unless service_instance.space
       return true if admin_user?
       FeatureFlag.raise_unless_enabled!(:service_instance_creation)
       return false if service_instance.in_suspended_org?
-      service_instance.space.has_developer?(context.user) && allowed?(service_instance)
+      service_instance.space && service_instance.space.has_developer?(context.user) && allowed?(service_instance)
     end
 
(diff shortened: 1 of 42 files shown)


ericpromislow:
Weird that code_climate complained about `shared_from_information` but not `enumerate_shared_to_information`, which is obviously more complex.  One thing though --
 we would rewrite 
```rescue CloudController::Errors::ApiError => e
if e.name == 'NotAuthorized'
  HTTP::NOT_FOUND
else
 raise e
end 
```
as 
```rescue CloudController::Errors::ApiError => e
return HTTP::NOT_FOUND if e.name == 'NotAuthorized'
raise
```
When you `raise` without the current exception variable you don't need to provide it.

PR #980, opened 2017-11-08
Add Validation that Service Instance Name Doesn't Clash with Shared Instances

jenspinney:
As an app dev (receiver), I shouldn't be able to create a service instance with the same name as an instance that has been shared with me. [#152314627](https://www.pivotaltracker.com/story/show/152314627)

**NOTE**: This PR builds on top of #979, which should be merged first. The actual changes on top of #979 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-service-instance-sharing-bound-app-count...cloudfoundry-incubator:pr-service-instance-sharing-name-collisions).

## What
This change adds validation to protect against a user seeing multiple service instances with the same name. When a user creates a service instance, we now check that it doesn't clash with service instances that have been shared into the active space.

This PR doesn't address the name collision in the opposite direction (i.e., someone shares a service instance into a space where that instance name is already taken). We have a separate story to address that: [#150801605](https://www.pivotaltracker.com/story/show/150801605). We also plan to block the ability to rename shared service instances in a different future story ([#152593218](https://www.pivotaltracker.com/story/show/152593218)).

Changes:
* Modified and expanded validation code for service instance model to include the restriction described above.

## PR 

Thanks, sapi (@jenspinney and @Samze)

### DIFF
--- a/spec/unit/controllers/services/service_instances_controller_spec.rb
+++ b/spec/unit/controllers/services/service_instances_controller_spec.rb
@@ -885,6 +885,32 @@ module VCAP::CloudController
             expect(last_response.status).to eq(400)
             expect(decoded_response['code']).to eq(60002)
           end
+
+          it 'does not allow a managed service instance with same name as a shared service instance' do
+            source_space = Space.make(organization: space.organization)
+            source_space.add_developer(developer)
+            service_instance = create_managed_service_instance(accepts_incomplete: 'false', space: source_space)
+            expect(last_response.status).to eq(201)
+
+            service_instance.add_shared_space(space)
+
+            create_managed_service_instance
+            expect(last_response.status).to eq(400)
+            expect(decoded_response['code']).to eq(60002)
+          end
+
+          it 'does not allow a user provided service instance with same name as a shared service instance' do
+            source_space = Space.make(organization: space.organization)
+            source_space.add_developer(developer)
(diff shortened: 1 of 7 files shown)


spec/unit/controllers/services/service_instances_controller_spec.rb
```diff
@@ -885,6 +885,53 @@ def stub_delete_and_return(status, body)
             expect(last_response.status).to eq(400)
             expect(decoded_response['code']).to eq(60002)
           end
+
+          context 'when a service instance share exists between spaces' do
```
anniesing: We got a bit confused initially reading through this because of the var name `space` since there's more than one space being referred to. We thought the var name `source_space` was great, so perhaps something like `target_space` would be more descriptive as well.

ablease: Hey @anniesing 

Thanks for noticing this, we totally agree that it would make more sense to refer to `target_space`.

Would you accept a PR later that addresses this? We are happy to create a chore in our backlog to look at this, and other test files we have changed, that might have vague variable names. 

Thanks, 

Sapi, (@deniseyu, alex)

ericpromislow: A future PR sounds great.

PR #981, opened 2017-11-08
App Dev Sharers Cannot Delete Shared Service Instances

Samze:
As an app dev (sharer), I cannot delete a service instance that I have shared which does not have bindings in another space [#152470931](https://www.pivotaltracker.com/story/show/152470931)
As an app dev (sharer), I cannot delete a service instance that I have shared which has bindings in another space[#151710830](https://www.pivotaltracker.com/story/show/151710830)

**NOTE**: This PR builds on top of #980, which should be merged first. The actual changes on top of #980 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-service-instance-sharing-name-collisions...cloudfoundry-incubator:pr-service-instance-sharing-deleting-shared-instances).

## What

This PR disables the deletion of service instances that have been shared to other spaces. This is similar validation to service instances that have bindings or have service keys. Users should unshare the service before issuing a delete. Note that if the service instance is deleted with the recursive query parameter then this validation is skipped.

Changes:
* Added to the validation in the service_instance_controller
* Added new user error message

## PR 

Thanks, sapi (@jenspinney and @Samze)

### DIFF
--- a/spec/unit/controllers/services/service_instances_controller_spec.rb
+++ b/spec/unit/controllers/services/service_instances_controller_spec.rb
@@ -2382,8 +2382,30 @@ module VCAP::CloudController
 
             expect(last_response).to have_status_code 400
             expect(last_response.body).to include 'ServiceIsShared'
-            expect(last_response.body).to include
-            'Service instances must be unshared before they can be deleted'
+            expect(last_response.body).to include(
+              'Service instances must be unshared before they can be deleted. ' \
+              "Unsharing #{service_instance.name} will automatically delete any bindings " \
+              'that have been made to applications in other spaces')
+          end
+
+          context 'and there are bindings to the shared instance' do
+            before do
+              ServiceBinding.make(
+                app: AppModel.make(space: space),
+                service_instance: service_instance
+              )
+            end
+
+            it 'should give the user an error' do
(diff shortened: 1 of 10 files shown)


spec/unit/controllers/services/service_instances_controller_spec.rb
```diff
@@ -885,6 +885,53 @@ def stub_delete_and_return(status, body)
...
             expect(decoded_response['code']).to eq(60002)
           end
+
+          context 'when a service instance share exists between spaces' do
+            let(:source_space) { Space.make(organization: space.organization) }
+            before do
+              source_space.add_developer(developer)
+
+              service_instance = create_managed_service_instance(accepts_incomplete: 'false', space: source_space)
+              service_instance.add_shared_space(space)
```
anniesing: **Copy from PR 980**: We got a bit confused initially reading through this because of the var name space since there's more than one space being referred to. We thought the var name source_space was great, so perhaps something like target_space would be more descriptive as well.

ablease: Hey @anniesing,

We replied to this, over on #980 

Thanks

alex and @deniseyu

PR #993, opened 2017-11-14
Prevent renaming a service instance to the name of one that has been shared into space

ablease:
As an app dev (receiver), I cannot rename a service instance to the same name of a service instance that has been shared with me. [#152848428](https://www.pivotaltracker.com/story/show/152848428)

**NOTE**: This PR builds on top of #992, which should be merged first. The actual changes on top of #992 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/cloudfoundry-incubator:pr-service-instance-sharing-no-sharing-to-own-space...cloudfoundry-incubator:pr-service-instance-sharing-no-renaming-to-shared-instance-name).

## What

This PR adds only some tests to lock in a workflow that is already enabled by [a previous story, #152314627](https://www.pivotaltracker.com/story/show/152314627).

## PR 

* [X] I have viewed signed and have submitted the Contributor License Agreement
* [X] I have made this pull request to the `master` branch
* [X] I have run all the unit tests using `bundle exec rake`
* [X] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats) on bosh lite

Thanks, sapi

PR #994, opened 2017-11-14
Users cannot share into a space where an existing service instance name would clash

ablease:
As an app dev (sharer), if I try to share a service instance into a space where a service instance exists with the same name, the share fails. [#152853193](https://www.pivotaltracker.com/story/show/152853193)

**NOTE**: This PR builds on top of #993, which should be merged first. The actual changes on top of #993 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-service-instance-sharing-no-renaming-to-shared-instance-name...cloudfoundry-incubator:pr-service-instance-sharing-prevent-sharing-if-will-cause-name-clash).

## What

This PR closes an edge case where a user could share `r1` into another space where a service instance called `r1` independently exists.

## Changes

* A 400 error will be returned with the message "A service instance called <service instance name> already exists in <space name>"
* Minor refactoring to service instance share action

## Why

The CLI will reconcile service names into GUIDs for the API call, so we need to prevent the case where multiple service instances with the same name exist across sharing spaces. If two spaces never interact via a service instance sharing relationship, they are still free to use the same names for their service instances.

## PR 

Thanks, sapi

### DIFF
--- a/app/actions/service_instance_share.rb
+++ b/app/actions/service_instance_share.rb
@@ -3,6 +3,12 @@ require 'repositories/service_instance_share_event_repository'
 module VCAP::CloudController
   class ServiceInstanceShare
     def create(service_instance, target_spaces, user_audit_info)
+      if service_instance.managed_instance?
+        unless service_instance.shareable?
+          raise CloudController::Errors::ApiError.new_from_details('ServiceShareIsDisabled', service_instance.service.label)
+        end
+      end
+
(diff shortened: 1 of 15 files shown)


elenasharma:
Hi @ablease - we accidentally merged this PR. Could you please open a new PR with the same changes?

PR #1024, opened 2017-12-05
Allow unsharing from a space without org visibilty

ablease:
As an app dev (sharer), I can unshare a service instance from a space where I do not have access.
[#153244755](https://www.pivotaltracker.com/story/show/153244755)

**NOTE**: This PR builds on top of #1023, which should be merged first. The actual changes on top of #1023 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-service-instance-sharing-fix-test-descriptions...cloudfoundry-incubator:pr-service-instance-sharing-shared-to-without-org-visibilty).

This PR address a bug that was discovered after this [PR](https://github.com/cloudfoundry/cli/pull/1281) was made to the CLI. When attempting to unshare a service instance from a space where the current user did not have access to that space, the unshare failed. 

Thanks, 

Sapi team (alex, @deniseyu, @jenspinney)

### DIFF
--- a/app/controllers/services/service_instances_controller.rb
+++ b/app/controllers/services/service_instances_controller.rb
@@ -122,6 +122,7 @@ module VCAP::CloudController
       validate_access(:read_for_update, service_instance)
       validate_access(:update, projected_service_instance(service_instance))
 
+      validate_name_update(service_instance)
       validate_space_update(related_objects[:space])
       validate_plan_update(related_objects[:plan], related_objects[:service], service_instance)
 
@@ -430,6 +431,14 @@ module VCAP::CloudController
 
     def validate_space_update(space)
       space_change_not_allowed! if space_change_requested?(request_attrs['space_guid'], space)
+    end
+
+    def validate_name_update(service_instance)
+      return unless request_attrs['name'] && service_instance.shared?
+
+      if request_attrs['name'] != service_instance.name
+        raise CloudController::Errors::ApiError.new_from_details('SharedServiceInstanceCannotBeRenamed')
+      end
     end
 
     def invalid_service_instance!(service_instance)
--- a/app/controllers/services/service_instances_controller.rb
+++ b/app/controllers/services/service_instances_controller.rb
@@ -361,6 +361,12 @@ module VCAP::CloudController
 
     private
 
+    class ServiceInstanceSharedToEagerLoader
+      def eager_load_dataset(spaces, _, _, _, _)
+        spaces.eager(:organization)
+      end
+    end
+
     class ServiceInstanceSharedToSerializer
       def initialize(service_instance)
         @service_instance = service_instance
@@ -374,7 +380,7 @@ module VCAP::CloudController
 
     def create_paginated_collection_renderer(service_instance)
       VCAP::CloudController::RestController::PaginatedCollectionRenderer.new(
-        VCAP::CloudController::RestController::SecureEagerLoader.new,
+        ServiceInstanceSharedToEagerLoader.new,
         ServiceInstanceSharedToSerializer.new(service_instance),
         {
           max_results_per_page: config.get(:renderer, :max_results_per_page),
--- a/spec/unit/controllers/services/service_instances_controller_spec.rb
+++ b/spec/unit/controllers/services/service_instances_controller_spec.rb
@@ -1742,10 +1742,27 @@ module VCAP::CloudController
             service_instance.add_shared_space(shared_to_space)
           end
 
-          context 'and a developer in the originating space tries to update the instance' do
+          context 'and a developer in the originating space tries to update the instance without renaming' do
             it 'updates successfully' do
               put "/v2/service_instances/#{service_instance.guid}", body
               expect(last_response).to have_status_code 201
+            end
+          end
+
+          context 'and a developer in the originating space tries to rename the instance' do
+            let(:body) do
+              {
+                name: 'dont-rename-me',
+                tags: []
+              }.to_json
+            end
+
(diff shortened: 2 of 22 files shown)


app/controllers/services/service_instances_controller.rb
```diff
@@ -347,6 +348,12 @@ def unbind_route(route_guid, instance_guid)
 
     private
 
+    class ServiceInstanceSharedToEagerLoader
```
Gerg: This class will not be private unless you do 
```rb
private_constant :ServiceInstanceSharedToEagerLoader
```

app/controllers/services/service_instances_controller.rb
```diff
@@ -360,7 +367,7 @@ def serialize(controller, space, opts, orphans=nil)
 
     def create_paginated_collection_renderer(service_instance)
       VCAP::CloudController::RestController::PaginatedCollectionRenderer.new(
-        VCAP::CloudController::RestController::SecureEagerLoader.new,
+        ServiceInstanceSharedToEagerLoader.new,
```
Gerg: This feels like a pretty fundamental change to service instances controller. Subverting the framework and how it expects to load associations could lead to confusing behavior later. I'm surprised this didn't break any existing tests.

Is there another way we could configure the framework to eager load organizations in addition to it's usual machinations?

app/controllers/services/service_instances_controller.rb
deniseyu: Thanks, we've made this change.

app/controllers/services/service_instances_controller.rb
deniseyu: The `create_paginated_collection_renderer` method, the `ServiceInstanceSharedToEagerLoader` and the `ServiceInstanceSharedToSerializer` are only used by the '/v2/service_instances/:guid/shared_to' endpoint. These are used to support paginated rendering using the custom `ServiceInstanceSharedToPresenter` presenter. We could not use the existing serialiser class as SharedTo does not have an associated model object. The eager loader is overridden in order to load the organisation model that is required by the presenter to display the org name.

These changes are confined to the `shared_to` endpoint, so it hasn't changed behaviour elsewhere in the controller.

app/controllers/services/service_instances_controller.rb
Samze: Changes for this included in #1033 and can be viewed here: https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/commit/ff8276e5d201e573df03640a0f6865288b2d55af#diff-9b86a08059c9bdc2b6875a70abf9ee29R366

spec/unit/controllers/services/service_instances_controller_spec.rb
```diff
@@ -4001,6 +4023,12 @@ def verify_forbidden(user)
 
       describe 'permissions' do
         let(:user) { User.make }
+        let(:other_org) { Organization.make }
```
elenasharma: Per the conversation that led to PR #1023, please prefer `target_org` over `other_org`; same for space.

spec/unit/controllers/services/service_instances_controller_spec.rb
```diff
@@ -4026,20 +4054,13 @@ def verify_forbidden(user)
 
               it "has a #{expected_status} http status code" do
                 get "/v2/service_instances/#{instance.guid}/shared_to"
-                expect(last_response.status).to eq(expected_status), "Expected #{expected_status}, got: #{last_response.status}, role: #{role}"
+                expect(last_response.status).to eq(expected_status), "Expected #{expected_status}, got: #{last_response.status}, role: #{role}, response: #{last_response.body}"
```
elenasharma: Is the `last_response.body` part intentional? We often use this during development for extra information, but typically don't commit uses of it.

spec/unit/controllers/services/service_instances_controller_spec.rb
```diff
@@ -4026,20 +4054,13 @@ def verify_forbidden(user)
...
               end
             end
           end
         end
 
         context 'when the user is a member of the org/space where the service instance was shared to' do
```
elenasharma: We'd expect to see a test case for the encountered bug, e.g. given two developers: a developer in `space-1` and `space-2` and a developer in `space-1`:
* the first developer shares service instance `SI` into `space-2`
* the second developer can unshare `SI` out of `space-2`

spec/unit/controllers/services/service_instances_controller_spec.rb
deniseyu: Thanks @elenasharma, it will be easier for us to fix this with a new commit on top rather than rebase everything, would you be open to merging this PR?

spec/unit/controllers/services/service_instances_controller_spec.rb
deniseyu: This was probably unintentional! We will remove it.

spec/unit/controllers/services/service_instances_controller_spec.rb
deniseyu: @elenasharma we're not clear what you mean here. This endpoint is not for unsharing, it's for viewing the space GUIDs where a service instance has been shared to. Developers in recipient spaces shouldn't have access to this, because then they'd see GUIDs for other spaces where they shouldn't normally have access.

We do test this workflow in the v3/controllers unit test for unsharing: https://github.com/cloudfoundry/cloud_controller_ng/blob/master/spec/unit/controllers/v3/service_instances_controller_spec.rb#L480-L505

If we've misunderstood what you meant, please let us know!

PR #1028, opened 2017-12-12
Only show service bindings if requester has read access to bound app

jenspinney:
As an app dev, I can only see bindings in spaces that I have access to for shared service instances [#153456495](https://www.pivotaltracker.com/story/show/153456495)

**NOTE**: This PR builds on top of #1027, which should be merged first. The actual changes on top of #1027 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-service-instance-sharing-allow-unshare-when-feature-flag-disabled...cloudfoundry-incubator:pr-service-instance-sharing-restrict-binding-visibility-to-app-visibility).

## Context
In our initial work for service instance sharing, we allowed service bindings to be viewed by anyone who has access to view the binding's service instance. That means anyone who can see the service instance because it was shared with them can also see the service bindings. We've now decided to restrict this so that a user can only see service bindings if they have read access to the app of the service binding.
This change was motivated by the behavior of the CLI during `cf service <service-name>`, where it reasonably assumes the app guids that come back from `/v2/space/:guid/service_instances/` can be queried for more information about the app.

## Impact
* Users who run the CLI command `cf service <service-instance-name>` for a service instance that has bindings to an app they don't have access to will no longer get a 403 Unauthorized error.
* API consumers who hit `GET /v2/service_bindings`, `GET /v2/space/:guid/service_bindings`, etc., will only see service bindings for which they have read access to the associated app.

## Code Changes:
* Make the service binding user visibility filter use the app's visibility filter, not the service instance's.
* Added a user visibility filter to `AppModel`. This is the same logic as the visibility filter used by the `ProcessModel`. (Since the service binding is attached to an `AppModel`, not a `ProcessModel`, we thought the easiest and cleanest way forward would be to add a user_visibility_filter to the AppModel. If this is a pattern we don't to continue into the v3 models, let us know.)

## PR 

Thanks,
SAPI Team (@jenspinney and @samze)

### DIFF
--- a/app/controllers/services/service_instances_controller.rb
+++ b/app/controllers/services/service_instances_controller.rb
@@ -122,6 +122,7 @@ module VCAP::CloudController
       validate_access(:read_for_update, service_instance)
       validate_access(:update, projected_service_instance(service_instance))
 
+      validate_name_update(service_instance)
       validate_space_update(related_objects[:space])
       validate_plan_update(related_objects[:plan], related_objects[:service], service_instance)
 
@@ -430,6 +431,14 @@ module VCAP::CloudController
 
     def validate_space_update(space)
       space_change_not_allowed! if space_change_requested?(request_attrs['space_guid'], space)
+    end
+
+    def validate_name_update(service_instance)
+      return unless request_attrs['name'] && service_instance.shared?
+
+      if request_attrs['name'] != service_instance.name
+        raise CloudController::Errors::ApiError.new_from_details('SharedServiceInstanceCannotBeRenamed')
+      end
     end
 
     def invalid_service_instance!(service_instance)
--- a/app/controllers/services/service_instances_controller.rb
+++ b/app/controllers/services/service_instances_controller.rb
@@ -361,6 +361,12 @@ module VCAP::CloudController
 
     private
 
+    class ServiceInstanceSharedToEagerLoader
+      def eager_load_dataset(spaces, _, _, _, _)
+        spaces.eager(:organization)
+      end
+    end
+
     class ServiceInstanceSharedToSerializer
       def initialize(service_instance)
         @service_instance = service_instance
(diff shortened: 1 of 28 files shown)


jenspinney:
Note: Travis CI test failure seems like a flake and unrelated to these code changes.

rainmaker:
Hi SAPI friends,

We're looking at your PR right now, and are a little bit confused about the expected initial behavior. You said: 

`In our initial work for service instance sharing, we allowed service bindings to be viewed by anyone who has access to view the binding's service instance. That means anyone who can see the service instance because it was shared with them can also see the service bindings. `

And you also said: 

`Users who run the CLI command cf service <service-instance-name> for a service instance that has bindings to an app they don't have access to will no longer get a 403 Unauthorized error.`

The first explanation implies that our pre-PR users should be able to see the service bindings in this scenario. The second explanation implies that the users should have been receiving a 403. We're seeing the former behavior right now. Your code changes seem clear to us, so we're looking to confirm the acceptance criteria for the feature as well.

Let us know what you think!

Raina and Maryam
CF CAPI Team

jenspinney:
Hi @rainmaker and Maryam,

Sorry for the confusion. The reason for receiving 403 when using the CLI command `cf service <service-name>` in the pre-PR state of the world is because the CLI essentially does the following:
1. Get me all service bindings for this space (i.e., `GET /v2/space/:guid/service_bindings`).
2. For each service binding in the returned list, grab the associated app's guid and look up more detailed information about that app, like the name, etc. (i.e., `GET /v2/apps/:guid`)

In the pre-PR code, step 1 returns all service bindings that are associated with a serivce instance that the user can view (including service instances that they've shared with other spaces where the user doesn't normally have read access). The CLI then tries to look up the details for the apps and returns a 403 error because it can't query the app details (the user doesn't have view permission in that space).

With the code in this PR, step 1 will no longer return those entries at all. The returned list of service bindings will only include bindings where the user should have enough permission to issue a `/v2/apps/:app_guid` request without getting a permissions failure.

Does that answer your question?

Thanks,
Jen

maryamklabib:
@jenspinney Thanks for the context, that's helpful!

PR #1029, opened 2017-12-12
Do not require service_instance_sharing feature flag in order to bind to shared service instances

jenspinney:
As an app dev (receiver), I can continue to bind apps to a service instance that has been shared with me when the feature flag is disabled. [#153438338](https://www.pivotaltracker.com/story/show/153438338)

**NOTE**: This PR builds on top of #1028, which should be merged first. The actual changes on top of #1028 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-service-instance-sharing-restrict-binding-visibility-to-app-visibility...cloudfoundry-incubator:pr-service-instance-sharing-bindable-share-zebra-feature%F0%9F%9A%A9).

## What
Previously, we restricted binding apps to shared service instances such that it was only allowed when the service_instance_sharing feature flag was enabled. For better UX in the situation where the feature flag was enabled, a service instance was shared, and then the feature flag was disabled, we now allow users to bind apps to shared spaces, even when the flag is disabled.

In PR #1027, we removed the requirement that the `service_instance_sharing` feature flag be enabled in order to unshare a service instance. This PR rounds out that user experience, so that users don't see orphaned service instances that may not bind apps to.

Changes:
* Remove feature flag checks at the model and action level when creating service bindings.
* Remove unnecessary setting of the feature flag at various points in the unit tests.

## PR 

Thanks, 
SAPI team (@jenspinney and @samze)

### DIFF
--- a/app/actions/service_binding_create.rb
+++ b/app/actions/service_binding_create.rb
@@ -63,8 +63,7 @@ module VCAP::CloudController
     end
 
     def bindable_in_space?(service_instance, app_space)
-      service_instance.space == app_space ||
-        (FeatureFlag.enabled?(:service_instance_sharing) && service_instance.shared_spaces.include?(app_space))
+      service_instance.space == app_space || service_instance.shared_spaces.include?(app_space)
     end
 
     def logger
--- a/app/controllers/services/service_instances_controller.rb
+++ b/app/controllers/services/service_instances_controller.rb
@@ -122,6 +122,7 @@ module VCAP::CloudController
       validate_access(:read_for_update, service_instance)
       validate_access(:update, projected_service_instance(service_instance))
 
+      validate_name_update(service_instance)
       validate_space_update(related_objects[:space])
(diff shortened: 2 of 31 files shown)


rainmaker:
Hi Jen,

We looked through your PR and noticed that the feature-flag itself is being left as-is, while you've modified several checks to ignore it. We're wondering why [the flag itself isn't being removed as well](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/models/runtime/feature_flag.rb#L22), to signify to users that something has changed. I could imagine a user disabling the flag with no actual results. Maybe we know that no one is depending on this flag being disabled? Mostly looking for context on that aspect, otherwise looks good 🐋 

cc @zrob for product context on how this usually goes?

Thanks!

Raina and Maryam (@maryamklabib)

maryamklabib:
Seems like the feature flag is still being used in one more place so makes sense why it wasn't totally removed. Merging!

PR #1037, opened 2017-12-15
When trying to share to multiple spaces, return better error message for each invalid space

deniseyu:
As a SpaceAuditor in the target space, when I POST /v3/service_instances/:guid/relationships/shared_spaces, the error message contains multiple errors if needed. [#153629300](https://www.pivotaltracker.com/story/show/153629300)

## What

Previously, when trying to share a service instance to multiple spaces, we would perform a bulk check on whether any of the specified spaces failed the read permissions validation. If it failed, we would return with a 422, listing all the non-readable spaces. If it succeeded, we would subsequently validate write permissions. If any space failed the write permissions check, we would return a 403 response code with no detailed error message. If someone tried to share into 50 spaces, and they had write permission on 49 of them, they would only see a 403 without knowing which space failed. This PR improves on this user experience.

We also now return 422 for validation failures on either read or write permissions for target spaces. 403s are used only for auth failures on the service instance itself.

This change consolidates the read and write permissions checks in order to return a single detailed error message explaining which validation each space failed on.

## PR 

### DIFF
--- a/app/controllers/v3/service_instances_controller.rb
+++ b/app/controllers/v3/service_instances_controller.rb
@@ -76,18 +76,16 @@ class ServiceInstancesV3Controller < ApplicationController
   private
 
   def check_spaces_exist_and_are_writeable!(service_instance, request_guids, found_spaces)
-    unreadable_space_guids = request_guids - found_spaces.map(&:guid)
-
     unreadable_spaces = found_spaces.reject do |space|
       can_read_space?(space)
     end
 
-    unreadable_space_guids += unreadable_spaces.map(&:guid)
-
     unwriteable_spaces = found_spaces.reject do |space|
-      can_write?(space.guid)
+      can_write_space?(space) || unreadable_spaces.include?(space)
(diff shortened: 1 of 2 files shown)


jenspinney:
Great suggestions, thanks @anniesing and @Gerg!

Cleaned up the method a bit using your suggestions. Let us know what you think :)

ericpromislow:
Changes look good, but two typos need fixing (could be done in a later PR): 'mulitple' appears in a spec goal (or whatever the argument to `it` is called), and "nonexistant" should be "nonexistent"

ericpromislow:
I tried running this live against a bosh-lite, and just get a generic NotAuthorized error message when I try to share a service from a space where a non-admin user is a SpaceDeveloper to other spaces.  Could you provide a script like the one we use in https://github.com/cloudfoundry/cloud_controller_ng/wiki/How-to-Make-a-Shareable-Service-Instance to test this live?

PR #1040, opened 2017-12-18
Return a more helpful error when recipient space developers try to update a shared service instance

deniseyu:
As an app dev (receiver), I see a relevant error message when I try to update a service instance that has been shared with me. [#152552023](https://www.pivotaltracker.com/story/show/152552023)

## What

When a service instance is shared, only developers in the source space have authorisation to update the service instance. Anyone else (space auditors, developers in target space, etc) will see a 403 Not Authorized when they attempt to run `cf update-service`.

In this PR we return a more helpful error message in the case when a space developer in a shared-to space attempts to run `cf update-service`, still with a 403 status code. Errors for other types of users remain unchanged.

**NOTE**: This PR builds on top of #1037, which should be merged first. The actual changes on top of #1037 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-service-instance-sharing-better-error-message-sharks...cloudfoundry-incubator:pr-service-instance-sharing-update-error).

## PR 

Thanks, sapi (@ablease and @deniseyu)

### DIFF
--- a/app/controllers/services/service_instances_controller.rb
+++ b/app/controllers/services/service_instances_controller.rb
@@ -119,6 +119,7 @@ module VCAP::CloudController
         raise CloudController::Errors::ApiError.new_from_details('UserProvidedServiceInstanceHandlerNeeded')
       end
 
+      validate_shared_space_updateable(service_instance)
       validate_access(:read_for_update, service_instance)
       validate_access(:update, projected_service_instance(service_instance))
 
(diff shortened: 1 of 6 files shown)


app/controllers/services/service_instances_controller.rb
```diff
@@ -119,6 +119,7 @@ def update(guid)
         raise CloudController::Errors::ApiError.new_from_details('UserProvidedServiceInstanceHandlerNeeded')
       end
 
+      validate_not_shared_space_update(service_instance)
```
anniesing: I have a bit of a nit pick around naming. In general, we prefer to use positive assertions rather than negatives (not, un, etc.) because it's easier for us to reason about. I propose we change this method name to something like `validate_shared_space_updateable` or `validate_shared_space_writeable`.

jenspinney: Yep, sounds much better.

jenspinney:
@anniesing Updated this PR to use the better method name you suggested. I also renamed the error message introduced in this PR to use the spelling "Updateable" instead of "Updatable" for consistency with the rest of this file (and for the sake for my own spelling nitpicking).

Thanks!

PR #1041, opened 2017-12-18
Return a more helpful error when recipient space developers try to delete a shared service instance

ablease:
**NOTE**: This PR builds on top of #1040, which should be merged first. The actual changes on top of #1040 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/cloudfoundry-incubator:pr-service-instance-sharing-update-error...cloudfoundry-incubator:pr-service-instance-sharing-delete-error).

As an app dev (receiver), I see a relevant error message when I try to delete a service instance that has been shared with me [#152552021](https://www.pivotaltracker.com/story/show/152552021)

## What

When a service instance is shared, only developers in the source space have authorisation to delete the service instance. Anyone else (space auditors, developers in target space, etc) will see a 403 Not Authorized when they attempt to run cf delete-service.

In this PR we return a more helpful error message in the case when a space developer in a shared-to space attempts to run cf delete-service, still with a 403 status code. Errors for other types of users remain unchanged.

## PR

Thanks, sapi (alex, @deniseyu)

### DIFF
--- a/app/controllers/services/service_instances_controller.rb
+++ b/app/controllers/services/service_instances_controller.rb
@@ -160,6 +160,7 @@ module VCAP::CloudController
         return [HTTP::NO_CONTENT, nil]
       end
 
+      validate_shared_space_deleteable(service_instance)
       validate_access(:delete, service_instance)
 
       unless recursive_delete?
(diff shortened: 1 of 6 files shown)


app/controllers/services/service_instances_controller.rb
```diff
@@ -159,6 +160,7 @@ def delete(guid)
         return [HTTP::NO_CONTENT, nil]
       end
 
+      validate_not_shared_space_delete(service_instance)
```
anniesing: Same naming nit pick as in PR 1040. Copy and pasted from that PR :-)

In general, we prefer to use positive assertions rather than negatives (not, un, etc.) because it's easier for us to reason about. I propose we change this method name to something like `validate_shared_space_deleteable`.

jenspinney:
@anniesing Thanks! Made the same changes here as for #1040.

PR #1051, opened 2018-01-04
Perm/send permissions

chendrix:
This sends along the appropriately named permission when creating the roles, which should finish the circuit. Now querying for whether or not someone has permission via the /v3 api endpoints should be correct when using the database or when checking perm.

### DIFF
--- a/app/controllers/v3/application_controller.rb
+++ b/app/controllers/v3/application_controller.rb
@@ -92,7 +92,9 @@ class ApplicationController < ActionController::Base
       perm_client,
       SecurityContext,
       configuration.get(:perm, :enabled),
-      configuration.get(:perm, :query_enabled))
+      configuration.get(:perm, :query_enabled),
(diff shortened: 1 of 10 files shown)


ericpromislow:
I get 38 failures/39 tests in `permissions/queryer_spec.rb`, all of the form:
```
 Failure/Error: experiment.run

      NoMethodError:
        undefined method `run' for #<VCAP::CloudController::Science::Experiment:0x00007fdffa31d648>
        Did you mean?  crun
```

chendrix:
@ericpromislow how did you run the tests? I'm not seeing that failure either locally or in any of our CI environments.

Did you do a bundle install before running the tests? The `run` method comes from a Module include that's part of a (relatively) newly introduced gem called Scientist.

lisamcho:
Looks good. I ran the tests locally and did not see any failures.

PR #1058, opened 2018-01-12
Record a service instance delete event when space is deleted

### LINKED ISSUE #974: Service instance delete event does not get recorded when space is deleted
## Issue

When a space or org where a service instance lives is deleted, an event is not logged for the service instance delete.

## Steps to Reproduce

1. Assuming you have a service called some-service with a plan called some-plan:  `cf create-space delete-me && cf target -s delete-me && cf create-service some-service some-plan si1 && cf create-service some-service some-plan si2`
1. Get the last timestamp of the last service instance delete event by running `export SAVED_TIMESTAMP=$(cf curl '/v2/events?q=type:audit.service_instance.delete&order-by=timestamp&order-direction=desc' | jq .resources[0].entity.timestamp)`
1. Verify that `cf curl "/v2/events?q=type:audit.service_instance.delete&q=timestamp>$SAVED_TIMESTAMP"` returns nothing. 
1. Delete si1 directly: `cf curl /v2/service_instances/$(cf service si --guid)?accepts_incomplete=true -X DELETE`.
1. Run `cf curl "/v2/events?q=type:audit.service_instance.delete&q=timestamp>$SAVED_TIMESTAMP"` and see that a new event was created.
1. Delete the space that si2 is in: `cf delete-space delete-me -f`.
1. Run `cf curl "/v2/events?q=type:audit.service_instance.delete&q=timestamp>$SAVED_TIMESTAMP"` and observe that no new event has been created for the deletion of si2.

## Expected result

There is a event of type audit.service_instance.delete corresponding to the deletion of service instance that was in a deleted space.

## Current result

There is no event for a service instance deleting via a cascade delete.

## Possible Fix

Log an event here: https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/actions/services/service_instance_delete.rb#L54

## Note
There are several places in the code that we do log an event for service instance delete, which may not be obvious without some digging. (For example, if the service broker responds to the deprovision request with data instead of the empty JSON body (`{}`) prescribed in the OSBAPI spec, you can get into a situation where the deletion event *is* logged, but only if the last_operation state eventually changes to `succeeded`. From our audit of the possible codepaths, the typical scenario where a cascade delete succeeds and the service broker responds as it should according to the spec, the event isn't logged.

After digging around the code, we came up with the following flow-chart describing when service instance deletes are recorded or not. Each decision point path represents a different codepath that's executed. 
image

Thanks,
Jen and @ablease

jenspinney:
## Summary
Fixes #974. With this change, audit events are correctly created when a service instance is deleted via its parent space deletion and the deprovision call to the broker is completed synchronously.

Tracker story (SAPI backlog): https://www.pivotaltracker.com/story/show/152631677
Tracker story (CAPI backlog): https://www.pivotaltracker.com/story/show/152542410

## Details
In investigating this bug, we familiarized ourselves with all the possible code paths that ultimately end up with a service instance being deleted. There are 3 different "choices" in the service deletion flow, resulting in 6 possible combinations. The 3 choices are:  
1. The service instance can be either deleted directly (via a `DELETE` request to `/v2/service_instances/:guid`, or deleted indirectly by deleting the parent space or org.
1. If a service instance is deleted directly, the request to CC can provide `async=true` or `async=false` as a query parameter. This determines whether the **cloud controller** responds synchronously or asynchronously. An async query here means the delete job get placed on the delayed jobs queue.
1. The **service broker** itself can choose whether to respond synchronously or asynchronously to the service instance deprovision request. An async response here means the broker responds to the CC with a 202, and the CC polls the broker and updates its db when the broker operation is finished.

So, the 6 combinations are:

| Deletion | CC response to CC client | broker response to CC |
| --- | --- | --- |
| Direct | sync | sync |
| Direct | async | sync |
| Direct | sync | async |
| Direct | async | async |
| Parent | sync | sync |
| Parent | sync | async |

Previous to this PR, the event logging for service instance deletion was scattered throughout each of these specific code paths. There was one bit of code responsible for logging the event when it was a direct deletion and the cc request was synchronous, another bit of code responsible for logging when it was a direct deletion and the cc request was async, and another bit of code responsible for logging whenever the broker response was async. If you draw these combinations out, you'll see it misses one of the 6 possible combinations: deletion of the parent when the broker responds synchronously.

This change not only fixes that gap, but also simplifies this logic significantly.  Now there are just two places where this event is logged: one place whenever the broker is synchronous and another whenever the broker is asynchronous. We've tested this change against all six combinations. 

In the course of doing this work, we found that the service_instance_deprovisioner had no tests, so we added a test file for that and backfilled all the tests there. By TDDing this file, we were able to delete some dead code from it.

Thanks,
SAPI team (Jen and @deniseyu)

### DIFF
--- a/app/actions/services/service_instance_delete.rb
+++ b/app/actions/services/service_instance_delete.rb
@@ -52,6 +52,7 @@ module VCAP::CloudController
 
         if attributes_to_update[:last_operation][:state] == 'succeeded'
           lock.unlock_and_destroy!
+          log_audit_event(service_instance)
         else
           lock.enqueue_unlock!(attributes_to_update, build_fetch_job(service_instance))
         end
@@ -86,6 +87,11 @@ module VCAP::CloudController
         @event_repository.user_audit_info,
         {},
       )
+    end
+
+    def log_audit_event(service_instance)
+      event_method = service_instance.managed_instance? ? :record_service_instance_event : :record_user_provided_service_instance_event
+      @event_repository.send(event_method, :delete, service_instance, {})
     end
   end
 end
--- a/app/controllers/services/lifecycle/service_instance_deprovisioner.rb
+++ b/app/controllers/services/lifecycle/service_instance_deprovisioner.rb
@@ -1,9 +1,7 @@
 module VCAP::CloudController
   class ServiceInstanceDeprovisioner
-    def initialize(services_event_repository, access_validator, logger)
+    def initialize(services_event_repository)
       @services_event_repository = services_event_repository
-      @access_validator = access_validator
-      @logger = logger
     end
 
     def deprovision_service_instance(service_instance, accepts_incomplete, async)
@@ -14,13 +12,10 @@ module VCAP::CloudController
 
       delete_job = build_delete_job(service_instance, delete_action)
 
-      enqueued_job = nil
-
       if async && !accepts_incomplete
-        enqueued_job = Jobs::Enqueuer.new(build_audit_job(service_instance, delete_job), queue: 'cc-generic').enqueue
+        enqueued_job = Jobs::Enqueuer.new(delete_job, queue: 'cc-generic').enqueue
       else
         delete_job.perform
-        log_audit_event(service_instance) unless service_instance.exists?
       end
 
       enqueued_job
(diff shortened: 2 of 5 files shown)


app/actions/services/service_instance_delete.rb
```diff
@@ -87,5 +88,10 @@ def build_fetch_job(service_instance)
         {},
       )
     end
+
+    def log_audit_event(service_instance)
+      event_method = service_instance.managed_instance? ? :record_service_instance_event : :record_user_provided_service_instance_event
```
jenspinney: Note: this function was moved from the service_instance_deprovisioner.rb. In this PR, the user provided service case is not tested (and it wasn't in the old code either). Test coverage for this line is added in PR #1059.

deniseyu:
Hey @zrob @Gerg, we'd love to get some feedback on this PR. Let us know if it'd be easier to hop on a call to explain the issues with the existing code paths. We're hoping that this can land in next week's release.

PR #1060, opened 2018-01-17
Prevent race condition when installing admin buildpacks

jandubois:
The cloud controller has 2 local workers that will add admin buildpacks during initial setup, and the `Sequel::Plugins::List` module to maintain the buildpack `position` "is not safe when concurrent modifications are made to the same list."

This commit adds a lock to prevent duplicate buildpack positions like this:

```
$ cf buildpacks
Getting buildpacks...

buildpack               position   enabled   locked   filename
staticfile_buildpack    1          true      false    staticfile_buildpack-v1.4.20-1847e4e.zip
java_buildpack          2          true      false    java-buildpack-v4.7-ad9eef6.zip
ruby_buildpack          2          true      false    ruby_buildpack-v1.7.7-5479a24.zip
nodejs_buildpack        3          true      false    nodejs_buildpack-v1.6.12-2127508.zip
...
```

### DIFF
--- a/app/jobs/runtime/buildpack_installer.rb
+++ b/app/jobs/runtime/buildpack_installer.rb
@@ -16,7 +16,11 @@ module VCAP::CloudController
 
           buildpack = Buildpack.find(name: name)
           if buildpack.nil?
-            buildpack = Buildpack.create(name: name)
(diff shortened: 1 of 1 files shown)


ericpromislow:
Looks good, bake & CATs passed.

PR #1064, opened 2018-01-22
Add stack association to buildpack model [#153256959]

idoru:
- Add "stack" to buildpack model
- At migration time, stacks.yml will be read (location from STACKS_YML env var)
  to determine what stack should be assigned to existing buildpacks
- Buildpacks are now unique over name AND stack
- Sets buildpack stack from manifest.yml in buildpack zip on creation
- Validate buildpack model stack against stack in buildpack zip manifest.yml
- Validate stack exists upon buildpack bits upload
- Include stack name in serialized buildpack filename
- Only provide buildpacks for the relevant stack to the staging container
- Handle buildpack stacks appropriately in the buildpack installer

NOTE: The API checkshum has changed due to adding stack as an input

Signed-off-by: Dave Goddard <dave@goddard.id.au>
Signed-off-by: Victoria Henry <vhenry@pivotal.io>
Signed-off-by: Jackson Feeny <jacksonfeeny@gmail.com>
Signed-off-by: Sam Coward <scoward@pivotal.io>

(see above)

Creating multiple buildpacks with the same name (as long as they have unique stacks).

Migration `20180102183100_add_stack_to_buildpack_table.rb` requires the `STACKS_YML` environment variable to be set to the path of the CC's `stacks.yml` so it can identify the default stack to be assigned to existing buildpacks during the migration. See this [capi-release PR](https://github.com/cloudfoundry/capi-release/pull/75)


idoru:
In this commit we had to add a new presenter class and it seems like there's already a bunch in here that have this `entity_hash` method which is causing the code climate issue. Is that going to be acceptable?

vendor/errors/v2.yml
```diff
@@ -839,9 +839,9 @@
   message: "Service broker returned dashboard client configuration without a dashboard URL"
 
 290001:
-  name: BuildpackNameTaken
-  http_code: 400
-  message: "The buildpack name is already in use: %s"
+  name: BuildpackNameStackTaken
+  http_code: 409
```
maryamklabib: Why did the http_code change?

lib/cloud_controller/upload_buildpack.rb
```diff
@@ -47,8 +50,31 @@ def upload_buildpack(buildpack, bits_file_path, new_filename)
       true
     end
 
+    def extract_stack_from_buildpack(bits_file_path)
+      bits_file_path = bits_file_path.path if bits_file_path.respond_to?(:path)
+      output, _, status = Open3.capture3('unzip', '-p', bits_file_path, 'manifest.yml')
```
maryamklabib: Since the second output is ignored, you could [use `capture2` instead](https://ruby-doc.org/stdlib-1.9.3/libdoc/open3/rdoc/Open3.html#method-c-capture2)

- @selzoc && maryam

spec/request/apps_spec.rb
```diff
@@ -14,14 +15,14 @@
...
-      buildpack      = VCAP::CloudController::Buildpack.make
+      buildpack      = VCAP::CloudController::Buildpack.make(stack: stack.name)
       create_request = {
         name: 'my_app',
         environment_variables: { open: 'source' },
         lifecycle: {
           type: 'buildpack',
           data: {
-            stack: nil,
+            stack: buildpack.stack,
```
maryamklabib: Can we have a test for the case where `nil` gets passed into the request as the stack name?

- @selzoc && maryam

maryamklabib:
@dgodd @jfeeny @tisvictress @idoru 

Just a ping in case y'all haven't seen our comments.

vendor/errors/v2.yml
dgodd: That seemed to us to be a better plan since the meaning changed. We are happy for you to leave it the same

lib/cloud_controller/upload_buildpack.rb
dgodd: certainly

db/migrations/20180102183100_add_stack_to_buildpack_table.rb
```diff
@@ -0,0 +1,30 @@
+require 'yaml'
+
+def default_stack
```
Gerg: Since there is some logic here, it would be nice to write a migration test. 

See examples here: https://github.com/cloudfoundry/cloud_controller_ng/tree/master/spec/migrations

db/migrations/20180102183100_add_stack_to_buildpack_table.rb
```diff
@@ -0,0 +1,30 @@
...
+
+Sequel.migration do
+  up do
+    alter_table(:buildpacks) do
+      add_column :stack, String, size: 255, null: true
+      drop_index :name, unique: true
+      add_index [:name, :stack], unique: true, name: :unique_name_and_stack
+    end
+
+    self['UPDATE buildpacks SET stack = ?', default_stack || 'unknown'].update
```
Gerg: We would prefer to not use raw SQL when possible. Is there a way to write this using the Sequel migration DSL?

app/fetchers/buildpack_lifecycle_fetcher.rb
```diff
@@ -4,14 +4,19 @@ class << self
...
       end
 
       private
 
-      def ordered_buildpacks(buildpack_names)
-        buildpacks = Buildpack.where(name: buildpack_names).all
+      ## NOTE: if a requested system buildpack is not on the requested stack,
+      ##       the BuildpackInfo object will have a name and not a record (even
+      ##       though the buildpack exists). At this point the error returned
+      ##       to the user will probably be VERY confusing
```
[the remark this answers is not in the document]
Gerg: We currently have cross teams planned with Routing and CLI, so it will be difficult to start another one in the near term.

app/jobs/runtime/buildpack_installer.rb
```diff
@@ -14,12 +14,12 @@ def perform
...
           logger.info "Installing buildpack #{name}"
 
-          buildpack = Buildpack.find(name: name)
+          buildpack = find_existing_buildpack
           if buildpack.nil?
             buildpacks_lock = Locking[name: 'buildpacks']
             buildpacks_lock.db.transaction do
               buildpacks_lock.lock!
-              buildpack = Buildpack.create(name: name)
+              buildpack = Buildpack.create(name: name, stack: 'unknown')
```
Gerg: Looks like stack already defaults to `'unknown'`. If you want to be explicit here, please pull out a constant.

app/jobs/runtime/buildpack_installer.rb
```diff
@@ -55,6 +57,22 @@ def buildpack_uploader
...
+
+        private
+
+        def find_existing_buildpack
+          stack = buildpack_uploader.extract_stack_from_buildpack(file)
+          return Buildpack.find(name: name, stack: stack) if stack.to_s != ''
+
+          buildpacks = Buildpack.where(name: name)
+          if buildpacks.count > 1
+            raise AmbiguousBuildpackException.new("Buildpack #{name} has #{buildpacks.count} stacks, not updated")
```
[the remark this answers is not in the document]
Gerg: This exception is 
1. Raised from a private method
1. Caught by the same class that raises it
1. Not re-raised

I think that this exception is only used for control flow in this class. I propose we either re-raise it (wrapped in an APIError) if it is an exceptional case or find another way of modeling the control flow without using exceptions.

app/jobs/runtime/buildpack_installer.rb
Gerg: I just noticed that the default is defined on the controller rather than the model. Either way, we shouldn't have the same string in multiple places.

app/presenters/v2/buildpack_presenter.rb
```diff
@@ -0,0 +1,17 @@
...
+  module Presenters
+    module V2
+      class BuildpackPresenter < DefaultPresenter
+        extend PresenterProvider
+
+        present_for_class 'VCAP::CloudController::Buildpack'
+
+        def entity_hash(controller, buildpack, opts, depth, parents, orphans=nil)
+          entity = super
+          entity['filename'] = "#{buildpack.filename} (#{buildpack.stack})".strip
```
Gerg: Why are we setting filename to include the stack? This feels like something that should happen in the CLI instead of the API.

lib/cloud_controller/diego/lifecycles/buildpack_lifecycle.rb
```diff
@@ -57,6 +57,13 @@ def app_stack
       @package.app.buildpack_lifecycle_data.try(:stack)
     end
 
+    def buildpack_stack
+      stacks = Buildpack.where(name: buildpacks_to_use).select(:stack).uniq
+      if stacks.length == 1
+        stacks.first
```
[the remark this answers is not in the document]
Gerg: I can't find the associated tests. I would expect them to be in `./spec/unit/lib/cloud_controller/diego/lifecycles/buildpack_lifecycle_spec.rb`

lib/cloud_controller/upload_buildpack.rb
```diff
@@ -47,8 +50,31 @@ def upload_buildpack(buildpack, bits_file_path, new_filename)
       true
     end
 
+    def extract_stack_from_buildpack(bits_file_path)
```
Gerg: This feels like a separate concern from uploading the buildpack. Since it is called by buildpack_installer and a private method in upload_buildpack, I would either move it or break it into its own class.

lib/cloud_controller/upload_buildpack.rb
Gerg: If possible, we would prefer to use the `rubyzip` gem instead of shelling out.

lib/cloud_controller/upload_buildpack.rb
```diff
@@ -47,8 +50,31 @@ def upload_buildpack(buildpack, bits_file_path, new_filename)
...
+      bits_file_path = bits_file_path.path if bits_file_path.respond_to?(:path)
+      output, _, status = Open3.capture3('unzip', '-p', bits_file_path, 'manifest.yml')
+      YAML.safe_load(output).dig('stack') if status.success?
+    end
+
     private
 
+    def determine_new_stack(buildpack, bits_file_path)
+      extracted_stack = extract_stack_from_buildpack(bits_file_path)
+      if extracted_stack.to_s != '' && !Stack.where(name: extracted_stack).first
```
Gerg: `extracted_stack.to_s != ''` can be `extracted_stack.present?`

Gerg: `!Stack.where(name: extracted_stack).first` can be `!Stack.where(name: extracted_stack).empty?` (this will stop Sequel from loading the whole model)

lib/cloud_controller/upload_buildpack.rb
```diff
@@ -47,8 +50,31 @@ def upload_buildpack(buildpack, bits_file_path, new_filename)
...
+        raise CloudController::Errors::ApiError.new_from_details('BuildpackStackDoesNotExist', extracted_stack)
+      end
+      new_stack = [extracted_stack, buildpack.stack, Stack.default.name].find { |s| s.to_s != '' && s.to_s != 'unknown' }
```
Gerg: The `s.to_s != '' && s.to_s != 'unknown'` can be moved into the Stack model and given a meaningful name.

lib/cloud_controller/upload_buildpack.rb
```diff
@@ -47,8 +50,31 @@ def upload_buildpack(buildpack, bits_file_path, new_filename)
...
+      if buildpack.stack != 'unknown' && buildpack.stack != new_stack
+        raise CloudController::Errors::ApiError.new_from_details('BuildpackStacksDontMatch', new_stack, buildpack.stack)
+      end
+
+      if buildpack.stack != new_stack && Buildpack.find(name: buildpack.name, stack: new_stack)
+        raise CloudController::Errors::ApiError.new_from_details('BuildpackNameStackTaken', buildpack.name, new_stack)
```
Gerg: This feels like duplicate logic for the model validation that checks for unique (buildpack, stack) pairs.

I've also noticed that we currently don't raise anywhere in this file currently (and explicitly catch Sequel errors). I suspect that we don't have a good reason for doing this, but it might make sense to think about how we handle exceptions in this file.

spec/request/apps_spec.rb
Gerg: I would test these permutations at the controller level. Request specs and generally happy path.

spec/support/test_zip.rb
```diff
@@ -1,14 +1,15 @@
...
+    Zip::File.open(zip_name, Zip::File::CREATE) do |zipfile|
+      file_count.times do |i|
+        zipfile.get_output_stream("ziptest_#{i}") do |f|
+          f.write('A' * file_size)
+        end
+      end
 
-    child = POSIX::Spawn::Child.new('zip', zip_name, *files.map(&:path))
-    child.status.exitstatus == 0 || raise("Failed zipping:\n#{child.err}\n#{child.out}")
+      blk.call(zipfile) if blk
```
Gerg: :+1: 😄

app/controllers/runtime/buildpacks_controller.rb
```diff
@@ -6,17 +6,18 @@ def self.dependencies
 
     define_attributes do
       attribute :name, String
+      attribute :stack, String, default: 'unknown'
```
Gerg: After reviewing this PR, I'm curious why we need the `'unknown'` magic string. It seems like we could just use `nil` and display `unknown` in the presentation layer.

Gerg: Also, do we care about the edge case where someone uses a stack named 'unknown'?

dgodd:
@gerg Can we schedule a time to talk on Monday when the main submitter of this code is back from holidays?

app/fetchers/buildpack_lifecycle_fetcher.rb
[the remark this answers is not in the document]
maryamklabib: The message was clear to me. I find tests to be valuable in terms of documenting desired/known behavior, in this case documenting what error users will see. Leaving comments in the code that describe behavior is a last resort. Main motivation being that tests go red when things change, whereas many a comment has become obsolete silently. Tests do a good job of showing rather than telling. Lastly, this particular case feels straightforward to set up, but correct me if I'm wrong.

app/jobs/runtime/buildpack_installer.rb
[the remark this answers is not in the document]
maryamklabib: A rewording to "update failed" is more clear to me, but I leave it you to determine what is most consistent with other buildpack errors.

vendor/errors/v2.yml
maryamklabib: How did the meaning change?

maryamklabib:
@dgodd @idoru sure send us an invite anytime and we can hop in a zoom. we are PST time zone.

app/presenters/v2/buildpack_presenter.rb
idoru: This was primarily to make things clearer to folks who might be using an older CLI which isn't aware of the multiple stack support -- otherwise they'll just see what appears to be duplicate buildpacks. If there's a better way we can handle this, we'd love to discuss.

db/migrations/20180102183100_add_stack_to_buildpack_table.rb
idoru: :+1: Will do.

db/migrations/20180102183100_add_stack_to_buildpack_table.rb
idoru: We'll give it a shot and reach out if we can't.

lib/cloud_controller/diego/lifecycles/buildpack_lifecycle.rb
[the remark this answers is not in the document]
idoru: Looks like there's just 1 missing test, we'll add.

lib/cloud_controller/upload_buildpack.rb
idoru: Ok sure.

lib/cloud_controller/upload_buildpack.rb
idoru: Cool, we will refactor to use rubyzip instead.

lib/cloud_controller/upload_buildpack.rb
idoru: Think you meant `Stack.where(name: extracted_stack).empty?` without the `!` but otherwise definitely makes sense.

lib/cloud_controller/upload_buildpack.rb
idoru: Agree with your comments prior and here regarding `unknown`. Will look at solving this, if not with `Stack.defined_name?` then something similar.

We'll go with `present?` since it's possible someone could inject an empty string as well.

spec/request/apps_spec.rb
idoru: And I think that buildpacks_controller_spec has a `#create defaults stack to unknown when nil` at https://github.com/cloudfoundry/cloud_controller_ng/pull/1064/commits/49b0a6676be3bb5f24831f8eeedc462372c9c375#diff-77441f6f04beceda9b6bb6c88e5a0643R64

vendor/errors/v2.yml
idoru: 400 seemed wrong, it's meant for when the request could not be understood by the server.
409 seemed more appropriate, which indicates that there is a conflict with the resource.

Will changing this cause problems? If so we are happy to revert.

app/fetchers/buildpack_lifecycle_fetcher.rb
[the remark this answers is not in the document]
idoru: So, we just gave this a test, and the error we see is:
```
Staging app and tracing logs...
Unexpected Response
Response code: 500
CC code:       0
CC error code: 
Request ID:    24dddc03-c91c-4fb8-4783-ac3b8568e49c::4ab4e09a-c73d-4466-a3e9-e5daffccf25e
Description:   {
  "description": "Stager error: no compiler defined for requested stack",
  "error_code": "CF-StagerError",
  "code": 170011
}
```

Coming back to comments made earlier by @dgodd : this error doesn't really inform the user well when they have specified a buildpack name, but there is no buildpack with that name for the relevant stack (either specified, or defaulted). 

As outsiders to this code base it appears to us that this error handling is interwoven with the lifecycle, and as such we would really want a cross team to improve this. I suppose this could be acceptable in a first cut, and we could attack this as an improvement later on.

app/controllers/runtime/buildpacks_controller.rb
idoru: It seemed clearer to use `unknown` as this makes it clearer that in the 2-phase process of creating the buildpack record and then later uploading the actual bits (and only during this 2nd phase we are able to properly ascertain what the stack really is) that there is an indeterminate state, rather than possibly non-intentionally not initializing it. We also didn't want the column to be optional in the database, and an empty string is just as magical and less obvious.

While the edge case could happen, the chances seem fairly remote. How far do we want to go to prevent something silly like this?

app/jobs/runtime/buildpack_installer.rb
idoru: @Gerg yeah having the defaults on the controller did seem odd, but `position`, `enabled` and `locked` are all defaulted here as well and it doesn't appear any sort of defaulting is done in the model.

We're not very familiar with Sequel, so if you can recommend a better practice, we would be happy to incorporate.

app/jobs/runtime/buildpack_installer.rb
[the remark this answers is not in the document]
idoru: @Gerg sure, we can remove the PBE. We don't want to re-raise the error since we don't want errors in buildpack installation to fail the deployment. We'd rather the operator see that a particular buildpack update failed and make fixes manually as needed.

app/jobs/runtime/buildpack_installer.rb
```diff
@@ -55,6 +57,22 @@ def buildpack_uploader
...
+          end
+          buildpacks.first
+        end
+
+        class AmbiguousBuildpackException < RuntimeError
```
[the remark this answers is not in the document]
idoru: This problem should go away since we're going to avoid processing by exception now.

maryamklabib:
@idoru Did you want a meeting about this PR, as @dgodd suggested?
Or do you all plan on pushing folow up commits?

--@selzoc and maryam

app/fetchers/buildpack_lifecycle_fetcher.rb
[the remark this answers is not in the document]
maryamklabib: @zrob Do you have thoughts on how this error message could be improved? Is it acceptable as is for an MVP?

-- @selzoc and maryam

vendor/errors/v2.yml
maryamklabib: I agree that 409 is more appropriate. Concern is that we break clients that rely on this behavior. @zrob thoughts?

idoru:
Yes please. I was feeling unwell today but it would be great if you have
time tomorrow or later this week during the EST overlap.

vendor/errors/v2.yml
zrob: i'm not quite clear on what's happening from this snippet, but roughly, we should not change names/codes of existing errors.   if a different error is appropriate we should add a new entry and use that.  as far as 409 -- we have moved towards a convention of using 422 for this type of error and relying more on the text to describe the case than the code.

app/fetchers/buildpack_lifecycle_fetcher.rb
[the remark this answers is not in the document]
idoru: @Gerg ^^

idoru:
Worth checking out this document which details the changes for both CC and CF CLI together https://docs.google.com/document/d/1hPvZuNsqk08ysw1wvmRkkqbr-6PFRxgLVgYEchf38z0/edit?usp=sharing

lib/cloud_controller/diego/lifecycles/buildpack_lifecycle.rb
```diff
@@ -57,6 +57,13 @@ def app_stack
       @package.app.buildpack_lifecycle_data.try(:stack)
     end
 
+    def buildpack_stack
+      stacks = Buildpack.where(name: buildpacks_to_use).select(:stack).map(&:stack).uniq
+      if stacks.length == 1
```
tcdowney: What should this return if `stacks.length` is > 1?

db/migrations/20180102183100_add_stack_to_buildpack_table.rb
```diff
@@ -0,0 +1,19 @@
+require 'yaml'
+
+Sequel.migration do
+  up do
+    alter_table(:buildpacks) do
+      add_column :stack, String, size: 255, null: true
+      drop_index :name, unique: true
```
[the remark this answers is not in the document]
tcdowney: According to the [Sequel docs](https://sequel.jeremyevans.net/rdoc/classes/Sequel/Schema/CreateTableGenerator.html#method-i-index), Sequel will attempt to use the same name it would have generated when adding the index initially. This should be fine for most cases. 

The naming logic looks like this:
https://github.com/jeremyevans/sequel/blob/7c7d14d00a2cfd3ccb7bfef7fc6f050a12e264dc/lib/sequel/database/schema_methods.rb#L775

I'm a bit worried that this will result in failed upgrades for some users, but hopefully that's just FUD since I can't think of a way around doing this.

idoru:
@tcdowney Thanks, I have incorporated your feedback in the latest revision of this PR.

lib/cloud_controller/diego/lifecycles/buildpack_lifecycle.rb
idoru: nil

db/migrations/20180102183100_add_stack_to_buildpack_table.rb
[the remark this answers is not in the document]
idoru: Weird. I checked the database before migrating and `SHOW INDEX on buildpacks` shows the name was `buildpacks_name_index`, but changing the migration to `drop_index name: :buildpacks_name_index` yields this error: 
`Sequel::DatabaseError: PG::UndefinedObject: ERROR:  index "buildpacks___name____buildpacks_name_index___index" does not exist`

So I have reverted to not passing the name here, but I'm happy to try other suggestions; our team is not experienced with Sequel.

selzoc:
Looking through the context both here and on slack, it seem like this PR is blocked on the merging of the [fix to rubyzip](https://github.com/rubyzip/rubyzip/pull/358) - I don't think we want to use a fork.  Another option that @Gerg noted would be for the stack to be passed as a parameter to the api instead of relying on unzipping stuff server-side.

Thanks,
Chris && Davis

julik:
I don't know if this might be relevant but you could use the ZIP reader we have in https://github.com/WeTransfer/zip_tricks as it opens these entries correctly

sclevine:
We would prefer to validate the stack on the server side. A buildpack that is associated with the wrong stack can have confusing and unexpected failure modes during app staging or launch, especially if the stacks are similar (ex. cflinuxfs2 vs. cflinuxfs3). We're also concerned about the behavior of the cf Java client.

We understand not wanting to merge this PR until the rubyzip PR is merged. We will wait on that PR for a few more days before considering other options.

elenasharma:
@sclevine Sounds good. We're good to wait on your PR to rubyzip/other potential options.

Gerg:
@sclevine Are you ready for us to look at this again?

sclevine:
@idoru @tylerphelan ^

idoru:
@Gerg All good now, except for https://codeclimate.com/github/cloudfoundry/cloud_controller_ng/lib/cloud_controller/buildpacks/stack_name_extractor.rb?from_sha=ad826cbb&to_sha=1a50e736

Thoughts?

db/migrations/20180404165800_assign_stacks_to_buildpacks.rb
```diff
@@ -0,0 +1,30 @@
+require 'yaml'
+
+def stacks_yml
+  stacks_yml_path = ENV.fetch('STACKS_YML', nil)
+  YAML.safe_load(File.read(stacks_yml_path)) if stacks_yml_path && File.exist?(stacks_yml_path)
+end
+
+def default_stack
+  stacks_yml['default'] if stacks_yml
```
lisamcho: The result of `stacks_yml` should be cached, even by doing something like:
```
@stacks_yml ||= begin
  stacks_yml_path = ENV.fetch('STACKS_YML', nil)
  YAML.safe_load(File.read(stacks_yml_path)) if stacks_yml_path && File.exist?(stacks_yml_path)
end
```

lisamcho: with @ericpromislow

lib/cloud_controller/upload_buildpack.rb
```diff
@@ -49,6 +55,27 @@ def upload_buildpack(buildpack, bits_file_path, new_filename)
...
+      if buildpack.errors.on(:stack).try(:include?, :buildpack_stack_does_not_exist)
+        raise CloudController::Errors::ApiError.new_from_details('BuildpackStackDoesNotExist', buildpack.stack)
+      end
+    end
+
+    def determine_new_stack(buildpack, bits_file_path)
+      extracted_stack = Buildpacks::StackNameExtractor.extract_from_file(bits_file_path)
+      new_stack = [extracted_stack, buildpack.stack, Stack.default.name].find(&:present?)
+
+      new_stack
```
lisamcho: There's no need to store the result of the `.find` in `new_stack`. Just return it.

With @ericpromislow

db/migrations/20180404165800_assign_stacks_to_buildpacks.rb
idoru: Can you help me understand what is the concern with reading a fairly small yaml file to get a single key a handful of times during a single migration?

Implementing your suggestion appears to introduce global state, breaking the tests. This is the diff of what I applied:

```
--- a/db/migrations/20180404165800_assign_stacks_to_buildpacks.rb
+++ b/db/migrations/20180404165800_assign_stacks_to_buildpacks.rb
@@ -1,8 +1,10 @@
 require 'yaml'
 
 def stacks_yml
-  stacks_yml_path = ENV.fetch('STACKS_YML', nil)
-  YAML.safe_load(File.read(stacks_yml_path)) if stacks_yml_path && File.exist?(stacks_yml_path)
+  @stacks_yml ||= begin
+    stacks_yml_path = ENV.fetch('STACKS_YML', nil)
+    YAML.safe_load(File.read(stacks_yml_path)) if stacks_yml_path && File.exist?(stacks_yml_path)
+  end
 end
 
 def default_stack
```

lib/cloud_controller/upload_buildpack.rb
idoru: :+1: Good spot. Updating code now...

lisamcho:
What happens if user upload hwc_buildpack as 'my-hwc-buildpack'
The data migration will assign it to the default stack rather than windows stack.

db/migrations/20180404165800_assign_stacks_to_buildpacks.rb
lisamcho: Not a dealbreaker. We only use that pattern when you new up a instance of a class so that the yaml  isn't cached forever.

sclevine:
It isn't possible to associate already-uploaded buildpacks with their intended stack, so this feature will require manual intervention by operators on foundations with more than one stack. We believe the HWC buildpack (uploaded as `hwc_buildpack` by default) is the only case where this affects an official CF buildpack, so we added this to the migration for convenience. We are aware that custom HWC buildpacks will need to be re-uploaded.

@zrob reminder that this should be called out loudly in the release notes

zrob:
@petergtz @suhlig @smoser-ibm 

Y'all should probably have been pulled into this earlier. This is a feature being introduced in cc that will need to work with bit service also.

PR #1062, opened 2018-01-18
Support new OSBAPI service retrievable definitions

ablease:
As a developer, I can find out whether a service supports retrieving service instances and bindings [#153738743](https://www.pivotaltracker.com/story/show/153738743)

## What
A proposed change to the OSBAPI spec introduces new endpoints for the service broker to return details about service instances and service bindings.

https://github.com/openservicebrokerapi/servicebroker/pull/333

To inform platforms that the broker supports this new endpoint, brokers can define the boolean properties, `instances_retrievable` and `bindings_retrievable` in the service definition of their service catalog.

## PR 

* [X] I have viewed signed and have submitted the Contributor License Agreement
* [X] I have made this pull request to the `master` branch
* [X] I have run all the unit tests using `bundle exec rake`
* [X] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats) on bosh lite

PR #1068, opened 2018-01-23
GET endpoint for service binding parameters

nmaslarski:
**NOTE**: This PR builds on top of #1062, which should be merged first. The actual changes on top of #1062 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-retrievable-service-instances-and-bindings...cloudfoundry-incubator:pr-fetching-service-binding-parameters).

- As a developer, I can call an endpoint to fetch service binding parameters for a broker that does not support this [#153738945](https://www.pivotaltracker.com/story/show/153738945)

- Fetching the parameters of a service bound to a user provided service results in a 500 [#154556208](https://www.pivotaltracker.com/story/show/154556208)

## What

This PR begins implementing the `GET` endpoints for retrieving services bindings and instances. This PR contains only the endpoint to `GET` service binding parameters; service instance parameters will be implemented as [a separate story](https://www.pivotaltracker.com/story/show/153739004).

422 responses are returned if a broker has not set `bindings_retrievable`, set it to false or is bound to a user provided service. 404 is returned if the service binding guid is invalid or if the guid belongs to a service key. The 200 case, when a service broker has `bindings_retrievable: true` returns an empty response for now but will be updated later.

## Changes

* Add `/v2/service_bindings/:guid/parameters` endpoint
* Add corresponding API docs

## PR 

Thanks, sapi (@nmaslarski and @Samze)

### DIFF
--- a/app/controllers/services/service_bindings_controller.rb
+++ b/app/controllers/services/service_bindings_controller.rb
@@ -25,6 +25,20 @@ module VCAP::CloudController
       object_renderer.render_json(self.class, obj, @opts)
     end
 
+    get '/v2/service_bindings/:guid/parameters', :parameters
+
+    def parameters(guid)
+      binding = find_guid(guid)
+      raise CloudController::Errors::ApiError.new_from_details('ServiceBindingNotFound', guid) unless binding.v2_app.present?
+
+      unless binding.service.bindings_retrievable
+        message = 'This service does not support fetching service binding parameters.'
+        raise CloudController::Errors::ApiError.new_from_details('UnprocessableEntity', message)
+      end
+
+      [HTTP::OK, {}]
+    end
+
     post path, :create
 
     def create
--- a/app/controllers/services/service_bindings_controller.rb
+++ b/app/controllers/services/service_bindings_controller.rb
@@ -31,7 +31,7 @@ module VCAP::CloudController
       binding = find_guid(guid)
       raise CloudController::Errors::ApiError.new_from_details('ServiceBindingNotFound', guid) unless binding.v2_app.present?
 
-      unless binding.service.bindings_retrievable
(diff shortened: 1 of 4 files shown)


Samze:
@Gerg We would like this reviewed as its the addition of the API/Docs for the `/v2/service_binding/:guid/parameters` endpoint.

Note: A subsequent PR change the 422 to a 400 to be more consistent with v2.

Gerg:
@zrob @anniesing

deniseyu:
### Please don't merge yet!

We need to address this issue first: https://www.pivotaltracker.com/story/show/154809864

app/controllers/services/service_bindings_controller.rb
```diff
@@ -25,6 +25,20 @@ def read(guid)
       object_renderer.render_json(self.class, obj, @opts)
     end
 
+    get '/v2/service_bindings/:guid/parameters', :parameters
+
+    def parameters(guid)
+      binding = find_guid(guid)
+      raise CloudController::Errors::ApiError.new_from_details('ServiceBindingNotFound', guid) unless binding.v2_app.present?
```
jenspinney: Is line 32 dead code? I couldn't get a controller unit test to fail when I deleted it in my local environment.

Opening up a chore to take a look (https://www.pivotaltracker.com/story/show/154875456). If it is unneeded, we'll open a PR to delete it down the line so we don't need to rebase all the PRs that build off of this one.

Samze: `/v2/service_bindings` only support the web process type and filters out all service bindings to any other type. After talking with Zach from CAPI he said that the general idea is that bindings should be per App, not per process and this guard is there on purpose. So we have kept this in and added a test for the line.

PR #1072, opened 2018-01-26
Add binding params object to GET /v2/service_bindings/:guid/parameters

nmaslarski:
**NOTE**: This PR builds on top of #1068, which should be merged first. The actual changes on top of #1068 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-fetching-service-binding-parameters...cloudfoundry-incubator:pr-fetch-binding-params).

- As a developer, I can fetch service binding parameters that I used to configure a service binding
[#153739071](https://www.pivotaltracker.com/story/show/153739071)

## What

In the previous story we implemented a bare-bones GET `/v2/service_bindings/:guid/parameters` endpoint. This story builds on top of that by actually adding the service binding parameters object to the response when the broker responds with a 200.

In cases when the broker responds with invalid JSON, the endpoint returns 502. 

## Changes

* Enhance `/v2/service_bindings/:guid/parameters` endpoint

## PR 

* [X] I have viewed signed and have submitted the Contributor License Agreement
* [X] I have made this pull request to the `master` branch
* [X] I have run all the unit tests using `bundle exec rake`
* [X] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats) on bosh lite

Thanks, sapi (@nmaslarski and @deniseyu)

PR #1073, opened 2018-01-26
Return 400 when fetching binding parameters from unsupported brokers

nmaslarski:
**NOTE**: This PR builds on top of #1072, which should be merged first. The actual changes on top of #1072 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-fetch-binding-params...cloudfoundry-incubator:pr-fetch-binding-params-from-unsupported-broker).

- As a developer, I can call an endpoint to fetch service binding parameters for a broker that does not support this and get a HTTP 400 [#154626124](https://www.pivotaltracker.com/story/show/154626124)

## What

When service brokers does not have `bindings_retrievable` set to `true` in the service catalog we want to return `HTTP status: 400`. This endpoint previously returned 422 in this scenario, but a 400 response is more consistent with other broker errors.

## PR 

Thanks, sapi (@nmaslarski and @deniseyu)

### DIFF
--- a/app/controllers/services/service_bindings_controller.rb
+++ b/app/controllers/services/service_bindings_controller.rb
@@ -39,7 +39,7 @@ module VCAP::CloudController
       client = VCAP::Services::ServiceClientProvider.provide(instance: binding.service_instance)
       resp = client.fetch_service_binding(binding)
 
-      [HTTP::OK, {}, resp['parameters'].to_json]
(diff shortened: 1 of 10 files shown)


Samze:
Once all other PRs that this PR depends on are merged the SAPI team can auto-merge this.

PR #1075, opened 2018-01-29
Improve handling of broker error responses

ablease:
**NOTE**: This PR builds on top of #1074, which should be merged first. The actual changes on top of #1074  can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-discover-service-binding-parameters-url...cloudfoundry-incubator:pr-handle-common-broker-errors).

[#153739167](https://www.pivotaltracker.com/story/show/153739167) As a developer, I see an error when a broker times out when fetching service binding parameters

## What

This PR improves handling of error responses that can be returned by the broker.  This is not only true for when cloud controller makes a request to fetch service binding parameters, but also for all other requests cc makes to the broker. 

## PR
* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `master` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [x] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats)

Thanks Sapiteam, (@georgi-lozev + @ablease)

PR #1083, opened 2018-02-07
Add GET /v2/service_instances/:guid/parameters endpoint

jenspinney:
### Relevant stories:
As a developer, I can call an endpoint to fetch service instance parameters for a broker that does not support this. [#150801660](https://www.pivotaltracker.com/story/show/150801660)
As a developer, I can fetch service instance parameters that I used to configure a service instance [#153739004](https://www.pivotaltracker.com/story/show/153739004)

**NOTE**: This PR builds on top of #1075, which should be merged first. The actual changes on top of #1075 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cloud_controller_ng_sapi/compare/pr-handle-common-broker-errors...cloudfoundry-incubator:pr-fetch-service-instance-parameters).

## What
This change introduces a new endpoint to allow CC clients to retrieve the parameters of a particular service instance. This allows CC consumers to take advantage of the new feature in the Open Service Broker API to retrieve the current values of parameters for a given service instance.

The equivalent work has already been implemented for service bindings, so this may feel slightly deja-vu-ish, but the difference here is that we're now implementing the functionality for service *instances*.

## Changes:
* Added the `GET /v2/service_instances/:guid/parameters` endpoint.
* Added a request spec
* Added a controller level specs.
* Verified that only users who have permission to read the service instance have the ability to get information from this endpoint.
* Added a new method to the service broker client to issue this request.
* Added a new error (ServiceFetchInstanceParametersNotSupported) when someone tries to query this endpoint, but the service in question doesn't have the `instances_retrievable` field set to true in its catalog.

## PR 

Thanks, 
SAPI team (@jenspinney and @georgi-lozev)

### DIFF
--- a/app/controllers/services/service_bindings_controller.rb
+++ b/app/controllers/services/service_bindings_controller.rb
@@ -39,7 +39,7 @@ module VCAP::CloudController
       client = VCAP::Services::ServiceClientProvider.provide(instance: binding.service_instance)
       resp = client.fetch_service_binding(binding)
 
-      [HTTP::OK, {}, resp['parameters'].to_json]
+      [HTTP::OK, {}, resp.fetch('parameters', {}).to_json]
     end
 
     post path, :create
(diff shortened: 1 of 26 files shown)


Samze:
When the preceding PRs have been merged sapi team are happy to merge this ourselves.

PR #1110, opened 2018-03-09
Do not allow Service Broker registration with basic auth in the broker_url

lurraca:
## Story

Do not allow broker registration that contains basic auth credentials in the URL [#155659461](https://www.pivotaltracker.com/story/show/155659461)

## Why

At the moment when a Service broker returns an invalid response, CC error handling logs the full URL that was used to register the SB. If that URL contains basic auth username:password, then this will be visible to the user whose request was answered with invalid response regardless of who registered the broker.

## What

Adds a validation for basic auth in the URL for the `ServiceBroker` model, if basic auth is included in the `broker_url` field the object fails validation.

Thanks, SAPI team

### DIFF
--- a/app/actions/services/mixins/service_broker_registration_error_parser.rb
+++ b/app/actions/services/mixins/service_broker_registration_error_parser.rb
@@ -8,6 +8,8 @@ module VCAP::CloudController
         CloudController::Errors::ApiError.new_from_details('ServiceBrokerUrlInvalid', broker.broker_url)
       elsif errors.on(:broker_url) && errors.on(:broker_url).include?(:unique)
         CloudController::Errors::ApiError.new_from_details('ServiceBrokerUrlTaken', broker.broker_url)
+      elsif errors.on(:broker_url) && errors.on(:broker_url).include?(:basic_auth)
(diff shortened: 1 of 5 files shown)


lurraca:
@zrob @Gerg This PR includes some small breaking changes, so it would be valuable that the CAPI team reviews it.

zrob:
Discussed with @mattmcneeney and @ablease.

- This does feel like a breaking change, but it is an acceptable one if the SAPI team feels it is worth it.
- SAPI to follow up on how this affects folks that already have brokers registered this way.  They should still be able to perform actions like catalog updates, etc.

georgi-lozev:
We've validated that users who have registered service brokers with credentials in the URL will still have their broker working, but will be enforced to remove the credentials from the URL upon next update.  We've also added additional test that covers the update service broker flow as a separate commit.

mattmcneeney:
Hey @zrob We confirmed that folks wouldn't end up in a sticky situation if they had already done this and that this would improve the security in the platform, so we decided to update and merge this PR. Let us know if you have any questions!

PR #1117, opened 2018-03-21
Allow devs to set custom DB connection info and still run units in parallel

jenspinney:
Previous to this PR, a developer working on cloud controller needed to have a root user in mysql with a password of `password` in order to run the unit tests in parallel.

A user wishing to use a different username and password for the unit test database could do so with `DB_CONNECTION_STRING`, but this would not allow the developer to run the units in parallel, which is the default `rake` behavior.

This change adds two new environment variables (`MYSQL_CONNECTION_PREFIX` and `POSTGRES_CONNECTION_PREFIX`) that a developer may set to override the default connection string prefix for mysql and postgres. Using these connection prefixes will still create and use databases called `cc_test` (serial) or `cc_test_<n>` (parallel).

Example usage:

    MYSQL_CONNECTION_PREFIX=mysql2://custom_user:custom_pass@localhost:3306 POSTGRES_CONNECTION_PREFIX=postgres://custom_user@localhost:5432 bundle exec rake

**Note**: You (capi team) may want to consider removing the mysql-travis db type and just using the MYSQL_CONNECTION_PREFIX for your travis CI runs.

* [n/a] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats)

Thanks,
SAPI Team (Jen and @lurraca)

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,23 +58,28 @@ keep it up to date, documenting the purpose of the various types of tests.
 By default `rspec` will randomly pick between postgres and mysql.
 
 It will try to connect to those databases with the following connection string:
-postgres: postgres://postgres@localhost:5432/cc_test
-mysql: mysql2://root:password@localhost:3306/cc_test
 
+* postgres: `postgres://postgres@localhost:5432/cc_test`
+* mysql: `mysql2://root:password@localhost:3306/cc_test`
+
+To specify a custom username, password, host, or port for either database type, you can override the default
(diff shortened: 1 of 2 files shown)


jenspinney:
@Gerg @zrob Since this isn't really services related, would love to get some eyes from capi on this to make sure it fits in with your idea of the dev workflow.

ericpromislow:
@jenspinney , thanks for solving this problem.

ericpromislow:
@jenspinney thanks for solving this problem

PR #1137, opened 2018-05-09
Fix vcap-request-id passing through to bits-service

petergtz:
https://www.pivotaltracker.com/n/projects/966314/stories/157261049

### DIFF
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -329,7 +329,8 @@ module CloudController
       BitsService::ResourcePool.new(
         endpoint: bits_service_options[:private_endpoint],
         request_timeout_in_seconds: config.get(:request_timeout_in_seconds),
-        ca_cert_path: bits_service_options[:ca_cert_path]
+        ca_cert_path: bits_service_options[:ca_cert_path],
+        vcap_request_id: VCAP::Request.current_id
       )
     end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -85,7 +85,7 @@ GEM
     backports (3.6.8)
     beefcake (1.0.0)
     bit-struct (0.16)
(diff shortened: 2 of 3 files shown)


lib/cloud_controller/dependency_locator.rb
```diff
@@ -329,7 +329,8 @@ def bits_service_resource_pool
       BitsService::ResourcePool.new(
         endpoint: bits_service_options[:private_endpoint],
         request_timeout_in_seconds: config.get(:request_timeout_in_seconds),
-        ca_cert_path: bits_service_options[:ca_cert_path]
+        ca_cert_path: bits_service_options[:ca_cert_path],
+        vcap_request_id: VCAP::Request.current_id
```
tcdowney: This works because this method is returning a new instance of the `BitsService::ResourcePool` every time it's invoked, but it worries me a bit. The `DependencyLocator` is a singleton and does a lot of memoization of it's clients (again don't think this is happening here) and `VCAP::Request.current_id` accesses a Thread local variable that should be different for each request.

Is there a way to set this on the Bits Service requests outside of the `DependencyLocator`?

tcdowney: Assuming there isn't, can you add some tests to ensure that the client is initialized with the correct vcap-request-id across different requests?

petergtz: @tcdowney I share your concern. I think the proper fix would be to make `ResourcePool` a long-living object instead of instantiating it on every request (which seems expensive anyway). Then we should probably pass `vcap-request-id` on the indiviual method calls of `ResourcePool`. But I think this is out of scope for this change. Maybe long-term we should overthink how the bits-service-client gets invoked by the Cloud Controller.

Therefore, I've added a test as you requested.

tcdowney:
Thanks @petergtz!

PR #1156, opened 2018-06-06
Return 502 if broker responds async to sync binding request

nmaslarski:
## Summary of the Change

On the `POST /v2/service_bindings` endpoint there is a flag called `accepts_incomplete`. This flag is passed down to the broker. If `accepts_incomplete` is set to `true` the broker may respond to the service binding create request with 202, which indicates that the binding will be created asynchronously. If `accepts_incomplete` is set to `false` the broker must create the service binding synchronously. If the broker is unable to create binding synchronously it is supposed to return 422.

This change addresses spec incompliant brokers which return 202 despite `accepts_incomplete` being set to `false`. In this case now we will return 502 back to the user.
We also send unbind request to the broker.

## Associated Story
[#158106741: An operator, I can update the broker poll interval while an async binding job is enqueued](https://www.pivotaltracker.com/story/show/158106741)

Thanks,
SAPI Team (@jenspinney  and @nmaslarski)

### DIFF
--- a/app/actions/service_binding_create.rb
+++ b/app/actions/service_binding_create.rb
@@ -9,6 +9,7 @@ module VCAP::CloudController
     class ServiceInstanceNotBindable < InvalidServiceBinding; end
     class ServiceBrokerInvalidSyslogDrainUrl < InvalidServiceBinding; end
     class ServiceBrokerInvalidBindingsRetrievable < InvalidServiceBinding; end
+    class ServiceBrokerRespondedAsyncWhenNotAllowed < InvalidServiceBinding; end
     class VolumeMountServiceDisabled < InvalidServiceBinding; end
     class SpaceMismatch < InvalidServiceBinding; end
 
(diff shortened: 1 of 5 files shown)


rboykin:
Is this consistent with how creation of a service `instance` works? If accepts_incomplete=false and the broker returns a 202 for a provision, what does cloud controller do now?

Samze:
Hey @rboykin,

If the users specifies accepts_incomplete=false and the broker returns 202 for provision/deprovision service instance then CC creates the entity and starts polling last operation and returns 202 to the client.

We believe this to be undesirable behaviour, so when we added the same functionality to bindings we wanted to error to clients instead of processing the request silently.

At the moment this is indeed inconsistent. We were worried that this was a breaking change and that  fixing this for service instances could break clients that are relying on the current behaviour. After discussing this with CAPI we decided that for now we will not introduce the same problem for bindings and leave the current behaviour for instances the same for now. We may fix the inconsistency in the future.

Sam & @ablease

Issue #929, opened 2017-10-05
Emit metrics / events for monitoring service brokers


henrytk:
## Issue

cloud_controller_ng does not emit metrics relating to interactions with service brokers. More specifically, it does not allow for the monitoring of failed interactions with service brokers.

## Context / Why

We want to be able to monitor and alert on failed service bindings, and possibly in the future monitor  other failed interactions with service brokers.

At first we are looking at alerting on 5xx errors received by Cloud Controller from any service broker. We have explored solutions, such as:

* having the brokers emit the metric: this does not cover situations where a request manages to get through to a broker which is completely down.
* parsing logs: this would fail silently if the messages failed to log or the message format changes.
* parsing service binding events from the CF API: not feasible. Events are not created for failed bindings. This is because the _creation_ of a service binding is the event, not the request to bind.

## Possible Fix

Two suggested solutions:

* emitting metrics relating to interactions with service brokers, specifically providing failure metrics for use in monitoring (4/5xx error responses from the different endpoints).
* extending the existing events to include requests for service bindings (and other operations) _and_ their outcome, not just the creation events.

Gerg:
Hey @henrytk.

I think emitting more metrics would be the best way to provide additional visibility into service broker interactions. What types of metrics would you be most interested in? How would you want them to be bucketed?

Thanks!

LeePorte:
Hi Gerg

To give you a little more context around this, we had an issue where a code change to one of our service brokers caused tenants to not be able to bind or unbind the backing service provided by that broker for existing instances, new instances were unaffected though. We managed to track this down after a while, however being able to trigger alerting of non normal status would have helped significantly.

Ideally focussing on the broker API would seem like a good place to start so emitting metrics and events regarding the different broker API operations.

 - Create
 - Bind
 - Unbind
 - Delete
 - Last Operation
 - Catalogue
 - Service ID
 - Plan name

Success, failure and time taken would be an excellent start. By having both the success and failure status would enable us to calculate success rates on which we can report.

If you are planning on adding events, it would be great if the success and error messages were passed.

In a normal situation all of these should succeed, it is only in a non-normal situation where failures would be expected. e.g. provider is in degraded service, code change has broken / is impacting on the time taken for an operation.

In terms of bucketing, grouping by broker or service name/UUID and a 1 minute interval for emitting the data would be perfect.

Thanks

Lee

henrytk:
Sounds good to me @LeePorte

[...]

PR #1165, opened 2018-06-19
Emit metrics / events for monitoring service brokers

henrytk:

Emit extra http status code metrics for certain API endpoints, namely service broker-related endpoints. HTTP requests with the following prefixes will send extra metrics:

/[version]/service_instances
/[version]/service_bindings
/[version]/service_keys

The emitted metric name is:

cc.requests.#{route}.#{method.downcase}.http_status.#{status.to_s[0]}XX

E.g. cc.requests.service_bindings.put.http_status.5XX

## An explanation of the use cases our change solves

Monitoring of service brokers to ensure operations are not failing, such as binding for example.

The current metrics only cover things internal to Cloud Foundry. These metrics provide the opportunity to monitor external components the platform relies on.

Associated issue: https://github.com/cloudfoundry/cloud_controller_ng/issues/929

---

### DIFF
--- a/lib/cloud_controller/metrics/request_metrics.rb
+++ b/lib/cloud_controller/metrics/request_metrics.rb
@@ -5,17 +5,28 @@ module VCAP::CloudController
     class RequestMetrics
       def initialize(statsd=Statsd.new)
         @statsd = statsd
+        @collect_metrics_for_routes = ['service_instances', 'service_bindings', 'service_keys']
       end
 
       def start_request
         @statsd.increment 'cc.requests.outstanding'
       end
 
-      def complete_request(status)
+      def complete_request(path, method, status)
         @statsd.batch do |batch|
           batch.decrement 'cc.requests.outstanding'
           batch.increment 'cc.requests.completed'
           batch.increment "cc.http_status.#{status.to_s[0]}XX"
+
+          pos = path.index('/', 1)
+          if !pos.nil?
+            path_without_version = path[pos + 1..-1]
+            @collect_metrics_for_routes.each do |route|
+              if path_without_version.start_with?(route)
+                batch.increment "cc.requests.#{route}.#{method.downcase}.http_status.#{status.to_s[0]}XX"
+              end
+            end
+          end
         end
       end
     end
--- a/lib/cloud_controller/metrics/request_metrics.rb
+++ b/lib/cloud_controller/metrics/request_metrics.rb
@@ -19,14 +19,12 @@ module VCAP::CloudController
           batch.increment "cc.http_status.#{status.to_s[0]}XX"
 
           pos = path.index('/', 1)
-          if !pos.nil?
-            path_without_version = path[pos + 1..-1]
-            @collect_metrics_for_routes.each do |route|
-              if path_without_version.start_with?(route)
-                batch.increment "cc.requests.#{route}.#{method.downcase}.http_status.#{status.to_s[0]}XX"
-              end
-            end
-          end
+          next if pos.nil?
+
+          path_without_version = path[pos + 1..-1]
+          @collect_metrics_for_routes.
+            select { |route| path_without_version.start_with? route }.
+            each { |route| batch.increment "cc.requests.#{route}.#{method.downcase}.http_status.#{status.to_s[0]}XX" }
         end
       end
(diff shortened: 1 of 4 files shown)


chrisfarms:
Hello, bump 😄 , is there anything we could do to improve this PR for consideration? 

... we think logging these metrics will be beneficial to anyone operating service brokers on their platforms, and are happy to work on it .... but if it is unlikely to get approved we shall have to look into an alternative way of monitoring when tenants are seeing errors returned to them during broker lifecycle methods.

keymon:
Hello, any update about this? See https://github.com/cloudfoundry/cloud_controller_ng/pull/1165#issuecomment-408859679

infews:
Hi @chrisfarms && @keymon,

Service operation failures are already emitted as logs.  We have some concerns about adding metrics on every request - currently there is no logic there.  Have you also taken a look at the service-related [audit events](https://docs.cloudfoundry.org/running/managing-cf/audit-events.html)?

Another option would be to monitor the service brokers themselves.

Thanks,
@selzoc && Davis

keymon:
This is unfortunate. Sorry not replying before, I was waiting for the people to work on the PR to answer.,

The alternative proposed is not as suitable as this PR, I explain below.

> Service operation failures are already emitted as logs. We have some concerns about adding metrics on every request - currently there is no logic there. Have you also taken a look at the service-related audit events?

Both logs and audit events requires more scaffolding to be able to action alerts on it, meanwhile metrics get feed directly into the monitoring system and it can configured as any other alerting and monitoring we do in the platform. 

> Another option would be to monitor the service brokers themselves.

That requires change each one of the brokers, and in some cases 3rd party brokers which we do not. Also it only catches errors caused in the broker itself. If the broker is not reachable because a networking issue, for instance, the error would pass undetected.

We do not think this PR causes too much overhead on the metrics system and it is handy for operators. Why do you not consider as a solution to this problem? Is there any restriction or problem you foresee?

tcdowney:
@mattmcneeney @crawsible 

Do you have any SAPI/CAPI product thoughts about the above?

keymon:
> Right now you are suggesting adding metrics to routes at the front of the CC, what would you want these metrics to be, counts of status codes, percentiles?

This PR explains clearly. A counter of the HTTP status codes for critical endpoints would be great, far enough to trigger an alarm so that the operator can investigate.

For instance, I have a counter from CF with `cc.requests.service_bindings.put.http_status.5XX` I can alert if the counter increases. It will tell me that something got wrong when the user was doing a binding.

Currently, we only rely on the users complaining about it. 

I am going to put a real case: we had an incident recently where some users were unable to do bind on a specific broker. The broker was not reporting errors. The CC was reporting API errors, but we have set some alert threshold as API errors can be anything. We did not notice the issue until the tenants complained.

If we had a metric  like `cc.requests.service_bindings.put.http_status.5XX` we could have alerted immediately.

williammartin:
I should have said, "in an ideal world, what would you want these metrics to be?". I can absolutely see the value in something that indicates health of service operations that hit external resources, I'm questioning whether an incrementing counter on the front of the CC across global service resources is the best way to expose the information about a single problematic broker.

keymon:
We could have a counter of errors creating, deleting, binding services or similar.

cwlbraa:
@williammartin : I think @infews and the rest of the CAPI team has technical concerns about how this is implemented, and I think you're expressing the same sentiments when you say things like

> would it make sense then to take the rest of the CC stack out of the picture and emit metrics closer to where the CC communicates with brokers

and

> I'm questioning whether an incrementing counter on the front of the CC across global service resources under only the v2 API is the best way to expose the information about a single problematic broker

@keymon would it be sufficient if the CC reported those counters not as "here's the error code we returned to the user" but rather as "here's the errors we encountered when communicating with brokers"?

mattmcneeney:
Hey @keymon Did you manage to read @cwlbraa's response above? Let us know what you think!

keymon:
Yes, I think that would suffice if I understood correctly.

Basically, of you provide a way to learn from metrics from firehose if the
operations with brokers are failing, that would be great.

Otherwise, we always need something else: ad-hoc monitoring per broker, log
parsing, etc

PR #1169, opened 2018-06-20
Call binmode in a separate statement

petergtz:

    Fix Tempfile usage when using bits-service-client.

    While Tempfile's binmode method, according to its implementation (https://ruby-doc.org/core-2.4.0/IO.html#method-i-binmode), returns the Tempfile object, it seems that garbage collection does not consider this a proper reference (note that the documentation also does not state that it is returning the Tempfile object). Hence, whenever garbage collection kicks in, it collects the package reference and with it, deletes the temporary file, which causes a failure further down the line when trying to work with the temp file. That's what we've seen in RelInt's Hermione env. The fact we haven't seen this anywhere else is probably because GC is undeterministic and simply wasn't executed in this specific case. Bottom line is, we used the Tempfile API wrong.

    [#157250835]

### DIFF
--- a/lib/cloud_controller/packager/bits_service_packer.rb
+++ b/lib/cloud_controller/packager/bits_service_packer.rb
@@ -19,7 +19,8 @@ module CloudController
       private
 
       def create_temp_file_with_content(content)
-        package = Tempfile.new('package.zip').binmode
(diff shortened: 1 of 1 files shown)


tcdowney:
Thanks @petergtz, looks good to us!

PR #1205, opened 2018-08-23
Enable pluggable container orchestration using OPI

gdankov:
### A short explanation of the proposed change

This PR introduces the option to enable OPI (Orchestrator Provider Interface) in the Cloud Controller, which adds the capability to swap the container orchestrator of Cloud Foundry. You can find more information about it on [eirini](https://github.com/cloudfoundry-incubator/eirini) and [eirini-release](https://github.com/cloudfoundry-incubator/eirini-release) repositories.

In order to integrate OPI with CC, we have implemented two clients &mdash; [Apps](https://github.com/mnitchev/cloud_controller_ng/blob/eirini/lib/cloud_controller/opi/apps_client.rb) and [Instances](https://github.com/mnitchev/cloud_controller_ng/blob/eirini/lib/cloud_controller/opi/instances_client.rb) &mdash; that can talk directly to the OPI component. These clients replace Diego's `BbsAppsClient` and `BbsInstancesClient` respectively when OPI is enabled.
Whether OPI is enabled or not, is controlled by configuration flags in [1](https://github.com/mnitchev/cloud_controller_ng/blob/eirini/config/cloud_controller.yml#L253-L255), [2](https://github.com/mnitchev/cloud_controller_ng/blob/eirini/lib/cloud_controller/config_schemas/api_schema.rb#L289-L292), [3](https://github.com/mnitchev/cloud_controller_ng/blob/eirini/lib/cloud_controller/config_schemas/clock_schema.rb#L162-L165), [4](https://github.com/mnitchev/cloud_controller_ng/blob/eirini/lib/cloud_controller/config_schemas/deployment_updater_schema.rb#L63-L66) and [5](https://github.com/mnitchev/cloud_controller_ng/blob/eirini/lib/cloud_controller/config_schemas/worker_schema.rb#L156-L159). When disabled, Diego is used as before.

This PR also introduces a small change to `BbsAppsClient`'s interface. Instead of accepting a recipe object created by `AppRecipeBuilder`, now it accepts the process object and uses `AppRecipeBuilder` internally. We think this creates a more generic interface, since the recipe creation is a Diego implementation detail which we do not use in the new clients.

Merging this PR enables the Eirini project to work without needing to maintain a Cloud Controller fork. The original proposal for Eirini is [here](https://docs.google.com/document/d/1qs6UQQDWMkfOpY19XqS3CfvI00jCns876TjplJ6E95s/edit#heading=h.poicq8c1xfqn).

/cc @julz @andrew-edgar @JulzDiverse @mnitchev @suhlig

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -93,8 +93,6 @@ end
 group :development do
   gem 'byebug'
   gem 'debase', '>= 0.2.2.beta14'
-  gem 'guard'
-  gem 'guard-rspec'
   gem 'listen'
   gem 'roodi'
   gem 'ruby-debug-ide', '>= 0.7.0.beta4'
--- a/Gemfile
+++ b/Gemfile
@@ -93,6 +93,8 @@ end
 group :development do
   gem 'byebug'
   gem 'debase', '>= 0.2.2.beta14'
+  gem 'guard'
+  gem 'guard-rspec'
   gem 'listen'
   gem 'roodi'
   gem 'ruby-debug-ide', '>= 0.7.0.beta4'
--- a/lib/cloud_controller/config_schemas/deployment_updater_schema.rb
+++ b/lib/cloud_controller/config_schemas/deployment_updater_schema.rb
@@ -160,7 +160,6 @@ module VCAP::CloudController
         }
         # rubocop:enable Metrics/BlockLength
       end
-      # rubocop:enable Metrics/BlockLength
 
       class << self
         def configure_components(config); end
--- a/lib/cloud_controller/config_schemas/deployment_updater_schema.rb
+++ b/lib/cloud_controller/config_schemas/deployment_updater_schema.rb
@@ -60,6 +60,11 @@ module VCAP::CloudController
             optional(:temporary_oci_buildpack_mode) => enum('oci-phase-1', NilClass),
           },
 
+          opi: {
+            enabled: bool,
+            url: String,
+          },
+
           default_app_memory: Integer,
           default_app_disk_in_mb: Integer,
           maximum_app_disk_in_mb: Integer,
@@ -153,6 +158,7 @@ module VCAP::CloudController
           statsd_host: String,
           statsd_port: Integer,
         }
+        # rubocop:enable Metrics/BlockLength
       end
       # rubocop:enable Metrics/BlockLength
 
--- a/lib/cloud_controller/config_schemas/deployment_updater_schema.rb
+++ b/lib/cloud_controller/config_schemas/deployment_updater_schema.rb
@@ -158,8 +158,8 @@ module VCAP::CloudController
           statsd_host: String,
           statsd_port: Integer,
         }
-        # rubocop:enable Metrics/BlockLength
       end
+      # rubocop:enable Metrics/BlockLength
 
       class << self
         def configure_components(config); end
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -26,6 +26,8 @@ require 'cloud_controller/bits_service_resource_pool_wrapper'
 require 'cloud_controller/packager/local_bits_packer'
 require 'cloud_controller/packager/bits_service_packer'
 require 'credhub/client'
+require 'cloud_controller/opi/apps_client'
+require 'cloud_controller/opi/instances_client'
 
 require 'bits_service_client'
 
@@ -73,7 +75,7 @@ module CloudController
     end
 
     def bbs_apps_client
-      @dependencies[:bbs_apps_client] || register(:bbs_apps_client, build_bbs_apps_client)
+      @dependencies[:bbs_apps_client] || register(:bbs_apps_client, build_apps_client)
     end
 
     def bbs_stager_client
@@ -385,15 +387,27 @@ module CloudController
       )
     end
 
+    def build_apps_client
+      if config.get(:opi, :enabled)
+        build_opi_apps_client
+      else
+        build_bbs_apps_client
+      end
+    end
+
+    def build_opi_apps_client
+      ::OPI::Client.new(config.get(:opi, :url))
+    end
+
     def build_bbs_apps_client
       bbs_client = ::Diego::Client.new(
         url: config.get(:diego, :bbs, :url),
         ca_cert_file: config.get(:diego, :bbs, :ca_file),
         client_cert_file: config.get(:diego, :bbs, :cert_file),
         client_key_file: config.get(:diego, :bbs, :key_file),
-      )
+    )
 
-      VCAP::CloudController::Diego::BbsAppsClient.new(bbs_client)
+      VCAP::CloudController::Diego::BbsAppsClient.new(bbs_client, config)
     end
 
     def build_bbs_task_client
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -87,7 +87,7 @@ module CloudController
     end
 
     def bbs_instances_client
-      @dependencies[:bbs_instances_client] || register(:bbs_instances_client, build_bbs_instances_client)
+      @dependencies[:bbs_instances_client] || register(:bbs_instances_client, build_instances_client)
     end
 
     def traffic_controller_client
@@ -419,6 +419,18 @@ module CloudController
       )
 
       VCAP::CloudController::Diego::BbsTaskClient.new(bbs_client)
+    end
+
+    def build_instances_client
+      if config.get(:opi, :enabled)
+        build_opi_instances_client
+      else
+        build_bbs_instances_client
+      end
+    end
+
+    def build_opi_instances_client
+      ::OPI::InstancesClient.new(config.get(:opi, :url))
     end
 
     def build_bbs_instances_client
--- a/lib/cloud_controller/opi/apps_client.rb
+++ b/lib/cloud_controller/opi/apps_client.rb
@@ -67,6 +67,7 @@ module OPI
         environment: hash_values_to_s(vcap_application(process)),
         instances: process.desired_instances,
         droplet_hash: process.current_droplet.droplet_hash,
+        droplet_guid: process.current_droplet.guid,
         health_check_type: process.health_check_type,
         health_check_http_endpoint: process.health_check_http_endpoint,
         health_check_timeout_ms: timeout_ms,
--- a/lib/cloud_controller/opi/apps_client.rb
+++ b/lib/cloud_controller/opi/apps_client.rb
@@ -1,0 +1,134 @@
+require 'uri'
+require 'httpclient'
+require 'multi_json'
+require_relative '../../vcap/vars_builder'
+require 'json'
+require 'ostruct'
+
+module OPI
+  class Client
+    def initialize(opi_url)
+      @client = HTTPClient.new(base_url: URI(opi_url))
+    end
+
+    def desire_app(process)
+      process_guid = process_guid(process)
+      path = "/apps/#{process_guid}"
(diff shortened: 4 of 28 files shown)


julz:
nag, nag :-)

tcdowney:
> nag, nag :-)

@julz we've scheduled a review of this PR in our regular backlog since it's pretty meaty. Our team should have some folks reviewing it in the next week or two. 😃 

https://www.pivotaltracker.com/n/projects/966314

spec/spec_helper.rb
```diff
@@ -91,6 +91,7 @@
       mocks.verify_partial_doubles = true
     end
     rspec_config.exclusion_filter = :perm
+    rspec_config.filter_run_excluding opi: true
```
selzoc: could this not be combined with the `perm` exclusion?

spec/unit/lib/cloud_controller/diego/bbs_apps_client_spec.rb
```diff
@@ -2,72 +2,95 @@
 
 module VCAP::CloudController::Diego
   RSpec.describe BbsAppsClient do
-    subject(:client) { BbsAppsClient.new(bbs_client) }
+    subject(:client) {         BbsAppsClient.new(bbs_client, config) }
```
selzoc: have you run `bake`?  the big whitespace here seems to indicate that `rubocop` wasn't run?

lib/cloud_controller/dependency_locator.rb
```diff
@@ -73,7 +75,7 @@ def copilot_client
     end
 
     def bbs_apps_client
-      @dependencies[:bbs_apps_client] || register(:bbs_apps_client, build_bbs_apps_client)
+      @dependencies[:bbs_apps_client] || register(:bbs_apps_client, build_apps_client)
```
selzoc: can you give us more details as to why you chose the dependency locator as the layer at which the bbs/opi swap occurs?

tcdowney: For example, we were wondering if some type of [backend](https://github.com/cloudfoundry/cloud_controller_ng/tree/master/lib/cloud_controller/backends) adapter would work better? Or a "bridge component"-esque proxy external to CC that would convert BBS requests to OPI/Eirini requests?

Not really prescribing anything here, just want to have a discussion around the design.

lib/cloud_controller/config_schemas/deployment_updater_schema.rb
```diff
@@ -153,8 +158,8 @@ class DeploymentUpdaterSchema < VCAP::Config
           statsd_host: String,
           statsd_port: Integer,
         }
+        # rubocop:enable Metrics/BlockLength
```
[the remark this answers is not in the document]
gdankov: It doesn’t seem to change anything. Our guess is that it was placed there by mistake. We’ll move it back.

Thanks,
Georgi && @suhlig

Gemfile
```diff
@@ -93,6 +93,8 @@ end
 group :development do
   gem 'byebug'
   gem 'debase', '>= 0.2.2.beta14'
+  gem 'guard'
```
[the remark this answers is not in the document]
gdankov: You are right, it is orthogonal. We added guard to help our development process. We will remove it.

Thanks,
Georgi && @suhlig

lib/cloud_controller/opi/apps_client.rb
```diff
@@ -0,0 +1,134 @@
...
+      desired_lrp_response = recursive_ostruct(JSON.parse(response.body))
+      desired_lrp_response.desired_lrp
+    end
+
+    def stop_app(process_guid)
+      path = "/apps/#{process_guid}/stop"
+      @client.put(path)
+    end
+
+    def bump_freshness; end
```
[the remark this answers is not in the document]
gdankov: Yes, it is. It is not yet implemented by OPI, but since it's always called by the [ProcessesSync](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/cloud_controller/diego/processes_sync.rb#L67) (as long as there are no Errors), we have to provide it.

Thanks,
Georgi && @suhlig

lib/cloud_controller/opi/apps_client.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    private
+
+    def desire_body(process)
+      timeout_ms = (process.health_check_timeout || 0) * 1000
+
+      body = {
+        process_guid: process_guid(process),
+        docker_image: process.current_droplet.docker_receipt_image,
+        start_command: process.command.nil? ? process.detected_start_command : process.command,
```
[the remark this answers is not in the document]
gdankov: Ok, we’ll change it.

lib/cloud_controller/opi/instances_client.rb
```diff
@@ -0,0 +1,74 @@
...
+        raise e
+      end
+      process_guid = resp_json['process_guid']
+      resp_json['instances'].map do |instance|
+        ActualLRP.new(ActualLRPKey.new(instance['index'], process_guid), instance['state'])
+      end
+    end
+
+    def desired_lrp_instance(process)
+      DesiredLRP.new('a')
```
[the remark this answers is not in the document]
gdankov: This method was stubbed because OPI does not currently support the notion of `isolation_segments` and at the same time CC expects from each process to always have a isolation segment &mdash; [link](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/cloud_controller/diego/reporters/instances_stats_reporter.rb#L40).
We'll modify the method to add a comment and make our intentions more explicit.

spec/integration/opi_instances_client_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      context 'when process guid does not exist' do
+        let(:process) { double(guid: 'jeff-goldblum') }
+
+        it 'raises an error' do
+          expect { client.lrp_instances(process) }.to raise_error(CloudController::Errors::NoRunningInstances, 'No running instances')
+        end
+      end
+    end
+
+    context 'Get an app' do
```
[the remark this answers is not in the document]
gdankov: We'll remove it.

spec/spec_helper.rb
gdankov: We’re not that familiar with the CC codebase and ruby and we’re not sure we understand what the perm exclusion is representing in this case. To us it looks like the perm tests are testing something completely different and in this case it wouldn’t make much sense to combine them. What do you think?

spec/unit/lib/cloud_controller/diego/bbs_apps_client_spec.rb
gdankov: We ran `bundle exec rake rubocop:auto_correct` multiple times, but this issue was not caught. We will remove the whitespaces manually.

lib/cloud_controller/dependency_locator.rb
gdankov: We wanted to be as least intrusive as possible and to make the minimum amount of changes needed to get OPI running from the CloudController side. Since most of the scheduler code in CC already works for us, we decided to reuse as much of it as possible, which meant that we only had to change the clients that directly communicate with the scheduler. It seems to us that if we create our own backend adapter we’ll have to re-implement a huge chunk of the Diego code that is already doing what we need.

Regarding the bridge-component idea, that would mean we’d have to convert the Diego recipes to OPI native requests. That approach would closely couple us with the Diego implementation, which is not what our goal is. Additionally, as a result of creating our OPI clients, we had to move the BBS recipe building logic (which is a BBS implementation detail) in the BBS clients, which creates a cleaner abstraction on this level in our opinion.

Note that we’re not that familiar with the CC code and we’re always open for a discussion and ideas. 🙂

julz:
Hi @cwlbraa! 

I've tried to give some high level answers here, but I'd love to hop on a call and have a higher bandwidth conversation if you'd like more detail, if y'all would be up for that

Answering same numbers as your questions:

1. This PR only does apps and instances. We do have a staging PR, which we're planning to submit separately. Right now tasks are out of scope for MVP of Eirini, but I'd expect tasks to happen soon after the first milestone. `cf push` does work (in fact we have most of the CATs are passing!), either using Diego for staging (as of this PR) or using our native stager (a follow-on PR).  Staging and all the CC stuff still works in terms of droplets and is pretty much unchanged (so things like rolling back droplet versions and patching should work fine), in the k8s backend we create a Docker Image for Kubernetes to use using a super-minimal registry-like endpoint (which we're in the midst of landing in bits-service) that serves an Image containing the current rootfs and the requested given droplet as layers. This way CC doesn't need to learn about images, and Kubernetes doesn't have to learn about droplets.

1. As part of the initial Incubation proposal for Eirini we decided, based on lessons learnt from the DEA->Diego transition, to take migration out of initial scope. So the idea for right now is you either use Eirini, or Diego, and you decide when you create the cluster. We might well decide to come back around and implement a migration path if the demand is there from people, but we intentionally haven't put much effort in for that yet. (I _think_ tho that the PR here would work out quite nicely for migration if/when we decide to do that, in a similar way to how the DEA->Diego transition was done, as it happens)

1. Yah, we considered moving Diego underneath the OPI boundary but decided against it for now for a couple reasons. Firstly, our intention is quite specifically _not_ to replicate all Diego functions in OPI to begin with: if you want all the bells and whistles, you should use Diego; if you want a cf push experience without learning a new scheduler, Eirini may be for you. Secondly, and more importantly, we committed as part of the incubation proposal to try really hard to avoid slowing Diego down or making it hard for Diego to add features, even if that means a bit of pain on our side keeping parity. By keeping the interfaces separate we make sure not to get in Diego's way, and we avoid having to map the whole surface area of Diego over to OPI prematurely. There're definitely some real trade-offs here, it's less "nice" until we unify it back up later, but fwiw I think (having been involved in a few transitions now, e.g. dea->diego, garden-linux->garden-runc, runc->containerd) it's the right trade-off given our current requirements and priorities

cwlbraa:
Hi @julz! The time zone shift is so brutal, otherwise we never would've immediately hopped on a call. We definitely should at some point in the future, but your prose is ever-so-helpful and explanatory that it may not be entirely necessary right now.  

>  By keeping the interfaces separate we make sure not to get in Diego's way, and we avoid having to map the whole surface area of Diego over to OPI prematurely.

Totally, and imho the whole Diego API interface with all its actions and whatnot is probably much more expressive than what the CC actually requires. I think the only one of my questions that's still unanswered is how Diego might conform to the OPI interface. It seems like, the way the interface is defined right now, someone would have to build a sort of opi-implementing stateless little translation server in front of the bbs.

julz:
Yah, very much spit-balling here but I think a decently plausible eventual state would be Diego moving behind the OPI backend with a stateless component that'd look suspiciously like the old nsync bridge (except without the sync loop) before we pushed that all back in to CC. Zach and I discussed a while back re-externalising the sync loop too (I personally think it was always kind of a mistake to conflate an api server and an orchestrator and nsync was a pretty good idea) with it pulling from a CC-based desired state api, which might also make stuff like container-networking/istio stuff a bit less coupled - in other words I think that could be justified quite separately from OPI. Needless to say I don't propose any of that in the short term, but I think the refactors seem like they'd work pretty cleanly down the road.

gdankov:
Hi @selzoc @tcdowney,

We have made the requested changes, but now the `snyk` integration fails and we don't have the permissions to see why. Can you please check and get back to us?

Thanks,
Georgi

/cc @mnitchev @suhlig

charleshansen:
@gdankov don't worry about Snyk. There was a thing with Nokogiri on our side, nothing to do with your PR. It should be fixed now even, but I don't know how to re-run the check other than pushing an extra commit.

PR #1222, opened 2018-09-27
Allows service brokers to be registered more than once with different names

lurraca:
This PR is the initial effort to solve the problem about multiple brokers registration within the same cf deployment. 
For more detailed description and a list of the considered solutions have  a look at the following [document](https://docs.google.com/document/d/1_OBnFCsL3ru43PEXocsCc3EuGaM0YLHjr0iAoXnakt4/edit#heading=h.wkjkf6foqxir).

As part of the PR we :
- remove uniqueness constraints on DB indexes and models on services related fields.
- add validation that prevents registering a broker whose catalog contains multiple services with the same service name

Related stories [#159950844](https://www.pivotaltracker.com/story/show/159950844), [#159950846](https://www.pivotaltracker.com/story/show/159950846)

BR,
Luis and @georgi-lozev on behalf of SAPI Team

### DIFF
--- a/app/models/services/service.rb
+++ b/app/models/services/service.rb
@@ -73,12 +73,7 @@ module VCAP::CloudController
       validates_presence :description, message: 'is required'
       validates_presence :bindable, message: 'is required'
       validates_url :info_url, message: 'must be a valid url'
-      validates_unique :unique_id, message: Sequel.lit('Service ids must be unique')
       validates_max_length 2048, :tag_contents, message: Sequel.lit("Service tags for service #{label} must be 2048 characters or less.")
-
-      validates_unique :label, message: Sequel.lit('Service name must be unique') do |ds|
-        ds.exclude(service_broker_id: nil)
-      end
     end
 
     serialize_attributes :json, :tags, :requires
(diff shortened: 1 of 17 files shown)


lurraca:
@Gerg @zrob This PR introduces a migration. If CAPI could take a look it would be appreciated.

georgi-lozev:
CC @mattmcneeney

williammartin:
This is currently failing a couple of codeclimate checks but neither are really valid failures. The DB migrations go untested reducing diff coverage, and they are similar, triggering a warning on potentially redundant code.

Samze:
When reviewing this commit this change will probably stand out to you:
```
Sequel.migration do
  up do
    if indexes(:services)[:services_label_provider_index]
      alter_table :services do
        drop_index :label, name: :services_label_provider_index
      end
    end
  end
```

You can read more about why we have to do this in the [updated wiki](https://github.com/cloudfoundry/cloud_controller_ng/wiki/CAPI-Migration-Style-Guide#known-differences-between-mysql-and-postgres). We felt that it was better to check if the index exists rather than switch on if we were running on MySQL, but we're open to other ideas.

PR #1227, opened 2018-10-04
Check for nil entries in bulk sync call

KauzClay:
[#160866520]

https://www.pivotaltracker.com/story/show/160866520
As noted in the above story, we observed a problem with route_syncer when a route was being deleted at the same time a bulk_sync was occurring. Basically, the bulk_sync would try to 'sync' a route that had already been deleted, causing some nil error.

To fix this, we added checks to remove any nil objects that were deleted before making the bulk_sync call on the copilot client.

### DIFF
--- a/lib/cloud_controller/copilot/adapter.rb
+++ b/lib/cloud_controller/copilot/adapter.rb
@@ -54,9 +54,16 @@ module VCAP::CloudController
 
         def bulk_sync(routes:, route_mappings:, processes:)
           copilot_client.bulk_sync(
-            routes: routes.map { |r| { guid: r.guid, host: r.fqdn, path: r.path } },
-            route_mappings: route_mappings.map { |rm| { capi_process_guid: rm.process.guid, route_guid: rm.route.guid, route_weight: rm.weight } },
-            capi_diego_process_associations: processes.map do |process|
+            routes: routes.compact.map { |r| { guid: r.guid, host: r.fqdn, path: r.path } },
+            route_mappings: route_mappings.compact.map do |rm|
+              next if rm.process.nil? || rm.route.nil?
+              {
+                capi_process_guid: rm.process.guid,
+                route_guid: rm.route.guid,
+                route_weight: rm.weight
+              }
+            end.compact,
+            capi_diego_process_associations: processes.compact.map do |process|
               {
                   capi_process_guid: process.guid,
                   diego_process_guids: [Diego::ProcessGuid.from_process(process)]
--- a/lib/cloud_controller/copilot/adapter.rb
+++ b/lib/cloud_controller/copilot/adapter.rb
@@ -52,23 +52,11 @@ module VCAP::CloudController
           end
         end
 
-        def bulk_sync(routes:, route_mappings:, processes:)
+        def bulk_sync(routes:, route_mappings:, capi_diego_process_associations:)
           copilot_client.bulk_sync(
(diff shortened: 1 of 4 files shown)


lib/cloud_controller/copilot/adapter.rb
```diff
@@ -54,9 +54,16 @@ def delete_capi_diego_process_association(process)
...
-            capi_diego_process_associations: processes.map do |process|
+            routes: routes.compact.map { |r| { guid: r.guid, host: r.fqdn, path: r.path } },
+            route_mappings: route_mappings.compact.map do |rm|
+              next if rm.process.nil? || rm.route.nil?
+              {
+                capi_process_guid: rm.process.guid,
+                route_guid: rm.route.guid,
+                route_weight: rm.weight
+              }
+            end.compact,
```
monamohebbi: The `.compact` at the end of the end block doesn't do anything. At this point it's acting on an array of hashes, and no hash is going to be considered nil.

monamohebbi: Hi @KauzClay 

We prefer to reject invalid values up front like so: ```
route_mappings: route_mappings.compact.reject { |rm| rm.process.nil? || rm.route.nil? }.map do |rm|
```
and then you don't need the `.compact` at the end of the loop.

**Comment by monamohebbi** [comment]
Hello @KauzClay 

Thanks for nailing down the problem.

Could you make the suggested change to the loop to reject nil values up front?

Also, TravisCI is failing, but most likely if you rebase your PR against master that will fix the problem.

Thanks,
@monamohebbi and @ericpromislow,  your CAPI Community Pair

File: lib/cloud_controller/copilot/adapter.rb [hunk 2]
```diff
@@ -54,9 +54,16 @@ def delete_capi_diego_process_association(process)
...
         def bulk_sync(routes:, route_mappings:, processes:)
           copilot_client.bulk_sync(
-            routes: routes.map { |r| { guid: r.guid, host: r.fqdn, path: r.path } },
-            route_mappings: route_mappings.map { |rm| { capi_process_guid: rm.process.guid, route_guid: rm.route.guid, route_weight: rm.weight } },
-            capi_diego_process_associations: processes.map do |process|
+            routes: routes.compact.map { |r| { guid: r.guid, host: r.fqdn, path: r.path } },
+            route_mappings: route_mappings.compact.map do |rm|
+              next if rm.process.nil? || rm.route.nil?
+              {
+                capi_process_guid: rm.process.guid,
```
tcdowney: Calling `rm.process` and `rm.route` actually does a database query under the hood. Since we're fetching all of these resources upfront in the calling class, we're going to be making **a lot** of queries against ccdb here in large environments ([see n+1 queries problem](https://semaphoreci.com/blog/2017/08/09/faster-rails-eliminating-n-plus-one-queries.html)).

If we only need the guids you could do something like `rm.process_guid`, but I'm not sure that will work with your existence check above.

I think a better solution would be to do this work up front in the calling class if possible. We should also implement some form of pagination there as well (might be out of scope for this part, but it needs to happen before this could go on a production environment)... 🙂

charleshansen:
@KauzClay there are some rubocop failures, but otherwise looks good. Can you fix those?

utako:
@charleshansen fixed the rubocop failures. should be good to go! we'll be addressing batching in [this story](https://www.pivotaltracker.com/story/show/161293119)

tcdowney:
Looks good, thanks everyone!

PR #1236, opened 2018-10-17
add batching to Copilot::Sync

utako:

  Add batching to Copilot::Sync per our discussion with @tcdowney and @cwlbraa 

  * Copilot::Sync only batches queries to the database.
  * It does not send batched resources to Copilot. Sending batched resources will be a bigger change and will be considered in the future.

[#161293119]

  When there are hella routes, route mappings and processes, database queries should be less expensive

https://github.com/cloudfoundry/cloud_controller_ng/pull/1227

### DIFF
--- a/lib/cloud_controller/copilot/sync.rb
+++ b/lib/cloud_controller/copilot/sync.rb
@@ -1,13 +1,15 @@
 module VCAP::CloudController
   module Copilot
     class Sync
+      BATCH_SIZE = 500
+
       def self.sync
         logger = Steno.logger('cc.copilot.sync')
         logger.info('run-copilot-sync')
 
-        routes = Route.eager(:domain).all
-        route_mappings = RouteMappingModel.eager(:process).all
-        web_processes = ProcessModel.where(type: ProcessTypes::WEB).all
+        routes = batch(route_batch_query)
+        route_mappings = batch(route_mappings_batch_query)
+        web_processes = batch(processes_batch_query)
 
         Adapter.bulk_sync(
           routes: routes.map { |r| { guid: r.guid, host: r.fqdn, path: r.path } },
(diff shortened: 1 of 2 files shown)


infews:
Did you take a look at the `SequelPaginator`? Sequel's datasets allow walking a query in pages/batches like you built up here.

utako:
> Did you take a look at the `SequelPaginator`? Sequel's datasets allow walking a query in pages/batches like you built up here.

No, but that looks appropriate. We wanted to adhere to the same querying pattern that was built in the Sync queries for tasks and LRPs. We could submit another PR to convert this and the rest of the batched queries to use the SequelPaginator for consistency. Happy for CAPI to do it, too. 😄

infews:
@tcdowney What do you think about @utako 's comment about query consistency?

charleshansen:
tldr, we could maybe make improvements, but I think we should merge the PR.

I looked at it again, you wouldn't actually use the `SequelPaginator` itself, but the query would be simpler to read if you used the Sequel pagination extension, like https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/cloud_controller/paging/sequel_paginator.rb#L12. On the other hand, it does make sense for the syncers to use the same batching strategy, as Utako said.

At this point, the benefit of any further refactoring is questionable and we should merge the PR unless there are other objections (I'm cool with the Proc usage at this point). If we decide we want to use sequel for pagination, we should do it everywhere and not just right here. That larger question can be discussed outside of this PR. 

Waiting for a +1 before actually merging...

tcdowney:
+1 to what @charleshansen said above. I think the Sequel Paginator suggestion that @infews had might be something we (CAPI) look into using across the sync jobs eventually, but not something I'd want to block this PR on (or would expect the Routing team to take on).

Looks good to me!

PR #1240, opened 2018-10-22
Service instance purger should record a purge audit event

### LINKED ISSUE #1208: Difficult to establish purge of Service Instances from Audit Events
## Issue
CF Audit Events don't differentiate between 'delete' and 'purge' scenarios. 

## Context

`cf purge-service-instance <instance_name>` triggers a purge of the service instance which leaves behind an orphaned deployment on Bosh. Since purge of a service instance results in a forced deletion, service broker does not get notified in this case. This causes orphaned resources on IaaS platform which contribute to stale resources piling up + cost. We @ SAP wanted to automatically cleanup/monitor such orphaned deployments. At the same time we want to be doubly sure that such deployments are only a result of CF purge and not created via some different mechanism.

As of now we are unable to figure out this information from CF Events since it reports purge scenarios with `audit.service.delete`. Additionally we also see a log in CF deployment VM on Bosh which mentions a pattern like `DELETE "/v2/service_instances/<GUID>?async=true&purge=true"`. However, depending on CF events for such information seems to be a better way to establish such user actions.

What should be the ideal approach to handle cleanup of such deployments:

i) Should cf purge-service-instance internally call a `cf delete-service` before going for the actual purge - so that in case broker is up and running, can ensure a proper cleanup ?

ii) Should cleanup of such deployments be handled by broker automatically ? 

iii) Should CF differentiate a purge action explicitly in the Audit Events ?

## Steps to Reproduce

`cf purge-service-instance <instance_name>`

## Expected result

`audit.service.purge` audit event in CF

## Current result

`audit.service.delete` audit event is seen

henryaj:
Previously we recorded an 'audit.service_instance.delete' event when a
service instance was purged, which makes it harder for operators to
distinguish between purge and delete.

fixes #1208 

cc @georgi-lozev

### DIFF
--- a/app/controllers/services/lifecycle/service_instance_purger.rb
+++ b/app/controllers/services/lifecycle/service_instance_purger.rb
@@ -29,7 +29,7 @@ module VCAP::CloudController
         end
 
         service_instance.destroy
-        @event_repository.record_service_instance_event('delete', service_instance, nil)
+        @event_repository.record_service_instance_event('purge', service_instance, nil)
       end
 
       logger.info("successfully purged service instance #{service_instance.guid}")
--- a/spec/unit/controllers/services/lifecycle/service_instance_purger_spec.rb
+++ b/spec/unit/controllers/services/lifecycle/service_instance_purger_spec.rb
@@ -14,11 +14,11 @@ module VCAP::CloudController
         expect(service_instance).not_to exist
       end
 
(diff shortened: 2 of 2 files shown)


henryaj:
Today's GitHub outage seems to have stopped this PR from triggering Travis et al. I'll close and re-open.

PR #1252, opened 2018-11-07
Fix ./bin/console script

Gerg:
- This script started depending on `DB_CONNECTION_STRING` to run locally
- My guess is this was broken by ccbf6c3266e2bf999bc0a64753b19d81c05f425f
- There is probably a more correct way to fix this, but this will work if you have mysql and the database `cc_test_` locally

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -2,5 +2,6 @@
 
 require File.expand_path('../config/boot', __dir__)
 ENV['RAILS_ENV'] ||= 'production'
+ENV['DB_CONNECTION_STRING'] ||= 'mysql2://root:password@localhost:3306/cc_test_'
 
 require 'cloud_controller/console'


tcdowney:
@Gerg does this still work with the `bin/console` scripts within our BOSH deployments? This seems like it might be a good candidate for an `.envrc` file. 🤔 💭

Gerg:
I just tried it on a bosh lite and it works.

bin/console
```diff
@@ -2,5 +2,6 @@
 
 require File.expand_path('../config/boot', __dir__)
 ENV['RAILS_ENV'] ||= 'production'
+ENV['DB_CONNECTION_STRING'] ||= 'mysql2://root:password@localhost:3306/cc_test_'
```
ericpromislow: Couple of issues here:
1. When I tried this, I needed to set the DB to `cc_test` instead of `cc_test_`
2. Your PR doesn't check to see if `MYSQL_CONNECTION_PREFIX` or `POSTGRES_CONNECTION_PREFIX` are set before assuming the standard DB_CONNECTION_PREFIX (see https://github.com/cloudfoundry/cloud_controller_ng/pull/1117 )

ericpromislow: Here's what I would do:
```
#!/usr/bin/env ruby

def ensure_db_connection_string
  ENV['DB_CONNECTION_STRING'] ||=
    if ENV['POSTGRES_CONNECTION_PREFIX']
      ENV['POSTGRES_CONNECTION_PREFIX'] + '/cc_test'
    elsif ENV['MYSQL_CONNECTION_PREFIX']
      ENV['MYSQL_CONNECTION_PREFIX'] + '/cc_test'
    else
      'mysql2://root:password@localhost:3306/cc_test'
    end
end

require File.expand_path('../config/boot', __dir__)
ENV['RAILS_ENV'] ||= 'production'
ensure_db_connection_string

require 'cloud_controller/console'
```

Don't know what's going on with that prefix. Does this change work for you?

@ericpromislow

selzoc:
@Gerg want to take a look at @ericpromislow 's suggestions?  Or is this something you no longer care about?

Thanks,
Chris and Eric

selzoc:
Fixed in 0f012e145326d66c8475bac968e136911532ce7d

PR #1257, opened 2018-11-16
Introduce configurable bbs client timeouts

petergtz:

Make the timeouts used by diego bbs client configurable.

We noticed that the hardcoded values of 10s is not enough in environments with more than 100,000 apps. This PR allows a configuration that can be adjusted to the specific environment.


lib/cloud_controller/config_schemas/api_schema.rb
```diff
@@ -274,6 +274,9 @@ class ApiSchema < VCAP::Config
               ca_file: String,
               cert_file: String,
               key_file: String,
+              optional(:connect_timeout) => Integer,
```
tcdowney: Is there a corresponding [capi-release](https://github.com/cloudfoundry/capi-release) PR to allow for these to be configurable from the BOSH job? I glanced briefly, but wasn't able to find one.

lib/cloud_controller/dependency_locator.rb
```diff
@@ -434,14 +413,23 @@ def build_opi_instances_client
...
+      connect_timeout = config.get(:diego, :bbs, :connect_timeout)
+      send_timeout = config.get(:diego, :bbs, :send_timeout)
+      receive_timeout = config.get(:diego, :bbs, :receive_timeout)
+
+      ::Diego::Client.new(
         url: config.get(:diego, :bbs, :url),
         ca_cert_file: config.get(:diego, :bbs, :ca_file),
         client_cert_file: config.get(:diego, :bbs, :cert_file),
         client_key_file: config.get(:diego, :bbs, :key_file),
+        connect_timeout: connect_timeout.nil? ? 10 : connect_timeout,
```
tcdowney: For new configurable properties like this we are defaulting them at the BOSH job spec level (like this [UAA timeout](https://github.com/cloudfoundry/capi-release/blob/972a14749c928de1283908023deac75f6708fcfc/jobs/cloud_controller_ng/spec#L655-L657) for example) since this gives CF platform operators a single place to look for defaults. I'd prefer for this to be defaulted there and this code could just read directly from the config.

lib/cloud_controller/config_schemas/api_schema.rb
petergtz: Ah, my mistake yes. Forgot to actually open that. Here it is: https://github.com/cloudfoundry/capi-release/pull/110

lib/cloud_controller/dependency_locator.rb
petergtz: The reason, why I provided this here is, so that the capi-release PR does not have to be synced with the CC sub-module bump. To avoid breaking the build, I could submit a follow-up PR which takes the default away here. Or do you accept a breakage, or a synced PR+sub-module bump?

tcdowney: Hmm that's a good point, but I don't think this particular set of changes should break the build if we merge the https://github.com/cloudfoundry/capi-release/pull/110/ PR first. In general though, our usual protocol is to merge them at roughly the same time and this usually results in no build breakages due to how our pipelines are structured.

I think it's simpler overall to just omit the defaults in this PR and have the pair that merges this take https://github.com/cloudfoundry/capi-release/pull/110 in to consideration.

petergtz: Okay. Will update the PR soon.

tcdowney:
@petergtz it looks like there are some unit test failures with the most recent commit. I think you'll need to add the new config properties to the test config fixture here:
https://github.com/cloudfoundry/cloud_controller_ng/blob/bda08a42825cf302b85072f4f6e5a6b4b7b39877/config/cloud_controller.yml#L247-L251

You'll also need to add the config to this schema: https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/cloud_controller/config_schemas/deployment_updater_schema.rb#L47

Thanks!

PR #1260, opened 2018-11-21
add container_purpose to container network info

ameowlia:
- so it can be known if a container is an app, staging, or an app task
- so network policies can be applied for running vs staging similar to ASGs
[#161904323]

Signed-off-by: Michael Oleske <moleske@pivotal.io>

### DIFF
--- a/lib/cloud_controller/diego/protocol.rb
+++ b/lib/cloud_controller/diego/protocol.rb
@@ -66,7 +66,7 @@ module VCAP::CloudController
           'etag'                            => process.updated_at.to_f.to_s,
           'allow_ssh'                       => process.enable_ssh,
           'ports'                           => OpenProcessPorts.new(process).to_a,
-          'network'                         => ContainerNetworkInfo.new(process.app, 'desired app message').to_h,
+          'network'                         => ContainerNetworkInfo.new(process.app, Protocol::ContainerNetworkInfo::APP).to_h,
           'volume_mounts'                   => AppVolumeMounts.new(process.app),
           'isolation_segment'               => VCAP::CloudController::IsolationSegmentSelector.for_space(process.space),
         }.merge(LifecycleProtocol.protocol_for_type(process.app.lifecycle_type).desired_app_message(process))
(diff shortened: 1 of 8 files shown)


ameowlia:
Hi Capi, 

We aren't sure when this file is used? https://github.com/cloudfoundry/cloud_controller_ng/blob/87e07f97be63dd272801556e44d90be4a1b3e033/lib/cloud_controller/diego/protocol.rb

It calls the network_info method that we altered, but we never saw it appear in our container metadata when we were running tests. 

We got everything else working the way we expected, but just wondering what that other case is? 

Thanks, 
Amelia and @xanderstrike

monamohebbi:
> we never saw it appear in our container metadata when we were running tests.

I think protocol.rb gets called via the runner and the runner is called via the process observer. That means if you scale, start, stop, enable ssh, etc, your test app you should be able to see your 'desired app message' field in there. 

Design-wise, ASGs don't seem to pass metadata around like this or declare their parent object type, it seems a bit unfortunate that y'alls ContainerNetworkInfo object needs to know what "type" its parent object is. It kind of denormalizes the api a bit.

On a style PR-mergability note, it'd be good to see the strings you're using `const`ed out somewhere as class or module variables.

XanderStrike:
Historically, capi kept track of both container state and which ASGs should be applied. We're trying to take that responsibility and give it to the policy-server who's in charge of C2C policies. That's why we never needed container state before, because y'all did all the logic. We'd prefer to control that logic and take it off your hands and have all the policies come from the same place, but to do that we need the container state.

We'll move those strings into constants defined under ContainerNetworkInfo, good advice thank you!

moleske:
So @XanderStrike, @monamohebbi, and @cwlbraa has conversation regarding the naming of the field.  The issue is we have a tautological field.  We're going to brainstorm on some naming ideas (like `egress_group`) and see if we can do better

XanderStrike:
We have renamed the field from `container_purpose` to `container_workload`, hopefully this is satisfactory!

lib/cloud_controller/diego/protocol.rb
```diff
@@ -66,7 +66,7 @@ def desire_app_message(process, default_health_check_timeout)
           'etag'                            => process.updated_at.to_f.to_s,
           'allow_ssh'                       => process.enable_ssh,
           'ports'                           => OpenProcessPorts.new(process).to_a,
-          'network'                         => ContainerNetworkInfo.new(process.app).to_h,
+          'network'                         => ContainerNetworkInfo.new(process.app, 'desired app message').to_h,
```
cwlbraa: this looks wrong to me, should this be `Protocol::ContainerNetworkInfo::APP`?

XanderStrike: you got it amigo

PR #1261, opened 2018-11-28
Rename ModelController run/enqueue delete method

henryaj:
The method `enqueue_deletion_job` technically only enqueues if the `async` param is passed - otherwise it runs synchronously. Renamed the method to be clearer.

https://www.pivotaltracker.com/story/show/159951480


charleshansen:
Seems like a good improvement to me. I never like methods that seem asynchronous but are occasionally synchronous.

PR #1299, opened 2019-02-27
Respect max polling duration for instance operations set by the broker

AartiKriplani:
This (OSBAPI specification change)[[openservicebrokerapi/servicebroker#627](https://github.com/openservicebrokerapi/servicebroker/pull/627)] introduces a  timeout value (at the catalog level for a service) a broker to specify a polling duration other than what the platform would default to.
This commit implements that change on the Cloud Controller API for operations on **service instances**.  Operations on bindings will be done as part of a following PR.

More details in the following stories:
https://www.pivotaltracker.com/story/show/163861066
https://www.pivotaltracker.com/story/show/163861093
https://www.pivotaltracker.com/story/show/163861101

Aarti, @williammartin , @nmaslarski

### DIFF
--- a/app/jobs/services/service_instance_state_fetch.rb
+++ b/app/jobs/services/service_instance_state_fetch.rb
@@ -43,6 +43,15 @@ module VCAP::CloudController
 
         private
 
+        def new_end_timestamp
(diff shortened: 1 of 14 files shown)


georgi-lozev:
We're don't think the codeclimate issues are relevant for this case. They don't seem to be compliant with our rubocop rules, which already pass, so we'll proceed with the merge.

PR #1310, opened 2019-03-15
Fix dummy desired LRP for opi

viovanov:
The current ccng is broken for opi/eirini, because of [this change](https://github.com/cloudfoundry/cloud_controller_ng/commit/118677a7d517e89c2222b9630215bbfb69662885#diff-87c4a5fb697b073910f9ee2e7c117737R73).
This will fix the problem.

ccng working with eirini


viovanov:
ping @ericpromislow

PR #1318, opened 2019-04-08
Allow cancel of the creation of an asynchronous service instance

AartiKriplani:
Before this commit CC was rejecting any request to a service instance if
there was ongoing operation in progress for it.
As part of [this OSBAPI spec](https://github.com/openservicebrokerapi/servicebroker/pull/581) a behaviour for the broker was defined for
situations where delete is sent while create is in progress:
It can cancel the create or it can reject the delete because of
the ongoing operation.

With this commit CC forwards the delete request to the broker
and acts accordingly to the response.
By allowing delete while create is in progress, there is a possiblity
that the delayed job for create runs at the same time as the delete
request and cause the instance to be in bad state. To avoid this race
condition, we have added a check on the delayed worker job to check the
last operation to be the same for which it was intended for (when
locking).

More details can be found [here](https://www.pivotaltracker.com/story/show/164046291).

### DIFF
--- a/app/jobs/services/service_instance_state_fetch.rb
+++ b/app/jobs/services/service_instance_state_fetch.rb
@@ -21,6 +21,8 @@ module VCAP::CloudController
           service_instance = ManagedServiceInstance.first(guid: service_instance_guid)
           return if service_instance.nil?
 
+          @intended_operation = service_instance.last_operation
+
           client = VCAP::Services::ServiceClientProvider.provide(instance: service_instance)
 
           last_operation_result = client.fetch_service_instance_last_operation(service_instance)
@@ -50,6 +52,8 @@ module VCAP::CloudController
         def update_with_attributes(last_operation, service_instance)
           ServiceInstance.db.transaction do
             service_instance.lock!
+            return unless @intended_operation == service_instance.last_operation
+
             service_instance.save_and_update_operation(
               last_operation: last_operation.slice(:state, :description)
             )
--- a/spec/unit/controllers/services/service_instances_controller_spec.rb
+++ b/spec/unit/controllers/services/service_instances_controller_spec.rb
@@ -3235,8 +3235,8 @@ module VCAP::CloudController
           end
         end
 
-        context 'and the instance operation is in progress' do
-          let(:last_operation) { ServiceInstanceOperation.make(state: 'in progress') }
+        context 'and the instance operation is update in progress' do
+          let(:last_operation) { ServiceInstanceOperation.make(state: 'in progress', type: 'update') }
           before do
             service_instance.service_instance_operation = last_operation
           end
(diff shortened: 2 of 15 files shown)


AartiKriplani:
@tcdowney @Gerg @zrob : maybe you guys want to take a look at this PR since it involves handling some concurrent requests/processes etc. Thanks!

AartiKriplani:
codeclimate issue is about long method (> 25 lines), which I think can be ignored, since its handling different error types and is not really easy to refactor or shorten.

spec/unit/controllers/services/service_instances_controller_spec.rb
```diff
@@ -3245,6 +3245,21 @@ def stub_delete_and_return(status, body)
             delete "/v2/service_instances/#{service_instance.guid}"
             expect(last_response.status).to eq 409
             expect(last_response.body).to match 'AsyncServiceInstanceOperationInProgress'
+            # expect(last_response.status).to eq(204)
```
monamohebbi: Can these be deleted?

app/jobs/services/service_instance_state_fetch.rb
```diff
@@ -50,6 +52,8 @@ def repository
         def update_with_attributes(last_operation, service_instance)
           ServiceInstance.db.transaction do
             service_instance.lock!
+            return unless @intended_operation == service_instance.last_operation
```
tcdowney: We're a little confused about the difference between `intended_operation` and `last_operation`. It looks like one is the last operation as reported by the broker and one is what we have recorded in CCDB? To accomplish this:

> By allowing delete while create is in progress, there is a possiblity
that the delayed job for create runs at the same time as the delete
request and cause the instance to be in bad state. To avoid this race
condition, we have added a check on the delayed worker job to check the
last operation to be the same for which it was intended for (when
locking).

Wondering if there's a way we can make it more clear that this is what we're trying to accomplish here. 🤔

tcdowney:
> To avoid this race
condition, we have added a check on the delayed worker job to check the
last operation to be the same for which it was intended for (when
locking)
What happens if the create job gets enqueued, but the delete happens first? Does that create job just fail?

monamohebbi:
Looks good! We left a few comments and questions that we could use some expanding on. Thanks!

Mona & @tcdowney

AartiKriplani:
@tcdowney : When the create job gets enqueued and  delete happens first, the create job just exits because the `last_operation` on the instance is not the one that ''this,, job was intended for. It checks for [terminal state](https://github.com/cloudfoundry/cloud_controller_ng/pull/1318/files/ff4ee3dbc85a755e762fc8d5ec03f0caef1c64ca#diff-ce736eaca5ebe0f8c0989d94731b08a7R31) and does not retry.

app/jobs/services/service_instance_state_fetch.rb
AartiKriplani: the `intended_operation` is the `last_operation` with which the job starts processing with. But meanwhile, (since the lock is released until we get a response from broker) if the `last_operation` changes , it just exits. We thought of names like `cached_operation` etc but did not really find any other name suitable. Suggestions are welcome :)

PR #1321, opened 2019-04-09
Add service_broker_name for space summary endpoint.

AartiKriplani:
This commit adds service_broker_name field for each service in the response of the space summary endpoint. i.e `/v2/spaces/:guid/summary`

More details can be found [here](https://www.pivotaltracker.com/story/show/163455079).

### DIFF
--- a/app/models/services/managed_service_instance.rb
+++ b/app/models/services/managed_service_instance.rb
@@ -61,6 +61,7 @@ module VCAP::CloudController
         'dashboard_url' => dashboard_url,
         'shared_from' => nil,
         'shared_to' => [],
+        'service_broker_name' => service_broker.name,
         'service_plan' => {
           'guid' => service_plan.guid,
           'name' => service_plan.name,
--- a/docs/v2/spaces/get_space_summary.html
+++ b/docs/v2/spaces/get_space_summary.html
@@ -264,6 +264,7 @@ Cookie: </pre>
         "created_at": "2016-06-08T16:41:28Z"
       },
       "dashboard_url": null,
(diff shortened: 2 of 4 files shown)


app/models/services/managed_service_instance.rb
```diff
@@ -61,6 +61,7 @@ def as_summary_json
         'dashboard_url' => dashboard_url,
         'shared_from' => nil,
         'shared_to' => [],
+        'service_broker_name' => service_broker.name,
```
tcdowney: Have we considered [eager loading](https://sequel.jeremyevans.net/rdoc/classes/Sequel/Model/Associations/DatasetMethods.html#method-i-eager) the service plan and service broker in the Space Summaries controller?

https://github.com/cloudfoundry/cloud_controller_ng/blob/0a30f3e5d6aaa5f2753bf6e536a1aa104de0d817/app/controllers/runtime/space_summaries_controller.rb#L53-L54

This will make an additional database query for each service in the response and I'm concerned about further slowing down an already slow endpoint. Maybe it won't work well though because this endpoint returns both `UserProvidedServiceInstance` and `ManagedServiceInstance`s and the user provided ones probably don't have the plan/broker association. 🤔

tcdowney: If this story is to prevent clients from making additional calls to get that information anyway, this is probably still faster and we could explore potential eager-loading independently.

waterlink: We’ve created a bug here for this: https://www.pivotaltracker.com/story/show/165345918

PR #1323, opened 2019-04-16
Allow cancel of creation of async service binding

nmaslarski:
Before this commit CC was rejecting any request to a service binding if
there was ongoing operation in progress for it.
In v2.15 of the OSBAPI spec a behaviour for the broker was defined for
situations where unbind is sent while bind is in progress:
It can cancel the bind or it can reject the unbind because of
the ongoing operation.

With this commit CC forwards the unbind request to the broker
and acts accordingly to the response.
By allowing unbind while bind is in progress, there is a possibility
that the delayed job for bind state fetch runs at the same time as the unbind
request and cause the binding to be in bad state. To avoid this race
condition, we have added a check on the delayed worker job to check the
last operation to be the same for which it was intended for (when
locking).

More information in the [story](https://www.pivotaltracker.com/story/show/164046295)

Best Regards
@blgm and Niki,
on behalf of the SAPI Team

### DIFF
--- a/app/actions/service_binding_delete.rb
+++ b/app/actions/service_binding_delete.rb
@@ -29,7 +29,10 @@ module VCAP::CloudController
       warnings_accumulator = []
       errors = each_with_error_aggregation(bindings_to_delete) do |service_binding|
         raise_if_instance_locked(service_binding.service_instance)
-        raise_if_binding_locked(service_binding)
+
+        if service_binding.operation_in_progress? && service_binding.service_binding_operation.type != 'create'
+          raise CloudController::Errors::ApiError.new_from_details('AsyncServiceBindingOperationInProgress', service_binding.app.name, service_binding.service_instance.name)
+        end
 
         broker_response = remove_from_broker(service_binding)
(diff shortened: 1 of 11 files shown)


nmaslarski:
@tcdowney @monamohebbi, Can you take a look at this PR. It's similar to #1318. We'd like a review on it because it adds complex concurrent operations.

williammartin:
See https://github.com/cloudfoundry/cloud_controller_ng/pull/1327 for similar review.

FlorentFlament:
We planned to merge this today, but didn't do it because the CAPI pipeline is currently red: https://capi.ci.cf-app.com/teams/main/pipelines/capi

tcdowney:
@FlorentFlament I did some digging think this was actually manually merged in 78b6ec90bc7fa1f8433bf82fda4fcbac86c0ab90. I'm not sure what's up with this diff in this PR -- can you confirm?

AartiKriplani:
@tcdowney : yes, looks like it's merged by @sannidhi , we weren't aware of it. Thanks for pointing it out, I'll close the PR now.

PR #1324, opened 2019-04-16
Ability to trigger an update (of maintenance_info) on service instance

AartiKriplani:
As part of the `update-one` feature users can test an individual instance of the service to gauge the difficulty and risk associated with upgrading that particular service before running an `upgrade-all` errand (on the broker)
This PR enables that feature by accepting PUT requests to `/v2/service_instances/:guid ` for a service instance with `{"maintenance_info": {"version":"2.0"}}` in the body and forwards it as a PATCH to the broker to trigger update/upgrade of service instance. 
And, additionally validate for outdated maintenance_info updates.

More info can be found [here](https://www.pivotaltracker.com/story/show/164961906).

### DIFF
--- a/spec/acceptance/broker_api_compatibility/broker_api_v2.15_spec.rb
+++ b/spec/acceptance/broker_api_compatibility/broker_api_v2.15_spec.rb
@@ -487,6 +487,7 @@ RSpec.describe 'Service Broker API integration' do
 
       before do
         setup_broker(catalog)
+        @broker = VCAP::CloudController::ServiceBroker.find guid: @broker_guid
       end
 
       it 'is saved with the service plan' do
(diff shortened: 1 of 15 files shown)


AartiKriplani:
codeclimate issues are due to comma at the last node of the hash and from what we've seen the team actually wants to allow this. Codeclimate needs to be fixed

spec/acceptance/broker_api_compatibility/broker_api_v2.15_spec.rb
```diff
@@ -498,6 +499,33 @@
...
+          expect(response).to have_status_code(202)
+          expect(
+            a_request(:patch, update_url_for_broker(@broker, accepts_incomplete: true)).with(
+              body: hash_including(maintenance_info: { version: '2.0' })
+            )
+          ).to have_been_made
+        end
+
+        context 'when the maintenance_info does not match the one from the service plan' do
+          it 'should forward the maintanance info to the broker' do
```
reneighbor: Should this be "should not forward"?

waterlink: @reneighbor Yes, you are right!

PR #1264, opened 2018-11-30
Add V3 list service brokers endpoint

henryaj:
This PR adds a `GET /v3/service_brokers` endpoint for listing service brokers registered with the platform. Documentation is included.

I've left the commits unsquashed as they stand independently of one another; happy to squish if you prefer.

Tracker stories for reference: [[1]](https://www.pivotaltracker.com/story/show/161809195) [[2]](https://www.pivotaltracker.com/story/show/162088299) [[3]](https://www.pivotaltracker.com/story/show/162088407)

cc @AartiKriplani @williammartin @ablease @georgi-lozev

PR #1338, opened 2019-05-06
Add V3 list service brokers endpoint

AartiKriplani:
This PR adds a `GET /v3/service_brokers` endpoint for listing service brokers registered with the platform. Documentation is included.

Tracker stories for reference: [[1]](https://www.pivotaltracker.com/story/show/161809195) [[2]](https://www.pivotaltracker.com/story/show/162088299) [[3]](https://www.pivotaltracker.com/story/show/162088407)

cc @AartiKriplani @williammartin @georgi-lozev @FlorentFlament

### DIFF
--- a/app/controllers/v3/root_controller.rb
+++ b/app/controllers/v3/root_controller.rb
@@ -19,6 +19,7 @@ class RootController < ActionController::Base
     links.merge!(create_link(:processes))
     links.merge!(create_link(:resource_matches, experimental: true))
     links.merge!(create_link(:routes, experimental: true))
+    links.merge!(create_link(:service_brokers))
     links.merge!(create_link(:service_instances))
     links.merge!(create_link(:spaces))
     links.merge!(create_link(:stacks))
--- a/app/controllers/v3/service_brokers_controller.rb
+++ b/app/controllers/v3/service_brokers_controller.rb
@@ -1,0 +1,23 @@
+require 'messages/service_brokers_list_message'
+require 'presenters/v3/service_broker_presenter'
+require 'fetchers/service_broker_list_fetcher'
+
+class ServiceBrokersController < ApplicationController
+  def index
+    message = ServiceBrokersListMessage.from_params(query_params)
+    invalid_param!(message.errors.full_messages) unless message.valid?
+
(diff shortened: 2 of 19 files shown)


AartiKriplani:
Unfortunately had to open a new one as I'm unable to reopen https://github.com/cloudfoundry/cloud_controller_ng/pull/1264

AartiKriplani:
Codeclimate issues can be ignored, we need to change the rules in the checker

nmaslarski:
Review is requested from the VAT team, don't merge before.

luan:
This looks exactly like I would've implemented that according to existing patterns.
The only thing that's missing is a request spec with all the permission checking that we do in other resources such as domains/routes: https://github.com/cloudfoundry/cloud_controller_ng/blob/master/spec/request/routes_spec.rb

psparks123:
Supraja and I agree with Luan, your implementation looks great overall! Here are a few suggestions/comments:
- While we were working on domains, we decided to no longer write controller specs and to instead use the request spec to test controller behavior including permissions. Refer to [this story](https://www.pivotaltracker.com/story/show/165075402) for the discussion regarding this decision. 
- For permissions testing in the request spec, we decided to use shared examples (https://github.com/cloudfoundry/cloud_controller_ng/blob/master/spec/request_spec_shared_examples.rb). You can find examples in the request specs for routes and domains. 
- For the domains fetcher, we didn't make a separate domain fetcher and domain list fetcher. We combined it into one DomainFetcher with a fetch method that fetches all the domains and then applies filters based on parameters specified in the message: (https://github.com/cloudfoundry/cloud_controller_ng/blob/2df0cbcd494f24766dfe396d1e92f07cb7b4ef52/app/fetchers/domain_fetcher.rb). This can be useful if there are complex permissions checking steps involved.

blgm:
- We have delete the controller spec as suggested, the tests that it contained are duplicated elsewhere
- Regarding the shared examples pattern for permissions testing, we think this is a good idea.  We think it makes sense to move to this pattern in a [future chore](https://www.pivotaltracker.com/story/show/165999827) and get this PR merged, as it's already an old PR.  We have renamed our original acceptance test to be a request spec, as it contains the right tests.

PR #1340, opened 2019-05-08
Add internal api docs

Gerg:
First attempt, probably not 100% complete or accurate

### DIFF
--- a/docs/internal/README.md
+++ b/docs/internal/README.md
@@ -62,7 +62,7 @@ We do not recommend using internal API endpoints for anything other than their i
 
 **Intended Consumer:** Diego running task or process
 
-**Auth Mechanism:** None
+**Auth Mechanism:** None (redirects to MTLS endpoint if CC is configured for MTLS)
 
 ### GET /internal/v4/droplets/:guid/:droplet_checksum/download
 **Description:** Permalink for downloading a droplet.
@@ -81,7 +81,7 @@ We do not recommend using internal API endpoints for anything other than their i
 ### GET /internal/v4/log_access/:app_guid
 **Description:** Check if a user has access to the log stream for an app
 
-**Intended Consumer:** Loggregator
+**Intended Consumer:** Loggregator (`scheduler` and `syslog-binding-cache`)
 
 **Auth Mechanism:** OAuth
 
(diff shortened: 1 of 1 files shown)


docs/internal/README.md
```diff
@@ -0,0 +1,130 @@
...
+**Intended Consumer:** Diego staging task
+
+**Auth Mechanism:** Basic Auth
+
+### GET /internal/v2/droplets/:guid/:droplet_checksum/download
+**Description:** (Deprecated) Permalink for downloading a droplet. Use `GET /internal/v4/droplets/:guid/:droplet_checksum/download` instead.
+
+**Intended Consumer:** Diego running task or process
+
+**Auth Mechanism:** None
```
tcdowney: Can you note that this endpoint just redirects to the `/internal/v4` mTLS authenticated one. It doesn't let you actually download the droplet without any auth.

docs/internal/README.md
```diff
@@ -0,0 +1,130 @@
...
+**Description:** Marks package as uploaded
+
+**Intended Consumer:** Bits Service
+
+**Auth Mechanism:** MTLS
+
+### GET /internal/v4/syslog_drain_urls
+**Description:** Return list of syslog drain urls from logging services
+
+**Intended Consumer:** Loggregator
```
tcdowney: Do we know what piece (these days at least) of the Loggregator subsystem actually hits this?

docs/internal/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+**Auth Mechanism:** MTLS
+
+### POST /internal/v4/tasks/:task_guid/completed
+**Description:** Marks task as complete
+
+**Intended Consumer:** Diego running task
```
tcdowney: Maybe clarify that this callback is for CF app tasks that run on Diego

https://v3-apidocs.cloudfoundry.org/version/3.71.0/index.html#tasks

docs/internal/README.md
tcdowney: I'm finding out here:
https://cloudfoundry.slack.com/archives/C02HCCXV5/p1558560940076900?thread_ts=1558560940.076900&cid=C02HCCXV5

docs/internal/README.md
Gerg: I think there is still technically aslo an unauthenticated path: https://github.com/cloudfoundry/cloud_controller_ng/blob/3a93d5fea615fa085b19c2c940c21ad0dfb05948/app/controllers/internal/download_droplets_controller.rb#L24

tcdowney:
Looks great! Thanks @Gerg. 😊 👍 👍

PR #1342, opened 2019-05-10
Increase instances of CC to three

Gerg:
Three nodes will allow CC nodes to maintain quorum even if there is a single-node outage

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,4 +1,4 @@
 ---
 applications:
-- instances: 1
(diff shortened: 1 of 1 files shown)


tcdowney:
Thanks for your PR @Gerg, I decided to take a different approach. See the following commit:
https://github.com/cloudfoundry/cloud_controller_ng/commit/5d24ea3017554536356cec8d412da6efe292579a

PR #1322, opened 2019-04-12
Add maintenance_info to the service plan

nmaslarski:
Cloud Controller validates that maintenance_info can be parsed as a hash
and stores it in the database as text.
When maintenance_info is not provided by the broker, Cloud Controller
presents it as an empty JSON object.

More information in the pivotal tracker [story](https://www.pivotaltracker.com/story/show/164957708)

Niki,
On behalf of SAPI Team

PR #1349, opened 2019-05-21
Add maintenance_info to the service_instance and enable update

nmaslarski:
This PR introduces the `maintenance_info` in the `service_instance` object.
The second PR enables update of that `maintenance_info`:
When a `PATCH` request is being sent to CC with `maintenance_info` in the body, it should be forwarded to the broker for hit to perform the requested maintenance.

More information [#165766356](https://www.pivotaltracker.com/story/show/165766356), [#165721596](https://www.pivotaltracker.com/story/show/165721596)

Best,
Niki on behalf of the SAPI Team

### DIFF
--- a/db/migrations/20190503090026_add_maintenance_info_to_service_instance.rb
+++ b/db/migrations/20190503090026_add_maintenance_info_to_service_instance.rb
@@ -1,0 +1,5 @@
+Sequel.migration do
+  change do
+    add_column :service_instances, :maintenance_info, :text, null: true
+  end
+end
--- a/app/actions/services/service_instance_create.rb
+++ b/app/actions/services/service_instance_create.rb
@@ -22,8 +22,10 @@ module VCAP::CloudController
         arbitrary_parameters: arbitrary_params
       )
 
+      service_instance_attributes = broker_response[:instance].merge({ maintenance_info: service_instance.service_plan.maintenance_info })
+
       begin
-        service_instance.save_with_new_operation(broker_response[:instance], broker_response[:last_operation])
+        service_instance.save_with_new_operation(service_instance_attributes, broker_response[:last_operation])
       rescue => e
         cleanup_instance_without_db(e, service_instance)
       end
--- a/app/actions/services/service_instance_update.rb
+++ b/app/actions/services/service_instance_update.rb
@@ -21,7 +21,10 @@ module VCAP::CloudController
 
       if update_broker_needed?(request_attrs, cached_service_instance['service_plan_guid'], service_instance)
         handle_broker_update(cached_service_instance, lock, previous_values, request_attrs, service_instance)
-        update_deferred_attrs(service_instance, service_plan_guid: request_attrs.fetch('service_plan_guid', false))
+        update_deferred_attrs(service_instance,
+                              service_plan_guid: request_attrs.fetch('service_plan_guid', false),
+                              maintenance_info: request_attrs.fetch('maintenance_info', false)
+                             )
       else
         lock.synchronous_unlock!
       end
@@ -76,7 +79,7 @@ module VCAP::CloudController
         accepts_incomplete: accepts_incomplete,
         arbitrary_parameters: request_attrs['parameters'],
         previous_values: previous_values,
-        maintenance_info: request_attrs['maintenance_info'],
+        maintenance_info: request_attrs['maintenance_info'] || service_plan.maintenance_info,
       )
 
       service_instance.last_operation.update_attributes(response[:last_operation])
@@ -88,9 +91,16 @@ module VCAP::CloudController
       err
     end
 
-    def update_deferred_attrs(service_instance, service_plan_guid:)
-      if service_plan_guid && !service_instance.operation_in_progress?
-        service_instance.update_service_instance(service_plan: ServicePlan.find(guid: service_plan_guid))
+    def update_deferred_attrs(service_instance, service_plan_guid:, maintenance_info:)
+      unless service_instance.operation_in_progress?
+        attrs_to_update = {}
+        if service_plan_guid
+          service_plan = ServicePlan.find(guid: service_plan_guid)
+          attrs_to_update[:service_plan] = service_plan
+          attrs_to_update[:maintenance_info] = service_plan.maintenance_info
+        end
+        attrs_to_update[:maintenance_info] = maintenance_info if maintenance_info
+        service_instance.update_service_instance(attrs_to_update)
       end
     end
 
--- a/app/controllers/services/validators/service_update_validator.rb
+++ b/app/controllers/services/validators/service_update_validator.rb
@@ -37,7 +37,7 @@ module VCAP::CloudController
       def validate_maintenance_info_update(service_plan, maintenance_info)
         return unless maintenance_info
 
-        if service_plan.maintenance_info.nil? || maintenance_info != service_plan.parsed_maintenance_info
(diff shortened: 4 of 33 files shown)


nmaslarski:
We require review on this PR, since it adds a migration to a table that may contain many entries.
We're not sure if the migration will be successful on huge CF Installations:?

db/migrations/20190503090026_add_maintenance_info_to_service_instance.rb
```diff
@@ -0,0 +1,5 @@
+Sequel.migration do
+  change do
+    add_column :service_instances, :maintenance_info, :text, null: true
```
ericpromislow: We have a policy in our [migration style guide](https://github.com/cloudfoundry/cloud_controller_ng/wiki/CAPI-Migration-Style-Guide#always-specify-a-size-for-a-string-and-dont-use-text) to not use text columns and to specify length of the string, something like:
```
add_column :service_instances, :maintenance_info, String, size: 4096, null: true
```
Feel free to replace the `4096` with a value appropriate to the application.

ericpromislow:
@nmaslarski When you say there could be many entries in this table, on what order of magnitude are we talking?

cc @sethboyles

ericpromislow:
Hi SAPI team, we've observed the following:

First, we added about 500,000 service instances associated to a single space.

This helped us find a capi-only bug, where validations were getting this count by doing
`space.managed_service_instances.count` which creates an array, which is expensive when you have 500K objects.  It should be `space.managed_service_instances_dataset.count` -- we'll fix that.

We then ran a small script which repeatedly picked a service instance, changed a field, and saved the changes. We ran this concurrently with a deploy of your code and observed this error in the small SI-update script:

```
Sequel::SerializationFailure: Mysql2::Error: WSREP detected deadlock/conflict and aborted the transaction. Try restarting the transaction
from /var/vcap/data/packages/cloud_controller_ng/bb1058db5b8e83ea0a52f625863c0810ee98950f/gem_home/ruby/2.4.0/gems/mysql2-0.4.10/lib/mysql2/client.rb:120:in `_query'
Caused by Mysql2::Error: WSREP detected deadlock/conflict and aborted the transaction. Try restarting the transaction
from /var/vcap/data/packages/cloud_controller_ng/bb1058db5b8e83ea0a52f625863c0810ee98950f/gem_home/ruby/2.4.0/gems/mysql2-0.4.10/lib/mysql2/client.rb:120:in `_query'
```

The migration completed successfully, but updates or inserts to the service_instances table may fail during the migration.  

### Methodology

We ran two scripts on the API node: 
```
bosh ssh api
sudo su -
source /var/vcap/jobs/cloud_controller_ng/bin/ruby-version.sh
```
Populator: 
```
num = 1_000_000
offset = ServiceInstance.count + 1 + 100
instance = ServiceInstance.first
num.times do |iterx|
guid = SecureRandom.uuid
id = iterx + offset
  ServiceInstance.insert(instance.values.delete_if {|k, _| [:id, :guid, :tags, :name].include?(k)}.
      merge(name: "name-#{iterx + offset}", guid: guid, id: id)  )
  if iterx % 100 == 0
    $stderr.print(".")
    if iterx % 7900 == 0
      $stderr.puts("")
    end
  end
end
```
Updater:
```
count = ServiceInstance.count
while true
  idx = rand(count)
  si = ServiceInstance.find(id: idx)
  $stderr.puts("Updating si #{idx}...")
  si.update(gateway_name: SecureRandom.uuid)
  sleep 0.1
end
```

To run a script, you need to run it in the console's context: 
`/var/vcap/jobs/cloud_controller_ng/bin/console  < SCRIPT`

@ericpromislow and @sethboyles - CAPI community team

georgi-lozev:
Hey @ericpromislow, @sethboyles,  thank you very much for the quick and detailed review.

When we added the migration we also had similar conversation, regarding text vs fixed size string and we reference the migration style guide as well.

There were two main reasons we decided to go with the text, despite the fact the guide was not if favor of that.
First one, was consistency. Just a few weeks ago we got a review/introduce a similar migration on service plan level and there are a couple of other similar examples introduce in the near past which follow this pattern. 
- https://github.com/cloudfoundry/cloud_controller_ng/pull/1322
- https://github.com/cloudfoundry/cloud_controller_ng/blob/master/db/migrations/20170712153045_add_instance_update_schema_to_service_plans.rb
- https://github.com/cloudfoundry/cloud_controller_ng/blob/master/db/migrations/20170602113045_add_instance_create_schema_to_service_plans.rb
- https://github.com/cloudfoundry/cloud_controller_ng/blob/master/db/migrations/20170724090255_add_binding_create_schema_to_service_plans.rb

Second reason was the ability to extend the payload in future without the need of additional migration. We're not sure what the fields in the `maintenance_info` are going to be. For now it's only `version`, but we expect to have more in future, based on the feedback we get. We also validate the size of the maintenance_info is not beyond some number on the app layer.

If you believe we should have this length constraint on the service instance level, then we could make a guess and go with some number there. Best case, we're going to chose a number that works, worst case we'll need to introduce a new migration in future. I hope that's it's possible to extend the length of an already existing column.

Regarding the service instances magnitude, what we could say today based on the latest information shared with the community on the last Summit in Philly. For a single CF deployment the numbers were around 60_000 service instances, so let's round to 100_000. The underlaying db backing the system is Postgres I believe, if it matters. 
I'm not quite sure how realistic is to have that number of instances under a single space.

I hope that clarifies the things a bit, just let us know if you think we should change the migration to a fixed size string and if it's also worth changing the others that we're aware of to follow the pattern.

tcdowney:
> There were two main reasons we decided to go with the text, despite the fact the guide was not if favor of that.
The main reason it is not preferred for us is due to the difference in behavior between Postgres and MySQL.

Can you add some sort of validations/constraints on it on the Cloud Controller side as a model validation? 16k characters maybe? I think that will help mitigate the Postgres vs MySQL concerns. A thing we've noticed is that users like to push these `text` columns to their limits and are surprised when they try to put in [over 64kb worth of text](https://stackoverflow.com/questions/6766781/maximum-length-for-mysql-type-text) and it fails.

We should have _some_ max encoded someone to give better errors than just generic `DatabaseError` and to help us reason about what might have been added in there.

georgi-lozev:
Hello, Tim.

We have a follow-up story(https://www.pivotaltracker.com/story/show/165560162) that adds a validation on the app layer for the size of the field. It should be ready for PR soon.
As part of it we decided to start with quite restrictive value(2k characters) and in case users complain increase it to something that makes more sense. Out main concern was that it's a field that is per service instance, so we didn't want to make it too large, thus vulnerable to some malicious broker.

AartiKriplani:
@tcdowney  : We've created a following [PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/1354) to add validations on the `maintenance_info` field, does this mitigate your concerns ?

tcdowney:
Thanks @AartiKriplani,  yeah that addresses my concerns. Based on the findings from @ericpromislow and @sethboyles I'm personally fine with this behavior:

> The migration completed successfully, but updates or inserts to the service_instances table may fail during the migration.

As long as the migration ultimately completes successfully, the unlikely event of individual service creation failures should be fine. We're mostly concerned with migrations that themselves fail and leave the database/CF upgrade in a bad spot.

georgi-lozev:
Thank you, @tcdowney 

Does that mean we're good to proceed and merge?

Can we do something to guard our migration to leave the database in bad spot?

tcdowney:
Sorry for being ambiguous. 🙂 Yes I think the migration should be fine and you're good to merge.

PR #1354, opened 2019-05-29
Validate maintanance info on service plan in the catalog

AartiKriplani:
maintenance_info
* must contain a version field that is a [semantic version string](https://semver.org/spec/v2.0.0.html)
* must not contain more than 2000 characters
* allows other fields for now, and stores them. But this might change soon

More info can be found [here](https://www.pivotaltracker.com/story/show/165560162)


AartiKriplani:
Do not merge before https://github.com/cloudfoundry/cloud_controller_ng/pull/1349

PR #1356, opened 2019-05-29
Add configuration for thin thread pool size

Gerg:
Not sure if configuring this is beneficial in any way. 

Nothing ventured, nothing gained!

### DIFF
--- a/config/bosh-lite.yml
+++ b/config/bosh-lite.yml
@@ -278,3 +278,5 @@ bits_service:
   private_endpoint: http://bits-service.service.cf.internal
   username: admin
   password: admin
+
+threadpool_size: 20
--- a/config/cloud_controller.yml
+++ b/config/cloud_controller.yml
@@ -409,3 +409,5 @@ locket:
   ca_file: 'spec/fixtures/certs/bbs_ca.crt'
   cert_file: 'spec/fixtures/certs/bbs_client.crt'
   key_file: 'spec/fixtures/certs/bbs_client.key'
+
+threadpool_size: 20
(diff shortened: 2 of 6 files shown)


georgi-lozev:
I tested the change on one of our `bosh-lite` environments and it has effect.
It seems to increase the number of concurrent API requests CC could handle which is beneficial in some worst case scenarios related to CC -> broker communication.

From my POV it's worth introducing such kind of configuration that will allow operators to tweak in case they feel some pains. 
I even could try to find some guinea pigs to play around with this property on a real CF deployment on relatively high scale.

tcdowney:
> It seems to increase the number of concurrent API requests CC could handle which is beneficial in some worst case scenarios related to CC -> broker communication.

Yeah I could definitely see it helping there since that's I/O bound. When Cloud Controllers are CPU bound though I think increasing the number of threads in the pool here will just put more pressure on the system and could potentially cause us additional [memory bloat](https://www.speedshop.co/2017/12/04/malloc-doubles-ruby-memory.html).

I'd be fine with making it configurable for folks to experiment with, though.

The capi-release PR does include sufficient warnings I think:
https://github.com/cloudfoundry/capi-release/pull/142/

tcdowney:
@zrob @Gerg @ScarletTanager 

Do any of you know how the number 20 was chosen to be the default originally?

Gerg:
I don't think anyone ever made a choice on the CC side. Looks like 20 comes from EventMachine by way of thin:

https://github.com/macournoyer/thin/blob/f83ab677f008afcfabcdfa8930225f6cea88f461/lib/thin/backends/base.rb#L27-L31

https://github.com/eventmachine/eventmachine/blob/537127fa22a058e3d7b248a214df210e7c77dc95/lib/eventmachine.rb#L1111

ScarletTanager:
I'll go with what @Gerg said.

PR #1359, opened 2019-06-03
Handle warnings while creating service broker

AartiKriplani:
This PR, introduces warnings header `X-Cf-Warnings` on a `v3` endpoint/controller for the first time. 

More info [#166186412](https://www.pivotaltracker.com/story/show/166186412)


waterlink:
@psparks123 

> VAT's been placing all the V3 actions directly in the app/actions directory rather than in app/actions/v3. We can discuss further whether we should do some moving around to either get rid of the v3 folder completely or to move everything into the existing v3 folder.

Makes sense, we’ll fix that.

> I saw that the warnings array check was being done at the application controller level -- I haven't seen ArgumentErrors being raised in any other controllers so I was wondering whether the array check can be done somewhere else? Like when warnings are being created?

This is an error not for the user/client, but for a developer to not accidentally pass different types that have `:each` defined on them.

> Like when warnings are being created?

Not sure how to do that. The point is that action code will create these arrays and return them to the controller, and controller will pass that to the `add_warning_headers` with that.

And actions should be creating arrays properly there. The check is only a sanity check that will inform the developer what they are doing wrong.

PR #1385, opened 2019-07-02
Use fully qualified column names in DB query

### LINKED ISSUE #1384: Filtering service plans by service broker id results in CF-DatabaseError
## Issue

`/v2/service_plans?q=service_broker_guid:<guid>` fails with `CF-DatabaseError`.

## Context

cloud_controller_v2 version 2.136.0

## Steps to Reproduce

Run `cf curl /v2/service_plans?q=service_broker_guid:<guid>`.

## Expected result

List of service plans for a specific service broker.

## Current result

```
HTTP/1.1 500 Internal Server Error

{
   "description": "Database error",
   "error_code": "CF-DatabaseError",
   "code": 10011
}
```

## Root Cause

The following database query is executed...
```
SELECT count(*) AS "count" FROM "service_plans"
  LEFT JOIN "services" ON ("services"."id" = "service_plans"."service_id")
  LEFT JOIN "service_brokers" ON ("service_brokers"."id" = "services"."service_broker_id")
  WHERE ((("public" IS TRUE) OR ("id" IN (<long list of IDs>)))
    AND ("active" IS TRUE) AND ("service_brokers"."guid" = '<guid>'))
  LIMIT 1
```
... and fails with `PG::AmbiguousColumn: ERROR: column reference "id" is ambiguous`.

Changing the query in psql leads to another error: `column reference "active" is ambiguous`.

blgm:
Fixed #1384

In #1384 a user described an error that occurs when querying service plans with a
service broker filter. The error were:

- PG::AmbiguousColumn: ERROR: column reference "id" is ambiguous
- PG::AmbiguousColumn: ERROR: column reference "active" is ambiguous

The captured DB query was:
```
SELECT count(*) AS "count" FROM "service_plans"
  LEFT JOIN "services" ON ("services"."id" = "service_plans"."service_id")
  LEFT JOIN "service_brokers" ON ("service_brokers"."id" = "services"."service_broker_id")
  WHERE ((("public" IS TRUE) OR ("id" IN (<long list of IDs>)))
    AND ("active" IS TRUE) AND ("service_brokers"."guid" = '<guid>'))
  LIMIT 1
```

It can be seen that the `active` and `id` columns are not qualified by
table name, so the DB is confused when doing a join where multiple
tables may contain the `active` and `id` columns.

The solution is to qualify the column names with the table name.

### DIFF
--- a/app/models/services/service_plan.rb
+++ b/app/models/services/service_plan.rb
@@ -96,7 +96,7 @@ module VCAP::CloudController
       included_ids = ServicePlanVisibility.visible_private_plan_ids_for_user(user).
                      concat(plan_ids_from_private_brokers(user))
 
-      Sequel.or({ public: true, id: included_ids }).&(active: true)
+      Sequel.or({ public: true, service_plans__id: included_ids }).&(service_plans__active: true)
     end
 
     def self.user_visibility_show_filter(user)
--- a/spec/unit/controllers/services/service_plans_controller_spec.rb
+++ b/spec/unit/controllers/services/service_plans_controller_spec.rb
@@ -309,14 +309,41 @@ module VCAP::CloudController
       end
 
       context 'as a space developer' do
+        let(:space) { Space.make }
+        let(:user) { User.make }
+        let(:different_service_plan) { ServicePlan.make }
+
+        before do
+          space.organization.add_user(user)
+          space.add_developer(user)
+        end
+
+        it 'can query by service broker guid using ":"' do
+          set_current_user(user)
+          service = @services[:public][0].service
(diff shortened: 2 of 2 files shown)


blgm:
Hi CAPI team.  We have made this change which fixes the bug #1384.  However we were wondering:
- do you know of a better fix for this?
- is there a concern that there might be other queries that use ambiguous column names that we haven't found yet?

selzoc:
@blgm 
* I think we have a good mix of what you've added, and stuff like 
```ruby
"#{ProcessModel.table_name}__guid".to_sym
```
which would handle the case where the table name changes?  but that's probably pretty unlikely
* almost certainly!

PR #1388, opened 2019-07-09
Upgrade Rubocop to latest version (0.72.0)

tcdowney:
I've noticed some significant performance improvements from this
upgrade! ☺️

Changelog:
https://github.com/rubocop-hq/rubocop/blob/v0.72.0/CHANGELOG.md

[#167174579](https://www.pivotaltracker.com/story/show/167174579)

Authored-by: Tim Downey <tdowney@pivotal.io>


.rubocop.yml
```diff
@@ -14,6 +14,9 @@ AllCops:
 Bundler/DuplicatedGem:
   Enabled: false
 
+Layout/AlignArguments:
+  Enabled: false
```
tcdowney: This cop results in 700+ changed lines. I agree with the cop in principle, but because we had made some weird indentation decisions in the past it's autocorrect tries to work with that and shoves everything over to the right.

i.e. something like this
```
ThingDoer.new(arg1,
  arg2,
  arg3
)
```

becomes
```
ThingDoer.new(arg1,
              arg2,
              arg3
)
```

when ideally (IMO) it would have been like this all along

```
ThingDoer.new(
  arg1,
  arg2,
  arg3
)
```

Wanted to discuss it with the team first.

.rubocop.yml
```diff
@@ -155,6 +158,9 @@ Naming/HeredocDelimiterNaming:
 Naming/PredicateName:
   Enabled: false
 
+Naming/RescuedExceptionsVariableName:
```
tcdowney: This one wants all rescued errors to be named `e` instead of something like `error` or `api_error` or whatever. Felt like an unnecessary update.

.rubocop.yml
```diff
@@ -178,10 +184,7 @@ Naming/UncommunicativeMethodParamName:
 Naming/VariableNumber:
   Enabled: false
 
-Performance/Casecmp:
```
tcdowney: This cop no longer exists as far as I can tell.

tcdowney:
Well CodeClimate certainly does not like this change! :)

Screen Shot 2019-07-09 at 10 40 45 AM

.rubocop.yml
cwlbraa: if rubymine had better rubocop integration I'd want this on, but since we only run it before committing, I think this is a wash and you've done the right thing in disabling it.

.rubocop.yml
ewrenn8: Agreed.

.rubocop.yml
ewrenn8: Looks good.

.rubocop.yml
ewrenn8: Couldn't have said it better.

Issue #1429, opened 2019-09-05
Fetching usage events after_guid when that guid has been pruned


lisamcho:
## Issue
As a usage service attempting to consume usage events and calculate durations that apps and services were running, we store the last fetched guid and then query the unfetched guids using the `after_guid` query parameter to the v2/app_usage_events and v2/service_usage_events endpoints.

When a CF deployment goes more than a month without pushing an app or creating a service event, then there are no new usage events.

This results in a BadQueryParameter error.

As a client, it's not clear what to do in this case, if we ignore the error and move on, then there's the risk that we missed events (because our app was down for some time and couldn't fetch events).

## Steps to Reproduce
cf curl v2/app_usage_events?after_guid=<pruned-event-guid>

## Expected result
We don't know what a good behavior would be in this case.

## Current result
This results in a BadQueryParameter error.

## Possible Fixes
1. The client (usage service) rescues and ignores the error and fetches the usage events with no query param. 
- Con: the badqueryparam error could be caused by something else on capi's side.
- Con: events can be missed if you fail to fetch events for >30 days. Currently this would result in an error, but it would be silently hidden with this fix, potentially resulting in corrupted data.

2. In Capi, when `after_guid` cannot be found, instead of returning BadQueryParameter error, return the first page of data as if the query param were not passed. 
- Con: events can be missed if you fail to fetch events for >30 days. Currently this would result in an error, but it would be silently hidden with this fix, potentially resulting in corrupted data.
- Con: you could query after_guid=<guid-with-typo> and it would return unexpected results.

3. Capi changes and always keeps at least one usage event for apps and services, even if it is older than the `cutoff_age_in_days`.
- Con: `cutoff_age_in_days` bosh properties are more complicated
^^ We like this option best. CC @jpalermo 

https://github.com/cloudfoundry/capi-release/blob/6059f17f7215919f0f9a23ffebea490002ab57e9/jobs/cloud_controller_clock/spec#L96-L101

4. Other ideas?

sethboyles:
Closing as https://github.com/cloudfoundry/cloud_controller_ng/pull/1430 is in 1.87

PR #1430, opened 2019-09-06
When pruning app and service usage events, ensure at least one record is kept

jpalermo:
This prevents the problem of api clients
getting an error when querying the endpoints with after_guid
and the whole table has been pruned.

Related issue: #1429

### DIFF
--- a/spec/unit/jobs/services/service_usage_events_cleanup_spec.rb
+++ b/spec/unit/jobs/services/service_usage_events_cleanup_spec.rb
@@ -5,6 +5,8 @@ module VCAP::CloudController
     RSpec.describe ServiceUsageEventsCleanup, job_context: :worker do
       let(:cutoff_age_in_days) { 30 }
       let(:logger) { double(Steno::Logger, info: nil) }
+      let!(:event_before_threshold) { ServiceUsageEvent.make(created_at: (cutoff_age_in_days + 1).days.ago) }
+      let!(:event_after_threshold) { ServiceUsageEvent.make(created_at: (cutoff_age_in_days - 1).days.ago) }
 
       subject(:job) do
         ServiceUsageEventsCleanup.new(cutoff_age_in_days)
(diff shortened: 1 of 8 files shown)


spec/unit/jobs/services/service_usage_events_cleanup_spec.rb
```diff
@@ -5,6 +5,9 @@ module Jobs::Services
     RSpec.describe ServiceUsageEventsCleanup, job_context: :worker do
       let(:cutoff_age_in_days) { 30 }
       let(:logger) { double(Steno::Logger, info: nil) }
+      let!(:event_before_threshold) { ServiceUsageEvent.make(created_at: (cutoff_age_in_days + 1).days.ago) }
+      let!(:event_after_threshold) { ServiceUsageEvent.make(created_at: (cutoff_age_in_days - 1).days.ago) }
+      let!(:last_event_that_is_always_kept) { ServiceUsageEvent.make }
```
selzoc: Is `last_event_that_is_always_kept` intended to be used in a test in either this or the other spec?

jpalermo: Hmm, no.

The tests even pass without the `last_event_that_is_always_kept` record, but you don't know if they're passing because `event_after_threshold` is actually after the threshold, or simply because it's the last record and we never delete that anymore.

Would it be more clear if I moved the event creation back into the individual tests? In that case `last_event_that_is_always_kept` is only really needed for the `keeps events created after the pruning threshold` test.

Or is there some other way to better show that intent?

I suppose we could add an explicit test in here to ensure that the last record is kept, although it feels a bit weird that all the edge cases are being tested at 3 different layers.

selzoc: I guess I don't really see any strong assertions around this behavior.  In all the specs I see, the assertions for the new tests are the same as the old tests, which is confusing.  Being very explicit about this unusual behavior would be good.

jpalermo: Ok, I cleaned everything up a bit. Tried doing a bit more, but then other tests started to get weird, so it's a bit of a middle ground.

PR #1431, opened 2019-09-07
Change include paths to match relationship names

Gerg:
Update current includes on v3 API to conform to https://docs.google.com/document/d/1iLsVoRm03Iy6rMWJGKENVo-2_3p2Ecr51AH5szmplF4/edit

Old include strings will still work, but are no longer documented.

`GET /v3/apps(/:guid)?include=org` is now `GET /v3/apps(/:guid)?include=space.organization`
`GET /v3/spaces(/:guid)?include=org` is now `GET /v3/apps(/:guid)?include=organization`

Also, specifying multi-step include paths will now automatically include intermediate resources. This is because the relationships on the intermediate resources are needed to link the included resources to the root resource. For example: `include=space.organization` will include spaces and organizations.

### DIFF
--- a/docs/v3/source/includes/concepts/_includes.md.erb
+++ b/docs/v3/source/includes/concepts/_includes.md.erb
@@ -1,7 +1,7 @@
 ## Include
 
 The `include` parameter allows clients to fetch resources and ensure that information about the parent object(s) is part of the response.
-For example, a response to `/v3/spaces/:guid?include=org` will contain detailed information about the space and its parent org.
+For example, a response to `/v3/spaces/:guid?include=organization` will contain detailed information about the space and its parent organization.
 
 Developers may choose to use the `include` feature to reduce the number of API calls. The include
 query param can be used with a single resource or a list of resources.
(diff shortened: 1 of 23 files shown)


docs/v3/source/includes/concepts/_includes.md.erb
```diff
@@ -25,7 +25,7 @@ Example request to apps resource to include parent orgs and spaces
 ```
 
 ```shell
-curl "https://api.example.org/v3/apps?include=org,space" \
+curl "https://api.example.org/v3/apps?include=space.organization" \
```
tcdowney: I think this section of the docs is supposed to demo including multiple resources. Should this be:

```
curl "https://api.example.org/v3/apps?include=space,space.organization" \
```

Gerg: `space.organization` will automatically include spaces also. This is because without the space, there will not be a way to determine which organization is associated with which app

docs/v3/source/includes/concepts/_includes.md.erb
```diff
@@ -1,7 +1,7 @@
 ## Include
 
 The `include` parameter allows clients to fetch resources and ensure that information about the parent object(s) is part of the response.
-For example, a response to `/v3/spaces/:guid?include=org` will contain detailed information about the space and its parent org.
+For example, a response to `/v3/spaces/:guid?include=organization` will contain detailed information about the space and its parent org.
```
selzoc:
```suggestion
For example, a response to `/v3/spaces/:guid?include=organization` will contain detailed information about the space and its parent organization.
```

PR #1447, opened 2019-09-17
Use a nested query to allow searching spaces with orgs & labels

ericpromislow:
[#168527952](https://www.pivotaltracker.com/story/show/168527952)

Authored-by: Eric Promislow <eric.promislow@suse.com>

### DIFF
--- a/app/fetchers/space_list_fetcher.rb
+++ b/app/fetchers/space_list_fetcher.rb
@@ -21,7 +21,11 @@ module VCAP::CloudController
       end
 
       if message.requested? :organization_guids
-        dataset = dataset.where(organization: Organization.where(guid: message.organization_guids))
+        dataset = if message.requested?(:label_selector)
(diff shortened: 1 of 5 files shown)


app/fetchers/space_list_fetcher.rb
```diff
@@ -21,7 +21,11 @@ def filter(message, dataset)
       end
 
       if message.requested? :organization_guids
-        dataset = dataset.where(organization: Organization.where(guid: message.organization_guids))
+        dataset = if message.requested?(:label_selector)
```
selzoc: I think it would be better to not have logic here that depends on another aspect of the message.  If you check out the AppListFetcher or PackageListFetcher, they both just use the map..flatten..map construct no matter what.

ericpromislow: but we've done that elsewhere.... Not that I agree with it. It was done in the spirit of making the smallest change that worked and not understanding the ORM well enough to improve it

PR #1451, opened 2019-09-26
Provide disk quota when desiring lrp from opi

mnitchev:
[#167725778](https://www.pivotaltracker.com/story/show/167725778)

Sends over the desired disk limit to eirini, so it can be properly set in the pod spec.

NOTE: We only ran the ones which work with Eirni. But since the change is only in the opi apps client, it should give us enough confidence.


mnitchev:
Closing in favour of #1474

Issue #1365, opened 2019-06-05
Tasks in pending state get start event


joyvuu-dave:
## Issue

The `v2/app_usage_events` API returns a start event for tasks in a pending state. If a client of cloud controller wishes to calculate the duration a task has been running, this results in an inaccurate answer. 

## Context

Normally, this isn't a problem, as typically tasks transition from the pending to running state quickly. However, in cases where a fast transition does not take place (typically an error), this means the runtime of a task as calculated by looking at the time of the start and stop events could be very off. Our system has sometimes had tasks stuck in a pending state for upwards of 90 days. 

## Steps to Reproduce
Create a task
<Somehow ensure that the task is still pending before the next step and not running yet>
curl `v2/app_usage_events` 

## Expected result
Should not return a TASK_STARTED event for the task.

## Current result
It is returning a TASK_STARTED event for the task

sethboyles:
I believe this was fixed by https://github.com/cloudfoundry/cloud_controller_ng/pull/1459.  Please reopen if that is not the case.

PR #1459, opened 2019-10-01
only create start usage events for RUNNING tasks

tcdowney:
Related issue: https://github.com/cloudfoundry/cloud_controller_ng/issues/1365 has more context. I've felt personally invested in this issue for a couple years now, so here's a PR. 🙂 

- [App Usage Events](https://docs.cloudfoundry.org/running/managing-cf/usage-events.html#app-usage-events) are used to inform consumers (such as billing
systems) of changes of app/task state that affect resource consumption
on the underlying platform
- A CF Task is not actually consuming resources until it is scheduled on
the runtime and transitions into the RUNNING state so this change
creates the `TASK_STARTED` event when the task is RUNNING instead of
PENDING
- A separate bug (https://www.pivotaltracker.com/story/show/152831505)
where the Task Syncer ignores PENDING tasks forever can cause tasks that
are never scheduled to linger in the PENDING state for a very long time.
This change mitigates the impact of that issue on billing.

### DIFF
--- a/app/models/runtime/task_model.rb
+++ b/app/models/runtime/task_model.rb
@@ -24,15 +24,14 @@ module VCAP::CloudController
     set_field_as_encrypted :environment_variables, column: :encrypted_environment_variables
     serializes_via_json :environment_variables
 
-    def after_create
-      super
-      create_start_event
-    end
-
     def after_update
       super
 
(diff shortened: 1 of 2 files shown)


sethboyles:
Looks great, thanks @tcdowney

PR #1472, opened 2019-10-11
Make label-selector parser friendlier

ericpromislow:
The current back-end is hard to use from the CLI.  Its parser is
too strict, and the error messages tend to leave command-line
users struggling to see what they need to fix.

* Accept ignorable white-space everywhere
* For syntax-errors, contextually highlight the first error
  so user can see what they need to fix
* Complain about some things that current implementation accepts,
  like the trailing comma in 'fish=scale,'

[#168488761](https://www.pivotaltracker.com/story/show/168488761)

For examples, use a v7 version of the CLI and target `pre-fix.capi.land`
and `post-fix.capi.land`

### pre fix:

```
$ cf7 apps --labels 'abc= def'
Getting apps in org org / space space as admin...

The query parameter is invalid: ' def' contains invalid characters
FAILED

$ cf7 apps --labels 'abc=def,'
Getting apps in org org / space space as admin...

No apps found
```

### post-fix

```
$ cf7 apps --labels 'abc= def'
Getting apps in org org / space space as admin...

No apps found

$ cf7 apps --labels 'abc=def,'
cf7 apps --labels 'abc=def,'
Getting apps in org org / space space as admin...

The query parameter is invalid: expecting a key or '!', got "abc=def,<<>>"
FAILED
```

Better usability

Really, I watched the other CLI devs try to use this, and they were all puzzled.

None

* [x ] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/CONTRIBUTING.md)

* [ x] I have run all the unit tests using `bundle exec rake`

* [x ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats)

### DIFF
--- a/app/controllers/v3/apps_controller.rb
+++ b/app/controllers/v3/apps_controller.rb
@@ -37,6 +37,7 @@ class AppsV3Controller < ApplicationController
   def index
     message = AppsListMessage.from_params(query_params)
     invalid_param!(message.errors.full_messages) unless message.valid?
+    # add_warning_headers(message.warnings) if message.warnings.size > 0
 
     dataset = if permission_queryer.can_read_globally?
                 AppListFetcher.new.fetch_all(message, eager_loaded_associations: Presenters::V3::AppPresenter.associated_resources)
--- a/app/controllers/v3/apps_controller.rb
+++ b/app/controllers/v3/apps_controller.rb
@@ -37,8 +37,6 @@ class AppsV3Controller < ApplicationController
   def index
     message = AppsListMessage.from_params(query_params)
     invalid_param!(message.errors.full_messages) unless message.valid?
-    # add_warning_headers(message.warnings) if message.warnings.size > 0
-
     dataset = if permission_queryer.can_read_globally?
                 AppListFetcher.new.fetch_all(message, eager_loaded_associations: Presenters::V3::AppPresenter.associated_resources)
               else
--- a/app/models/helpers/metadata_helpers.rb
+++ b/app/models/helpers/metadata_helpers.rb
@@ -1,24 +1,6 @@
 module VCAP::CloudController
   class MetadataHelpers
     KEY_SEPARATOR = '/'.freeze
-    # REQUIREMENT_SPLITTER = /(?:\(.*?\)|[^,])+/.freeze
-    # KEY_CHARACTERS = %r{[\w\-\.\_\/]+}.freeze
-    #
-    # IN_PATTERN = /\A(?<key>.*?) in \((?<values>.*)\)\z/.freeze                     # foo in (bar,baz)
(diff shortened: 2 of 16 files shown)


app/controllers/v3/apps_controller.rb
```diff
@@ -37,6 +37,7 @@ class AppsV3Controller < ApplicationController
   def index
     message = AppsListMessage.from_params(query_params)
     invalid_param!(message.errors.full_messages) unless message.valid?
+    # add_warning_headers(message.warnings) if message.warnings.size > 0
```
selzoc: Is this meant to be commented-out?

app/models/helpers/metadata_helpers.rb
```diff
@@ -1,24 +1,24 @@
...
+    # NOT_EXISTS_PATTERN = /\A!(?<key>#{KEY_CHARACTERS})(?<values>)\z/.freeze        # !foo
+    #
+    # REQUIREMENT_OPERATOR_PAIRS = [
+    #   { pattern: IN_PATTERN, operator: :in },
+    #   { pattern: NOT_IN_PATTERN, operator: :notin },
+    #   { pattern: EQUAL_PATTERN, operator: :equal },
+    #   { pattern: NOT_EQUAL_PATTERN, operator: :not_equal },
+    #   { pattern: EXISTS_PATTERN, operator: :exists }, # foo
+    #   { pattern: NOT_EXISTS_PATTERN, operator: :not_exists },
+    # ].freeze
```
selzoc: another big block of commented-out code

app/controllers/v3/apps_controller.rb
ericpromislow: I should delete it and make it a separate story. Currently there are no warnings, but what's an error in a set-metadatum should just be a warning in a query.

app/models/helpers/metadata_helpers.rb
ericpromislow: oh blast it, thx

app/controllers/v3/apps_controller.rb
ericpromislow: On second thought, a query like `!<10 MB long name>` should continue to raise an error rather than a warning because executing a lot of such queries in parallel would create a DOS. Better to continue to error-out on any invalid name.

selzoc:
@ericpromislow there's a new cla process for the cloudfoundry foundation: https://lists.cloudfoundry.org/g/cf-dev/message/8728

tcdowney:
Here's the code for the k8s API selector parser:

https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/selector.go

It would be cool if most selectors that work in `kubectl` worked for `cf`. Maybe some type of conformance tests that run against both? Just some food for thought + inspiration for potential error messages.

ericpromislow:
Time to pull this out of consideration. I have the branch if someone ever complains about how the CLI reports errors involving labels and annotations.

PR #1482, opened 2019-10-29
Cut oversize error messages down so they'll fit in a 16K database field

ericpromislow:
Error handling can break with very large error messages, and it appears as if no error was encountered.  

This PR does two things:

1.) If a raw error message is too large, try cutting the backtrace down 1/2 at a time until the raw message will fit in a 16K field.

2.) If #1 fails, remove blocks, largest first, favoring ends of arrays, until the YAML-encoding of the object will fit in the field.

Four service-broker-related tests were breaking on travis because travis runs its tests from a path that is slightly longer than for concourse and the dev machines. This wasn't noticed until some change was made that created full error messages that exceeded the limits on travis, but not elsewhere.

### DIFF
--- a/app/jobs/pollable_job_wrapper.rb
+++ b/app/jobs/pollable_job_wrapper.rb
@@ -19,8 +19,20 @@ module VCAP::CloudController
       end
 
       def error(job, exception)
-        api_error = convert_to_v3_api_error(exception)
-        save_error(api_error, job)
+        begin
+          api_error = convert_to_v3_api_error(exception)
+          save_error(api_error, job)
+        rescue Sequel::DatabaseError
+          if (exception.backtrace.size rescue 0) > 0
+            exception.backtrace.slice!((exception.backtrace.size / 2)..-1)
+            retry
+          else
+            raise
+          end
+        end
+      rescue StandardError => ex
+        logger.error("can't yaml-encode exception #{exception}: #{ex.message}")
+        raise
       end
 
       def failure(job)
@@ -63,6 +75,10 @@ module VCAP::CloudController
         find_pollable_job(job).each do |pollable_job|
           pollable_job.update(state: new_state)
         end
+      end
+
+      def logger
+        @logger ||= Steno.logger('cc.pollable.job.wrapper')
       end
     end
   end
--- a/spec/unit/jobs/pollable_job_wrapper_spec.rb
+++ b/spec/unit/jobs/pollable_job_wrapper_spec.rb
@@ -143,25 +143,20 @@ module VCAP::CloudController::Jobs
       end
 
       context 'with a big message' do
-        # postgres complains with 15,824
-        # mysql complains with 15,825, so test for failure at that point
-        # bracket the values with 100 each way to allow for differences on travis
+        # postgres complains with 15,826
+        # mysql complains with 15,828, so test for failure at that point
 
         it 'squeezes just right one in' do
-          raise BigException.new(message: 'x' * 15_823)
-        rescue BigException => exception
           expect {
-            pollable_job.error(job, exception)
+            pollable_job.error(job, BigException.new(message: 'x' * 15_825))
           }.to_not raise_error
         end
 
         it 'gives up' do
-          raise BigException.new(message: 'x' * 15_825)
-        rescue BigException => exception
           pg_error = /value too long for type character varying/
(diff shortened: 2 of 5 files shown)


waterlink:
LGTM! 👍

app/jobs/pollable_job_wrapper.rb
```diff
@@ -36,7 +37,11 @@ def after(job)
       def convert_to_v3_api_error(exception)
         v3_hasher = V3ErrorHasher.new(exception)
         error_presenter = ErrorPresenter.new(exception, Rails.env.test?, v3_hasher)
-        YAML.dump(error_presenter.to_hash)
+        res = YAML.dump(error_presenter.to_hash)
+        return res if res.size <= 16_000 || exception.backtrace.empty?
```
sethboyles: Think `bytesize` would be a little safer in case of unicode characters

sethboyles: or is the db column support unicode and arbitrarily limit on number of characters?

ericpromislow: Yeah, good point. I'll get that in,after cli standup

app/jobs/pollable_job_wrapper.rb
```diff
@@ -59,6 +64,18 @@ def persist_warnings(job)
       def save_error(api_error, job)
         find_pollable_job(job).each do |pollable_job|
           pollable_job.update(cf_api_error: api_error)
+        rescue Sequel::DatabaseError => ex
+          begin
+            m = /value too long for type character varying\((\d+)\)/.match(ex.message)
+            limit = m ? m[1].to_i - 1 : 15_999
```
sethboyles: what are the cases where the character size limit is not 16,000?

ericpromislow: don't know.  With the change in `convert_to_v3_api_error` I no longer was hitting any of the exceptions in `save_error`, but left those exception handlers in so they could be logged. If you think they should be dropped (and the yaml truncator files removed) I'm ok with that

sethboyles: Yeah, if this code isn't getting hit anymore, I think it makes sense to remove it.  Seems like this code would only get hit if someone manually altered the DB.

One possible alternative that would combine the two approaches--attempt to save the job, catch the error, and only then cut the backtrace in half.  That way we wouldn't need to know the magic `16000` number

ericpromislow: I might have something, that still calls for the yaml truncator....

ericpromislow: Yup, did that more or less. The refactor puts the exception backtrace cutting and yaml-checking in one place, although only the postgres error message gives the max field width. Note that the travis PR is now failing because of an error in master (discussed on the vat channel), but the push itself is fine

ericpromislow: I dropped the whole yaml truncator because it isn't needed for this case, and probably will never be. No one's going to emit an exception with a 12K message.

spec/unit/jobs/pollable_job_wrapper_spec.rb
```diff
@@ -119,6 +123,56 @@ module VCAP::CloudController::Jobs
...
     end
 
+    describe 'error' do
+      let(:job) { double(job_name_in_configuration: 'my-job', max_attempts: 2, perform: nil, guid: '15') }
+      let!(:actual_pollable_job) { VCAP::CloudController::PollableJobModel.create(delayed_job_guid: job.guid) }
+
+      context 'with a big backtrace' do
+        it 'culls it down' do
+          raise BigException.new
+        rescue BigException => exception
```
sethboyles: are you raising an error just to get an instance of an exception?  I think you can do this more explicitly like this:
```
exception = BigException.new
exception.set_backtrace(long_string)
```

ericpromislow: I learn something with every tweak. This also removes the dependency on different path lengths between travis and other environments, so no need to test for a range of expected array sizes -- the final exception size will always be 8 (start with 32 * 1000-char backtrace + the rest of the body, about 200 chars. Divide the backtrace until the full size < 16000 chars. Meaning we can stop with an 8-line backtrace.

PR #1488, opened 2019-11-06
Remove workflows section of the docs

Gerg:
We haven't had a reason to expand this documentation and the information currently in it is confusing and out of date. These types of docs should probably go in the official CF docs instead.


tcdowney:
LGTM

Please sign the CLA, though.

PR #1491, opened 2019-11-14
Update random route generation to add 2 letter suffix

jpalermo:
Decreases odds of collisions being reported by users.

We've updated the random route generation in cli v6 to match this new behavior.

[#169141312](https://www.pivotaltracker.com/story/show/169141312)

### DIFF
--- a/lib/cloud_controller/random_route_generator.rb
+++ b/lib/cloud_controller/random_route_generator.rb
@@ -7,7 +7,10 @@ module VCAP::CloudController
     end
 
     def route
-      @adjective_noun_generator.generate
+      ascii_letter_a = 97
+      first_letter = (rand(26) + ascii_letter_a).chr
+      second_letter = (rand(26) + ascii_letter_a).chr
+      @adjective_noun_generator.generate + "-#{first_letter}#{second_letter}"
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/cloud_controller/random_route_generator.rb
```diff
@@ -7,7 +7,10 @@ def initialize
     end
 
     def route
-      @adjective_noun_generator.generate
+      ascii_letter_a = 97
+      first_letter = (rand(26) + ascii_letter_a).chr
+      second_letter = (rand(26) + ascii_letter_a).chr
+      @adjective_noun_generator.generate + "-#{first_letter}#{second_letter}"
```
tcdowney: Why not use something like `SecureRandom.hex` and append its output to the route?

```
SecureRandom.hex(3)
=> "19c483"
```

jpalermo: Letters give us a bigger random space to use. Although using 3 hex characters would give us more than 2 letters.

We thought about using letters and numbers, but didn't want the possibility of confusing `1` with `l` or `0` with `O`. And numbers in hostnames are kind of weird too, not illegal or anything, but somewhat unusual.

We had quite a few discussions about it, originally just adding a 3rd word seemed like a sane idea. Although we've already had complaints from users about long app names getting rejected when the random part of the route makes the route too long and they get an error back.

At the end of the day, `hex(3)` is probably fine too. Any solution will work as long as it's not too long, not too confusing, and not likely to add swear words back into the mix.

PR #1500, opened 2019-11-21
V3 service_offerings visibility for space-scoped offerings

blgm:
- service offerings from space-scoped can be seen
- split the ServiceOfferingsFetcher so that plan visibility is handled
with a new ServicePlanVisibilityFetcher
- now respects the `hide_marketplace_from_unauthenticated_users` feature flag

[more details in story #169482815](https://www.pivotaltracker.com/story/show/169482815)

### DIFF
--- a/app/controllers/v3/service_offerings_controller.rb
+++ b/app/controllers/v3/service_offerings_controller.rb
@@ -1,37 +1,52 @@
 require 'fetchers/service_offerings_fetcher'
+require 'fetchers/service_plan_visibility_fetcher'
 require 'presenters/v3/service_offering_presenter'
 
 class ServiceOfferingsController < ApplicationController
   def show
     guid = hashed_params[:guid]
 
-    offering = if !current_user
-                 ServiceOfferingsFetcher.fetch_one_anonymously(guid)
-               elsif permission_queryer.can_read_globally?
-                 ServiceOfferingsFetcher.fetch_one(guid)
-               else
-                 ServiceOfferingsFetcher.fetch_one(guid, org_guids: permission_queryer.readable_org_guids)
-               end
+    not_authenticated! if !current_user && VCAP::CloudController::FeatureFlag.enabled?(:hide_marketplace_from_unauthenticated_users)
 
+    offering, space, public = ServiceOfferingsFetcher.fetch(guid)
     service_offering_not_found! if offering.nil?
 
-    presenter = Presenters::V3::ServiceOfferingPresenter.new(offering)
-    render status: :ok, json: presenter.to_json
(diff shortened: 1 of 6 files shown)


spec/request/service_offerings_spec.rb
```diff
@@ -67,7 +70,21 @@
...
+      it_behaves_like 'permissions for single object endpoint', COMPLETE_PERMISSIONS
+
+      context 'when the hide_marketplace_from_unauthenticated_users feature flag is enabled' do
+        before do
+          VCAP::CloudController::FeatureFlag.create(name: 'hide_marketplace_from_unauthenticated_users', enabled: true)
+        end
+
+        let(:expected_codes_and_responses) do
+          h = Hash.new(successful_response)
+          h['unauthenticated'] = { code: 401 }
```
winnab: Should this be a 401 or a 404?

blgm: The V2 behaviour is to be 401, so it's consistent with V2 for this to be 401: https://github.com/cloudfoundry/cloud_controller_ng/blob/13fa61ceba7d731adb57c6dd3664f7c815875bd1/app/controllers/services/services_controller.rb#L30-L31

app/controllers/v3/service_offerings_controller.rb
```diff
@@ -1,37 +1,52 @@
...
-    super
+    action_name == 'show' ? false : super
   end
 
   def enforce_read_scope?
-    return false if action_name == 'show'
+    action_name == 'show' ? false : super
+  end
+
+  private
```
sethboyles: this should be moved up to L#21 I think, since the `enforce_` methods are private in the application controller

app/fetchers/service_plan_visibility_fetcher.rb
```diff
@@ -0,0 +1,15 @@
+module VCAP::CloudController
+  class ServicePlanVisibilityFetcher
+    class << self
+      def service_plans_visible_in_orgs?(service_plan_guids, readable_org_guids)
+        empty = ServicePlanVisibility.dataset.
+                left_join(:organizations, id: Sequel[:service_plan_visibilities][:organization_id]).
+                left_join(:service_plans, id: Sequel[:service_plan_visibilities][:service_plan_id]).
+                where { (Sequel[:service_plans][:guid] =~ service_plan_guids) & (Sequel[:organizations][:guid] =~ readable_org_guids) }.
```
reid47: What is this line doing? What are those `=~` operators in this context?

blgm: The `=~` operator is documented here: https://sequel.jeremyevans.net/rdoc/files/doc/dataset_filtering_rdoc.html

We felt that it was more readable than the alternative, but are open to changing it if folks are more familiar with other ways of doing it.

PR #1502, opened 2019-11-22
Change label selector query logic to allow complex queries

sethboyles:

* [ ] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/CONTRIBUTING.md)

* [ ] I have made this pull request to the `master` branch

### DIFF
--- a/app/fetchers/app_revisions_fetcher.rb
+++ b/app/fetchers/app_revisions_fetcher.rb
@@ -10,7 +10,7 @@ module VCAP::CloudController
       if message.requested?(:label_selector)
         dataset = LabelSelectorQueryGenerator.add_selector_queries(
           label_klass: RevisionLabelModel,
-          resource_dataset: dataset,
+          resource_dataset: dataset.qualify,
           requirements: message.requirements,
           resource_klass: RevisionModel,
         )
--- a/app/fetchers/domain_fetcher.rb
+++ b/app/fetchers/domain_fetcher.rb
@@ -17,7 +17,7 @@ module VCAP::CloudController
           [:id, readable_shared_private_domains_filter]
         ]).sql_boolean
 
-        Domain.where(user_visible_domains)
+        Domain.where(user_visible_domains).qualify
       end
 
       def fetch(message, readable_org_guids)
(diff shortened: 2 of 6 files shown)


sethboyles:
@ericpromislow:

@reneighbor and I took your tests to try to come up with a more general solution so we could fix this across controllers more easily.  The weirdness is around the sequel `qualify` method:

```
0> resource_dataset.sql
=> "SELECT * FROM \"routes\" WHERE ((\"guid\" IN ('route-in-org-guid', 'route-in-other-org-guid', 'guid-1', 'guid-2')) AND (\"routes\".\"space_id\" IN (SELECT \"spaces\".\"id\" FROM \"spaces\" WHERE ((\"guid\" IN ('839a4ce2-ce77-4b57-9c2b-ead835a69d3d')) AND (\"spaces\".\"id\" IS NOT NULL)))))"
0> resource_dataset.qualify(:routes).sql
=> "SELECT \"routes\".* FROM \"routes\" WHERE ((\"routes\".\"guid\" IN ('route-in-org-guid', 'route-in-other-org-guid', 'guid-1', 'guid-2')) AND (\"routes\".\"space\" IN (SELECT * FROM \"spaces\" WHERE (\"guid\" IN ('839a4ce2-ce77-4b57-9c2b-ead835a69d3d')))))"
```


Let me know what you think.

sethboyles:
I might reach out to Jeremy Evans tommorow to see if there is another way

sethboyles:
I spoke with Jeremy Evans and this is the solution he recommends.  We should be using `qualify` before we join, not after.

sethboyles:
> jeremyevans  10:58:27
sbboyles: Note that in general, you should always qualify before you join, not after.  The purpose of qualifying is to take an unqualified dataset and qualify it so that if you join you don't end up with an ambiguous reference
sbboyles  10:59:17
That is what I thought--I am just trying to understand how changing `space_id` to `space` is disambiguating anything
Or what the intention is there
jeremyevans  11:01:59
sbboyles: In order for the model association dataset filtering to work, you need to have :space be a plain symbol.  When you call dataset.quality, it changes :space to a Sequel object (Sequel[:routes][:space]).  At that point the model association dataset filtering code does not recognize it as an association

PR #1513, opened 2019-12-12
Add link for deployment cancel action

Gerg:
Only visible if the deployment is cancelable

### DIFF
--- a/app/actions/deployment_cancel.rb
+++ b/app/actions/deployment_cancel.rb
@@ -11,7 +11,7 @@ module VCAP::CloudController
       def cancel(deployment:, user_audit_info:)
         deployment.db.transaction do
           deployment.lock!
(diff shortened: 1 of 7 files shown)


app/actions/deployment_cancel.rb
```diff
@@ -30,9 +30,7 @@ def cancel(deployment:, user_audit_info:)
       private
 
       def valid_state?(deployment)
```
selzoc: I feel like i would just get rid of this method if we're delegating to the `DeploymentModel`

sethboyles: Agreed.  Especially if the parameter here isn't being checked.

sethboyles: ^^

selzoc:
Going to close/reopen this to trigger our new checks

PR #1515, opened 2019-12-12
Remove deployment state from API

Gerg:
It's currently marked as deprecated in the docs. Not sure what the blockers are for removing it, but it might make sense to remove when GAing deployments.

I removed state from the presenter, but did not touch any of the internal logic that currently depends on deployment state. Not sure if y'all were thinking of doing a larger, internal refactor as part of removing the field. Examples where it is used:
- https://github.com/cloudfoundry/cloud_controller_ng/blob/8af9682bcb0510538567f9113eb719fd8b95c636/lib/cloud_controller/deployment_updater/dispatcher.rb#L14-L15
- https://github.com/cloudfoundry/cloud_controller_ng/blob/8af9682bcb0510538567f9113eb719fd8b95c636/app/actions/deployment_cancel.rb#L33-L35

### DIFF
--- a/app/presenters/v3/deployment_presenter.rb
+++ b/app/presenters/v3/deployment_presenter.rb
@@ -10,7 +10,6 @@ module VCAP::CloudController::Presenters::V3
         guid: deployment.guid,
         created_at: deployment.created_at,
         updated_at: deployment.updated_at,
-        state: deployment.state,
         status: {
           value: deployment.status_value,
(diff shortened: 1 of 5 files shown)


sethboyles:
Hm, it might be nice to refactor all the internal logic depending on state before/when we merge this...

selzoc:
@Gerg do you mine rebasing this?

PR #1550, opened 2020-01-23
Implement GET /v3/service_plans

blgm:
- and support for "hide_marketplace_from_unauthenticated_users" feature flag
- also, add documentation for GET Service plans

https://www.pivotaltracker.com/story/show/170725626
https://www.pivotaltracker.com/story/show/170725626

Co-authored-by: Marcela Campo <mcampo@pivotal.io>

### DIFF
--- a/app/models/services/service_plan.rb
+++ b/app/models/services/service_plan.rb
@@ -134,6 +134,12 @@ module VCAP::CloudController
       service.bindable?
     end
 
+    def plan_updateable?
+      return plan_updateable unless plan_updateable.nil?
+
+      !!service.plan_updateable
+    end
+
     def service_broker
       service.service_broker if service
     end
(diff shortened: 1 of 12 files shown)


app/models/services/service_plan.rb
```diff
@@ -134,6 +134,12 @@ def bindable?
       service.bindable?
     end
 
+    def plan_updateable?
```
blgm: This looks a bit odd.  Is there a clearer way to describe it?

app/presenters/v3/service_offering_presenter.rb
```diff
@@ -42,12 +42,13 @@ def service_offering
         end
 
         def shareable(metadata)
-          metadata.nil? ? false : metadata['shareable'] == true
+          metadata['shareable'] == true
```
blgm: If `service_offering.metadata = "null"` then `JSON.parse(service_offering.metadata)` will be `nil`, and there will be an exception

FelisiaM: This should return an empty object rather than a nil so it should not be a problem anymore

PR #1553, opened 2020-02-04
Change the metadata representation

paroxp:
At the moment, an API is returning a string field:

```
{
  "metadata": "\n{\"test\": \"yes\"}"
}
```

This means the client would need to parse the string into an object in
order to use it. It's not practical and provides an overhead.

We can remove the JSON parsing which will cause the object to expand
further into:

```
{
  "metadata": {"test": "yes"}
}
```

I haven't written any tests nor signed CLA, just creating a cause for discussion.


paroxp:
Looks like it's constantly changing. Looks good at the current. Thanks!

PR #1557, opened 2020-02-14
Return empty Loggregator envelope if Logcache is unavailable

heycait:
[#170267522]

Co-authored-by: David Timm <dtimm@pivotal.io>

This will allow `cf push` and `cf app` to work even if log-cache is unavailable or removed.

If log-cache is rolling, users cannot `cf push` apps.
https://github.com/cloudfoundry/log-cache-release/issues/25#issuecomment-548294752

N/A

### DIFF
--- a/lib/logcache/client.rb
+++ b/lib/logcache/client.rb
@@ -1,4 +1,5 @@
 require 'logcache/logcache_egress_services_pb'
+require 'logcache/v2/envelope_pb'
 
 module Logcache
   class Client
@@ -18,7 +19,7 @@ module Logcache
     end
 
     def container_metrics(source_guid:, envelope_limit: DEFAULT_LIMIT, start_time:, end_time:)
-      with_request_error_handling do
+      with_request_error_handling(source_guid) do
         service.read(
           Logcache::V1::ReadRequest.new(
             source_id: source_guid,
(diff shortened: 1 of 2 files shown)


heycait:
The `bundle exec rake spec:all` tests seems to be flakey. I didn't change anything and the times it ran, it failed for different reasons...

https://travis-ci.org/cloudfoundry/cloud_controller_ng/jobs/650578418?utm_medium=notification&utm_source=github_status

jspawar:
Hey @heycait,

Had some questions around the specifics of the failure modes of a client's interaction with the log-cache server. In the instance motivating this particular PR where we have no server to provide container metrics, can we safely assume we will always receive a gRPC status code of 14? How does the client distinguish between a non-existent/unresponsive server and any other errors from that server? Does anything else return a status code of 14? 

Reason we are clarifying is that we are planning to incorporate your suggested changes behind a temporary feature flag which will allow operators to opt into this behavior of returning empty container metrics and want to ensure that we are capturing only the minimal set of failures behind that flag. We would like to prevent these empty metrics from becoming the default failure response going forward.

Thanks,
Jwal + @christarazi

jspawar:
Working containing your PR: https://github.com/cloudfoundry/cloud_controller_ng/tree/wip-empty-container-metrics-support

heycait:
Hi @jspawar @christarazi 

Yes, when there is no server available the client will return a status code of 14 which corresponds to Unavailable. The status codes come from when you use the LogCache client (which is a Ruby gRPC client associated with our gRPC server) to connect to the gRPC server. 

Here's some resources for the status codes that get returned from gRPC:
* https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/errors.rb
* https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

I can't speak to the specifics of how gRPC does this, but they do have status codes for "Not Found", "Deadline Exceeded", "Permission Denied", etc. I hope this addresses the concerns you mentioned.

Ideally, we would like this to be the default behavior and not captured under a flag as `cf push` still successfully pushes an application even if it can't gather the metrics for it, so the failure there is a bit misleading. 

Perhaps @hev can speak more about that from the product side.

hev:
Some context on this circuit breaker. 

* This originally came away from several page incidents on PWS for cf push failures. Container metrics used a legacy system on dopplers and it frequently timed out the request and a circuit breaker was instrumented on the metric system that created this behavior. 
* As we transitioned to a new system (log-cache) the circuit breaker has been both on and off at certain times. It doesn't seem like reports of cf push failures have been as prevalent with it off, but they do still coming up. 
* There is currently a need for a circuit breaker in PAS for k8s (but we actually will likely be past that need very soon). 

It seems fine to feature flag it in for now as I think moving this circuit breaker up stream is the improvement here.

jspawar:
Hey @hev @heycait,

Thank you very much for clarifying and adding the additional context. We'll move forward with your PR and keeping it behind our feature flag then. 

@jspawar @christarazi

jspawar:
Merged in with our changes: https://github.com/cloudfoundry/cloud_controller_ng/commit/df477fa16e8edbd2d4391e20e7be4402cc7beb43

PR #1560, opened 2020-02-21
make cc block until ccdb is current or newer than local migrations

cwlbraa:
following advice from https://srcco.de/posts/kubernetes-liveness-probes-are-dangerous.html

make sure that your Readiness Probe includes database initialization/migration
the simplest way to achieve this is to start the HTTP server listening only after the initialization finished,
i.e. instead of changing the health check status, just don't start the web server until the DB migration is complete

this also makes capi more conformant with https://github.com/cloudfoundry-incubator/kubernetes-guidelines/blob/master/DETAILED.md#rollback

[#171394379]

fixes https://github.com/cloudfoundry/capi-release/issues/165

### DIFF
--- a/config/cloud_controller.yml
+++ b/config/cloud_controller.yml
@@ -2,6 +2,7 @@
 local_route: 127.0.0.1
 external_port: 8181
 tls_port: 8182
+readiness_port: 9999
 pid_filename: /tmp/cloud_controller.pid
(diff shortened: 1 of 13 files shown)


cwlbraa:
the first bit of this track of work has been merged out-of-band.

PR #1561, opened 2020-02-21
Remove old migrations specs

sethboyles:
Removing old migrations specs here as they do not provide us much value and can be slow (particularly slow on my 2015 macbook air).

I tried squashing these migrations as well using Sequel's [schema dump feature:](https://sequel.jeremyevans.net/rdoc/files/doc/migration_rdoc.html#label-Dumping+the+current+schema+as+a+migration) but, as the docs note, Sequel cannot correctly dump constraints.  it seems to dump some defaults and constraints correctly (like defaulting to a timestamp), but columns with a default of an integer, for example, won't be dumped correctly.

If we really wanted to do this we would probably have to instead do two dumps, one using MySql's and one using Postgres's, and loading that instead.  In short... probably not worth the effort :)


cwlbraa:
I think it'd be worth a try just to see how much faster test setup is in MySQL...

jspawar:
Going to merge because these specs aren't really used anymore. 

Re: squashing the migrations, we will create a chore in our backlog or maybe another GitHub issue to address how to/if we will do that.

Sound good? @sethboyles @cwlbraa

PR #1575, opened 2020-03-17
Remove the Customer IP from the info log message in the request logs

MMisoch:
This is a small change that removes the IPs in the log messages that are printed in the `info` log level in the request logs. We really want to keep the logs for as long as possible since we need them for debugging. However we can not do that because there is a policy in place that stops us from keeping specific customer information for longer than 3 days. Because the IPs in the log messages are customer specific we can not keep our logs longer than 3 days which really hurts debugging for now.

Remove the customer IPs from the log level `info` messages. 

We can not have the IP of customers in the log messages as it violates our corporate rules

I run `bundle exec rspec spec/middleware/request_logs_spec.rb` specifically

### DIFF
--- a/middleware/request_logs.rb
+++ b/middleware/request_logs.rb
@@ -10,11 +10,10 @@ module CloudFoundry
         request = ActionDispatch::Request.new(env)
 
         @logger.info(
-          sprintf('Started %<method>s "%<path>s" for user: %<user>s, ip: %<ip>s with vcap-request-id: %<request_id>s at %<at>s',
+          sprintf('Started %<method>s "%<path>s" for user: %<user>s with vcap-request-id: %<request_id>s at %<at>s',
             method: request.request_method,
             path: request.filtered_path,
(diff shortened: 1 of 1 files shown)


cwlbraa:
> there is a policy in place that stops us from keeping specific customer information for longer than 3 days

MMisoch:
It is regulated by the GDPR. We want to have the Debug Information (logs) for at least 2 weeks which is not allowed here if they contain customer specific IP-Address data.

cwlbraa:
awesome, thank you for the info @MMisoch and the PR. We'll probably work to merge @FloThinksPi 's PR over this one, so I'm going to close this out. If you have improvements over the other PR feel free to reopen! 

closing in favor of
https://github.com/cloudfoundry/cloud_controller_ng/pull/1585

MMisoch:
Sure, that was the plan from our side after having some more discussion on the topic.

PR #1576, opened 2020-03-18
Keep AppUsageEvent records for running apps

joyvuu-dave:
**Proposed Change**
Maintain AppUsageEvent records for running apps

**Use Cases**
By default, AppUsageEvents are purged every 30 days. This means
applications that want to track application usage in CloudFoundry must be started within
30 days of a foundation starting. The only way for applications to track
usage after 30 days is for the app to first make a `purge_and_seed`
request. This has the effect of pruning all exising AppUsageEvents and creating fake start events
for any currently running applications.

By saving the AppUsageEvents of running applications, applications that
begin tracking usage after 30 days will not have to make a
`purge_and_seed` request. Their knowledge of running applications will
be accurate and multiple applications that wish to track usage can do so
without effecting other applications that are already doing so.

### DIFF
--- a/app/repositories/app_usage_event_repository.rb
+++ b/app/repositories/app_usage_event_repository.rb
@@ -145,7 +145,7 @@ module VCAP::CloudController
       end
 
       def delete_events_older_than(cutoff_age_in_days)
-        Database::OldRecordCleanup.new(AppUsageEvent, cutoff_age_in_days, keep_at_least_one_record: true).delete
+        Database::OldRecordCleanup.new(AppUsageEvent, cutoff_age_in_days, keep_at_least_one_record: true, keep_running_app_records: true).delete
       end
 
       private
--- a/lib/database/old_record_cleanup.rb
+++ b/lib/database/old_record_cleanup.rb
@@ -2,13 +2,15 @@ require 'database/batch_delete'
 
 module Database
   class OldRecordCleanup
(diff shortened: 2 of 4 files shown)


app/repositories/app_usage_event_repository.rb
```diff
@@ -145,7 +145,7 @@ def purge_and_reseed_started_apps!
       end
 
       def delete_events_older_than(cutoff_age_in_days)
-        Database::OldRecordCleanup.new(AppUsageEvent, cutoff_age_in_days, keep_at_least_one_record: true).delete
+        Database::OldRecordCleanup.new(AppUsageEvent, cutoff_age_in_days, keep_at_least_one_record: true, keep_running_app_records: true).delete
```
tcdowney: Have y'all done any performance testing around the impact this might have on Cloud Controllers on foundations with a large number of running apps? I think even with the thirty-day expiry for events large environments can get pretty bogged-down.

@zrob might have some more historical context here.

joyvuu-dave: I see there were some previous stories around performance issues while deleting old records, and I'll investigate the impact my PR might have. Thanks for sharing that context.

joyvuu-dave:
We've identified an issue with the current implementation. Because deletes are done in batches, we could end up deleting the stop event in one pass, and then not knowing that we should delete the corresponding start event in a subsequent pass. 

Working on a different approach.

Issue #1568, opened 2020-03-05
Customer IPs in Cloud Controller log messages


MMisoch:
As already shortly discussed in the call with @selzoc and other members of the capi team yesterday we would like to have an improvement on the logging of the cloud controller

## Issue

In the log messages of the cloud controller there are customer IPs logged

## Context

On the info log level the IPs of the request are contained in the log message.
However we are (by policy) not allow that customer IPs are logged and persisted for longer than 3 days. The problem we got now is that we want to keep the rest of the logs for a longer period than 3 days because of debugging reasons. There are a few possible solutions we came up with:
See possible fix section

## Steps to Reproduce

See cloud controller request logs:

> "Started DELETE \"/v2/service_instances/XXXX?async=true\" for user: XXXX, **ip: 35.198.152.XX** with vcap-request-id: XXX::XXX at 2020-02-19 13:36:49 UTC"

## Expected result

Having the same logs without the IPs so they can be kept longer than 3 days.

## Current result

Logs with customer IPs must be deleted after 3 days because of policy reasons

## Possible Fix

- Remove IPs from logging in general
- Move IP logging from info to debug log level
- Remove the last octet from the IPs so that they are anonymous
- Config switch for enabling/disabling IP logging

Changes here could do the trick:
https://github.com/cloudfoundry/cloud_controller_ng/blob/44e2d6cf46f9d2b451d3ff57600cc1e049f6be28/middleware/request_logs.rb#L12-L19

MMisoch:
I created a PR with a possible fix for this Issue:
https://github.com/cloudfoundry/cloud_controller_ng/pull/1575

aashah:
PR is merged. Closing this issue. Please re-open/comment if there's more!

Thanks for opening the issue.

PR #1585, opened 2020-03-25
Add config flag for GDPR compliance. Remove last octet from ips in request-logs accordingly.

FloThinksPi:

IP Adresses now get their last octet cut of and replaced by 0 if the
new config option logging.gdpr_compliant is true. E.g 192.168.1.80
will be logged as 192.168.1.0 which provides sufficent anonymization
to comply with european data protection regulations.

If the flag is false or not present, CC will have the old behaviour
which means logging the full ips.

This flag can also be used in other places that one may find needs
adoptions to comply to EU GDPR.

It provides a way to comply to the european data protection regulations.
All EU Customers of Pivotal that save their logs would benefit from this change.
The IP is specifically called as Personal information: https://gdpr.eu/recital-30-online-identifiers-for-profiling-and-identification/
So either one has to:
- Not save it at all
- Request formal approval from all users for saving their ips
- Anonymise the IP so it can not be tracked down to an individual

This change enables customers to use the third option as the first ones are not practicable.

Followup of: #1575 
Issue: #1568 
Slack discussion: https://cloudfoundry.slack.com/archives/C07C04W4Q/p1584635872138900

Could not get CATS to run(having no bosh-lite atm). Would be thankful if some pivot can run it against one of you bosh-lites.

### DIFF
--- a/config/cloud_controller.yml
+++ b/config/cloud_controller.yml
@@ -89,6 +89,7 @@ logging:
   file: /tmp/cloud_controller.log
   level: debug2
   syslog: vcap.example
+  anonymize_ips: false
 
 telemetry_log_path: spec/artifacts/cloud_controller_telemetry.log
 
--- a/middleware/request_logs.rb
+++ b/middleware/request_logs.rb
@@ -1,3 +1,5 @@
+require 'ipaddr'
+
 module CloudFoundry
   module Middleware
     class RequestLogs
@@ -6,17 +8,31 @@ module CloudFoundry
         @logger = logger
       end
 
+      def anonymize_ip(request_ip)
+        # Remove last octet of ip if EU GDPR compliance is needed
+        ip = IPAddr.new(request_ip) rescue nil
+        if VCAP::CloudController::Config.config.get(:logging, :anonymize_ips) && ip.nil?.!
+           if ip.ipv4?
+             ip.to_string.split('.')[0...-1].join('.') + '.0'
+           else
+             ip.to_string.split(':')[0...-5].join(':') + ':0000:0000:0000:0000:0000'
+           end
+         else
+           ip.to_string rescue request_ip
+         end
+      end
+
       def call(env)
         request = ActionDispatch::Request.new(env)
 
         @logger.info(
           sprintf('Started %<method>s "%<path>s" for user: %<user>s, ip: %<ip>s with vcap-request-id: %<request_id>s at %<at>s',
-            method: request.request_method,
-            path: request.filtered_path,
-            user: env['cf.user_guid'],
-            ip: request.ip,
-            request_id: env['cf.request_id'],
-            at: Time.now.utc)
+                  method: request.request_method,
+                  path: request.filtered_path,
+                  user: env['cf.user_guid'],
+                  ip: anonymize_ip(request.ip),
+                  request_id: env['cf.request_id'],
+                  at: Time.now.utc)
         )
 
         status, headers, body = @app.call(env)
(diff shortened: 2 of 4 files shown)


config/cloud_controller.yml
```diff
@@ -89,6 +89,7 @@ logging:
   file: /tmp/cloud_controller.log
   level: debug2
   syslog: vcap.example
+  gdpr_compliant: false
```
selzoc: Making the flag imply that turning it to true makes the cloud controller fully gdpr-compliant worries me a bit.

middleware/request_logs.rb
```diff
@@ -9,12 +9,15 @@ def initialize(app, logger)
       def call(env)
         request = ActionDispatch::Request.new(env)
 
+        # Remove last octet of ip if EU GDPR compliance is needed
+        ip = VCAP::CloudController::Config.config.get(:logging, :gdpr_compliant) ? request.ip.split('.')[0...-1].join('.') + '.0' : request.ip
```
selzoc: This looks like it'll only work with ipv4 - maybe worth looking into a gem that also supports ipv6 to do this as many folks are probably anonymizing IP addresses in ruby for gdpr compliance. (I found https://github.com/ankane/ip_anonymizer after a brief search)

selzoc: I would probably also favor this construct over a ternary:
```ruby
ip = if VCAP::CloudController::Config.config.get(:logging, :gdpr_compliant)
         <redact ip>
      else
        <unredacted-ip>
```

config/cloud_controller.yml
FloThinksPi: Hm yea i just wanted to have a parameter that can be reused in multiple parts in the logging code for the same purpose. E.g. we also found some emails in the logs which also have to be hidden to comply, but we are looking at one thing at a time currently.

middleware/request_logs.rb
FloThinksPi: @selzoc Thank you for the hint. We changed the implementation quite a bit to allow for v6 adresses. We ditched third party libs to not introduce new dependencies as the task is quite simple to adress with the stdlib of ruby. Also using you recommendation for the variable assignment : ).

If we find we need this in other parts of the CC we will think of pulling the code to anonymize the ip out, somwhere else to make it reusable. For now i think we are good keeping it in request_logs.

config/cloud_controller.yml
FloThinksPi: Ah and we have it under the logging node on purpose to indicate it is just anonymizing logs so the can be stored safely and still comply to gdpr. In contrast to other parts of the cc (like database user table etc.) there to keep emails and ips in logs for the CC to do its work. For saving personal data in CC for the purpose of providing the service the gdpr does not limit this as long as you tell the customer. So imho its just the logging that currently is not optimal and thus the config parameter is located there :).

selzoc: 😊 I totally did not expand the preview to see it was under logging 👍

middleware/request_logs.rb
selzoc: I still think it would be clearer to pull it out into a function.  But yeah I guess no need for a gem!

FloThinksPi: @selzoc as i am still not that familiar with the codebase, where would such a function be placed in the coding ? Maybe lib/utils ?

selzoc: I just meant extracting lines 15-24 into a function within the same class, just to clean up the `call` function a bit.

FloThinksPi: Done : )

FloThinksPi:
@selzoc can you review again,  also squashed all changes into one commit so history is more clean once merged.

aashah:
Thanks for the PR!

PR #1588, opened 2020-04-06
Implement the OPI task client

danail-branekov:
Implement task creation in the OPI task client. Note that we do not support other task operations yet, such as `cancel`, i.e. tasks CATS tests would fail if run against CF+Eirini deployment.

The task client `desire_task` method has its signature changed to
receive the task rather than the task definition. This caused the bbs
task client to build the task definition via an injected recipe builder.
This also brought some changes in the dependency locator wiring.

[#171785772]

Co-authored-by: Julian Skupnjak <skupnjak@de.ibm.com>
Co-authored-by: Georgi Sabev <gcapizzi@pivotal.io>
Co-authored-by: Danail Branekov <danailster@gmail.com>
Co-authored-by: Giuseppe Capizzi <gcapizzi@pivotal.io>

### DIFF
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -432,11 +432,19 @@ module CloudController
     end
 
     def build_opi_task_client
-      ::OPI::TaskClient.new(config)
+      ::OPI::TaskClient.new(config, build_task_completion_callback_generator, VCAP::CloudController::Diego::TaskEnvironmentVariableCollector)
     end
 
     def build_bbs_task_client
-      VCAP::CloudController::Diego::BbsTaskClient.new(build_bbs_client)
+      VCAP::CloudController::Diego::BbsTaskClient.new(config, build_bbs_client, build_task_recipe_builder)
+    end
+
+    def build_task_recipe_builder
+      VCAP::CloudController::Diego::TaskRecipeBuilder.new
+    end
+
+    def build_task_completion_callback_generator
+      VCAP::CloudController::Diego::TaskCompletionCallbackGenerator.new(config)
     end
 
     def build_instances_client
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -432,19 +432,11 @@ module CloudController
     end
 
     def build_opi_task_client
-      ::OPI::TaskClient.new(config, build_task_completion_callback_generator, VCAP::CloudController::Diego::TaskEnvironmentVariableCollector)
+      ::OPI::TaskClient.new(config, VCAP::CloudController::Diego::TaskEnvironmentVariableCollector)
     end
 
     def build_bbs_task_client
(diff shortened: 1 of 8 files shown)


lib/cloud_controller/dependency_locator.rb
```diff
@@ -432,11 +432,19 @@ def build_task_client
...
-      ::OPI::TaskClient.new(config)
+      ::OPI::TaskClient.new(config, build_task_completion_callback_generator, VCAP::CloudController::Diego::TaskEnvironmentVariableCollector)
     end
 
     def build_bbs_task_client
-      VCAP::CloudController::Diego::BbsTaskClient.new(build_bbs_client)
+      VCAP::CloudController::Diego::BbsTaskClient.new(config, build_bbs_client, build_task_recipe_builder)
+    end
+
+    def build_task_recipe_builder
```
aashah, line 442: Would prefer to see `build_task_recipe_builder` simply be instantiated within `BbsTaskClient`. In this way, it mirrors the same setup as `BbsAppsClient` instantiating the `AppRecipeBuilder`.

lib/cloud_controller/dependency_locator.rb
```diff
@@ -432,11 +432,19 @@ def build_task_client
...
+      VCAP::CloudController::Diego::TaskRecipeBuilder.new
+    end
+
+    def build_task_completion_callback_generator
+      VCAP::CloudController::Diego::TaskCompletionCallbackGenerator.new(config)
```
aashah, line 447: This is something that is used in the Diego workflow as well, and is instantiated inside the method here - https://github.com/cloudfoundry/cloud_controller_ng/blob/7a977015f82c2799c6c84906812bb7fa38c09c92/lib/cloud_controller/diego/task_recipe_builder.rb#L19

I feel like we should pick one style over another (instantiating directly within the class versus dependency injected)

gdankov:
@aashah Followed your recommendations and pushed a commit with these changes.

Kind regards,
Georgi && @kieron-pivotal

georgethebeatle:
We need the task name as well for this story(https://www.pivotaltracker.com/story/show/171785787) so we sneaked it in this PR to minimise the back and forth. Hope you don't mind :)

sethboyles:
When talking with the Release Integration team about another feature, they expressed concerns around integrating features that don't have the cleanup part of the lifecycle implemented.  For example, if I'm understanding correctly, with this PR if you delete an app with running tasks, those tasks will be left running.  I think we would prefer to have `cancel` be done as either part of this PR or a following PR so we can avoid the issue. 

Any chance you all will be able to get to the `cancel` part of this feature soon?

selzoc:
As we discussed in our sync, we will merge this after the delete task is implemented, whether that's in another pr or this one

danail-branekov:
Hello, we have implemented the task cancellation bit in Eirini. We managed to successfully execute task creation/termination CATs successfully having this PR applied. Therefore, please proceed with reviewing.

Thank you!

PR #1596, opened 2020-04-22
Clarify plan_updateable overrides

gberche-orange:

The OSB API states regarding a service offering metadata that its updateable flag can be overriden in plans.

https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#service-offering-object

> Response Field | Type | Description
> -- | -- | --
> [...]
> plan_updateable | boolean | Whether the Service Offering supports upgrade/downgrade for Service  Plans by default. Service Plans can override this field (see Service Plan). Please note that the misspelling of the attribute plan_updatable as plan_updateable  was done by mistake. We have opted to keep that misspelling instead of  fixing it and thus breaking backward compatibility. Defaults to false.
> 

https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#service-plan-object

> Response Field | Type | Description
> -- | -- | --
> [...]
> plan_updateable | boolean | Whether the Plan supports upgrade/downgrade/sidegrade to another  version. This field is OPTIONAL. If specified, this takes precedence  over the Service Offering's plan_updateable field. If not  specified, the default is derived from the Service Offering. Please note  that the attribute is intentionally misspelled as plan_updateable for legacy reasons.
> 

Unfortunately the CC API reuses the two fields, but isn't precise enough to state that plan.updateable override serviceoffering.plan_updateable

Clarify the expected CC behavior, to avoid CC API clients to make wrong assumptions, see related https://github.com/cloudfoundry/cf-java-client/issues/1044

### DIFF
--- a/docs/v3/source/includes/experimental_resources/service_offerings/_object.md.erb
+++ b/docs/v3/source/includes/experimental_resources/service_offerings/_object.md.erb
@@ -22,7 +22,7 @@ Name | Type | Description
 **broker_catalog** | _object_ | This object contains information obtained from the Service Broker Catalog.
 **broker_catalog.id** | _string_ | The identifier that the Service Broker provided for this Service Offering.
 **broker_catalog.metadata** | _object_ | Additional information provided by the Service Broker.
-**broker_catalog.features.plan_updateable** | _boolean_ | Whether the Service Offering supports upgrade/downgrade for Service Plans by default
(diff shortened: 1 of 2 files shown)


blgm:
I have resolved this in alternative commit because:
- Subsequent changes to the files meant that conflicts had to be resolved
- I wanted to tweak the wording so that it was clear that the property can only be set in the catalog and not via the CC API.

Thank you highlighting the issue @gberche-orange

gberche-orange:
thanks @blgm !

PR #1601, opened 2020-04-29
Create consistent AppUsageEvent records

joyvuu-dave:
When a user makes a POST to
/v2/app_usage_events/destructively_purge_all_and_reseed_started_apps,
all existing AppUsageEvent records are purged from the database and new
AppUsageEvent records are created for any running processes. This action
is taken by clients who may have missed previous events that have now
been pruned from the database as part of routeine mainteance and now
need a new starting spot for tracking usage.

All AppUsageEvent records created outside of a purge and reseed action
will have their parent_app_name and parent_app_guid fields set. This
change makes AppUsageEvent recoreds created as a result of a purge
consistent with all other records in that respect.

Not having the parent_app_name and parent_app_guid fields set was
causing 3rd party usage monitoring systems that expected their presence
to break.

### DIFF
--- a/app/repositories/app_usage_event_repository.rb
+++ b/app/repositories/app_usage_event_repository.rb
@@ -110,6 +110,8 @@ module VCAP::CloudController
           space_name:                         "#{Space.table_name}__name".to_sym,
           org_guid:                           "#{Organization.table_name}__guid".to_sym,
           created_at:                         Sequel.datetime_class.now,
+          parent_app_name:                    :parent_app__name,
+          parent_app_guid:                    :parent_app__guid,
         }
 
         latest_package_query = PackageModel.select(:app_guid).select_append { max(id).as(:id) }.group(:app_guid)
--- a/spec/unit/repositories/app_usage_event_repository_spec.rb
+++ b/spec/unit/repositories/app_usage_event_repository_spec.rb
@@ -546,6 +546,13 @@ module VCAP::CloudController
             expect(AppUsageEvent.last).to match_app(process)
           end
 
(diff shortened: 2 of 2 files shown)


sannidhi:
Thanks for the PR @joyvuu-dave. We will take a look and merge once checks complete.

sannidhi:
ah you're getting robocoped @joyvuu-dave 
https://travis-ci.org/github/cloudfoundry/cloud_controller_ng/jobs/681201719

Looks like its just a trailing whitespace but it will fail the units.

app/repositories/app_usage_event_repository.rb
```diff
@@ -110,6 +110,8 @@ def purge_and_reseed_started_apps!
           space_name:                         "#{Space.table_name}__name".to_sym,
           org_guid:                           "#{Organization.table_name}__guid".to_sym,
           created_at:                         Sequel.datetime_class.now,
+          parent_app_name:                    :parent_app__name,
+          parent_app_guid:                    "#{ProcessModel.table_name}__guid".to_sym,
```
sethboyles, line 114: should this be `#{AppModel.table_name}__guid`?

joyvuu-dave: I don't think the reference to AppModel resolves (at least it didn't when I tried). However, it looks like we can just reference parent_app__guid directly. I updated the PR and also added a test to confirm parent_app_guid and app_guid end up being equal.

joyvuu-dave: One last change -- I updated the test to show that `parent_app_name` and `parent_app_guid` are set to the values of the App name and guid.

PR #1603, opened 2020-05-04
Give option to cleanup all orphaned blobs

### LINKED ISSUE #1602: Unbounded growth of disk usage for singleton-blobstore
## Issue

Takes more than a week to delete orphaned blobs

## Context

We deploy (and deploy often) a set of large apps via pipelines. This makes the blobstore grow reasonably quickly.

## Steps to Reproduce

Create lots of big apps.

## Expected result

That a daily (or otherwise) scheduled task would remove orphaned blobs from the blobstore, or that at minimum it be possible for to us to configure (via a BOSH property) a more frequent run of such a task.

## Current result

The cleanup schedule task has non-configurable frequency and runs at 1:30am daily.

And - it has been built so that it cleans up a different section of the blobstore depending on the day of the week.

This make it impossible to run a full cleanup (if needed) without waiting up to a week.

<https://github.com/cloudfoundry/cloud_controller_ng/blob/1ebab1cbb5270a3d51c0a098a37cd9e8ca0f721d/app/jobs/runtime/orphaned_blobs_cleanup.rb#L126>

## Possible Fix

I'll see if I can PR something. This current behaviour is unusual and surprising (most particularly the day of week aspect).

ae-govau:
Fixes #1602.

Contributing draft PR. Feel free to pick up and implement a better approach.

### DIFF
--- a/app/jobs/runtime/orphaned_blobs_cleanup.rb
+++ b/app/jobs/runtime/orphaned_blobs_cleanup.rb
@@ -19,7 +19,7 @@ module VCAP::CloudController
             return
           end
 
-          day_of_week = Time.now.wday
+          day_of_week = config.get(:disable_partial_blob_cleanup) ? -1 : Time.now.wday
           cleanup(day_of_week)
         end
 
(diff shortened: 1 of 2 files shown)


matt-royal:
Hi @ae-govau. It looks like your changes caused some tests to fail due to the new required config option. You can likely make them pass by updating the `config/cloud_controller.yml` file to include the new option.

ae-govau:
I'll just close it out. The bigger issue, as it turned out, in the original issue, is that the app bits cache grows unbounded.

PR #1604, opened 2020-05-06
Ignore updates of buildpack (e.g. second cf push)

kramerul:

When running a `cf push` twice with a `manifest.yml` which includes a `buildpacks` setting, the second push fails with `undefined method 'buildpacks='`

This also relates to https://github.com/cloudfoundry/cf-for-k8s/issues/88 which was closed.
The current behaviour is not consitent. The first push succeeds and the second fails. It should either fail in both cases or succeed in both cases. We would vote for the later.

  Not require for this simple change.

### DIFF
--- a/app/models/runtime/kpack_lifecycle_data_model.rb
+++ b/app/models/runtime/kpack_lifecycle_data_model.rb
@@ -38,8 +38,12 @@ module VCAP::CloudController
       []
     end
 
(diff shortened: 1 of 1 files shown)


sethboyles:
Also fails with `cf push appname -b buildpack`.  It's interesting that we accept buildpack arguments at all when they do nothing, but I agree we should at least be consistent.  Thanks for the PR!

Issue #1567, opened 2020-03-04
[Feature Request]: Status in Revisions


JenGoldstrich:
Currently the Revisions object in CAPI does not return information about whether or not it is a deployable resource or not.  I think this is a good opportunity for the API to improve the Revision experience.  I would propose that when querying for an apps revision that each Revision include a status field, below are a few potential statuses.

- Deployed -> This revision is currently live 
- Not-Deployed -> This revision is not currently live but its droplet is still alive and we can roll back to this revision
- Expired -> This revision's droplet has expired and can no longer be rolled back to

This could also be bundled with a flag on /v3/apps/:guid/revisions to query for only non-expired revisions.

sannidhi:
@JenGoldstrich We have made code changes here: https://github.com/cloudfoundry/cloud_controller_ng/pull/1611 

Is this enough to resolve the issue? What we have is the `deployable` boolean which indicates that it is staged but does not indicate if it is expired or not deployed. Seth has also made additional suggestions in the issue - feel free to give us feedback on it. 

Thanks!
Sannidhi & Jwal

JenGoldstrich:
Hey @sannidhi !!

Yes this is fantastic, I will close this issue!

Thanks!
Jenna

PR #1611, opened 2020-05-15
Add deployable property to revisions endpoints.

sethboyles:
Meant to resolve https://github.com/cloudfoundry/cloud_controller_ng/issues/1567

@ssisil @selzoc  @JenGoldstrich   Wanted to get feedback on this solution since it's a little different from the issue's suggested solution. 

Instead of a `status` property, this PR just surfaces a `deployable` property which indicates whether or not the revision can be rolled back to (i.e. the droplet is in the STAGED state).

We were hesitant to introduce `state: DEPLOYED` because that implies around whether or not the process is actually started.  Two revisions may also be `DEPLOYED` at the same time (during rolling deployments)--it doesn't necessarily indicate which revision is the "current" revision.  So we weren't sure if we should go down the path of figuring out if we needed to add states like `DEPLOYED_AND_CURRENT`, `DEPLOYED_AND_NOT_CURRENT`, `state: NOT_DEPLOYED_AND_CURRENT`, etc if being able to tell if a revision could be used to create a deployment would was the main use case here.  

We can also consider adding a `current` and `deployed` flag.

TODO:
- [x] documentation

### DIFF
--- a/app/fetchers/app_revisions_fetcher.rb
+++ b/app/fetchers/app_revisions_fetcher.rb
@@ -1,10 +1,14 @@
 module VCAP::CloudController
   class AppRevisionsFetcher
     def self.fetch(app, message)
-      dataset = RevisionModel.where(app_guid: app.guid)
+      dataset = RevisionModel.where(Sequel[:revisions][:app_guid] => app.guid)
 
       if message.requested?(:versions)
-        dataset = dataset.where(app_guid: app.guid, version: message.versions)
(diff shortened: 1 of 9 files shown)


sethboyles:
Hey @bm5w, sorry for the inconvenience--I cannot merge this unless you sign the CLA.  Would you like to sign it?  If you don't want to, we can re-do this work in a new PR.

bm5w:
@sethboyles I filled out the CLA. Let me know if I need to do something else.

PR #1622, opened 2020-05-22
Remove experimental tag from label selectors

Gerg:
I don't remember why they weren't GA'd with the rest of metadata. Is there any remaining work blocking GAing label selectors?


selzoc:
@Gerg this is now out of date due to a doc styling refactor

PR #1623, opened 2020-05-22
Remove experimental tag from includes

Gerg:
These parameters have been around for a while and are used by the CLI.

Slack discussion: https://cloudfoundry.slack.com/archives/CCRLX88DD/p1589818952216600?thread_ts=1589356454.154100&cid=CCRLX88DD


selzoc:
@Gerg this is now out of date due to a doc styling refactor

PR #1692, opened 2020-06-15
V3 update maintenance info

pivotal-marcela-campo:
https://www.pivotaltracker.com/story/show/171702179

### DIFF
--- a/app/actions/service_instance_update_managed.rb
+++ b/app/actions/service_instance_update_managed.rb
@@ -46,8 +46,10 @@ module VCAP::CloudController
       parameters_changed = message.requested?(:parameters)
       service_plan_changed = message.service_plan_guid &&
         message.service_plan_guid != service_instance.service_plan.guid
+      maintenance_info_changed = message.maintenance_info_version &&
+        message.maintenance_info_version != service_instance.maintenance_info&.symbolize_keys[:version]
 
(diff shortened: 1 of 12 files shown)


app/actions/service_instance_update_managed.rb
```diff
@@ -129,5 +132,24 @@ def raise_if_bind_inconsistency!(service_instance, message)
...
+
+    def raise_if_invalid_maintenance_info_change!(service_instance, message)
+      return unless message.maintenance_info_version
+
+      raise_if_unsupported_by_current_plan!(service_instance, message)
+      raise_if_version_mismatch!(service_instance, message)
+    end
+
+    def raise_if_unsupported_by_current_plan!(service_instance, message)
+      return if service_instance.service_plan.maintenance_info
```
blgm, line 144: This checks MI against the current plan, whereas it's more correct to check against the plan you're moving to. But this story is just about updating the MI by itself, so this is ok for now.

pivotal-marcela-campo: hmmm.. anyway, apparently you cannot change plan and maintenance_info at the same time as per [#172861407](https://www.pivotaltracker.com/story/show/172861407). Will check what v2 does anyway.

PR #1697, opened 2020-06-18
Validate when updating maintenance_info alongside other fields at the

pivotal-marcela-campo:
same time.

[#172861407](https://www.pivotaltracker.com/story/show/172861407)

### DIFF
--- a/app/jobs/v3/update_service_instance_job.rb
+++ b/app/jobs/v3/update_service_instance_job.rb
@@ -27,13 +27,15 @@ module VCAP::CloudController
 
           client = VCAP::Services::ServiceClientProvider.provide({ instance: service_instance })
 
+          maintenance_info = updated_maintenance_info(message, service_instance, service_plan)
+
           broker_response, err = client.update(
             service_instance,
             service_plan,
             accepts_incomplete: false,
             arbitrary_parameters: message.parameters || {},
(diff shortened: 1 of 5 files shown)


app/jobs/v3/update_service_instance_job.rb
```diff
@@ -41,7 +41,7 @@ def perform
           updates = message.updates.tap do |u|
             u[:service_plan_guid] = service_plan.guid
             u[:dashboard_url] = broker_response[:dashboard_url] if broker_response.key?(:dashboard_url)
-            u[:maintenance_info] = message.maintenance_info if message.maintenance_info
+            u[:maintenance_info] = message.maintenance_info || service_plan.maintenance_info
```
FelisiaM, line 44: I think there is a risk that the maintenance info will be changed even when the user is changing just parameters?
Am I right in thinking that the service plan ill be set even when the service plan is not being updated?

Scenario I am thinking:
SI is created with MI version 1.0.0
Plan is updated to MI version 1.0.0 
User requests parameter changes only
As a side effect the MI is also updated - which should not happen

pivotal-marcela-campo: Good catch @FelisiaM!!! Will change that. Thanks!

PR #1708, opened 2020-06-29
v3(services): delete user-provided SI

kirederik:
[#171726153](https://www.pivotaltracker.com/story/show/171726153)

### DIFF
--- a/app/actions/service_instance_delete_user_provided.rb
+++ b/app/actions/service_instance_delete_user_provided.rb
@@ -1,0 +1,27 @@
+module VCAP::CloudController
+  class ServiceInstanceDeleteUserProvided
+    class AssociationNotEmptyError < StandardError; end
+
+    def initialize(event_repo)
+      @service_event_repository = event_repo
+    end
+
+    def delete(service_instance)
+      association_not_empty! if service_instance.has_bindings? || service_instance.has_keys?
+
+      service_instance.db.transaction do
+        service_instance.lock!
+        service_instance.destroy
+        service_event_repository.record_user_provided_service_instance_event(:delete, service_instance, {})
+      end
+    end
+
+    private
+
+    def association_not_empty!
+      raise AssociationNotEmptyError.new
+    end
+
+    attr_reader :service_event_repository
+  end
+end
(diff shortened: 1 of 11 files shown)


app/controllers/v3/service_instances_controller.rb
```diff
@@ -102,6 +103,24 @@ def update
...
+    case service_instance
+    when UserProvidedServiceInstance
+      service_event_repository = VCAP::CloudController::Repositories::ServiceEventRepository::WithUserActor.new(user_audit_info)
+      ServiceInstanceDeleteUserProvided.new(service_event_repository).delete(service_instance)
+      head :no_content
+    else
+      head :not_implemented
+    end
+  rescue ServiceInstanceDeleteUserProvided::AssociationNotEmptyError
+    raise associations_not_empty!
```
butzopower, line 121:
```suggestion
    associations_not_empty!
```

app/actions/service_instance_delete_user_provided.rb
```diff
@@ -0,0 +1,29 @@
...
+    def delete(service_instance)
+      association_not_empty! if service_instance.has_bindings? || service_instance.has_keys?
+
+      service_instance.db.transaction do
+        service_instance.lock!
+        service_instance.destroy
+        service_event_repository.record_user_provided_service_instance_event(:delete, service_instance, {})
+      end
+
+      nil
```
blgm, line 18: Do you think we should be explicitly returning `nil` as a general style?

app/actions/service_instance_delete_user_provided.rb
```diff
@@ -0,0 +1,29 @@
...
+    def initialize(event_repo)
+      @service_event_repository = event_repo
+    end
+
+    def delete(service_instance)
+      association_not_empty! if service_instance.has_bindings? || service_instance.has_keys?
+
+      service_instance.db.transaction do
+        service_instance.lock!
+        service_instance.destroy
```
blgm, line 14: I think a `delete()` would be safer, but it's not worth changing.

app/actions/service_instance_delete_user_provided.rb
kirederik: Probably not -- will remove that

app/actions/service_instance_delete_user_provided.rb
kirederik: `destroy` runs any hooks (like `before_destroy` and `after_destroy`) that may have been defined in the model -- it also performs validations and etc. `delete` runs the SQL DELETE and that's it. In general (looking through other delete actions) we use `destroy` instead of `delete`

PR #1710, opened 2020-06-30
v3(services): cannot delete shared SIs

kirederik:
or when there are route bindings present

[#171726582](https://www.pivotaltracker.com/story/show/171726582)


app/models/services/service_instance.rb
```diff
@@ -217,6 +217,10 @@ def has_keys?
       !service_keys.empty?
     end
 
+    def has_routes?
+      !routes.empty?
```
butzopower, line 221:
```suggestion
      routes.present?
```

Alternative way provided by ActiveSupport, FYI.

spec/unit/actions/service_instance_delete_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+        expect {
+          service_instance.reload
+        }.to raise_error(Sequel::Error, 'Record not found')
+        expect(VCAP::CloudController::ServiceInstanceLabelModel.where(service_instance: service_instance)).to be_empty
+        expect(VCAP::CloudController::ServiceInstanceAnnotationModel.where(service_instance: service_instance)).to be_empty
+      end
+
+      it 'fails to delete managed service instances' do
+        instance = VCAP::CloudController::ManagedServiceInstance.make
+        expect { subject.delete(instance) }.to raise_error(ServiceInstanceDelete::NotImplementedError)
```
butzopower, line 48: Should this also check the record still exists and wasn't accidentally deleted?  Guess that's hard to make fail 🤔, though maybe it would have failed before switching on the type?

spec/request/service_instances_spec.rb
```diff
@@ -2304,6 +2332,29 @@ def expect_metadata(instance, annotations: [], labels: [])
...
+  end
+
+  def enable_sharing!
+    feature_flag = VCAP::CloudController::FeatureFlag.where(name: 'service_instance_sharing').first
+    if feature_flag.nil?
+      VCAP::CloudController::FeatureFlag.make(name: 'service_instance_sharing', enabled: true)
+    else
+      feature_flag.enabled = true
+      feature_flag.save
+    end
```
butzopower, line 2355:
```suggestion
    VCAP::CloudController::FeatureFlag.
      find_or_create(name: 'service_instance_sharing') { |ff| ff.enabled = true }.
      update(enabled: true)
```

This was a gratuitous reason to stretch the Rubydoc muscle and see what I can remember about how the language works.  Feel free to ignore this one.

kirederik: No that's great :D

spec/unit/actions/service_instance_delete_spec.rb
kirederik: Yeah, will do that. The intention was for this to be really short-lived, since the very next change will remove this code but we never know 1) how long that will take and 2) when CAPI might be released.

PR #1711, opened 2020-06-30
Remove vendor directory

Gerg:
- Relocate errors to their own top-level directory
- This is legacy from when https://github.com/cloudfoundry-attic/errors
was a submodule of CC
- Vendor directory hides where error files live, especially since some
editors don't index it


tcdowney:
> Vendor directory hides where error files live, especially since some
editors don't index it

This explains why I would have trouble finding them sometimes. Wow! 😳

/lgtm

PR #1717, opened 2020-07-02
V3 update async si 171670119

pivotal-marcela-campo:
update MSI when broker response is async.

extracted shared_examples for create/update to take care of last operation polling testing

error scenario when last_operation is 'failed' but success http error code raises a new API error instead of generic one. That does not provide much value but it matches what was done for create.

### DIFF
--- a/app/jobs/v3/update_service_instance_job.rb
+++ b/app/jobs/v3/update_service_instance_job.rb
@@ -41,6 +41,7 @@ module VCAP::CloudController
 
           if service_instance.last_operation.state == 'succeeded'
             update_service_instance(broker_response, maintenance_info)
+            logger.info("Service instance update complete #{service_instance_guid}")
             finish
           end
         rescue => err
@@ -57,8 +58,6 @@ module VCAP::CloudController
           logger.info("Service instance update failed: #{service_instance.last_operation.description}")
           operation_failed!(service_instance.last_operation.description)
         end
-
-        logger.info("Service instance update complete #{service_instance_guid}")
       end
 
       def handle_timeout
@@ -104,7 +103,7 @@ module VCAP::CloudController
           maintenance_info: maintenance_info,
           name: message.requested?(:name) ? message.name : service_instance.name,
         )
-        raise err if err # TODO: create rewrites this errors with an api error keeping the message
+        raise err if err
 
         service_instance.save_with_new_operation({}, broker_response[:last_operation])
       end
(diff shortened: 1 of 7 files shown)


app/jobs/v3/update_service_instance_job.rb
```diff
@@ -62,6 +52,22 @@ def perform
           })
           raise err
         end
+
+        if service_instance.last_operation.state == 'failed'
+          logger.info("Service instance update failed: #{service_instance.last_operation.description}")
+          operation_failed!(service_instance.last_operation.description)
+        end
+
+        logger.info("Service instance update complete #{service_instance_guid}")
```
blgm, line 61: Is the update really complete, or is it just this job?

app/jobs/v3/update_service_instance_job.rb
```diff
@@ -86,7 +92,22 @@ def display_name
...
+        @broker_response, err = client.update(
+          service_instance,
+          service_plan,
+          accepts_incomplete: true,
+          arbitrary_parameters: message.parameters || {},
+          previous_values: previous_values,
+          maintenance_info: maintenance_info,
+          name: message.requested?(:name) ? message.name : service_instance.name,
+        )
+        raise err if err # TODO: create rewrites this errors with an api error keeping the message
```
blgm, line 107: When should this TODO be done?

errors/v2.yml
```diff
@@ -348,6 +348,11 @@
   http_code: 400
   message: "The service broker reported an error during provisioning: %s"
 
+60031:
```
kirederik, line 351: I think we've been trying to not add more errors in the v2.yml file? I think the proposed errors refactor are introducing some "generic" errors that can be reused. Maybe we can use the "UnableToPerform" error instead of this one? See https://github.com/cloudfoundry/cloud_controller_ng/blob/b9d9f1a8adb01a0165fc2279e3ff10c8af55b349/errors/v2.yml#L71

errors/v2.yml
```diff
@@ -348,6 +348,11 @@
...
+  name: ServiceInstanceUpdateFailed
+  http_code: 400
+  message: "The service broker reported an error during update: %s"
+
```
kirederik, line 355: I think we've been trying to not add more errors in the v2.yml file? I think the proposed errors refactor are introducing some "generic" errors that can be reused. Maybe we can use the "UnableToPerform" error instead of this one? See https://github.com/cloudfoundry/cloud_controller_ng/blob/b9d9f1a8adb01a0165fc2279e3ff10c8af55b349/errors/v2.yml#L71-L74

app/jobs/v3/update_service_instance_job.rb
pivotal-marcela-campo: thanks for spotting this.

errors/v2.yml
pivotal-marcela-campo: changing it

app/jobs/v3/update_service_instance_job.rb
pivotal-marcela-campo: removed

PR #1729, opened 2020-07-07
v3(services): deleting msi synchronously

kirederik:
[finishes #171726168](https://www.pivotaltracker.com/story/show/171726168)

### DIFF
--- a/app/actions/v3/service_instance_delete.rb
+++ b/app/actions/v3/service_instance_delete.rb
@@ -1,9 +1,11 @@
+require 'jobs/v3/delete_service_instance_job'
+
 module VCAP::CloudController
   module V3
     class ServiceInstanceDelete
       class AssociationNotEmptyError < StandardError; end
+
       class InstanceSharedError < StandardError; end
-      class NotImplementedError < StandardError; end
 
       def initialize(event_repo)
         @service_event_repository = event_repo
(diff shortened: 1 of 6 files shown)


app/jobs/v3/delete_service_instance_job.rb
```diff
@@ -0,0 +1,93 @@
...
+      private
+
+      attr_reader :service_instance_guid
+
+      def service_instance
+        ManagedServiceInstance.first(guid: @service_instance_guid)
+      end
+
+      def execute_request(client)
+        broker_response = client.send(
```
butzopower, line 54:
```suggestion
        broker_response = client.public_send(
```
Haven't had a look at how we use `send` elsewhere, but `public_send` makes sure you don't accidentally call private methods.

app/actions/v3/service_instance_delete.rb
```diff
@@ -14,20 +16,38 @@ def delete(service_instance)
...
+      def synchronous_destroy(service_instance, lock)
+        lock.unlock_and_destroy!
+        service_event_repository.record_user_provided_service_instance_event(:delete, service_instance, {})
+        nil
+      end
+
+      def asynchronous_destroy(service_instance)
+        delete_job = V3::DeleteServiceInstanceJob.new(
+          service_instance.guid,
+          :deprovision,
```
butzopower, line 43: It feels a bit like we are exposing the details of the Job by providing the method we are going to send to the broker client within the action.  It also feels a bit weird since the method that will be called on the client is being put into the database through the job queueing.

I think you hinted at a potential refactor across all these jobs, so it makes sense to keep this for now and maybe after the refactoring there might be an obvious way to expose this detail outside the job?

kirederik: Yeah, this is pretty much a work in progress at this point and not quite there yet. I'll keep your point in mind when consolidating the jobs 🤔

PR #1732, opened 2020-07-08
Add ADR 9 to supercede ADR 6

matt-royal:
This came out of https://www.pivotaltracker.com/story/show/173630293 after we decided that ADR 6 should not be implemented as originally written.

@jspawar and @njbennett also have context on this discussion

### DIFF
--- a/decisions/0009-reorganize-code-for-interacting-with-kubernetes.md
+++ b/decisions/0009-reorganize-code-for-interacting-with-kubernetes.md
@@ -1,0 +1,98 @@
+# 9. Reorganize code for interacting with kubernetes
+
+Date: 2020-07-08
+
+## Status
+
+Proposed
+
+Supercedes [6. Refactoring Interactions with Kubernetes Resources](0006-refactoring-interactions-with-kubernetes-resources.md)
+
+## Context
+
+A previous ADR (#6) proposed a new structure for our interactions with
+Kubernetes resources and the k8s API. When we attempted to implement the
+decisions made there we discovered some problems with the proposal that made it
+difficult to implement. Below are the two goals of ADR 6 and the problems they
+present.
+
+### Translate an object from the CC database
+The initial proposal assumed that we could directly translate from Cloud
+Controller models into k8s resources. While this is true for creating new
+resources, it doesn't appear to be true when we want to update an existing
+resource.
+
+[This line](https://github.com/cloudfoundry/cloud_controller_ng/blob/31a6e5c13b953cf47d443f2a7ce31aa43df4ccfa/lib/cloud_controller/kpack/stager.rb#L18)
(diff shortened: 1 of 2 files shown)


decisions/0009-reorganize-code-for-interacting-with-kubernetes.md
```diff
@@ -0,0 +1,98 @@
...
+     API and handle error translation
+   - Current examples of classes that do this: 
+     - [`Kpack::Stager`](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/cloud_controller/kpack/stager.rb)
+     - [`VCAP::CloudController::KpackBuildpackListFetcher`](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/fetchers/kpack_buildpack_list_fetcher.rb)
+     - [`Kubernetes::RouteResourceManager`](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/app/fetchers/kpack_buildpack_list_fetcher.rb) (formerly `RouteCrdClient`)
+
+3. Move all k8s-specific classes under a single namespace (e.g. `Kubernetes` or
+   `VCAP::CloudController::K8s`)
+4. Move all classes that are specific to bosh/VM-based deployments of CF into
+   their own namespace as well (e.g. `Vm`, `Bosh`, `Classic`, or `Legacy`)
```
cwlbraa, line 91: Point 4 here is pretty broad. I'm in favor of it aspirationally, but I don't think it can be done all in one change. 

Some of these "legacy" classes live under `VCAP::CloudController::Diego`, but a lot of the code in that module is actually still in use in cf-for-k8s despite the absence of Diego. I think all of the top level `::Diego` code is BOSH only. Attempting to untangle all that when we don't have context or eyes on it currently might be risky.

matt-royal: Thank you for the context. I definitely think of this as a nice-to-have and less important than the other items here. That said, untangling what "Diego" code is actually generic code would make the codebase clearer, so may be worthwhile.

matt-royal:
I reached out to our CLA manager to resolve the CLA issue. I'm not part of the VMware org, despite being an employee.

sweinstein22:
Hi @matt-royal , @JenGoldstrich and I are working on moving our default branch from master to main, so we are going to close this issue out. If you could please reopen it against the new main branch we'd appreciate it.

Thanks,
Sarah and Jenna

jspawar:
Closing because consensus was iirc consensus was never reached and we never implemented this proposal fully (also we later moved from writing proposals as RFCs instead of ADRs so this should be closed out anyway I guess). We will probably not get around to doing this refactor either, so no need to keep around.

PR #1749, opened 2020-07-21
v3(services): fix job state when deleting orphans

kirederik:
[#171727934](https://www.pivotaltracker.com/story/show/171727934)

### DIFF
--- a/app/jobs/pollable_job_wrapper.rb
+++ b/app/jobs/pollable_job_wrapper.rb
@@ -13,9 +13,14 @@ module VCAP::CloudController
       # use custom hook as Job does not have the guid field populated during the normal `enqueue` hook
       def after_enqueue(job)
         if existing_guid && (existing = PollableJobModel.find(guid: existing_guid))
+          state = if @handler.respond_to?(:pollable_job_state)
+                    @handler.pollable_job_state
+                  else
+                    PollableJobModel::POLLING_STATE
+                  end
           existing.update(
             delayed_job_guid: job.guid,
-            state: PollableJobModel::POLLING_STATE,
+            state: state,
             operation: @handler.display_name,
             resource_guid: @handler.resource_guid,
             resource_type: @handler.resource_type
--- a/app/jobs/pollable_job_wrapper.rb
+++ b/app/jobs/pollable_job_wrapper.rb
@@ -13,11 +13,7 @@ module VCAP::CloudController
       # use custom hook as Job does not have the guid field populated during the normal `enqueue` hook
       def after_enqueue(job)
         if existing_guid && (existing = PollableJobModel.find(guid: existing_guid))
-          state = if @handler.respond_to?(:pollable_job_state)
-                    @handler.pollable_job_state
-                  else
-                    PollableJobModel::POLLING_STATE
-                  end
+          state = @handler.try(:pollable_job_state) || PollableJobModel::POLLING_STATE
           existing.update(
             delayed_job_guid: job.guid,
             state: state,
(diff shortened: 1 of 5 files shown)


spec/unit/jobs/pollable_job_wrapper_spec.rb
```diff
@@ -59,6 +59,20 @@ class BigException < StandardError
           expect(job_record.resource_type).to eq('droplet')
           expect(job_record.cf_api_error).to be_nil
         end
+
+        context 'when the job defines its state' do
+          before do
+            job.define_singleton_method(:pollable_job_state) do
+              'ABRACADABRA'
+            end
```
butzopower, line 67: I changed this from `job.class.define_method` to `job.define_singleton_method` as jobs of this class are actually used in other tests in the file and defining the method can break those tests if they run in a different order.  `define_singleton_method` will define the method only for that specific instance (ruby is nuts).

spec/unit/jobs/v3/delete_service_instance_job_spec.rb
```diff
@@ -171,17 +192,6 @@ module V3
...
-      describe '#pollable_job_state' do
-        it 'returns POLLING' do
-          expect(subject.pollable_job_state).to eq(PollableJobModel::POLLING_STATE)
-        end
-
-        it 'returns PROCESSING when the request has failed' do
-          subject.instance_variable_set(:@request_failed, true)
-          expect(subject.pollable_job_state).to eq(PollableJobModel::PROCESSING_STATE)
-        end
-      end
```
butzopower: I switched this to be a bit more behavioral / triggered off of how perform works.

spec/unit/jobs/v3/delete_service_instance_job_spec.rb
```diff
@@ -46,7 +62,12 @@ module V3
 
             expect(subject.instance_variable_get(:@attempts)).to eq(1)
             expect(subject.instance_variable_get(:@first_time)).to eq(true)
-            expect(subject.instance_variable_get(:@request_failed)).to eq(true)
```
butzopower: I think I put in the original instance variable gets in this test, and it actually feels a bit dirty to reach into the object and break encapsulation.  I might have a think on how to fix the others as well (for instance, maybe attempts doesn't matter because we show that it fails after 3 tries).

app/jobs/pollable_job_wrapper.rb
```diff
@@ -13,11 +13,7 @@ def initialize(handler, existing_guid: nil)
       # use custom hook as Job does not have the guid field populated during the normal `enqueue` hook
       def after_enqueue(job)
         if existing_guid && (existing = PollableJobModel.find(guid: existing_guid))
-          state = if @handler.respond_to?(:pollable_job_state)
-                    @handler.pollable_job_state
-                  else
-                    PollableJobModel::POLLING_STATE
-                  end
+          state = @handler.try(:pollable_job_state) || PollableJobModel::POLLING_STATE
```
butzopower: Rails actually extended `Object` with `try` since this is pretty common to want to do.

spec/unit/jobs/pollable_job_wrapper_spec.rb
```diff
@@ -62,7 +62,7 @@ class BigException < StandardError
 
         context 'when the job defines its state' do
           before do
-            job.class.define_method(:pollable_job_state) do
+            job.define_singleton_method(:pollable_job_state) do
```
butzopower: I changed this from `job.class.define_method` to `job.define_singleton_method` as jobs of this class are actually used in other tests in the file and defining the method can break those tests if they run in a different order. `define_singleton_method` will define the method only for that specific instance (ruby is nuts).

PR #1750, opened 2020-07-21
v3(services): Return credential binding info for both key and app bindings

butzopower:
[#173148626](https://www.pivotaltracker.com/story/show/173148626)


app/models/services/service_credential_binding.rb
```diff
@@ -0,0 +1,44 @@
...
+    module ServiceCredentialBindingViews
+      module Types
+        SERVICE_KEY = 'key'.freeze
+        SERVICE_BINDING = 'app'.freeze
+      end
+
+      SERVICE_KEY = ServiceKey.select(
+        :guid,
+        Sequel.as(Types::SERVICE_KEY, :type),
+        Sequel.as(Sequel.cast(:service_instance_id, String), :service_instance_id)
```
butzopower, line 12: This bit was pretty weird, and only cropped up when I ran this against MySQL.  MySQL will actually turn the field name into "CAST service_instance_id AS char" or something similar, so the `as` remaps it back to being `service_instance_id`

PR #1754, opened 2020-07-23
v3(services): Add request test to ensure service credential binding sharing permissions

butzopower:
[#173473714](https://www.pivotaltracker.com/story/show/173473714)


spec/request/service_credential_bindings_spec.rb
```diff
@@ -63,7 +63,8 @@
   end
 
   describe 'GET /v3/service_credential_bindings/:app_guid' do
```
kirederik, line 65: This commit hasn't changed this, but should it be `binding_guid`?

butzopower: `:app_guid` is to distinguish between the other way this endpoint can be used e.g. `:key_guid`.

PR #1758, opened 2020-07-28
v3(services): shared_spaces can include fields

kirederik:
Valid fields are:
* space: name, guid, relationships.organization
* space.organization: name, guid

[#171954418](https://www.pivotaltracker.com/story/show/171954418)

### DIFF
--- a/app/decorators/field_service_instance_organization_decorator.rb
+++ b/app/decorators/field_service_instance_organization_decorator.rb
@@ -12,9 +12,9 @@ module VCAP::CloudController
       @fields = fields[:'space.organization'].to_set.intersection(self.class.allowed)
     end
 
-    def decorate(hash, service_instances)
+    def decorate(hash, resources)
       hash[:included] ||= {}
-      spaces = service_instances.map(&:space).uniq
+      spaces = resources.map { |r| r.try(:space) || r }.uniq
       orgs = spaces.map(&:organization).uniq
 
       hash[:included][:organizations] = orgs.sort_by(&:created_at).map do |org|
--- a/docs/v3/source/includes/resources/service_instances/_list_shared_spaces.md.erb
+++ b/docs/v3/source/includes/resources/service_instances/_list_shared_spaces.md.erb
@@ -25,6 +25,19 @@ This endpoint lists the spaces that the service instance has been shared to.
 #### Definition
 `GET /v3/service_instances/:guid/relationships/shared_spaces`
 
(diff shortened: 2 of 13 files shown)


app/decorators/field_service_instance_organization_decorator.rb
```diff
@@ -12,9 +12,9 @@ def initialize(fields)
       @fields = fields[:'space.organization'].to_set.intersection(self.class.allowed)
     end
 
-    def decorate(hash, service_instances)
+    def decorate(hash, resources)
       hash[:included] ||= {}
-      spaces = service_instances.map(&:space).uniq
+      spaces = resources.map { |r| r.try(:space) || r }.uniq
```
pivotal-marcela-campo, line 17: I am not a fan of this in either of the two decorators. It looks to me more like accidental duplication and would have probably written them again as we might want to include different fields for space vs shared space/org vs shared org in the future. I am ok keeping this anyway, but would split them out as soon as they start deviating from each other.

docs/v3/source/includes/resources/service_instances/_list_shared_spaces.md.erb
```diff
@@ -25,6 +25,19 @@ This endpoint lists the spaces that the service instance has been shared to.
 #### Definition
```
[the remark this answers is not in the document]
kirederik: Will add that, thanks!

app/decorators/field_service_instance_organization_decorator.rb
kirederik: Yeah, it's a bit weird, I agree. We can probably write something better that does not involve duplicating this code entirely.

I don't see the divergence of fields as a problem though. The message we create in the controller should determine what each method can include, returning earlier if the request includes non-allowed fields (that is, if the message is invalid).

You can see that the `relationships_shared_spaces`, `index` and `show` all use this same decorator, but not necessarily allow the same fields to be included.

pivotal-marcela-campo: hmm, you are right. I always forget it is restricted in the message anyway. And also agree we can write something better than duplicating the whole code as well. The only thing I can think of at the moment is inheritance and having a get_spaces method overriden by subclasses. It feels overkilling though.

kirederik: I'll merge this for now and we can always get back 👀

PR #1761, opened 2020-07-29
v3(services): Return list of service credential bindings

butzopower:
[#173171550](https://www.pivotaltracker.com/story/show/173171550)

### DIFF
--- a/app/controllers/v3/service_credential_bindings_controller.rb
+++ b/app/controllers/v3/service_credential_bindings_controller.rb
@@ -1,20 +1,43 @@
 require 'fetchers/service_credential_binding_fetcher'
+require 'fetchers/service_credential_binding_list_fetcher'
+require 'presenters/v3/service_credential_binding_presenter'
 
 class ServiceCredentialBindingsController < ApplicationController
-  before_action :ensure_service_credential_binding_exists!
-  before_action :ensure_user_has_access!
+  def index
+    results = list_fetcher.fetch(space_guids: space_guids)
+
+    presenter = Presenters::V3::PaginatedListPresenter.new(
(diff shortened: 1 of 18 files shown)


config/routes.rb
```diff
@@ -185,7 +185,9 @@
   delete '/service_bindings/:guid', to: 'service_bindings#destroy'
 
   # service_credential_bindings
-  get '/service_credential_bindings/:guid', to: 'service_credential_bindings#show'
+  resources :service_credential_bindings,
```
FelisiaM, line 188: Would be nice to show this in the VAT call tomorrow just to get an idea why it is not done this way for the other APIs

spec/unit/presenters/v3/service_credential_binding_presenter_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'presenters/v3/service_credential_binding_presenter'
+
+module VCAP
+  module CloudController
+    describe Presenters::V3::ServiceCredentialBindingPresenter do
```
FelisiaM, line 5: This is failing to run. I think you might need `RSpec.describe `

app/controllers/v3/service_credential_bindings_controller.rb
```diff
@@ -1,20 +1,43 @@
...
+    if permission_queryer.can_read_globally?
+      :all
+    else
+      permission_queryer.readable_space_guids
+    end
+  end
+
+  def pagination_options
+    query_params_with_order_by = query_params.reverse_merge(order_by: :created_at)
+    MetadataListMessage.from_params(query_params_with_order_by, []).pagination_options
```
FelisiaM, line 36: This is redundant at this point since we are not returning any metadata yet.

PR #1769, opened 2020-08-04
v3(services): GET credential binding object response

kirederik:


docs/v3/source/includes/api_resources/_service_bindings.erb
```diff
@@ -2,15 +2,20 @@
 {
   "guid": "dde5ad2a-d8f4-44dc-a56f-0452d744f1c3",
   "type": "app",
```
FelisiaM, line 4: This is the file that refers to the existing implementation of v3 service bindings. The one you want to edit is `docs/v3/source/includes/api_resources/_service_credential_bindings.erb`

PR #1775, opened 2020-08-06
introduce Kubernetes::UpdateReapplyClient

cwlbraa:
- consistently manages get-update-retry-on-409 loops
- adds retryable as a gem
- adds retries to image and custombuilder updates in the kpack stager
- replaces handrolled retry implementation in RouteResourceManager
- should fix some samus flakes like https://capi.ci.cf-app.com/teams/main/pipelines/capi/jobs/samus-cf-for-k8s/builds/1006


cwlbraa:
merged out-of-band!

PR #1780, opened 2020-08-11
v3(services): only redact credentials when updating upsi when actually provided

butzopower:
[#174144371](https://www.pivotaltracker.com/story/show/174144371)

### DIFF
--- a/spec/unit/messages/service_instance_update_user_provided_message_spec.rb
+++ b/spec/unit/messages/service_instance_update_user_provided_message_spec.rb
@@ -1,4 +1,4 @@
-require 'spec_helper'
+require 'lightweight_spec_helper'
 require 'messages/service_instance_update_user_provided_message'
 
 module VCAP::CloudController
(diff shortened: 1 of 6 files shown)


spec/unit/messages/service_instance_update_user_provided_message_spec.rb
```diff
@@ -99,9 +99,21 @@ module VCAP::CloudController
     end
 
     describe '.audit_hash' do
-      it 'produces a redacted audit hash' do
-        expected = body.dup.tap { |b| b[:credentials] = '[PRIVATE DATA HIDDEN]' }
-        expect(message.audit_hash).to match(expected.with_indifferent_access)
+      context 'when credentials are passed in' do
+        it 'produces a redacted audit hash' do
+          expected = body.dup.tap { |b| b[:credentials] = '[PRIVATE DATA HIDDEN]' }
```
pivotal-marcela-campo, line 104: I tend to avoid doing this kind of things in tests as It is almost as duplicating the real code. I prefer to just build the hash manually.
Anyways, this is very simple so happy with it.

spec/unit/messages/service_instance_update_user_provided_message_spec.rb
```diff
@@ -99,9 +99,21 @@ module VCAP::CloudController
...
+          expect(message.audit_hash).to match(expected.with_indifferent_access)
+        end
+      end
+
+      context 'when no credentials are passed in' do
+        let(:body_without_credentials) { body.without(:credentials) }
+        let(:message) { described_class.new(body_without_credentials) }
+
+        it 'does not add a redacted credentials key' do
+          expected = body.without(:credentials)
```
pivotal-marcela-campo, line 114: Didn't know `without` existed!

spec/unit/messages/service_instance_update_user_provided_message_spec.rb
butzopower: Ya, not the biggest fan of this either, but I was just keeping the existing test mostly unchanged and just wrapping it in a context to mirror the case where no credentials are passed.

spec/unit/messages/service_instance_update_user_provided_message_spec.rb
butzopower: It's in `ActiveSupport` so it's available in a Rails app but not out of the box ruby.

PR #1787, opened 2020-08-13
Add audit events for Apps#show_env and Revisions#show_environment_variables

tlwr:

Add two audit events:
- `audit.app.show-env`
- `audit.app.revision.show-environment-variables`

Environment variables frequently contain sensitive information, and so we should audit access, even when viewing.

As a security analyst,
I want to audit when developers access environment variables,
so that I can hold them to account

As an operator,
I want to record access to environment variables,
so my users can hold me to account

### DIFF
--- a/app/controllers/v3/apps_controller.rb
+++ b/app/controllers/v3/apps_controller.rb
@@ -261,6 +261,8 @@ class AppsV3Controller < ApplicationController
 
     FeatureFlag.raise_unless_enabled!(:space_developer_env_var_visibility)
 
+    Repositories::AppEventRepository.new.record_app_show_environment_variables(app, user_audit_info)
+
     render status: :ok, json: Presenters::V3::AppEnvironmentVariablesPresenter.new(app)
   end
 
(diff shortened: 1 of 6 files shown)


Gerg:
Hey @tlwr,

From a high level, I think this makes sense and will be a valuable addition. What would you think about changing the types of the audit events to something like `audit.app.environment_variables.show` and `audit.revisions.environment_variables.show`. I think this will be more in line with [existing naming conventions](https://docs.cloudfoundry.org/running/managing-cf/audit-events.html#types).

Also, I suspect you may have overlooked the `GET /v3/apps/:guid/environment_variables` endpoint, which can also be used to view environment variables.

Thanks,
Greg

tlwr:
All sensible points, thanks especially for pointing out `GET /v3/apps/:guid/environment_variables` I shall make the changes

tlwr:
I've fixed up the PR so that the following audit events are generated:

1. `audit.app.environment.show`
1. `audit.app.environment_variables.show`
1. `audit.app.revision.environment_variables.show`

where the difference between 1. and 2. is that 1. includes VCAP_SERVICES and 2. does not

PR #1789, opened 2020-08-18
v3(services): introduce service route binding endpoint

blgm:
[#174063444](https://www.pivotaltracker.com/story/show/174063444)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -189,6 +189,11 @@ Rails.application.routes.draw do
             param: :guid,
             only: [:show, :index]
 
+  # service_route_bindings
+  resources :service_route_bindings,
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -189,6 +189,11 @@
             param: :guid,
             only: [:show, :index]
 
+  # service_route_bindings
+  resources :service_route_bindings,
+    param: :guid,
```
butzopower, line 194: This is probably fine to keep since there's likely a show or update or delete in the future, but just so you know, this is the key for the resource (opposed to resources/list) endpoints which by convention is keyed off as `:id` in the params.

PR #1792, opened 2020-08-18
Service instance sharing  tests

FelisiaM:
[#174254053](https://www.pivotaltracker.com/story/show/174254053)

### DIFF
--- a/spec/request/service_instances_spec.rb
+++ b/spec/request/service_instances_spec.rb
@@ -670,183 +670,6 @@ RSpec.describe 'V3 service instances' do
     end
   end
 
-  describe 'GET /v3/service_instances/:guid/relationships/shared_spaces' do
-    let(:user_header) { headers_for(user) }
-    let(:instance) { VCAP::CloudController::ManagedServiceInstance.make(space: space) }
-    let(:other_space) { VCAP::CloudController::Space.make }
-
-    before(:each) do
-      share_service_instance(instance, other_space)
-    end
-
-    it 'returns a list of space guids where the service instance is shared to' do
-      set_current_user_as_role(role: 'space_developer', org: space.organization, space: space, user: user)
-      get "/v3/service_instances/#{instance.guid}/relationships/shared_spaces", nil, user_header
-      expect(last_response.status).to eq(200)
-
-      expected_response = {
-        data: [{ guid: other_space.guid }],
-        links: {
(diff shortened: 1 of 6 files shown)


spec/request/service_instances_spec.rb
```diff
@@ -3360,57 +3185,511 @@ def create_bindings(instance, space:, count:)
...
+        include(
+          {
+            'detail' => 'Service instance not found',
+            'title' => 'CF-ResourceNotFound'
+          })
+      )
+    end
+
+    describe 'invalid request body' do
+      context 'when request is not a valid relationship' do
```
kirederik, line 3286: maybe "when the request body is invalid"?

spec/request/service_instances_spec.rb
```diff
@@ -3360,57 +3185,511 @@ def create_bindings(instance, space:, count:)
...
+          {
+            'data' => [
+              { 'guid' => no_access_target_space.guid },
+              { 'guid' => target_space_1.guid }
+            ]
+          }
+        end
+
+        it 'responds with 422' do
+          api_call.call(space_dev_headers)
```
kirederik, line 3369: one thing it's not clear from this test is whether the service instance was shared or not with the space the user has access to. Should we add another `it` to cover that (either `it 'does not share with any of the spaces'` or `it 'shares with the other spaces'` or something like that)

spec/request/service_instances_spec.rb
```diff
@@ -3495,51 +3774,6 @@ def enable_sharing!
     let!(:service_instance2) { VCAP::CloudController::ManagedServiceInstance.make(space: space, name: 'redis') }
     let!(:service_instance3) { VCAP::CloudController::ManagedServiceInstance.make(space: another_space, name: 'mysql') }
 
-    describe 'POST /v3/service_instances/:guid/relationships/shared_spaces' do
```
kirederik, line 3498: I guess we want to delete the entire `unrefactored` describe that this one is in?

FelisiaM: still have delete tests to finish  so will remove unrefactored and the controller unit test after that

spec/request/service_instances_spec.rb
FelisiaM: added an expectation to check that it was not shared at all

spec/request/service_instances_spec.rb
FelisiaM: I have removed all POST related tests

PR #1795, opened 2020-08-19
v3(bindings): allow the binding SI to be included

kirederik:
Get and list endpoints

[#173172605](https://www.pivotaltracker.com/story/show/173172605)

### DIFF
--- a/spec/unit/decorators/include_binding_service_instance_decorator_spec.rb
+++ b/spec/unit/decorators/include_binding_service_instance_decorator_spec.rb
@@ -1,0 +1,42 @@
+require 'spec_helper'
+require 'decorators/include_binding_service_instance_decorator'
+
+module VCAP::CloudController
+  RSpec.describe IncludeBindingServiceInstanceDecorator do
+    subject(:decorator) { described_class }
+    let(:bindings) { [ServiceBinding.make, ServiceBinding.make, ServiceBinding.make] }
+    let(:instances) {
+      bindings.map { |b| Presenters::V3::ServiceInstancePresenter.new(b.service_instance).to_hash }
+    }
+
+    it 'decorates the given hash with service instances from bindings' do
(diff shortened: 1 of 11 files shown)


spec/unit/decorators/include_binding_service_instance_decorator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    it 'does not overwrite other included fields' do
+      dict = { foo: 'bar', included: { fruits: ['tomato', 'banana'] } }
+      hash = subject.decorate(dict, bindings)
+      expect(hash[:foo]).to eq('bar')
+      expect(hash[:included][:service_instances]).to match_array(instances)
+      expect(hash[:included][:fruits]).to match_array(['tomato', 'banana'])
+    end
+
```
blgm, line 26: The implementation has code to de-duplicate service instances using `.uniq`, and I don't see a test that drives that implementation. Should we have one, or does the cost outweigh the benefit?

blgm (approved):

spec/unit/decorators/include_binding_service_instance_decorator_spec.rb
kirederik: It's indirectly (or even accidentally I'd say) tested in the request spec (there are 4 bindings originating from 2 instances). There's no harm in making it explicit and it should be simple enough to add in the decorator test.

kirederik: Interesting enough, the `.uniq` seems to be unnecessary to guarantee uniqueness in the returned result. I guess it's building a slightly better SQL query (`guid in ("a","b")` rather than `guid in ("a","b","b")`). I'm inclined to remove the `.uniq` and leave the new test... 🤔

PR #1797, opened 2020-08-19
v3(services): test order_by for services resources

blgm:
- implemented order_by name where it was not implemented already except
for service offerings where it is complex due to the DB column being
called `label` rather than `name`
- improved the pagination shared example to check links

### DIFF
--- a/app/messages/service_credential_bindings_list_message.rb
+++ b/app/messages/service_credential_bindings_list_message.rb
@@ -26,7 +26,7 @@ module VCAP::CloudController
     end
 
     def valid_order_by_values
-      super << :name
+      super + [:name]
     end
   end
 end
--- a/app/messages/service_credential_bindings_list_message.rb
+++ b/app/messages/service_credential_bindings_list_message.rb
@@ -24,5 +24,9 @@ module VCAP::CloudController
     def self.from_params(params)
       super(params, ARRAY_KEYS.map(&:to_s))
     end
+
+    def valid_order_by_values
+      super << :name
+    end
   end
 end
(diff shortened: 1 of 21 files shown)


spec/request/service_brokers_spec.rb
```diff
@@ -90,13 +90,17 @@
...
     before do
       stub_request(:get, 'http://example.org/broker-url-v2/v2/catalog').
         to_return(status: 200, body: catalog('v2-broker').to_json, headers: {})
       stub_uaa_for('v2-broker')
     end
 
-    let(:api_call) { lambda { |user_headers| get '/v3/service_brokers', nil, user_headers } }
+    it_behaves_like 'paginated response', '/v3/service_brokers' do
+      let!(:resources) { (0..1).collect { VCAP::CloudController::ServiceBroker.make } }
+    end
```
butzopower, line 103:
```suggestion
      let!(:resources) { Array.new(2) { VCAP::CloudController::ServiceBroker.make } }
    end
```

Not sure if this is what you are trying to do?

app/messages/service_credential_bindings_list_message.rb
```diff
@@ -24,5 +24,9 @@ class ServiceCredentialBindingsListMessage < ListMessage
     def self.from_params(params)
       super(params, ARRAY_KEYS.map(&:to_s))
     end
+
+    def valid_order_by_values
+      super << :name
```
butzopower, line 29:
```suggestion
      super + [:name]
```

Not sure if this will actually matter since I think we freeze most constants, but `<<` modifies and return new value while `+` will just return a new value.

butzopower (approved):
Some minor changes suggested, but functionally it's good to go!

PR #1798, opened 2020-08-20
v3(bindings): Get details of a credential binding

kirederik:
[#173529480](https://www.pivotaltracker.com/story/show/173529480)
[#173510836](https://www.pivotaltracker.com/story/show/173510836)

### DIFF
--- a/app/controllers/v3/service_credential_bindings_controller.rb
+++ b/app/controllers/v3/service_credential_bindings_controller.rb
@@ -1,6 +1,7 @@
 require 'fetchers/service_credential_binding_fetcher'
 require 'fetchers/service_credential_binding_list_fetcher'
 require 'presenters/v3/service_credential_binding_presenter'
+require 'presenters/v3/service_credential_binding_details_presenter'
 require 'messages/service_credential_bindings_list_message'
 require 'messages/service_credential_bindings_show_message'
 require 'decorators/include_binding_app_decorator'
@@ -31,6 +32,13 @@ class ServiceCredentialBindingsController < ApplicationController
     ensure_service_credential_binding_is_accessible!
 
     render status: :ok, json: serialized(message)
+  end
+
+  def details
+    ensure_service_credential_binding_is_accessible!
+    ensure_binding_details_are_accessible!
+
+    render status: :ok, json: credential_binding_details
   end
 
   private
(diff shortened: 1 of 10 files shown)


app/controllers/v3/service_credential_bindings_controller.rb
```diff
@@ -65,10 +73,18 @@ def serialized(message)
...
 
+  def credential_binding_details
+    Presenters::V3::ServiceCredentialBindingDetailsPresenter.new(service_credential_binding).to_hash
+  end
+
   def ensure_service_credential_binding_is_accessible!
     not_found! unless service_credential_binding_exists?
   end
 
+  def ensure_binding_details_are_accessible!
```
blgm, line 84: I'm generally a bit fan of using small functions to reduce cognitive load. For this function, my first thought was "what does accessible mean?", and I actually find the implementation easier to understand than the function name. But that's just my opinion and other people might feel differently.

spec/request/service_credential_bindings_spec.rb
```diff
@@ -442,6 +442,99 @@ def check_filtered_bindings(*bindings)
     end
   end
 
+  describe 'GET /v3/service_credential_bindings/:app_guid/details' do
```
blgm, line 445: I'd prefer `binding_guid` or `app_binding_guid` since it's no an App guid.

spec/unit/fetchers/service_credential_binding_list_fetcher_spec.rb
```diff
@@ -18,15 +18,37 @@ module CloudController
...
-        let!(:key_binding) { VCAP::CloudController::ServiceKey.make(service_instance: instance) }
-        let!(:app_binding) { VCAP::CloudController::ServiceBinding.make(service_instance: instance, name: Sham.name) }
+
+        let(:key_details) {
+          {
+            credentials: '{"some":"key"}'
+          }
+        }
+
+        let(:app_details) {
```
blgm, line 28: Looks like an app binding, rather than an App

spec/unit/presenters/v3/service_credential_binding_details_presenter_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+module VCAP
+  module CloudController
+    RSpec.describe Presenters::V3::ServiceCredentialBindingDetailsPresenter do
+      let(:instance) { ServiceInstance.make(guid: 'instance-guid') }
+      let(:app) { AppModel.make(guid: 'app-guid', space: instance.space) }
+      let(:json_creds) { { password: 'super secret avocado toast' }.to_json }
+      let(:credential_binding) do
+        ServiceBinding.make(
+          name: 'some-name',
```
blgm, line 12: Doesn't the blueprint fill in `name` and `guid` if they are not set? If not, then we should fix the blueprint

blgm (approved):
Since we are doing more in this area, I'm happy for this to be merged as is

PR #1803, opened 2020-08-24
Remove unused RingBuffer class

Gerg:
Call was removed in 25efe93a2cb43562e096b6694dfbb52b05a5a2c0 as part of removing varz


lib/cloud_controller.rb
```diff
@@ -8,6 +8,8 @@
 
 require 'uaa/token_coder'
 
+module VCAP; end
```
tcdowney, line 11: What's this for?

Gerg: Previously, RingBuffer was the first file loaded with the `VCAP` module defined:
- https://github.com/cloudfoundry/cloud_controller_ng/blob/df570cbf20aea26406ec9678e5e848a156a81dc6/lib/cloud_controller.rb#L11
- https://github.com/cloudfoundry/cloud_controller_ng/blob/df570cbf20aea26406ec9678e5e848a156a81dc6/lib/sinatra/vcap.rb#L1
Removing RingBuffer has the side effect of causing `http_constants.rb` to fail loading because `VCAP` is undefined.
- https://github.com/cloudfoundry/cloud_controller_ng/blob/df570cbf20aea26406ec9678e5e848a156a81dc6/lib/sinatra/vcap.rb#L2
- https://github.com/cloudfoundry/cloud_controller_ng/blob/df570cbf20aea26406ec9678e5e848a156a81dc6/lib/vcap/rest_api.rb#L1
- https://github.com/cloudfoundry/cloud_controller_ng/blob/df570cbf20aea26406ec9678e5e848a156a81dc6/lib/vcap/rest_api/http_constants.rb#L1

Gerg: So, rather than shifting module definitions around for what random file gets to be the first one to define VCAP, I pulled it up to the top level.

PR #1806, opened 2020-08-25
v3(services): route binding to managed service instance

blgm:
[#174088304](https://www.pivotaltracker.com/story/show/174088304)

### DIFF
--- a/app/jobs/v3/services/create_route_binding_job.rb
+++ b/app/jobs/v3/services/create_route_binding_job.rb
@@ -33,7 +33,8 @@ module VCAP::CloudController
 
       def perform
         precursor = RouteBinding.first(guid: @precursor_guid)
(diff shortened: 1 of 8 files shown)


app/jobs/v3/services/create_route_binding_job.rb
```diff
@@ -0,0 +1,41 @@
...
+        @precursor_guid
+      end
+
+      def resource_type
+        'service_route_binding'
+      end
+
+      def perform
+        precursor = RouteBinding.first(guid: @precursor_guid)
+        action = V3::ServiceRouteBindingCreate.new(@user_audit_info)
```
butzopower: Ideally the perform constructs the EventRepo here rather than pull it off the queue, as that can cause issues.  It probably means we need to enqueue the actual user audit info, though I'm not sure how complex that data is.

butzopower (approved):

PR #1809, opened 2020-08-26
V3 services route binding async

blgm:


app/presenters/v3/service_route_binding_presenter.rb
```diff
@@ -7,34 +7,53 @@ module V3
...
               relationships: relationships,
               links: links
             }
           end
 
           private
 
+          def binding
+            @resource
+          end
```
butzopower, line 23: Big fan of this.

app/models/services/route_binding.rb
```diff
@@ -33,8 +35,36 @@ def required_parameters
...
+        save_changes
+
+        if self.last_operation
+          self.last_operation.destroy
+        end
+
+        # it is important to create the service route binding operation with the service binding
+        # instead of doing self.service_route_binding_operation = x
+        # because mysql will deadlock when requests happen concurrently otherwise.
+        RouteBindingOperation.create(last_operation.merge(route_binding_id: self.id))
```
butzopower, line 63: I would probably rename the 2nd argument `new_operation` so that you don't have to deal with the overlap of `last_operation` and sometimes refer to `self.last_operation` when you mean the old one.

The only reason I spotted this was because it was drilled into me in my rubyist days that `self` is the default receiver, and so at first I assumed it was redundant here.

spec/unit/models/services/route_binding_spec.rb
```diff
@@ -52,5 +52,150 @@ module VCAP::CloudController
...
+
+      context 'when saving the binding operation fails' do
+        before do
+          allow(RouteBindingOperation).to receive(:create).and_raise(Sequel::DatabaseError, 'failed to create new-binding operation')
+        end
+
+        it 'should rollback the binding' do
+          expect { route_binding.save_with_new_operation({}, { state: 'will fail' }) }.to raise_error(Sequel::DatabaseError)
+          expect(RouteBinding.count).to eq(0)
+        end
```
butzopower, line 99: Not the most egregious and probably something we do in many tests in the codebase, but stubbing `create` leaks quite a bit of implementation detail (for instance, now you can't refactor to `#.new` -> `.save` if you needed to for whatever reason).

It's a bit more leaking the business / DB constraints, but I'd probably try and set the operation up with data that will actually cause an error to be raised.

butzopower (approved):
Two non-functional changes proposed, feel free to ignore them!

PR #1828, opened 2020-09-03
v3(bindings): validations for UPSI ↔ App

kirederik:
[#174035434](https://www.pivotaltracker.com/story/show/174035434)

### DIFF
--- a/app/controllers/v3/service_credential_bindings_controller.rb
+++ b/app/controllers/v3/service_credential_bindings_controller.rb
@@ -2,14 +2,15 @@ require 'fetchers/service_credential_binding_fetcher'
 require 'fetchers/service_credential_binding_list_fetcher'
 require 'presenters/v3/service_credential_binding_presenter'
 require 'presenters/v3/service_credential_binding_details_presenter'
-require 'messages/service_credential_bindings_list_message'
-require 'messages/service_credential_bindings_show_message'
+require 'messages/service_credential_binding_list_message'
+require 'messages/service_credential_binding_show_message'
+require 'messages/service_credential_binding_create_message'
 require 'decorators/include_binding_app_decorator'
 require 'decorators/include_binding_service_instance_decorator'
 
 class ServiceCredentialBindingsController < ApplicationController
   def index
-    message = ServiceCredentialBindingsListMessage.from_params(query_params)
+    message = ServiceCredentialBindingListMessage.from_params(query_params)
     invalid_param!(message.errors.full_messages) unless message.valid?
 
     results = list_fetcher.fetch(space_guids: space_guids, message: message)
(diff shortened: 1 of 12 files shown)


app/messages/service_credential_binding_create_message.rb
```diff
@@ -0,0 +1,33 @@
+module VCAP::CloudController
+  class ServiceCredentialBindingCreateMessage < MetadataBaseMessage
```
FelisiaM, line 2: I think it might be a while before we implement metadata for bindings so having this here and documented when we will not be using might be confusing. I didn't find it mentioned in any tests so it should be easy to remove it for the moment and not mention it in the docs yet.

app/messages/service_credential_binding_create_message.rb
```diff
@@ -0,0 +1,33 @@
...
+    register_allowed_keys [:type, :name, :relationships]
+    validates_with NoAdditionalKeysValidator, RelationshipValidator
+    validates :type, allow_blank: false, inclusion: {
+      in: %w(key app),
```
FelisiaM, line 6: If we don't include this here at all we won't need to think about it until key creation is implemented (after app bindings for MSI). Might make the logic on the controller easier? (not a big deal anyway)

app/controllers/v3/service_credential_bindings_controller.rb
```diff
@@ -26,14 +27,32 @@ def index
...
 
     render status: :ok, json: serialized(message)
   end
 
+  def create
+    message = ServiceCredentialBindingCreateMessage.new(hashed_params[:body])
+    unprocessable!(message.errors.full_messages) unless message.valid?
+
+    service_instance = VCAP::CloudController::ServiceInstance.first(guid: message.service_instance_guid)
+    unprocessable!("Service instance '#{message.service_instance_guid}' not found") unless !service_instance.blank? && can_read_from_space?(service_instance.space)
```
FelisiaM, line 43: Isn't this ths same as `unless service_instance && can_read_from_space?(service_instance.space)`

FelisiaM, line 43: Route bindings create return a slightly different message from the this. 

> `The service instance could not be found: #{guid}"`

Would be nice to have consistent messages from both endpoints, but I'm not sure which one is the most consistent

FelisiaM (changes requested):

app/messages/service_credential_binding_create_message.rb
kirederik: Fair point, will remove it

app/messages/service_credential_binding_create_message.rb
kirederik: The story asks for rejection of `type=key` when creating a binding from a UPSI, so I guess we need this? It would make the code simpler if `type` wasn't required at this point at all 🤔

app/controllers/v3/service_credential_bindings_controller.rb
kirederik: For this particular line, yes. It's a bit safer to call the method though (we might change how we fetch the service instance in following stories [although unlikely :P]). I will change it from `!blank?` to `present?` though

PR #1833, opened 2020-09-07
v3(bindings): UPSI can be bound to apps

kirederik:
[#174063082](https://www.pivotaltracker.com/story/show/174063082)

### DIFF
--- a/app/actions/service_credential_binding_create.rb
+++ b/app/actions/service_credential_binding_create.rb
@@ -17,8 +17,7 @@ module VCAP::CloudController
           name: name,
           app: app,
           type: 'app',
-          credentials: service_instance.credentials,
-          syslog_drain_url: service_instance.syslog_drain_url
+          credentials: {}
         }
 
         ServiceBinding.new(**binding_details).tap do |b|
@@ -31,11 +30,15 @@ module VCAP::CloudController
         end
       rescue Sequel::ValidationFailed => e
         already_bound! if e.message =~ /The app is already bound to the service/
-        raise e
+        raise UnprocessableCreate(e.full_message)
       end
 
       def bind(binding)
-        binding.save_with_new_operation({ type: 'create', state: 'succeeded' })
+        client = VCAP::Services::ServiceClientProvider.provide(instance: binding.service_instance)
+        details = client.bind(binding, arbitrary_parameters: {}, accepts_incomplete: false)
+
+        binding.save_with_new_operation({ type: 'create', state: 'succeeded' }, attributes: details[:binding])
+
         event_repository.record_create(binding, @user_audit_info, manifest_triggered: false)
       rescue => e
         binding.save_with_new_operation({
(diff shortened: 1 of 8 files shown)


app/actions/service_credential_binding_create.rb
```diff
@@ -0,0 +1,68 @@
...
+  module V3
+    class ServiceCredentialBindingCreate
+      class UnprocessableCreate < StandardError
+      end
+
+      def initialize(user_audit_info)
+        @user_audit_info = user_audit_info
+      end
+
+      def precursor(service_instance, app: nil, name: nil)
```
FelisiaM, line 11: I find this name not that descriptive of what it does. It just shows when it does it. 
It's really doing "validate and create binding" but that doesn't help. We might need to think about this later

app/actions/service_credential_binding_create.rb
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        ServiceBinding.new(**binding_details).tap do |b|
+          b.save_with_new_operation(
+            {
+              type: 'create',
+              state: 'in progress',
+            }
+          )
+        end
```
FelisiaM, line 31: I am a bit confused as to why this is doing the binding and also setting "create in progress" when it is already storing the details of the binding (cred and sys drain log). 

Would it be better to not store the cred and sys log drain here, but use the client.bind as part of the `bind` (where client is` VCAP::Services::ServiceBrokers::UserProvided::Client`)? 
That way `bind` is doing the actual binding and the client knows what properties are needed for a bind

app/actions/service_credential_binding_create.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+
+      def initialize(user_audit_info)
+        @user_audit_info = user_audit_info
+      end
+
+      def precursor(service_instance, app: nil, name: nil)
+        not_supported! unless service_instance.user_provided_instance?
+        app_is_required! unless app.present?
+
```
FelisiaM, line 14: Not sure what type of validations go here instead of in the controller
e.g route does space mismatch check here rather than in the controller.

kirederik: I personally find weird that route binding is doing all of those validations in the action and not in the controller. Looking at other controllers (like service instances) I find most of the validation to be executed in the controller itself rather than in the action.

In this particular case, my reasoning here was to test the pre-requisites for the `precursor` method rather than the "validity" of the binding. That is, for this method to be executed, an app must be provided and the SI must be a UPSI.

app/actions/service_credential_binding_create.rb
kirederik: It depends what we understand as a binding. One could say that as soon as that there's a row in the service_bindings table we have a binding. I see the point of maybe storing the creds only at `bind`. I didn't feel the need to add the ServiceBrokers::UserProvided::Client as a dependency (I actually didn't know it existed). I'll take a look at it.

app/actions/service_credential_binding_create.rb
kirederik: I agree, I don't like this name. I thought about calling it `create_binding` -- that's what it's actually doing. I chose to keep it consistent with the route bindings though. I guess we should change it to a name that works for both bindings?

Note that, as mentioned in another comment below, this particular method is NOT running validations.

PR #1835, opened 2020-09-09
V3 route binding last operation polling 174398276

pivotal-marcela-campo:
[#174398276](https://www.pivotaltracker.com/story/show/174398276)

Fixes 

1. Fix job endpoint to look up for correct table for route_bindings, GET v3/jobs/<route-binding-job-guid> now works. I added a special case although it could have also been fixed by changing the value of resource_type in the job class. I chose this path so as to keep consistency and store always the resource name as we do with the rest.

2. Fix error codes handling according to osbapi spec, only 200 and 400 are valid responses from last operation for a create request

### DIFF
--- a/app/actions/service_route_binding_create.rb
+++ b/app/actions/service_route_binding_create.rb
@@ -47,7 +47,10 @@ module VCAP::CloudController
 
       def poll(binding)
         client = VCAP::Services::ServiceClientProvider.provide(instance: binding.service_instance)
-        details = client.fetch_service_binding_last_operation(binding)
+
+        details = fetch_last_operation(client, binding)
+        return PollingNotComplete.new unless details
+
         attributes = {}
 
         complete = details[:last_operation][:state] == 'succeeded'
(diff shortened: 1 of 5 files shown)


app/actions/service_route_binding_create.rb
```diff
@@ -108,6 +100,19 @@ class BindingNotRetrievable < StandardError; end
 
       attr_reader :service_event_repository
 
+      def fetch_last_operation(client, binding)
+        client.fetch_service_binding_last_operation(binding)
+      rescue VCAP::Services::ServiceBrokers::V2::Errors::ServiceBrokerBadResponse,
```
FelisiaM, line 105: I think we are not being consistent on where these errors are caught. For provision/deprovision/bind/unbind these are caught in the client. For Last operations polling (for SI and SB) we are doing it on the action. This might have to do with the fact that LO errors are interpreted differently for create and delete (or update). Starting to wonder if we should instead have several  `fetch_service_binding_last_operation` in the client that are specific to create/update/delete polling and that way there is only one place where we define what is considered as an error from OSBAPI. Thinking to create a chore for this though

FelisiaM (approved):

app/actions/service_route_binding_create.rb
pivotal-marcela-campo: There must be some inconsistency, when comparing behaviour for bindings and SI provisioning/update I found SI doesn't respect fully the OSBAPI spec. So I think we should create a explore for that first, as I didn't check all scenarios.  That will throw some light on it as well.
Also, the flow is quite convoluted, and I think that probably v2 itself doesn't completely respect osbapi 2.15. So it's worth assessing that as well and decide if we want to fix for both or diverge and create new client methods to make it compliant in a simpler way

PR #1837, opened 2020-09-10
v3(services): services route binding includes

butzopower:
[#174272947](https://www.pivotaltracker.com/story/show/174272947)

### DIFF
--- a/app/controllers/v3/service_route_bindings_controller.rb
+++ b/app/controllers/v3/service_route_bindings_controller.rb
@@ -50,10 +50,19 @@ class ServiceRouteBindingsController < ApplicationController
       paginated_result: SequelPaginator.new.get_page(route_bindings, message.try(:pagination_options)),
       path: '/v3/service_route_bindings',
       message: message,
+      decorators: decorators(message)
     )
   end
 
   private
+
+  AVAILABLE_DECORATORS = [
+    IncludeBindingServiceInstanceDecorator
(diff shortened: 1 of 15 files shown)


app/messages/service_route_binding_show_message.rb
```diff
@@ -0,0 +1,17 @@
+require 'messages/list_message'
+
+module VCAP
+  module CloudController
+    class ServiceRouteBindingShowMessage < BaseMessage
+      QUERY_PARAMS = %w[
+        include
+      ].freeze
+
+      register_allowed_keys QUERY_PARAMS.map(&:to_sym)
```
FelisiaM, line 10: The message is missing the validation the values allowed for `include`. The same for the list message as well

app/controllers/v3/service_route_bindings_controller.rb
```diff
@@ -50,11 +57,21 @@ def index
...
       message: message,
+      decorators: decorators(message)
     )
   end
 
   private
 
+  AVAILABLE_DECORATORS = [
+    IncludeBindingServiceInstanceDecorator,
+    IncludeRouteDecorator
```
FelisiaM, line 68: Would probably rename this to `IncludeBindingRouteDecorator`to be in line with other decorators

FelisiaM (changes requested):

PR #1844, opened 2020-09-14
emit pending/failed job metrics even when queue is empty

tcdowney:
Historically metrics would only be emitted when there were pending/failed jobs for a given queue in the database at the time the `PeriodicUpdater` triggered. This means that for quick moving queues, the metric may never get emitted.

For pull-based metrics aggregators like Prometheus this causes even bigger problems. Metrics would appear "stuck" at the value the queue had the last time the `PeriodicUpdater` ran when the queue was non-empty.

Here's some examples:

## Before
Screen Shot 2020-09-14 at 3 10 21 PM

Notice how Prometheus was not able to scrape the metric until it was non-zero (2) and that the metric never goes back down.

## After
Screen Shot 2020-09-14 at 2 41 32 PM

Notice how Prometheus is able to scrape zero-valued metrics and the queue returns back to zero.

### Additional Context
Since each `cf-api-server` is scraped at a different time, I think the best way to actually track this metric is with the `max` promql function. E.g.

```
max(cc_job_queue_length_cc_cf_api_server_5b94765866_qql2x)
```

Screen Shot 2020-09-14 at 3 25 16 PM

[CAPI Story #174516310](https://www.pivotaltracker.com/story/show/174516310)

### DIFF
--- a/lib/cloud_controller/metrics/periodic_updater.rb
+++ b/lib/cloud_controller/metrics/periodic_updater.rb
@@ -8,6 +8,9 @@ module VCAP::CloudController::Metrics
       @updaters = updaters
       @log_counter = log_counter
       @logger = logger
+      @known_job_queues = {
+        VCAP::CloudController::Jobs::Queues.local(VCAP::CloudController::Config.config).to_sym => 0
+      }
     end
 
     def setup_updates
@@ -75,10 +78,12 @@ module VCAP::CloudController::Metrics
 
       total                      = 0
       pending_job_count_by_queue = jobs_by_queue_with_count.each_with_object({}) do |row, hash|
+        @known_job_queues[row[:queue].to_sym] = 0
         total += row[:count]
         hash[row[:queue].to_sym] = row[:count]
       end
 
+      pending_job_count_by_queue.reverse_merge!(@known_job_queues)
       @updaters.each { |u| u.update_job_queue_length(pending_job_count_by_queue, total) }
     end
 
@@ -93,10 +98,12 @@ module VCAP::CloudController::Metrics
 
       total                = 0
       failed_jobs_by_queue = jobs_by_queue_with_count.each_with_object({}) do |row, hash|
+        @known_job_queues[row[:queue].to_sym] = 0
         total += row[:count]
         hash[row[:queue].to_sym] = row[:count]
       end
 
+      failed_jobs_by_queue.reverse_merge!(@known_job_queues)
       @updaters.each { |u| u.update_failed_job_count(failed_jobs_by_queue, total) }
     end
 
--- a/spec/unit/lib/cloud_controller/metrics/periodic_updater_spec.rb
+++ b/spec/unit/lib/cloud_controller/metrics/periodic_updater_spec.rb
@@ -241,6 +241,38 @@ module VCAP::CloudController::Metrics
         allow(updater2).to receive(:update_job_queue_length)
       end
 
+      context 'when local queue has pending jobs' do
+        it 'emits the correct count' do
+          Delayed::Job.enqueue(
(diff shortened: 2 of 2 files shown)


lib/cloud_controller/metrics/periodic_updater.rb
```diff
@@ -75,10 +78,12 @@ def update_job_queue_length
 
       total                      = 0
       pending_job_count_by_queue = jobs_by_queue_with_count.each_with_object({}) do |row, hash|
+        @known_job_queues[row[:queue].to_sym] = 0
         total += row[:count]
         hash[row[:queue].to_sym] = row[:count]
       end
 
+      pending_job_count_by_queue = @known_job_queues.merge(pending_job_count_by_queue)
```
cwlbraa, line 86: the name makes this opaque to me... before doing this merge pending_job_count_by_queue is different how? what do we accomplish by merging them?

cwlbraa (changes requested):
The tests and commit make it pretty clear to me that the goal here is to correctly emit 0 when there are no pending jobs, but the production code is pretty opaque and shadows variable names in a way that doesn't help. 

How does merging the counted queues into known_job_queues help here? Is it just keeping track of all the queues and `||= 0`ing them at the instance variable level?

lib/cloud_controller/metrics/periodic_updater.rb
tcdowney: @known_job_queues is building up a persistent hash of queues that have been seen before with 0 as a default value since if the queue isn't returned by the database query we know the count is 0.

This ends up looking like:

`{"some-queue" => 0, "some-other-queue" => 0}`

The pending_job_count_by_queue hash only contains things that have non-zero counts. It ends up looking something like:

`{"some-other-queue" => 1}`

The merge operation will merge the two and return a new hash. The values from the argument hash take precedent.

So it becomes `{"some-queue" => 0, "some-other-queue" => 1}`

I realize though that since we're using Rails / ActiveSupport I could use `reverse_merge!` (https://apidock.com/rails/v5.2.3/Hash/reverse_merge%21) which will make the calling hash's values take precedence. This might make this look more intuitive.

So instead we'd have like:

```ruby
pending_job_count_by_queue.reverse_merge!(@known_job_queues)
```

There are also more verbose ways of doing this. This just felt more Rubyish.

tcdowney: Here's the Set approach:

```ruby
    def update_job_queue_length
      jobs_by_queue_with_count = Delayed::Job.where(attempts: 0).group_and_count(:queue)

      total                      = 0
      pending_job_count_by_queue = jobs_by_queue_with_count.each_with_object({}) do |row, hash|
        @known_job_queues.add(row[:queue].to_sym)
        total += row[:count]
        hash[row[:queue].to_sym] = row[:count]
      end

      @known_job_queues.each do |queue|
        pending_job_count_by_queue[queue] ||= 0
      end

      @updaters.each { |u| u.update_job_queue_length(pending_job_count_by_queue, total) }
    end
```

cwlbraa: `reverse_merge!` is best IMO, it disambiguates the structure of of the final hash and shows that the transformation is being done in-place.

tcdowney: Switched to the `reverse_merge!` version:
https://github.com/cloudfoundry/cloud_controller_ng/commit/0fddfd413e89f5c14fe9f71241628e7aeadb68c7#diff-e263d41f0b0a3927a046dcd948d895f6R106

cwlbraa (approved):
LGTM!

tcdowney:
MySQL failure was an unrelated services test flake.

Issue #1848, opened 2020-09-11
[sapi] 1.93.0 regression. Unable to update meta-data to a service instance being async provisionned


gberche-orange:
## Issue

As a cf user
* in order to attach metadata to a service instance that takes hours to create
* I need the `PATCH  /v3/service_instances/guid` endpoint to accept update to metadata although the service instance is still in progress

This behavior regressed with capi 1.93.0 

## Context

This scenario also affects tooling that provision service instances and need to attach metadata to them (this issue was detected through https://github.com/orange-cloudfoundry/osb-cmdb/issues/65 )

## Steps to Reproduce

```
cf cs service async-plan myservice
guid=$(cf s myservice --guid)
CF_TRACE=true cf curl -X PATH /v3/service_instances/${guid} -d '{
    "metadata": {
      "annotations": {
        "note": "detailed information"
      },
      "labels": {
        "key": "value"
      }
    }
  }'
```

## Expected result

`HTTP/1.1 200 OK`

## Current result

starting with 1.93.0 (and likely related support for async brokers at https://www.pivotaltracker.com/n/projects/2105761/stories/171649250 ) this fails with

`CF-AsyncServiceInstanceOperationInProgress(60016): An operation for service instance f0a8f886-76a2-49d8-80eb-5f7230e5f8fe is in progress.`

## Possible Fix

Allow metadata-only update to a service instance (or service binding) despite the status "in progress"

kirederik:
Thanks @gberche-orange!

This issue is now fixed and will be included in the next capi-release. We'll close the issue once it's released

gberche-orange:
Thanks a lot @kirederik for the prompt fix, I'll watch the capi release repo to test iit.

kirederik:
CAPI 1.99.0 is released and should contain the fix for this!

I'll close it for now but please reopen it if you find any other issues!

Issue #1847, opened 2020-09-15
capi 1.97.0 breaking change: "cf set-label service-instance" swallows service instance status and message


gberche-orange:
## Issue

As a cf-user
* in order to manage service instances with labels and annotations
* I need the service instance status and message to not be altered after metadata are assigned

Rationale: labels and annotations can be assigned to service instances by different personas such as automated tooling (for use cases such as recharging, alarming, etc...) which should be independent of the service instance resource lifecycle (which is kept in sync with the vision exposed by the associated service broker)

## Context

Support for attaching metadata has been around for a while now, and enable users and tooling to attach metadata to cloudfoundry resources, including service instances.

Starting with recent capi version (likely 1.93.0) with work on service instances in v3 api, the equivalent of a `cf set-label service-instance` has the side effect of resetting the service instance status and message fields.

## Steps to Reproduce

* deploy overview broker from https://github.com/mattmcneeney/overview-broker
* edit catalog to leave only small plan without associated schema
* configure broker with async failure response
![image](https://user-images.githubusercontent.com/4748380/93106321-6b1d9800-f6b0-11ea-937a-207fe7589470.png)

```
cf cs overview-service small gberche3
Creating service instance gberche3 in org osb-cmdb-services-acceptance-tests / space development as gberche...
OK

Create in progress. Use 'cf services' or 'cf service gberche3' to check operation status.
guillaume@guillaume-dev-box:~/code/sec-group-broker-filter$ (master) cf service gberche3
Showing info of service gberche3 in org osb-cmdb-services-acceptance-tests / space development as gberche...

name:             gberche3
service:          overview-service
tags:             
plan:             small
description:      Provides an overview of any service instances and bindings that have been created by a platform.
documentation:    
dashboard:        https://overview-broker.redacted-domain.org/dashboard?time=2020-09-14T15:06:12.514Z
service broker:   overview-broker

Showing status of last operation from service gberche3...

status:    create failed
message:   Operation failed
started:   2020-09-14T15:06:12Z
updated:   2020-09-14T15:06:16Z

There are no bound apps for this service.

There is no upgrade available for this service.
```

then update metadata for this service
```
CF_TRACE=true cf curl -X PATCH /v3/service_instances/$(cf service gberche3 --guid) -d '{
    "metadata": {
      "annotations": {
        "note": "detailed information"
      },
      "labels": {
        "key": "value"
      }
    }
  }'
```

## Expected result

[...]

PR #1850, opened 2020-09-16
v3(service instance): allow metadata updates while create is in progress

kirederik:
[#174825693](https://www.pivotaltracker.com/story/show/174825693)
[#174825537](https://www.pivotaltracker.com/story/show/174825537)

Solves #1847
Solves #1848

### DIFF
--- a/spec/unit/actions/service_instance_update_managed_spec.rb
+++ b/spec/unit/actions/service_instance_update_managed_spec.rb
@@ -66,14 +66,17 @@ module VCAP::CloudController
 
           expect(service_instance.name).to eq('different-name')
           expect(service_instance.tags).to eq(%w(accounting couchbase nosql))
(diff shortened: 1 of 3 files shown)


spec/unit/actions/service_instance_update_managed_spec.rb
```diff
@@ -388,6 +408,90 @@ module VCAP::CloudController
...
+                                                                    alpha: 'beta',
+                                                                    'pre.fix/to_delete': nil,
+                                                                }
+                                                            }
+                                                        })
+          expect { action.update(service_instance, msg) }.not_to raise_error
+
+          service_instance.reload
+
+          expect(service_instance.labels.map { |l| { prefix: l.key_prefix, key: l.key_name, value: l.value } }).to match_array([
```
pivotal-marcela-campo, line 475: There's a expect_metadata method in this file that could be used here I think

pivotal-marcela-campo (changes requested):

PR #1851, opened 2020-09-17
v3(services): ensure MSI <=> app bindings are valid before creating

butzopower:
[#174088711](https://www.pivotaltracker.com/story/show/174088711)

### DIFF
--- a/app/actions/service_credential_binding_create.rb
+++ b/app/actions/service_credential_binding_create.rb
@@ -1,17 +1,22 @@
+require 'repositories/service_binding_event_repository'
+require 'services/service_brokers/service_client_provider'
+
 module VCAP::CloudController
   module V3
     class ServiceCredentialBindingCreate
       class UnprocessableCreate < StandardError
       end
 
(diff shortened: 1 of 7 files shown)


kirederik (changes requested):
I think we haven't added a request test to validate shared spaces. I can see we have one in the action that tests it will raise an `Unimplemented` error. There's an argument that we don't need to kinda duplicate the tests everywhere, but I wanted to raise this up in case it wasn't a conscious decision.

butzopower:
I purposely left out the request spec for that since it doesn't really feel like something we should document yet in the code.  It's not really intended that we throw a 501, so I kind of leave it to the story that will deal with sharing to add it.  I guess potentially this story enables binding to shared rather than "validates" against it?

I also didn't test disabling volume bindings in the request specs, partially because I didn't know how, but also because I'm not really sure how happy path the request tests should be for validations.  It sorta feels like we should have at least one API test that shows that a validation error maps to a 422, but the rest feel like they can be covered in the appropriate validation logic specs.

PR #1861, opened 2020-09-23
v3(services): update the platform broker poll interval while an async…

pivotal-marcela-campo:
… job is enqueued

[#174398272](https://www.pivotaltracker.com/story/show/174398272)

This just improves testing overall by making sure the updated interval is actually being used at runtime.

It also allows for platform default value to be updated in the unlikely scenario that we changed the platform value right after changing the polling interval from code. At the moment this scenario is very unlikely because it's the small interval between getting the response from the broker with a retry-after and re-scheduling the job, so I would be okay rolling back that change if we think it makes the code more complicated, but it makes for a more predictable behaviour.

### DIFF
--- a/app/jobs/reoccurring_job.rb
+++ b/app/jobs/reoccurring_job.rb
@@ -30,7 +30,8 @@ module VCAP::CloudController
       end
 
       def polling_interval_seconds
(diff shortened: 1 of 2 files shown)


kirederik (approved):
Sounds good!

PR #1863, opened 2020-09-23
v3(services): all bindings should have last_operation

blgm:
When V3 bindings work is complete all binding types should have a
last_operation. However bindings created in V2 will not have this. To
make the output JSON for V3 GET endpoints consistent, we should patch up
missing last_operation objects, rather than present `null` which can be
harder to parse.

[#171843320](https://www.pivotaltracker.com/story/show/171843320)

### DIFF
--- a/app/presenters/mixins/last_operation_helper.rb
+++ b/app/presenters/mixins/last_operation_helper.rb
@@ -1,0 +1,21 @@
+module VCAP::CloudController::Presenters::Mixins
+  module LastOperationHelper
+    def last_operation(resource)
+      last_operation = resource.try(:last_operation)
+
+      if last_operation
+        last_operation.to_hash
+      else
+        # Bindings created using V2 may not have a last operation, so we make
+        # a best attempt so that the output JSON can be parsed consistently
+        {
+          type: 'create',
+          state: 'succeeded',
+          description: '',
+          created_at: resource.try(:created_at),
+          updated_at: resource.try(:updated_at),
+        }
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


app/presenters/mixins/last_operation_helper.rb
```diff
@@ -0,0 +1,21 @@
+module VCAP::CloudController::Presenters::Mixins
+  module LastOperationHelper
+    def last_operation(resource)
+      last_operation = resource.try(:last_operation)
```
pivotal-marcela-campo, line 4: If I understand correctly `resource.try` is safe in the case `resource` is `nil`. Can that really happen?

pivotal-marcela-campo: oh, just checked and it also swallows no_method error so I guess that's why is here

app/presenters/mixins/last_operation_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+          type: 'create',
+          state: 'succeeded',
+          description: '',
+          created_at: resource.try(:created_at),
+          updated_at: resource.try(:updated_at),
+        }
+      end
+    end
+  end
+end
```
pivotal-marcela-campo, line 21: I think there are alternative ways of writing this same code
e.g. 
last_operation = resource.try(:last_operation)&.to_hash || {the hash here}

pivotal-marcela-campo (approved):
Looks good. There is only one suggestion but it's fine to merge anyway.

PR #1864, opened 2020-09-24
V3 route binding operation in progress

pivotal-marcela-campo:
[#174540070](https://www.pivotaltracker.com/story/show/174540070)

### DIFF
--- a/app/models/services/route_binding.rb
+++ b/app/models/services/route_binding.rb
@@ -1,3 +1,5 @@
+require 'cloud_controller/process_route_handler'
+
 module VCAP::CloudController
   class RouteBinding < Sequel::Model
     plugin :after_initialize
--- a/app/repositories/app_usage_event_repository.rb
+++ b/app/repositories/app_usage_event_repository.rb
@@ -1,4 +1,5 @@
 require 'database/old_record_cleanup'
+require 'cloud_controller/url_secret_obfuscator'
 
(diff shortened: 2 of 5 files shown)


app/models/services/route_binding.rb
```diff
@@ -1,3 +1,5 @@
+require 'cloud_controller/process_route_handler'
```
blgm, line 1: Does this relate to the change in spec helper?

app/repositories/app_usage_event_repository.rb
```diff
@@ -1,4 +1,5 @@
 require 'database/old_record_cleanup'
+require 'cloud_controller/url_secret_obfuscator'
```
blgm, line 2: Does this relate to the change in spec helper?

blgm (approved):
There are a couple of new `requires`s that are not immediately obvious, but may relate to the change in spec helper. If they are intentional fixes that I'm fine with that, but just wanted to highlight them in case they are the result of the IDE doing strange things.

app/models/services/route_binding.rb
pivotal-marcela-campo: yes!

app/repositories/app_usage_event_repository.rb
pivotal-marcela-campo: yes!

PR #1881, opened 2020-10-02
v3(services): async delete route binding

blgm:
[#174383882](https://www.pivotaltracker.com/story/show/174383882)

### DIFF
--- a/spec/unit/actions/service_route_binding_delete_spec.rb
+++ b/spec/unit/actions/service_route_binding_delete_spec.rb
@@ -5,13 +5,13 @@ module VCAP::CloudController
   module V3
     RSpec.shared_examples 'successful delete' do
       it 'deletes the binding' do
-        subject
+        perform_action
 
         expect(RouteBinding.all).to be_empty
       end
 
       it 'creates an audit event' do
-        subject
+        perform_action
 
         expect(event_repository).to have_received(:record_service_instance_event).with(
           :unbind_route,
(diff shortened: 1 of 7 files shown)


spec/unit/actions/service_route_binding_delete_spec.rb
```diff
@@ -3,6 +3,48 @@
 
 module VCAP::CloudController
   module V3
+    RSpec.shared_examples 'successful delete' do
+      it 'deletes the binding' do
+        subject
```
kirederik, line 8: I confess I find the use of `subject` this way (here and in a few other tests) a bit weird. 

Maybe that's because I'm used to thinking about `subject` either as a value or as an instance of the described class. In the docs, they mostly used it that way (see [implicit](https://relishapp.com/rspec/rspec-core/v/3-9/docs/subject/implicitly-defined-subject) and [explicit](https://relishapp.com/rspec/rspec-core/v/3-9/docs/subject/explicit-subject) `subject`)

There's an [example](https://relishapp.com/rspec/rspec-core/v/3-9/docs/subject/explicit-subject#the-%60subject%60-is-memoized-within-an-example-but-not-across-examples) where `subject` is used in a similar way, but the documentation recommends against it (or it recommends against mutating stuff on a `subject` call, it's a bit unclear).

In my view, a better approach would be to create the class in the `subject` and execute the function in a `before(:each)` (there's some cases where this is not entirely possible, like when exceptions are thrown).

kirederik (changes requested):
See comment above; feel free to ignore it as well :)

spec/unit/actions/service_route_binding_delete_spec.rb
blgm: I've pushed a change that uses a named subject, that I think makes things clearer. It's actually quite hard to use a `before do` because quite a lot of tests need to do some setup before performing the action. What do you think?

kirederik: Cool, that's fine 👍

kirederik (approved):

PR #1886, opened 2020-10-07
Add ability to disable staging through `PUT /v2/apps/:guid`

tcdowney:
This PR introduces a new, temporary config property called `temporary_disable_v2_staging` that can be used to disable package staging through the v2 APIs. The property will be removed once the [v2 APIs are fully disabled](https://docs.google.com/document/d/1KFZogeeexOqFf13oKHloe2QAorLh9OqwQHp8JvBl9lY/edit), hence the `temporary_` prefix.

In `capi-k8s-release` this property will default to `true`. We do not validate that the v2 APIs work in `cf-for-k8s` so disabling this can help reinforce their lack of support and encourage users to move over to clients that use the v3 APIs.

Additionally, this fixes an issue where the `PUT /v2/apps/:guid` API would not detect that a package uploaded to the package registry needed staging. This resulted in `cf push` commands that used the `/v2/apps/:guid` endpoint to hang until staging timed out.

CAKE Story: [#175157530](https://www.pivotaltracker.com/story/show/175157530)

### DIFF
--- a/app/actions/v2/app_update.rb
+++ b/app/actions/v2/app_update.rb
@@ -110,6 +110,10 @@ module VCAP::CloudController
       end
 
       def prepare_to_stage(app)
+        if v2_api_staging_disabled?
+          raise CloudController::Errors::ApiError.new_from_details('UnprocessableEntity', 'Staging through the v2 API is disabled')
+        end
+
         app.update(droplet_guid: nil)
       end
 
@@ -172,6 +176,10 @@ module VCAP::CloudController
 
       def staging_necessary?(process, request_attrs)
         request_attrs.key?('state') && process.needs_staging?
+      end
+
+      def v2_api_staging_disabled?
+        !!VCAP::CloudController::Config.config.get(:temporary_disable_v2_staging)
       end
     end
   end
--- a/app/models/runtime/process_model.rb
+++ b/app/models/runtime/process_model.rb
@@ -184,7 +184,7 @@ module VCAP::CloudController
       return nil if package.nil?
 
       if package.bits?
-        package.package_hash
+        package.checksum_info[:value]
       elsif package.docker?
         package.image
       end
@@ -485,7 +485,7 @@ module VCAP::CloudController
     end
 
     def needs_staging?
-      package_hash && !staged? && started?
+      package_hash.present? && !staged? && started?
     end
 
     def staged?
--- a/spec/unit/actions/app_start_spec.rb
+++ b/spec/unit/actions/app_start_spec.rb
@@ -89,17 +89,24 @@ module VCAP::CloudController
               state:   DropletModel::STAGED_STATE,
             )
           end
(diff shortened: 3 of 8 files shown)


spec/unit/actions/app_start_spec.rb
```diff
@@ -89,17 +89,24 @@ module VCAP::CloudController
               state:   DropletModel::STAGED_STATE,
             )
           end
-          let(:package) { PackageModel.make(app: app, package_hash: 'some-awesome-thing', state: PackageModel::READY_STATE) }
+          let(:package) do
+            PackageModel.make(
+              app: app,
+              package_hash: 'some-sha1',
+              sha256_checksum: 'some-sha256',
+              state: PackageModel::READY_STATE
```
tcdowney, line 97: One thought I had while working on this was what if we could key off of the `PackageModel::READY_STATE` instead of the presence of checksums for know when an app is ready to be staged in v2?

I'm unsure of if there are other implications around doing that in v2 though...
Or if `READY` truly means "ready" in this case

app/models/runtime/process_model.rb
```diff
@@ -184,7 +184,7 @@ def package_hash
       return nil if package.nil?
 
       if package.bits?
-        package.package_hash
+        package.checksum_info[:value]
       elsif package.docker?
         package.image
```
tcdowney, line 189: The fact that a method called `package_hash` can return a docker image address makes me not worried that it can now return a sha1 or a sha256 checksum.

spec/unit/actions/app_start_spec.rb
cwlbraa: I think principal-of-smallest change is probably best here, and sticking with using the checksums fits with that.

cwlbraa (commented):
is it possible to make the api_schema field optional & bool with a default?

app/actions/v2/app_update.rb
```diff
@@ -173,6 +177,10 @@ def is_web_process_deploying(mass_assign, process)
       def staging_necessary?(process, request_attrs)
         request_attrs.key?('state') && process.needs_staging?
       end
+
+      def v2_api_staging_disabled?
+        !!VCAP::CloudController::Config.config.get(:temporary_disable_v2_staging)
```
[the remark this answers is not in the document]
tcdowney: Yes, I want to coerce nil into false since this is an optional config property. In the schema it is set to be a `bool`, so I don't expect there to be any strings.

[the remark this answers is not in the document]
tcdowney: Regarding your comment above:

> is it possible to make the api_schema field optional & bool with a default?

We traditionally did this by making it have a default value in `capi-release`, we currently don't have a generic solution for defaulting config right now in CCNG as far as I know.

cwlbraa (approved):

tcdowney:
Rebased and merged manually. Closing this PR.

PR #1897, opened 2020-10-12
V3 service credential bindings rework 174119280

FelisiaM:
[#174119280](https://www.pivotaltracker.com/story/show/174119280)

### DIFF
--- a/app/actions/service_credential_binding_create.rb
+++ b/app/actions/service_credential_binding_create.rb
@@ -55,6 +55,18 @@ module VCAP::CloudController
         event_repository.record_create(binding, @user_audit_info, @audit_hash, manifest_triggered: false)
       end
 
+      def save_incomplete_binding(binding, broker_operation)
+        binding.save_with_attributes_and_new_operation(
+          {},
+          {
+            type: 'create',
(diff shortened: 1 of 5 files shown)


app/actions/service_credential_binding_create.rb
```diff
@@ -55,6 +55,18 @@ def complete_binding_and_save(binding, binding_details, last_operation)
...
+      def save_incomplete_binding(binding, broker_operation)
+        binding.save_with_attributes_and_new_operation(
+          {},
+          {
+            type: 'create',
+            state: 'in progress',
+            broker_provided_operation: broker_operation
+          }
+        )
+        event_repository.record_start_create(binding, @user_audit_info, @audit_hash, manifest_triggered: false)
```
blgm, line 67: Ideally a function called `save...` would only save things, and not also create audit events. In future we might rename this to something like `record_binding_start()`.

blgm (approved):

app/actions/service_credential_binding_create.rb
FelisiaM: Good point - I will update the method name in the next story

PR #1903, opened 2020-10-14
v3(services) Make sure the LO state is failed when job fails

FelisiaM:
Sometimes the Job can fail for unexpected reasons. In any failure we should make sure to leave the LO state as 'failed'

[#175264299](https://www.pivotaltracker.com/story/show/175264299)

### DIFF
--- a/app/jobs/v3/create_binding_async_job.rb
+++ b/app/jobs/v3/create_binding_async_job.rb
@@ -7,7 +7,7 @@ require 'jobs/v3/create_service_binding_job_factory'
 module VCAP::CloudController
   module V3
     class CreateBindingAsyncJob < Jobs::ReoccurringJob
-      class BindingGone < CloudController::Errors::ApiError; end
+      class BindingNotFound < CloudController::Errors::ApiError; end
 
       def initialize(type, precursor_guid, parameters:, user_audit_info:, audit_hash:)
         super()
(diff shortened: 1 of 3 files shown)


app/jobs/v3/create_binding_async_job.rb
```diff
@@ -87,6 +92,19 @@ def handle_timeout
         )
       end
 
+      def save_failure(error_message)
```
blgm, line 95: The resource can be updated by the action or the job, which I think is confusing. I think it would be better if all the updating of the resource were handled in the action.

blgm (changes requested):
The resource is accessed by both the job and the action. The read access is historical, but probably an anti-pattern. Adding write access in this change makes it harder to reason about what the action does and what the job does.

app/jobs/v3/create_binding_async_job.rb
```diff
@@ -70,9 +72,12 @@ def perform
         if polling_status[:retry_after].present?
           self.polling_interval_seconds = polling_status[:retry_after]
         end
+      rescue BindingGone => e
```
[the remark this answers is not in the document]
FelisiaM: Updated this to BindingNotFound - that's a more wider-used terminology for these errors

app/jobs/v3/create_binding_async_job.rb
FelisiaM: I agree- will put a refactoring in the backlog to see how the updates of the bindings (here and in handle timeout) can be removed from the job

https://www.pivotaltracker.com/story/show/175283473

PR #1904, opened 2020-10-15
v3(bindings): can create bindings with params

kirederik:
[#175203762](https://www.pivotaltracker.com/story/show/175203762)


blgm (approved):
Looks OK, but tests failed

PR #1909, opened 2020-10-15
fix docker app deletion when using a package registry

tcdowney:
- the package.bits_image_reference method returns an error when the
package type is 'docker'
- there is no need to clean up the package registry for a docker app so
we can skip it

fixes https://github.com/cloudfoundry/capi-k8s-release/issues/84
[#175295965](https://www.pivotaltracker.com/story/show/175295965)

### DIFF
--- a/app/actions/package_delete.rb
+++ b/app/actions/package_delete.rb
@@ -8,12 +8,8 @@ module VCAP::CloudController
       packages = Array(packages)
 
       packages.each do |package|
-        package_delete = if VCAP::CloudController::Config.config.package_image_registry_configured?
-                           Jobs::Kubernetes::RegistryDelete.new(package.bits_image_reference)
-                         else
-                           Jobs::Runtime::BlobstoreDelete.new(package.guid, :package_blobstore)
-                         end
-        Jobs::Enqueuer.new(package_delete, queue: Jobs::Queues.generic).enqueue
+        package_src_delete_job = create_package_source_deletion_job(package)
+        Jobs::Enqueuer.new(package_src_delete_job, queue: Jobs::Queues.generic).enqueue if package_src_delete_job
         package.destroy
 
         Repositories::PackageEventRepository.record_app_package_delete(
@@ -22,6 +18,18 @@ module VCAP::CloudController
       end
 
       []
+    end
+
+    private
+
+    def create_package_source_deletion_job(package)
+      return Jobs::Runtime::BlobstoreDelete.new(package.guid, :package_blobstore) unless package_registry_configured?
+
+      package.bits? ? Jobs::Kubernetes::RegistryDelete.new(package.bits_image_reference) : nil
+    end
+
+    def package_registry_configured?
+      VCAP::CloudController::Config.config.package_image_registry_configured?
     end
   end
 end
--- a/app/jobs/runtime/delete_expired_package_blob.rb
+++ b/app/jobs/runtime/delete_expired_package_blob.rb
@@ -17,11 +17,7 @@ module VCAP::CloudController
           package = PackageModel.find(guid: package_guid)
           return unless package
 
-          if VCAP::CloudController::Config.config.package_image_registry_configured?
-            VCAP::CloudController::Jobs::Kubernetes::RegistryDelete.new(package.bits_image_reference).perform
-          else
-            BlobstoreDelete.new(package_guid, :package_blobstore).perform
-          end
+          create_package_source_deletion_job(package)&.perform
           package.update(package_hash: nil, sha256_checksum: nil)
         end
 
@@ -35,6 +31,18 @@ module VCAP::CloudController
 
         def logger
           @logger ||= Steno.logger('cc.background')
+        end
+
+        private
+
+        def package_registry_configured?
+          VCAP::CloudController::Config.config.package_image_registry_configured?
+        end
+
+        def create_package_source_deletion_job(package)
+          return Jobs::Runtime::BlobstoreDelete.new(package.guid, :package_blobstore) unless package_registry_configured?
+
+          package.bits? ? Jobs::Kubernetes::RegistryDelete.new(package.bits_image_reference) : nil
         end
       end
     end
--- a/spec/unit/actions/package_delete_spec.rb
+++ b/spec/unit/actions/package_delete_spec.rb
@@ -24,18 +24,32 @@ module VCAP::CloudController
             TestConfig.override(packages: { image_registry: { base_path: 'hub.example.com/user' } })
           end
 
-          it 'schedules a job to the delete the blobstore item' do
-            expect {
-              package_delete.delete(package)
-            }.to change {
-              Delayed::Job.count
-            }.by(1)
+          context 'when the package type is bits' do
+            it 'schedules a job to the delete the package from the container registry' do
+              expect {
+                package_delete.delete(package)
+              }.to change {
+                Delayed::Job.count
+              }.by(1)
 
-            job = Delayed::Job.last
-            expect(job.handler).to include('VCAP::CloudController::Jobs::Kubernetes::RegistryDelete')
-            expect(job.handler).to include(package.bits_image_reference(digest: false))
-            expect(job.queue).to eq(Jobs::Queues.generic)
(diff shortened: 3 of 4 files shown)


cwlbraa (commented):
this looks fine to me, but I also don't understand how this wasn't failing with docker packages prior and `BlobstoreDelete.new(package_guid, :package_blobstore).perform`

spec/unit/actions/package_delete_spec.rb
```diff
@@ -24,18 +24,32 @@ module VCAP::CloudController
...
-            expect(job.guid).not_to be_nil
+              job = Delayed::Job.last
+              expect(job.handler).to include('VCAP::CloudController::Jobs::Kubernetes::RegistryDelete')
+              expect(job.handler).to include(package.bits_image_reference(digest: false))
+              expect(job.queue).to eq(Jobs::Queues.generic)
+              expect(job.guid).not_to be_nil
+            end
+          end
+
+          context 'when the package type is docker' do
```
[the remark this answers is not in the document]
cwlbraa: does the old task perform idempotent deletes, but the new one does not?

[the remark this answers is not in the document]
tcdowney: My guess is this code was protecting it:
https://github.com/cloudfoundry/cloud_controller_ng/blob/68aa90a221e56d159586f7cd51449567ff2357c1/app/jobs/runtime/blobstore_delete.rb#L18-L19

I think the intent of it is to not try and delete something that was already deleted... but it probably shields docker packages as well.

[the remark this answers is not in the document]
tcdowney: We were hoping to get this through by EOD for the cf-for-k8s 1.0 cutoff, so I was hesitant to change the behavior there.

app/actions/package_delete.rb
```diff
@@ -23,5 +19,17 @@ def delete(packages)
 
       []
     end
+
+    private
+
+    def create_package_source_deletion_job(package)
```
njbennett, line 25: These ternaries are a little hard for me to read but I think I've worked it out:

If a package registry isn't configured, it returns a BlobstoreDelete job
If a package registry is configured *and* there are package bits, it returns a RegistryDelete job
If a package registry is configured but there are no package bits, it returns nil

app/actions/package_delete.rb
```diff
@@ -23,5 +19,17 @@ def delete(packages)
...
+      return Jobs::Runtime::BlobstoreDelete.new(package.guid, :package_blobstore) unless package_registry_configured?
+
+      package.bits? ? Jobs::Kubernetes::RegistryDelete.new(package.bits_image_reference) : nil
```
njbennett, line 28: What's the difference between Jobs::Kubernetes and Jobs::Runtime?

app/jobs/runtime/delete_expired_package_blob.rb
```diff
@@ -17,11 +17,7 @@ def perform
           package = PackageModel.find(guid: package_guid)
           return unless package
 
-          if VCAP::CloudController::Config.config.package_image_registry_configured?
-            VCAP::CloudController::Jobs::Kubernetes::RegistryDelete.new(package.bits_image_reference).perform
-          else
-            BlobstoreDelete.new(package_guid, :package_blobstore).perform
-          end
+          create_package_source_deletion_job(package)&.perform
```
njbennett, line 20: Is the & here marking this as a block that gets passed to perform or... something else?

I'm looking at (this article on the ampersand operator)[https://medium.com/@sologoubalex/parameter-with-ampersand-operator-in-ruby-6a6a7fd666d5] but it doesn't look like an exact match for this syntax.

app/jobs/runtime/delete_expired_package_blob.rb
```diff
@@ -36,6 +32,18 @@ def max_attempts
...
           @logger ||= Steno.logger('cc.background')
         end
+
+        private
+
+        def package_registry_configured?
+          VCAP::CloudController::Config.config.package_image_registry_configured?
+        end
+
+        def create_package_source_deletion_job(package)
```
njbennett, line 42: Why is this same method defined in two different places?

spec/unit/actions/package_delete_spec.rb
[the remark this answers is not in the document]
cwlbraa: I wouldn't change the behavior, but a different way of framing the bug is to say that the RegistryDelete job is insufficiently idempotent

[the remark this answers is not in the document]
tcdowney: > RegistryDelete job is insufficiently idempotent

The `RegistryDelete` job is idempotent. This error is actually being raised by the `package.bits_image_reference` method since there is no such thing as a `bits_image_reference` for a docker package.

If that method returned an empty string in this case we would then handle that in the `RegistryDelete` action and not call the registry-buddy... but ideally I think neither the `BlobstoreDelete` nor `RegistryDelete` should be called in the case of docker apps cause we _know_ there is no source code to delete.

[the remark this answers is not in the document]
tcdowney: To summarize that a bit more succinctly...

`package.bits_image_reference` currently raises an error. We could have that method return `""` or `nil` in the docker package case.

The caveat is it kind of masks the fact that we don't really need to delete anything from a container registry for docker packages.

When @matt-royal and I were pairing on this code originally we valued the explicitness of the error raising... but now I have no strong feelings either way.

njbennett (commented):
Oh I see I have to actually submit a review

app/jobs/runtime/delete_expired_package_blob.rb
tcdowney: This is the "safe navigation" operator that was introduced in Ruby 2.3 (I think).

https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

An alternative way of writing this would be:
```ruby
package_src_delete_job = create_package_source_deletion_job(package)
package_src_delete_job.perform if package_src_delete_job
```

But I liked the way this read and felt very Rubyish. 😂 

I can change it to a more explicit form if you'd like.

cwlbraa: safe cuz it's nil-safe :)

spec/unit/actions/package_delete_spec.rb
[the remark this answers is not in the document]
cwlbraa: > package.bits_image_reference currently raises an error

ahhh by the time I had dug this far I had lost sight of this

cwlbraa (approved):

app/jobs/runtime/delete_expired_package_blob.rb
tcdowney: They're private methods for each class.

I introduced them to take the focus off of the messy conditionals in the `perform` method. My hope was that it would make it clearer what was going on in `perform`. 

However, I made them private because I don't want them to become part of the public interface of these jobs and want them to remain free to drift over time.

At one point while extracting this logic out of the main `perform` methods these actually had different code in them. They only became the same when I got fancy and wanted to use the `&.` operator. 🙂

app/actions/package_delete.rb
tcdowney: We had put this class in a `jobs/kubernetes` folder and by convention the Ruby module namespacing should match the folder structure (we break this all over the place in Cloud Controller).

@matt-royal also has some hopes and dreams for a `kubernetes` module namespace and this is one step closer to that.

app/actions/package_delete.rb
tcdowney: Yeah, that's right. One minor clarification though is that instead of "there are package bits" it is actually "this package is a bits package" (as opposed to a "docker package").

I'd really like the logic here to be:

* If `package.bits?`
  * If package registry configured: return `RegistryDelete`
  * Else return `BlobstoreDelete` job
* Else return nil

But the original behavior was to always return a `BlobstoreDelete` job and I was scared to change that behavior. See the convo with Connor above: https://github.com/cloudfoundry/cloud_controller_ng/pull/1909#discussion_r505910827

PR #1913, opened 2020-10-19
V3 handle last operation on unbind route service 174748133

FelisiaM:
[#174748133](https://www.pivotaltracker.com/story/show/174748133)


app/jobs/v3/delete_route_binding_job.rb
```diff
@@ -38,7 +38,9 @@ def resource_type
 
       def perform
         binding = route_binding
-        gone! unless binding
+        if binding.nil?
+          return finish
+        end
```
FelisiaM, line 43: While writing the tests I noticed we are failing when binding is not there. Given this is a delete should we just succeed the job?

kirederik: Yes. Technically it should go into the job right (as it should exit from the controller), but makes no harm in adding this here

app/jobs/v3/delete_route_binding_job.rb
```diff
@@ -47,24 +49,43 @@ def perform
         if @first_time
           @first_time = false
           delete_result = action.delete(binding, async_allowed: true)
-          return finish if delete_result.is_a? V3::ServiceRouteBindingDelete::DeleteComplete
+          if delete_result[:finished]
+            finish
```
kirederik, line 53: If you don't return here the function will proceed (therefore executing the polling on line 57). Is this intentional?

app/jobs/v3/delete_route_binding_job.rb
```diff
@@ -47,24 +49,43 @@ def perform
...
-          unless polling_status.retry_after.nil?
-            self.polling_interval_seconds = polling_status.retry_after.to_i
-          end
+        end
+
+        if polling_status[:retry_after].present?
+          self.polling_interval_seconds = polling_status[:retry_after]
         end
       rescue => e
+        if route_binding.reload.last_operation.state != 'failed'
```
kirederik, line 66: You could technically skip this and always save "state failed" 🤔

app/jobs/v3/delete_route_binding_job.rb
FelisiaM: hmm - I thought that was returning. There is a test [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/3bea98c52a0680498b9bfadcc683749df0a555a5/spec/unit/jobs/v3/delete_route_binding_job_spec.rb#L54-L71) checks that. 
Does that look right?

app/jobs/v3/delete_route_binding_job.rb
```diff
@@ -38,7 +38,9 @@ def resource_type
 
       def perform
         binding = route_binding
-        gone! unless binding
+        if binding.nil?
```
kirederik, line 41: a guard clause maybe? :D

app/jobs/v3/delete_route_binding_job.rb
FelisiaM: the intention here was to not change the description - as set by the previous steps. Happy to change it if you think it makes sense.

E.g poll can through LastOperationFailed and also sets the state to failed with the right description.

app/actions/service_route_binding_delete.rb
```diff
@@ -4,9 +4,14 @@ class ServiceRouteBindingDelete
...
-      DeleteStarted = Struct.new(:operation).freeze
-      DeleteInProgress = Struct.new(:retry_after).freeze
+
+      DeleteStatus = Struct.new(:finished, :operation).freeze
+      DeleteStarted = ->(operation) { DeleteStatus.new(false, operation) }
+      DeleteComplete = DeleteStatus.new(true, nil).freeze
+
+      PollingStatus = Struct.new(:finished, :retry_after).freeze
+      PollingFinished = PollingStatus.new(true, nil).freeze
+      ContinuePolling = ->(retry_after) { PollingStatus.new(false, retry_after) }
```
kirederik, line 14: Starting to think we may want to extract this into a common place (create has similar structs)... But its fine as is for now

app/actions/service_route_binding_delete.rb
```diff
@@ -18,13 +23,12 @@ def delete(binding, async_allowed:)
         operation_in_progress! if binding.service_instance.operation_in_progress?
 
         result = send_unbind_to_broker(binding)
-        case result
-        when DeleteStarted
-          update_last_operation(binding, operation: result[:operation])
-        when DeleteComplete
+        if result[:finished]
           perform_delete_actions(binding)
+          return result
```
kirederik, line 28: Maybe an `if ... else ... end` would work better here?

app/jobs/v3/delete_route_binding_job.rb
kirederik: Hum... that test looks wrong 🙈

It's probably passing because we never call `subjet.perform` in the second `it` block

app/jobs/v3/delete_route_binding_job.rb
kirederik: Fair enough. I was assuming that whatever is raising, is doing it with the correct error message. But maybe not?

kirederik (changes requested):
See comments :D

app/jobs/v3/delete_route_binding_job.rb
FelisiaM: you're right - surpriseee
I fixed it

app/jobs/v3/delete_route_binding_job.rb
FelisiaM: done

kirederik (approved):

PR #1417, opened 2019-08-21
Add docker support in Eirini

akshaymankar:
* A short explanation of the proposed change:

This PR sends data for docker lifecycle on app_desire.
We have also refactored the way lifecycle specific data was being sent, so it is clearer which lifecycle is being followed.

* An explanation of the use cases your change solves

Users of eirini who want to push docker images.

* Links to any other associated PRs
N/A

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `master` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [x] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats)
  **NOTE**: We only ran the ones which work with Eirni. But since the change is only in the opi apps client, it should give us enough confidence.

PR #1917, opened 2020-10-20
Introduce "kpack" Droplets and Support Custom Start Commands

tcdowney:
This PR introduces the concept of "kpack" droplets by associating the existing `kpack_lifecycle_data` for a build with its droplet instead of creating "docker" droplets. It allows us to distinguish kpack-built droplets from regular "docker" droplets and means we can now support custom start commands by invoking them with the CNB launcer: `/cnb/lifecycle/launcher`.

Additionally it includes a migration to convert existing kpack-built "docker" droplets into "kpack" droplets by associating their `kpack_lifecycle_data` with the "docker" droplet.

This change addresses the following issues:
* https://github.com/cloudfoundry/capi-k8s-release/issues/72
* https://github.com/cloudfoundry/cf-for-k8s/issues/502

CAKE Story: https://www.pivotaltracker.com/story/show/175081086

### DIFF
--- a/app/actions/droplet_create.rb
+++ b/app/actions/droplet_create.rb
@@ -63,6 +63,20 @@ module VCAP::CloudController
       droplet
     end
 
+    def create_kpack_droplet(build)
+      droplet = droplet_from_build(build)
+
+      DropletModel.db.transaction do
+        droplet.save
+        droplet.kpack_lifecycle_data = build.kpack_lifecycle_data
+      end
+
+      droplet.reload
+      Steno.logger('build_completed').info("droplet created: #{droplet.guid}")
+      record_audit_event(droplet, build.package, user_audit_info_from_build(build))
+      droplet
+    end
+
     private
 
     def droplet_from_build(build)
--- a/db/migrations/20201020204056_index_kpack_lifecycle_data_droplet_guid.rb
+++ b/db/migrations/20201020204056_index_kpack_lifecycle_data_droplet_guid.rb
@@ -1,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table :kpack_lifecycle_data do
+      add_index :droplet_guid, name: :kpack_lifecycle_droplet_guid_index
+    end
+  end
+
+  down do
+    alter_table :kpack_lifecycle_data do
+      drop_index :droplet_guid, name: :kpack_lifecycle_droplet_guid_index
+    end
+  end
+end
--- a/lib/cloud_controller/opi/apps_client.rb
+++ b/lib/cloud_controller/opi/apps_client.rb
@@ -112,8 +112,8 @@ module OPI
       end
 
       def to_hash
-        command = if @process.command.presence
-                    ['/bin/sh', '-c', "#{CNB_LAUNCHER_PATH} #{@process.command}"]
+        command = if @process.started_command.presence
+                    ['/bin/sh', '-c', "#{CNB_LAUNCHER_PATH} #{@process.started_command}"]
                   else
                     []
                   end
--- a/lib/cloud_controller/opi/apps_client.rb
+++ b/lib/cloud_controller/opi/apps_client.rb
@@ -104,10 +104,35 @@ module OPI
       end
     end
 
+    class KpackLifecycle
+      CNB_LAUNCHER_PATH = '/cnb/lifecycle/launcher'.freeze
+
+      def initialize(process)
+        @process = process
+      end
+
+      def to_hash
+        command = if @process.command.presence
+                    ['/bin/sh', '-c', "#{CNB_LAUNCHER_PATH} #{@process.command}"]
+                  else
+                    []
+                  end
+        {
+          docker_lifecycle: {
+            command: command,
+            image: @process.desired_droplet.docker_receipt_image,
+          }
+        }
+      end
+    end
+
     def lifecycle_for(process)
-      if process.app.droplet.lifecycle_type == VCAP::CloudController::Lifecycles::DOCKER
+      case process.app.droplet.lifecycle_type
+      when VCAP::CloudController::Lifecycles::DOCKER
         DockerLifecycle.new(process)
-      elsif process.app.droplet.lifecycle_type == VCAP::CloudController::Lifecycles::BUILDPACK
+      when VCAP::CloudController::Lifecycles::KPACK
+        KpackLifecycle.new(process)
+      when VCAP::CloudController::Lifecycles::BUILDPACK
         BuildpackLifecycle.new(process)
       else
         raise("lifecycle type `#{process.app.lifecycle_type}` is invalid")
--- a/spec/unit/actions/droplet_create_spec.rb
+++ b/spec/unit/actions/droplet_create_spec.rb
@@ -220,5 +220,72 @@ module VCAP::CloudController
         end
       end
     end
+
+    describe '#create_kpack_droplet' do
+      let!(:kpack_lifecycle_data) { KpackLifecycleDataModel.make(build: build) }
+
+      it 'sets it on the droplet' do
+        expect {
+          droplet_create.create_kpack_droplet(build)
+        }.to change { [DropletModel.count, Event.count] }.by([1, 1])
+
+        droplet = DropletModel.last
+
+        expect(droplet.state).to eq(DropletModel::STAGING_STATE)
+        expect(droplet.app).to eq(app)
+        expect(droplet.package).to eq(package)
+        expect(droplet.build).to eq(build)
+        expect(droplet.kpack?).to be true
+
+        kpack_lifecycle_data.reload
+        expect(kpack_lifecycle_data.droplet).to eq(droplet)
+
+        event = Event.last
+        expect(event.type).to eq('audit.app.droplet.create')
+        expect(event.actor).to eq('schneider')
+        expect(event.actor_type).to eq('user')
+        expect(event.actor_name).to eq('bob@loblaw.com')
+        expect(event.actor_username).to eq('bobert')
+        expect(event.actee).to eq(app.guid)
+        expect(event.actee_type).to eq('app')
+        expect(event.actee_name).to eq(app.name)
+        expect(event.timestamp).to be
+        expect(event.space_guid).to eq(app.space_guid)
+        expect(event.organization_guid).to eq(app.space.organization.guid)
+        expect(event.metadata).to eq({
+          'droplet_guid' => droplet.guid,
+          'package_guid' => package.guid,
+        })
+      end
+
+      context 'when the build does not contain created_by fields' do
+        let(:build) do
+          BuildModel.make(
+            app: app,
+            package: package,
+          )
+        end
+
+        it 'sets the actor to UNKNOWN' do
+          expect {
+            droplet_create.create_kpack_droplet(build)
+          }.to change { [DropletModel.count, Event.count] }.by([1, 1])
+
+          droplet = DropletModel.last
+          expect(droplet.build).to eq(build)
+
+          event = Event.last
+          expect(event.type).to eq('audit.app.droplet.create')
+          expect(event.actor_type).to eq('user')
+          expect(event.actor).to eq('UNKNOWN')
+          expect(event.actor_name).to eq('')
+          expect(event.actor_username).to eq('')
+          expect(event.metadata).to eq({
+            'droplet_guid' => droplet.guid,
+            'package_guid' => package.guid,
+          })
+        end
+      end
+    end
   end
 end
--- a/app/actions/droplet_update.rb
+++ b/app/actions/droplet_update.rb
@@ -9,7 +9,7 @@ module VCAP::CloudController
 
         if message.requested?(:image)
           raise InvalidDroplet.new('Droplet image can only be updated on staged droplets') unless droplet.staged?
-          raise InvalidDroplet.new('Images can only be updated for docker droplets') unless droplet.docker?
+          raise InvalidDroplet.new('Images can only be updated for docker droplets') unless droplet.has_docker_image?
 
           droplet.docker_receipt_image = message.image
         end
--- a/app/models/runtime/droplet_model.rb
+++ b/app/models/runtime/droplet_model.rb
@@ -31,10 +31,15 @@ module VCAP::CloudController
       class: 'VCAP::CloudController::BuildpackLifecycleDataModel',
       key: :droplet_guid,
       primary_key: :guid
+    one_to_one :kpack_lifecycle_data,
+      class: 'VCAP::CloudController::KpackLifecycleDataModel',
+      key: :droplet_guid,
+      primary_key: :guid
     one_to_many :labels, class: 'VCAP::CloudController::DropletLabelModel', key: :resource_guid, primary_key: :guid
     one_to_many :annotations, class: 'VCAP::CloudController::DropletAnnotationModel', key: :resource_guid, primary_key: :guid
 
     add_association_dependencies buildpack_lifecycle_data: :destroy
+    add_association_dependencies kpack_lifecycle_data: :destroy
     add_association_dependencies labels: :destroy
     add_association_dependencies annotations: :destroy
 
@@ -82,6 +87,14 @@ module VCAP::CloudController
 
     def docker?
       lifecycle_type == DockerLifecycleDataModel::LIFECYCLE_TYPE
+    end
+
+    def kpack?
+      lifecycle_type == KpackLifecycleDataModel::LIFECYCLE_TYPE
+    end
+
+    def has_docker_image?
+      docker? || kpack?
     end
 
     def docker_ports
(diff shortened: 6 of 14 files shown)


db/migrations/20201020182259_backfill_kpack_lifecycle_data_association_to_droplets.rb
```diff
@@ -0,0 +1,18 @@
+Sequel.migration do
```
tcdowney, line 1: I hate making this sort of migration -- lots of potential for performance and data integrity issues -- so would appreciate a close look!

🕵️

db/migrations/20201020182259_backfill_kpack_lifecycle_data_association_to_droplets.rb
```diff
@@ -0,0 +1,18 @@
...
+  up do
+    unassociated_kpack_lifecycle_data = self[:kpack_lifecycle_data].select(:id, :build_guid).where(droplet_guid: nil).exclude(build_guid: nil)
+
+    unassociated_kpack_lifecycle_data.each do |kld|
+      droplet = self[:droplets].select(:guid).where(build_guid: kld[:build_guid]).first
+
+      # there may not have been a droplet created yet for the build
```
tcdowney, line 8: I think there is an unlikely (?) race condition here where the `ccdb-migrate` job runs this but there are still `cf-api-servers` out there with the old code making "docker" droplets. If there are any apps in this state, they won't be able to get stack updates until the app is repushed.

I'm not sure if there is a way around that apart from either:
1. Reverting https://github.com/cloudfoundry/cloud_controller_ng/pull/1917/commits/91769482589608ed0ae664da9e8b33b70c4f29ff and always supporting the ability to update image for "docker" and "kpack" droplets.
2. Running the migration again as part of a future release.

lib/cloud_controller/opi/apps_client.rb
```diff
@@ -104,10 +104,35 @@ def to_hash
...
+      end
+
+      def to_hash
+        command = if @process.command.presence
+                    ['/bin/sh', '-c', CNB_LAUNCHER_PATH, @process.command]
+                  else
+                    []
+                  end
+        {
+          docker_lifecycle: {
```
jspawar, line 121: I suspect this was intentional and fine, but for my knowledge, why does this remain `docker_lifecycle` instead of something new like `kpack_lifecycle`? Is this because Eirini itself is looking for specific information under that key when it desires a LRP?

lib/cloud_controller/opi/apps_client.rb
```diff
@@ -104,10 +104,35 @@ def to_hash
...
+    class KpackLifecycle
+      CNB_LAUNCHER_PATH = '/cnb/lifecycle/launcher'.freeze
+
+      def initialize(process)
+        @process = process
+      end
+
+      def to_hash
+        command = if @process.command.presence
+                    ['/bin/sh', '-c', CNB_LAUNCHER_PATH, @process.command]
```
tcdowney, line 116: I think this command should actually be:
```ruby
['/bin/sh', '-c', "#{CNB_LAUNCHER_PATH} #{@process.command}"]
```

where the launcher and specified command are a single string that's passed to the shell.

The app I had tested this with was working, but I think it may have been running the images default entrypoint? I don't know... a bit confused by that. Anyways going to add a commit for this.

tcdowney: https://github.com/cloudfoundry/cloud_controller_ng/pull/1917/commits/6a32abbd93ca77604f2657347cdb6522b647f857

lib/cloud_controller/opi/apps_client.rb
```diff
@@ -104,10 +104,35 @@ def to_hash
...
 
+    class KpackLifecycle
+      CNB_LAUNCHER_PATH = '/cnb/lifecycle/launcher'.freeze
+
+      def initialize(process)
+        @process = process
+      end
+
+      def to_hash
+        command = if @process.command.presence
```
tcdowney, line 115: Does anyone know why we use `@process.command` here instead of `@process.started_command`?

https://github.com/cloudfoundry/cloud_controller_ng/blob/bd3d96ca9a959936c6f655049e155b3edc7a7c8b/lib/cloud_controller/diego/buildpack/lifecycle_protocol.rb#L74

I'm having trouble understanding how this picks of start commands for other processes that are defined in the `Procfile`.

lib/cloud_controller/opi/apps_client.rb
tcdowney: Yeah, exactly.

Eirini is looking for either a `DockerLifecycle` or a `BuildpackLifecycle` (which I guess is only used in `kube-cf`):
https://github.com/cloudfoundry-incubator/eirini/blob/ee87b6612c6832f26a42a55ee309c222eee36bcb/bifrost/convert.go#L140

db/migrations/20201020182259_backfill_kpack_lifecycle_data_association_to_droplets.rb
```diff
@@ -0,0 +1,18 @@
...
+      next if droplet.nil?
```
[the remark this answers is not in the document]
tcdowney: Yeah I think that's the same as what I mentioned in this comment:
https://github.com/cloudfoundry/cloud_controller_ng/pull/1917#discussion_r508857218

I'm not sure if there is a way around it apart from either:

* Reverting 9176948 and always supporting the ability to update image for "docker" and "kpack" droplets.
* Running the migration again as part of a future release (this requires people to upgrade to this version first -- if they jump and skip this version then same problem)
* A Kubernetes job that runs on every cf-for-k8s upgrade and migrates them which we eventually delete? 🤔
* We accept it as an edge case and document the workaround (repush) if stack updates fail for an app

db/migrations/20201020182259_backfill_kpack_lifecycle_data_association_to_droplets.rb
tcdowney: Resolving this, let's move discussion to this thread:
https://github.com/cloudfoundry/cloud_controller_ng/pull/1917#discussion_r508889764

db/migrations/20201020182259_backfill_kpack_lifecycle_data_association_to_droplets.rb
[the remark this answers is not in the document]
cwlbraa: > I think there is an unlikely (?) race condition here where the ccdb-migrate job runs this but there are still cf-api-servers out there with the old code making "docker" droplets. If there are any apps in this state, they won't be able to get stack updates until the app is repushed (repushing is a fairly easy mitigation IMO).

> Scenario I'm imagining would require one cf-api-server pod to rollout with new changes and one to remain with old changes. cf push gets its request routed to the old one => droplet eventually created without associated kpack_lifecycle_data

This is almost certainly possible, and was not something I had considered when Tim and I talked about this earlier. One thing that we've never been able to build on BOSH that would totally fix this type of deploy scenario is a post-deploy migration job in addition to the pre-deploy one. If we had that running its own, idempotent post-deploy migrations, we could clean up after ourselves in scenarios like this and maybe even use it implement proper [expand-and-contract parallel-change migrations](https://martinfowler.com/articles/evodb.html).

> We accept it as an edge case and document the workaround (repush) if stack updates fail for an app

I think we're still early enough in the life of cf-for-k8s that this is acceptable, particularly if you consider this relative to how broken other things are.

lib/cloud_controller/opi/apps_client.rb
tcdowney: Heh when I switch to `process.started_command` it doesn't work cause we run into a new bug. I don't think our build reconciler is copying over detected start commands correctly.

https://github.com/cloudfoundry/capi-k8s-release/issues/88

db/migrations/20201020182259_backfill_kpack_lifecycle_data_association_to_droplets.rb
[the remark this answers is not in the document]
tcdowney: > I think we're still early enough in the life of cf-for-k8s that this is acceptable, particularly if you consider this relative to how broken other things are

I think it may be cheaper to just leave this in now versus doing it later and risking the context being lost. 🤷 

Cause of issues like https://github.com/cloudfoundry/capi-k8s-release/issues/88 where we're literally missing data on some of the droplets, I think folks may have to repush anyway.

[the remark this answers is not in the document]
tcdowney: Actually, as I think about this more... we might be able to use the fact that historical kpack-built droplets are "docker" to help resolve that issue (https://github.com/cloudfoundry/capi-k8s-release/issues/88).

It will be difficult to know when it is safe for a "kpack" droplet to use `@process.started_command` if some "kpack" droplets have the incomplete command, but if we fix that issue at the same time as we introduce "kpack" droplets then we can assume that all of them have complete detected start commands. 🤔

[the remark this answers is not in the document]
cwlbraa: > It will be difficult to know when it is safe for a "kpack" droplet to use @process.started_command if some "kpack" droplets have the incomplete command, but if we fix that issue at the same time as we introduce "kpack" droplets then we can assume that all of them have complete detected start commands. 🤔

This is a good point. It's a kinda risky two-birds-with-one-stone way of fixing these bugs, but it sounds like it'd work.

Effectively what you're saying is _dont_ turn any docker droplets into kpack droplets. Instead, require repushing. 

This way, we'll know that kpack droplets have usable start commands and the migration won't break any apps that just happen to work in the current, weird state. Otherwise, switching to use `started_command` could break apps on migration because we weren't using them before and were likely saving the wrong data.

lib/cloud_controller/opi/apps_client.rb
jspawar: Should we teach eirini to speak `KpackLifecycle`? Can merge this and revisit after if we really want to

db/migrations/20201020182259_backfill_kpack_lifecycle_data_association_to_droplets.rb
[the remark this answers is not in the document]
tcdowney: > Otherwise, switching to use started_command could break apps on migration because we weren't using them before and were likely saving the wrong data.

Yep, when I tried doing this in this branch `dora` would just run `bundle` and install gems instead of `bundle rackup`.

lib/cloud_controller/opi/apps_client.rb
tcdowney: IMO we shouldn't until there is a reason to. I feel like at this point if Eirini doesn't need to distinguish between an app using a "regular" Docker image and an app using a kpack-built Docker image it shouldn't have to.

lib/cloud_controller/opi/apps_client.rb
tcdowney: I put out a PR to fix the detected start commands we're saving on droplets:
https://github.com/cloudfoundry/capi-k8s-release/pull/90

Once that is merged I can revert the "docker" -> "kpack" migration and switch this over to using `@process.started_command`.

cwlbraa (changes requested):
This looks good, but given what we've learned about the current state of PATCH droplet.process_types, the migration to turn docker droplets into kpack droplets and then use those seems likely to break apps that just happen to work currently.

spec/unit/actions/droplet_create_spec.rb
```diff
@@ -220,5 +220,72 @@ module VCAP::CloudController
...
+
+      context 'when the build does not contain created_by fields' do
+        let(:build) do
+          BuildModel.make(
+            app: app,
+            package: package,
+          )
+        end
+
+        it 'sets the actor to UNKNOWN' do
```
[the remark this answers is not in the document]
tcdowney: existing behavior 👻

app/actions/droplet_create.rb
```diff
@@ -63,6 +63,20 @@ def create_buildpack_droplet(build)
...
 
+    def create_kpack_droplet(build)
+      droplet = droplet_from_build(build)
+
+      DropletModel.db.transaction do
+        droplet.save
+        droplet.kpack_lifecycle_data = build.kpack_lifecycle_data
+      end
+
+      droplet.reload
```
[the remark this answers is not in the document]
tcdowney: > assign the lifecycle data... why does this happen after save?

The droplet in this case hasn't been persisted yet (and I think doesn't have a `guid`) until it's saved. It's kinda tricky, but `droplet.kpack_lifecycle_data = build.kpack_lifecycle_data` actually sets the `kpack_lifecycle_data.droplet_guid` field in a different table.

The `droplet.reload` I assume is for the events... I actually just saw that the regular buildpack droplet create was doing this and figured it was important for that.

db/migrations/20201020182259_backfill_kpack_lifecycle_data_association_to_droplets.rb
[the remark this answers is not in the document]
tcdowney: I rebased on master and removed the migration as part of that. It's no longer present in this commit:
https://github.com/cloudfoundry/cloud_controller_ng/pull/1917/commits/87d0369f72ccd411b6ce2551d42105b1d9e6a48a

db/migrations/20201020182259_backfill_kpack_lifecycle_data_association_to_droplets.rb
tcdowney: This migration is now gone. Resolving.

lib/cloud_controller/opi/apps_client.rb
tcdowney: That's merged, so I switched this over:
https://github.com/cloudfoundry/cloud_controller_ng/pull/1917/commits/0a55501093d6a10472274f8bc134dbf4bbe60f9f

tcdowney:
The unit test failures are unrelated. It looks like these have been failing on Github Actions on other PRs (with unrelated changes) as well. :/
E.g. https://github.com/cloudfoundry/cloud_controller_ng/pull/1920/checks?check_run_id=1287920643

db/migrations/20201020204056_index_kpack_lifecycle_data_droplet_guid.rb
```diff
@@ -0,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table :kpack_lifecycle_data do
+      add_index :droplet_guid, name: :kpack_lifecycle_droplet_guid_index
```
cwlbraa, line 4: This strikes me as an obvious improvement, but I've never thought about indices on `lifecycle_data` tables prior to now. It seems likely that we've missed some other indices on this table in the past `lifecycle_data` links itself to all sorts of other objects...

lib/cloud_controller/opi/apps_client.rb
```diff
@@ -104,10 +104,35 @@ def to_hash
...
+    class KpackLifecycle
+      CNB_LAUNCHER_PATH = '/cnb/lifecycle/launcher'.freeze
+
+      def initialize(process)
+        @process = process
+      end
+
+      def to_hash
+        command = if @process.started_command.presence
+                    ['/bin/sh', '-c', "#{CNB_LAUNCHER_PATH} #{@process.started_command}"]
```
cwlbraa, line 116: 🤔  is the `bin/sh` necessary here? it's safe to leave it in but it's kinda interesting to think about.

app/actions/droplet_create.rb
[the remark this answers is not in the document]
cwlbraa: > droplet.kpack_lifecycle_data = build.kpack_lifecycle_data actually sets the kpack_lifecycle_data.droplet_guid field in a different table.

lol this is definitely why people don't like ORMs

cwlbraa (approved):
lgtm

db/migrations/20201020204056_index_kpack_lifecycle_data_droplet_guid.rb
tcdowney: I put a little blurb in the commit message about this one:

```
- add index to kpack_lifecycle_data.droplet_guid
  - this is important since droplets find their associated lifecycle
  through this column
  - a similar index exists on buildpack_lifecycle_data
```

I think it's important if we want the `droplet.kpack?` type of methods to be performant.

lib/cloud_controller/opi/apps_client.rb
tcdowney: Yeah, I thought that as well. I went with this cause it's what we were doing for docker droplets, but 🤷 .
https://github.com/cloudfoundry/cloud_controller_ng/pull/1417/files#diff-738bfbb749a5827ff834ca07b3594487e3f321066cf64d20ff4bdf7359137f4bR74

I can ask the Eirini folks.

jspawar (approved):
ya lgtm

PR #1921, opened 2020-10-21
Add basic Honeycomb instrumentation

njbennett:
When `honeycomb.write_key` is present, emit events for every incoming
HTTP request, except to the `/healthz` endpoint.

Authored-by: Nat Bennett <nbennett@pivotal.io>


config/initializers/honeycomb.rb
```diff
@@ -0,0 +1,39 @@
+require 'honeycomb-beeline'
+
+module CCInitializers
```
cwlbraa, line 3: what's up with the module here? does this run at process startup or something?

tcdowney:
Hmm, it looks like the postgres tests are failing with errors like this:

```
An error occurred while loading ./spec/acceptance/async_bindings_spec.rb.
Failure/Error: CCInitializers.send(method, @config_hash)
```

matt-royal:
It looks like the `honeycomb-beeline` gem is Apache 2 licensed, so there shouldn't be any licensing issues with adding this gem.

config/initializers/honeycomb.rb
njbennett: Yeah, this is our version of a standard Rails pattern. The NewRelic integration has a slightly more involved version where it only runs the initializers at all if NewRelic is enabled.

njbennett:
Force pushed some changes to the commit. Now we check only for the top-level honeycomb key, rather than a sub-key, because the call to get the subkey value with [] fails when the top level key returns nil.

lib/cloud_controller/rack_app_builder.rb
```diff
@@ -21,6 +21,7 @@ def build(config, request_metrics)
         use CloudFoundry::Middleware::Cors, config.get(:allowed_cors_domains)
         use CloudFoundry::Middleware::VcapRequestId
         use CloudFoundry::Middleware::NewRelicCustomAttributes if config.get(:newrelic_enabled)
+        use Honeycomb::Rack::Middleware, client: Honeycomb.client if config.get(:honeycomb, :write_key)
```
cwlbraa, line 24: i feel like this ought to also be conditional on `config.get(:honeycomb)`

njbennett: Yes. Updated.

tcdowney (approved):
LGTM apart from the CLA bit 🤔

njbennett:
Github not being aware that I've signed the CLA doesn't seem to stop me from merging things so 😬

PR #1933, opened 2020-10-29
test: combine methods of testing metadata

blgm:
Removes to previous patterns for testing metadata and replaces them
with a new, more fluent method.

[#173082202](https://www.pivotaltracker.com/story/show/173082202)


spec/support/matchers/have_annotations.rb
```diff
@@ -0,0 +1,43 @@
...
+      expected = expected.map do |annotation|
+        {
+          prefix: annotation.with_indifferent_access[:prefix],
+          key: annotation.with_indifferent_access[:key],
+          value: annotation.with_indifferent_access[:value],
+        }
+      end
+
+      expect(actual_annotations).to match_array(expected)
+    else
```
pivotal-marcela-campo, line 23: can these two paths be tested by match_array? if expected is empty then I would assume `expect(actual_annotations).to match_array([])` only passes if actual_annotations is empty. And also would assume that `map` on an empty array would yield an empty array as well. so the whole if/else could be simplified to just one path.

spec/support/matchers/have_annotations.rb
```diff
@@ -0,0 +1,43 @@
...
+      "Annotations not expected: got #{actual_annotations}"
+    end
+  end
+
+  failure_message_when_negated do
+    if expected.any?
+      "Annotations unexpectedly match: got #{actual_annotations}"
+    else
+      "Annotations found when not expected: got #{actual_annotations}"
```
pivotal-marcela-campo, line 40: not have_annotations([]) seems like the message should actually be "Expected annotations but got []" if I am understanding correctly the usage of these method in the matchers design.

spec/support/matchers/have_labels.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  failure_message_when_negated do
+    if expected.any?
+      "Labels unexpectedly match: got #{actual_labels}"
+    else
+      "Labels found when not expected: got #{actual_labels}"
+    end
+  end
+end
```
pivotal-marcela-campo, line 43: same feedback as for annotations

pivotal-marcela-campo (changes requested):

spec/support/matchers/have_annotations.rb
blgm: I think this is a Ruby-ism because you can write:
```ruby
expect(thing).to have_annotations
```
Which you would expect to *pass* when there are any annotations. I wasn't able to differentiate between:
```ruby
expect(thing).to have_annotations
expect(thing).to have_annotations([])
```

The assumption is that if you want to assert on there being no annotations then you would write
```ruby
expect(thing).not_to have_annotations
```

That's the reason it appears unusual, and I wasn't able to find a better way. However I can at least comment.

blgm: I've improved the messages and added a comment, so hopefully future maintainers will be clear on the original intention

pivotal-marcela-campo (approved):

spec/support/matchers/have_annotations.rb
pivotal-marcela-campo: Thanks for the explanation on the usage of any? on the matcher context. The message is fine then.

PR #1938, opened 2020-10-30
Upgrade middleman et al. for v3 docs

Gerg:
Probably not ready to merge yet.

Related to https://github.com/cloudfoundry/cloud_controller_ng/pull/1743 and https://github.com/cloudfoundry/cloud_controller_ng/pull/1782.

Partially based on https://github.com/slatedocs/slate/commit/7f88aafcba83e6125ae7496495eccae3fb35f4d1

I was able to run the following:
- `npm start` / `gulp webserver` / `gulp` / `bundle exec middleman server`
- `gulp build` / `bundle exec middleman build`
- `gulp checkdocs` / `gulp checkV2docs` / `gulp checkV3docs`


docs/v3/source/includes/api_resources/_security_groups.erb
```diff
@@ -41,12 +41,12 @@
 }
 <% end %>
 
-<% content_for :paginated_security_group, File.read(__dir__ + '/examples/security_group_paginated.json') %>
+<% content_for :paginated_security_group, File.read('source/includes/api_resources/examples/security_group_paginated.json') %>
```
Gerg, line 44: These were probably the weirdest change needed. It seems like the rendering context for templates is different in middleman 4, so `__dir__` doesn't return the correct path.

docs/v3/config.rb
```diff
@@ -15,21 +15,25 @@
 set :images_dir, 'images'
 set :fonts_dir, 'fonts'
 
+ignore 'includes/api_resources/examples/*.json'
```
Gerg, line 18: These json files are consumed with rendering templates, so don't need to be included in built resources.

reid47 (approved):
love it

docs/v3/config.rb
```diff
@@ -15,21 +15,25 @@
...
+
 # Activate the syntax highlighter
 activate :syntax
 
+activate :sprockets
+
+activate :autoprefixer do |config|
+  config.browsers = ['last 2 version', 'Firefox ESR']
```
tcdowney, line 26: What's this do.

tcdowney (approved):
lgtm

docs/v3/config.rb
reid47: activates the sprockets

reid47: and autoprefixer will ensure your CSS is compatible with older browsers (or whichever browsers you configure here). See https://autoprefixer.github.io/

Gerg: Sprockets is a ruby asset pipeline. It compiles all the js-like and css-like stuff and shoves it into your html in a suitable manor. I think it was always used behind the scenes in middleman, but in 4+ you need to manually activate it or the docs lose style and scripts.

The autoprefixer blocked moved from the :build block below. I'm assuming it adds all the needed browser-specific css prefixes.

tcdowney: I mean specifically the `'last 2 version'` 😛

reid47: For those following along at home, see https://github.com/browserslist/browserslist – this configuration syntax is shared by autoprefixer and lots of front-end-y tools

PR #1945, opened 2020-11-05
V3 add audit events for route bindings 175228173

FelisiaM:


app/actions/service_credential_binding_create.rb
```diff
@@ -84,7 +84,7 @@ def all_space_guids(service_instance)
       end
 
       def event_repository
-        Repositories::ServiceBindingEventRepository
+        Repositories::ServiceGenericBindingEventRepository.new('service_binding')
```
kirederik, line 87: Should the parameter be a constant defined somewhere in the `ServiceGenericBindingEventRepository`? I can see a world where we decide to update it to something different and have to remember to change in multiple places 🤔

app/actions/service_credential_binding_delete.rb
```diff
@@ -21,7 +21,7 @@ def perform_start_delete_actions(binding)
       end
 
       def event_repository
-        Repositories::ServiceBindingEventRepository
+        Repositories::ServiceGenericBindingEventRepository.new('service_binding')
```
kirederik, line 24: Everytime this method is called a new `ServiceGenericBindingEventRepository` will be created. It seems it's only being called once in the action, so not a big deal, but may be interesting to cache this one?

app/actions/service_route_binding_delete.rb
```diff
@@ -4,27 +4,29 @@
...
         binding.notify_diego
       end
 
-      def record_audit_event(binding)
-        service_event_repository.record_service_instance_event(
-          :unbind_route,
-          binding.service_instance,
-          { route_guid: binding.route.guid },
-        )
+      def perform_start_delete_actions(binding)
```
kirederik, line 28: I found this rename slightly obscure... `record_audit_event` is a little more explicit than `perform_start_delete_actions`. The latter makes me believe there are some things we need to do before deleting when in fact we are only logging the audit event. 

I can see how this is better in terms of symmetry with the `perform_delete_actions` though.

kirederik (approved):

app/actions/service_route_binding_delete.rb
FelisiaM: The idea here is that every binding can choose to do whatever they need as part of recording an async operation. It just happens to be just auditing for now. 
I thought this naming would keep it open to extentions

PR #1958, opened 2020-11-16
Validate POST request for key type credential bindings

FelisiaM:
Extracted the Create service credential binding message into two specific one for app and key type

This completes scenario 1 to 5 of 
[#174145595](https://www.pivotaltracker.com/story/show/174145595)

### DIFF
--- a/app/actions/service_credential_binding_key_create.rb
+++ b/app/actions/service_credential_binding_key_create.rb
@@ -1,0 +1,41 @@
+module VCAP::CloudController
+  module V3
+    class ServiceCredentialBindingKeyCreate
+      class UnprocessableCreate < StandardError
+      end
+
+      def precursor(service_instance, volume_mount_services_enabled: false)
+        if service_instance.managed_instance?
+          service_not_bindable! unless service_instance.service_plan.bindable?
+          service_not_available! unless service_instance.service_plan.active?
+          volume_mount_not_enabled! if service_instance.volume_service? && !volume_mount_services_enabled
+          operation_in_progress! if service_instance.operation_in_progress?
+        else
+          key_not_supported_for_user_provided_service!
+        end
+      end
+
+      private
+
+      def key_not_supported_for_user_provided_service!
+        raise UnprocessableCreate.new("Service credential bindings of type 'key' are not supported for user-provided service instances.")
+      end
+
+      def operation_in_progress!
+        raise UnprocessableCreate.new('There is an operation in progress for the service instance.')
+      end
+
+      def service_not_bindable!
+        raise UnprocessableCreate.new('Service plan does not allow bindings.')
+      end
+
+      def service_not_available!
+        raise UnprocessableCreate.new('Service plan is not available.')
+      end
+
+      def volume_mount_not_enabled!
+        raise UnprocessableCreate.new('Support for volume mount services is disabled.')
+      end
+    end
+  end
+end
--- a/app/actions/service_credential_binding_key_create.rb
+++ b/app/actions/service_credential_binding_key_create.rb
@@ -23,7 +23,6 @@ module VCAP::CloudController
       private
 
       def validate!(service_instance)
-
         if service_instance.managed_instance?
           service_not_bindable! unless service_instance.service_plan.bindable?
           service_not_available! unless service_instance.service_plan.active?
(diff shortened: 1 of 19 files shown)


app/messages/service_credential_app_binding_create_message.rb
```diff
@@ -0,0 +1,20 @@
+module VCAP::CloudController
+  class ServiceCredentialAppBindingCreateMessage < ServiceCredentialBindingCreateMessage
```
pivotal-marcela-campo, line 2: I find this a bit confusing. The base class validates that the type is either 'app' or 'key' and then the subclass can only be one of those types. Also not a big fan of having `if type==app` scattered throughout the controller. I am wondering if we just should create a simple message to extract the type and then take completely different paths depending on that. So we can create a specific message and do any validations/action calls that we need from them on thinking just on one kind of resource. There is no much reuse of code provided by the inheritance in the messages to actually care too much for it.

app/messages/service_credential_key_binding_create_message.rb
```diff
@@ -0,0 +1,13 @@
+module VCAP::CloudController
+  class ServiceCredentialKeyBindingCreateMessage < ServiceCredentialBindingCreateMessage
+    validates :name, string: true, presence: true
+
+    def relationships_message
+      @relationships_message ||= Relationships.new(relationships&.deep_symbolize_keys)
+    end
+
+    class Relationships < ServiceCredentialBindingCreateMessage::Relationships
+      validates_with NoAdditionalKeysValidator
```
pivotal-marcela-campo, line 10: I don't think we gain much from inheriting here. For me it makes the code more difficult to follow and we are just reusing a couple of configuration lines.

pivotal-marcela-campo (approved):

app/messages/service_credential_app_binding_create_message.rb
FelisiaM: The reason I like the reduplication here is not about the code repetition but because these messages should always be kept in sync and are represented as the same thing to API client. E.g a future change on SI relationship should affect both key and app type, hence it makes sense to keep the SI extracted. 
I think this is not the perfect pattern, but it is following the same pattern as the service instance create message and I think at this point in the project it might be more beneficial to have similar patterns between al endpoint.

app/actions/service_credential_binding_key_create.rb
```diff
@@ -0,0 +1,41 @@
+module VCAP::CloudController
+  module V3
+    class ServiceCredentialBindingKeyCreate
+      class UnprocessableCreate < StandardError
+      end
+
+      def precursor(service_instance, volume_mount_services_enabled: false)
+        if service_instance.managed_instance?
```
pivotal-marcela-campo, line 27: I'd like more sth like 
```
key_not_supported_for_user_provided_service! unless service_instance.managed_instance?

service_not_bindable! unless service_instance.service_plan.bindable?...
...
```
I find it more readable

app/actions/service_credential_binding_key_create.rb
```diff
@@ -0,0 +1,41 @@
...
+          service_not_bindable! unless service_instance.service_plan.bindable?
+          service_not_available! unless service_instance.service_plan.active?
+          volume_mount_not_enabled! if service_instance.volume_service? && !volume_mount_services_enabled
```
pivotal-marcela-campo: Is this validation needed here? It is not in the AC and not tested either

app/actions/service_credential_binding_key_create.rb
```diff
@@ -4,19 +4,35 @@ class ServiceCredentialBindingKeyCreate
...
+      def precursor(service_instance, name)
+        validate!(service_instance)
+
+        binding_details = {
+          service_instance: service_instance,
+          name: name,
+          credentials: {}
+        }
+
+        ServiceKey.db.transaction do
```
pivotal-marcela-campo, line 16: should we do this in the model to keep it closer to what we do for credential bindings?

app/actions/service_credential_binding_key_create.rb
pivotal-marcela-campo: I see this was fixed in another commit

app/actions/service_credential_binding_key_create.rb
pivotal-marcela-campo: We'll refactor this later

app/actions/service_credential_binding_key_create.rb
pivotal-marcela-campo: We'll refactor this later

PR #1960, opened 2020-11-17
Remove app apply manifest endpoint

Gerg:
- Endpoint was experimental and has been replaced by the space-level
equivalent
- Remove corresponding docs
- Merge named_app_manifest_message into app_manifest_message since there
is no longer a way to apply a manifest without app names


sweinstein22:
Looks good to me!

PR #1961, opened 2020-11-18
v3(services): Only update expected binding fields when creating a

pivotal-marcela-campo:
binding

OSBAPI spec outlines the broker expected response for a create binding
and indicates that any other field return by brokers should be ignored.
v2 has this behaviour but was missing in v3 and was causing failures
when the broker response includes attributes that are not part of the CC
DB. This is so we don't try to update any other fields.

[#175662475](https://www.pivotaltracker.com/story/show/175662475)

### DIFF
--- a/spec/request/service_credential_bindings_spec.rb
+++ b/spec/request/service_credential_bindings_spec.rb
@@ -1383,17 +1383,8 @@ RSpec.describe 'v3 service credential bindings' do
                   syslog_drain_url: syslog_drain_url,
                   credentials: credentials,
                   parameters: parameters,
-                  service_id:'extra-field-service_id',
-                  plan_id:'extra-field-plan_id',
-                  name: 'extra-field-name',
-                  bind_resource:{
-                    app_guid:'extra-field',
-                    space_guid:'extra-field'
-                  },
-                  context:
-                    {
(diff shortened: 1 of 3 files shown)


spec/request/service_credential_bindings_spec.rb
```diff
@@ -1392,7 +1392,18 @@ def check_filtered_bindings(*bindings)
...
+                plan_id:'extra-field-plan_id',
+                name: 'extra-field-name',
+                bind_resource:{
+                  app_guid:'extra-field',
+                  space_guid:'extra-field'
+                },
+                context:
+                  {
+                    platform:'extra-field',
+                  }
```
FelisiaM, line 1406: I think for the purpose of this test all we need is just an extra field `non_osbapi_param: 'foo'` even though having all the other ones doesn't harm

FelisiaM (approved):

spec/request/service_credential_bindings_spec.rb
pivotal-marcela-campo: I've changed it to just two fields: one that doesn't exist in our side so we verify it doesn't raise, and another one that does exist but we are not supposed to update to verify we also ignore those.

PR #1976, opened 2020-11-26
V3 create key binding sync 174147679

pivotal-marcela-campo:
[#174147679](https://www.pivotaltracker.com/story/show/174147679)

### DIFF
--- a/app/actions/v3/service_binding_create.rb
+++ b/app/actions/v3/service_binding_create.rb
@@ -74,13 +74,12 @@ module VCAP::CloudController
           }
         )
 
-        notify(binding)
+        post_bind_action(binding)
 
         event_repository.record_create(
           binding,
           @user_audit_info,
-          @audit_hash,
-          manifest_triggered: false
+          @audit_hash
         )
       end
 
@@ -105,10 +104,10 @@ module VCAP::CloudController
             broker_provided_operation: broker_operation
           }
         )
-        event_repository.record_start_create(binding, @user_audit_info, @audit_hash, manifest_triggered: false)
+        event_repository.record_start_create(binding, @user_audit_info, @audit_hash)
       end
 
-      def notify(binding); end
+      def post_bind_action(binding); end
 
       def bindings_retrievable?(binding)
         binding.service_instance.service.bindings_retrievable
--- a/app/actions/v3/service_binding_create.rb
+++ b/app/actions/v3/service_binding_create.rb
@@ -64,6 +64,26 @@ module VCAP::CloudController
         )
       end
 
+      def complete_binding_and_save(binding, binding_details, last_operation)
+        binding.save_with_attributes_and_new_operation(
+          binding_details.symbolize_keys.slice(*permitted_binding_attributes),
+          {
+            type: 'create',
+            state: last_operation[:state],
+            description: last_operation[:description],
+          }
+        )
+
+        notify(binding)
+
+        event_repository.record_create(
+          binding,
+          @user_audit_info,
+          @audit_hash,
+          manifest_triggered: false
+        )
+      end
+
       def save_last_operation(binding, details)
         binding.save_with_attributes_and_new_operation(
           {},
@@ -75,6 +95,20 @@ module VCAP::CloudController
           }
         )
       end
+
+      def save_incomplete_binding(binding, broker_operation)
+        binding.save_with_attributes_and_new_operation(
+          {},
+          {
+            type: 'create',
+            state: 'in progress',
+            broker_provided_operation: broker_operation
+          }
+        )
+        event_repository.record_start_create(binding, @user_audit_info, @audit_hash, manifest_triggered: false)
+      end
+
+      def notify(binding); end
 
       def bindings_retrievable?(binding)
         binding.service_instance.service.bindings_retrievable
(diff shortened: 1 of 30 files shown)


app/controllers/v3/service_credential_bindings_controller.rb
```diff
@@ -120,13 +120,16 @@ def parameters
...
 
-    head :not_implemented
-    return
+    case service_instance
+    when ManagedServiceInstance
+      pollable_job_guid = enqueue_key_job(binding.guid, message)
+      head :accepted, 'Location' => url_builder.build_url(path: "/v3/jobs/#{pollable_job_guid}")
+    else
+      head :not_implemented
+    end
```
FelisiaM, line 132: Keys cannot be created for UPSI (the precursor will return error on validation) so there is no need for the switch here

app/models/services/service_key.rb
```diff
@@ -79,8 +81,44 @@ def logger
...
-      false
+      !!service_key_operation && service_key_operation.state == 'in progress'
+    end
+
+    def required_parameters
+      nil
+    end
+
+    def save_with_attributes_and_new_operation(attributes, operation)
+      save_with_new_operation(operation, attributes: attributes)
```
FelisiaM, line 102: I think `save_with_attributes_and_new_operation ` is the standard method that routes and service bindings are using from the actions. Might be worth having only that one and inlining `save_with_new_operation` (which is inherited from the initial implementation of service bindings)

app/controllers/v3/service_credential_bindings_controller.rb
pivotal-marcela-campo: Thanks for spotting this!

app/models/services/service_key.rb
pivotal-marcela-campo: we are currently using both for some reason throughout the bindings actors/actions. I'll add a chore to take a look after this.

app/actions/v3/service_binding_create.rb
```diff
@@ -64,6 +64,26 @@ def save_failed_state(binding, e)
...
+        binding.save_with_attributes_and_new_operation(
+          binding_details.symbolize_keys.slice(*permitted_binding_attributes),
+          {
+            type: 'create',
+            state: last_operation[:state],
+            description: last_operation[:description],
+          }
+        )
+
+        notify(binding)
```
FelisiaM, line 77: I'm wondering if there is a better name for this that would make it even more decoupled from the specific action - but I cannot think of one. Maybe just `notify_diego` might be at least more descriptive

FelisiaM: maybe something like "post_bind_action" is generic enough to express that this action can be specific and different for each binding

app/actions/v3/service_binding_create.rb
```diff
@@ -76,6 +96,20 @@ def save_last_operation(binding, details)
...
+      def save_incomplete_binding(binding, broker_operation)
+        binding.save_with_attributes_and_new_operation(
+          {},
+          {
+            type: 'create',
+            state: 'in progress',
+            broker_provided_operation: broker_operation
+          }
+        )
+        event_repository.record_start_create(binding, @user_audit_info, @audit_hash, manifest_triggered: false)
```
FelisiaM, line 108: nit: not manifest_triggered is by default false so no need to define it here (and in complete_binding_and_save )

app/actions/v3/service_binding_create.rb
pivotal-marcela-campo: yeah, I didn't want to mention `diego` at all. I think `post_bind_action` is a good enough hook name, much better than notify.

app/controllers/v3/service_credential_bindings_controller.rb
```diff
@@ -181,6 +179,18 @@ def enqueue_bind_job(binding_guid, message)
     pollable_job.guid
   end
 
+  def enqueue_key_job(binding_guid, message)
+    bind_job = VCAP::CloudController::V3::CreateBindingAsyncJob.new(
+      :key,
```
FelisiaM, line 184: `key` arg is the only difference between this and `enqueue_bind_job` maybe just extract that as a param and use `enqueue_bind_job(key, binidng_guid, message)`

lib/services/service_brokers/v2/client.rb
```diff
@@ -94,21 +94,22 @@ def create_service_key(key, arbitrary_parameters: {})
     end
 
     def bind(binding, arbitrary_parameters: {}, accepts_incomplete: false)
-      path              = service_binding_resource_path(binding.guid, binding.service_instance.guid, accepts_incomplete: accepts_incomplete)
-      body              = {
+      path = service_binding_resource_path(binding.guid, binding.service_instance.guid, accepts_incomplete: accepts_incomplete)
+      key_required_parameters = { credential_client_id: @config.get(:cc_service_key_client_name) } if binding.is_a?(VCAP::CloudController::ServiceKey)
```
FelisiaM, line 98: This line made me realise that the abstraction of the `required_parameters` as a method in the model might not quite right if we cant use it for all three cases. Do you see any issues with accessing the config in the ServiceKey model?

```
def required_parameters
      { credential_client_id: VCAP::CloudController::Config.config.get(:cc_service_key_client_name) }
end
```

That would remove the need for the client to know the specific of the Service Key

FelisiaM (changes requested):

lib/services/service_brokers/v2/client.rb
pivotal-marcela-campo: I don't see any issues, it's just a pattern that we don't seem to be following elsewhere. It seems to me weird that a model class can be configured in any way. I also didn't like the required_parameters in the model, as that means the model does know something about OSBAPI requests, which breaks encapsulation. It is cleaner for me to leave those things to the client. Our problem is that our client is a big ball of mud.

FelisiaM (approved):

PR #1979, opened 2020-12-01
V3 key bindings om 

pivotal-marcela-campo:
[175742023](https://www.pivotaltracker.com/story/show/175742023)

Only the [last commit](https://github.com/cloudfoundry/cloud_controller_ng/pull/1979/commits/475649c31807834b2518cd2db3a25c5a435bb556) needs review. The rest are part of another story

### DIFF
--- a/spec/request/service_credential_bindings_spec.rb
+++ b/spec/request/service_credential_bindings_spec.rb
@@ -1479,7 +1479,7 @@ RSpec.describe 'v3 service credential bindings' do
 
     context 'creating a credential binding as a key' do
       let(:service_instance) { VCAP::CloudController::ManagedServiceInstance.make(space: space, **service_instance_details) }
-      let(:binding_name) { 'some-key-name' }
+      let(:binding_name) { Sham.name }
       let(:create_body) {
         {
           type: 'key',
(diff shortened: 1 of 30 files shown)


spec/request/service_credential_bindings_spec.rb
```diff
@@ -1156,6 +1156,9 @@ def check_filtered_bindings(*bindings)
         describe 'a successful creation' do
           let(:syslog_drain_url) { 'http://syslog.example.com/wow' }
           let(:route_service_url) { 'http://route.example.com/wow' }
+          # this is because Orphan mitigation (orphan_mitigator.cleanup_failed_key) for keys does not currently support accepts_incomplete
```
FelisiaM, line 1159: I found the comments a bit difficult to understand without having a lot of OM deep understanding and I am wondering if that would be the case for future us as well.
I think we have now the opportunity and time to unify all OM for route, bindings and keys in the client code so why not do it. - It should not affect anything regarding v2 keys given that v3 is using different client method.
That way we also get rid of the difference between `cleanup_failed_key` and `cleanup_failed_bind` in the client.

PR #1984, opened 2020-12-02
make credential_client_id optional when creating service keys

tcdowney:
- makes the `cc_service_key_client_name` and
`cc_service_key_client_secret` config properties optional to accomodate
environments that do not have a platform CredHub deployed
- updates the services api client to only send the optional
`credential_client_id` bind_resource param when those config properties
are present
- see: https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#bind-resource-object

Addresses https://github.com/cloudfoundry/capi-k8s-release/issues/101

Tracker Story: [#175562836](https://www.pivotaltracker.com/story/show/175562836)

### DIFF
--- a/lib/services/service_brokers/v2/client.rb
+++ b/lib/services/service_brokers/v2/client.rb
@@ -71,10 +71,11 @@ module VCAP::Services::ServiceBrokers::V2
       body              = {
         service_id:    key.service.broker_provided_id,
         plan_id:       key.service_plan.broker_provided_id,
-        bind_resource: { credential_client_id: @config.get(:cc_service_key_client_name) },
+        bind_resource: {},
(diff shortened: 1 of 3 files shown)


lib/services/service_brokers/v2/client.rb
```diff
@@ -327,6 +327,14 @@ def fetch_service_binding(service_binding)
 
     private
 
+    def binding_required_parameters(binding)
```
tcdowney, line 330: I had to pull this out into a method cause Rubocop started complaining about cyclomatic complexity. 😞

pivotal-marcela-campo (approved):

FelisiaM (approved):

PR #2031, opened 2021-01-05
V3 service cred binding metadata 174260138

FelisiaM:


app/controllers/v3/service_credential_bindings_controller.rb
```diff
@@ -119,7 +119,7 @@ def parameters
 
   def create_key_binding(message, service_instance)
     action = V3::ServiceCredentialBindingKeyCreate.new(user_audit_info, message.audit_hash)
-    binding = action.precursor(service_instance, message.name)
+    binding = action.precursor(service_instance, message.name, message: message)
```
blgm, line 122: Perhaps it's not necessary to send `message` and `message.name`

spec/request/service_credential_bindings_spec.rb
```diff
@@ -1068,6 +1072,11 @@ def check_filtered_bindings(*bindings)
             get "/v3/service_credential_bindings/#{@binding_guid}", {}, admin_headers
             expect(last_response).to have_status_code(200)
             expect(parsed_response).to match_json_response(binding_response)
+
+            binding = VCAP::CloudController::ServiceBinding.last
+            expect(binding).to be
```
blgm, line 1077: What does this line do?

blgm (approved):

app/controllers/v3/service_credential_bindings_controller.rb
FelisiaM: Fixed this in latest commit

spec/request/service_credential_bindings_spec.rb
FelisiaM: It checks that it is not nil, similar to `expect(binding).to_not be_nil`
On failure it errors like this:
```
Failure/Error: expect(binding).to be
       expected nil to evaluate to true
```

I can see that it's not widely used, so I will change the occurrences to `to_not be_nil` just to not introduce more ways to do that same thing

PR #2034, opened 2021-01-06
v3(services): unbind error should include app name

blgm:
- when an app name is available, it should be in the unbind error message
- message was chosen to match the V2 message

[#176089624](https://www.pivotaltracker.com/story/show/176089624)

### DIFF
--- a/app/actions/v3/service_instance_delete.rb
+++ b/app/actions/v3/service_instance_delete.rb
@@ -153,7 +153,7 @@ module VCAP::CloudController
           unless result[:finished]
             polling_job = DeleteBindingJob.new(type, binding.guid, user_audit_info: service_event_repository.user_audit_info)
             Jobs::Enqueuer.new(polling_job, queue: Jobs::Queues.generic).enqueue_pollable
-            unbinding_operation_in_progress!
+            unbinding_operation_in_progress!(type, binding)
           end
         rescue => e
           errors << e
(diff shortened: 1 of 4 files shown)


app/actions/v3/service_instance_delete.rb
```diff
@@ -194,8 +194,15 @@ def operation_in_progress!
...
 
-      def unbinding_operation_in_progress!
-        raise UnbindingOperatationInProgress.new("An unbinding operation for a service binding of service instance #{service_instance.name} is in progress.")
+      def unbinding_operation_in_progress!(type, binding)
+        raise UnbindingOperatationInProgress.new(
+          case type
+          when :credential
+            "An operation for the service binding between app #{binding.app.name} and service instance #{service_instance.name} is in progress."
+          else
+            "An unbinding operation for a service binding of service instance #{service_instance.name} is in progress."
```
pivotal-marcela-campo, line 203: I would make the two messages similar for consistency. So for `key bindings` I would say `An operation for the service binding of service instance  #{service_instance.name} is in progress.`

pivotal-marcela-campo: Also, instead of passing `type` as an argument we can use `is_a` on the `binding` parameter as we have the whole object.

pivotal-marcela-campo (changes requested):

pivotal-marcela-campo (approved):

PR #2038, opened 2021-01-07
V3 delete metadata from service cred binding 174260622

FelisiaM:
[#174260622](https://www.pivotaltracker.com/story/show/174260622)

### DIFF
--- a/spec/request/service_credential_bindings_spec.rb
+++ b/spec/request/service_credential_bindings_spec.rb
@@ -434,14 +434,18 @@ RSpec.describe 'v3 service credential bindings' do
     end
 
     describe 'key credential binding' do
+      let(:labels) { { foo: 'bar' } }
+      let(:annotations) { { baz: 'wow' } }
       let(:key) do
         VCAP::CloudController::ServiceKey.make(service_instance: instance) do |binding|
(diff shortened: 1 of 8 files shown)


spec/unit/models/services/service_binding_spec.rb
```diff
@@ -637,13 +637,17 @@ def new_model
     describe '#destroy' do
       it 'cascade deletes all ServiceBindingOperations for this binding' do
```
blgm, line 638: A more generic description would be better now, for instance "it cascades deletion of related dependencies"

spec/request/service_credential_bindings_spec.rb
```diff
@@ -1868,13 +1873,19 @@ def check_filtered_bindings(*bindings)
       end
 
       context 'app binding' do
-        it_behaves_like 'service credential binding delete endpoint', 'service_binding', VCAP::CloudController::ServiceBinding
+        it_behaves_like 'service credential binding delete endpoint', 'service_binding',
```
blgm, line 1876: I have a preference for `'service_binding'` to be on it's own line as it highlights that it's an argument like the others

blgm (approved):

PR #2051, opened 2021-01-12
v3(services): reject binding parameters for user-provided service

blgm:
[#175515392](https://www.pivotaltracker.com/story/show/175515392)

### DIFF
--- a/app/controllers/v3/service_credential_bindings_controller.rb
+++ b/app/controllers/v3/service_credential_bindings_controller.rb
@@ -207,9 +207,8 @@ class ServiceCredentialBindingsController < ApplicationController
   end
 
   def check_parameters_support(service_instance, message)
-    unless service_instance.is_a?(VCAP::CloudController::ManagedServiceInstance)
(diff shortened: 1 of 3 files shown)


app/controllers/v3/service_credential_bindings_controller.rb
```diff
@@ -205,6 +206,16 @@ def enqueue_unbind_job(type, binding_guid)
     pollable_job.guid
   end
 
+  def check_parameters_support(service_instance, message)
+    unless service_instance.is_a?(VCAP::CloudController::ManagedServiceInstance)
```
pivotal-marcela-campo, line 210: Being that the error message in the exception thrown is specific for user-provided service instances, and that's the only other ServiceInstance type we have, I think this would be more readable if it was:
```
if service_instance.is_a?(VCAP::CloudController::UserProvidedServiceInstance) && message.requested?(:parameters)
       raise exception here
end
```

pivotal-marcela-campo (changes requested):

pivotal-marcela-campo (approved):

PR #2060, opened 2021-01-14
v3(service) document changes of service instances in v3 vs v2

FelisiaM:
[#175575306](https://www.pivotaltracker.com/story/show/175575306)


docs/v3/source/includes/upgrade_guide/changed_resources/_service_instances_in_v3.md
```diff
@@ -0,0 +1,26 @@
...
+
+When operating on service instances of type `user-provided` the API will respond synchronously for all operations.
+When the service instance type is `managed` the API will respond asynchronously and the operation might include communicating to the service broker. Read more about async responses [here](#asynchronous-operations).
+
+#### Listing bindings
+
+In v2, there were specific endpoints `/v2/service_instances/:guid/service_bindings` and `/v2/user_provided_service_instances/:guid/service_bindings` 
+to retrieve the bindings for managed and user-provided service instances.
+
+In v3, service bindings of a service instance can be retrieved by filtering the `v3/service_credential_binding` by `service_instance_guids`.
```
pivotal-marcela-campo, line 24: v3/service_credential_binding should be plural

pivotal-marcela-campo (approved):
I wonder if it makes sense to also add a note in the upgrade guide about the existence of the `fields` query parameters. Just to call attention to it.

FelisiaM:
> I wonder if it makes sense to also add a note in the upgrade guide about the existence of the `fields` query parameters. Just to call attention to it.

I saw there is a section about the summary endpoints. That might be a better place to mention the fields since that is exactly the behaviour it replaces. We have something on the backlog to update that.

PR #2065, opened 2021-01-15
v3(services): service broker client can take user_guid

blgm:
The service broker client reads the user_guid from the SecurityContext
when run from a controller. When run in a job, the SecurityContext is
not present. The client API now will now consistently accept an optional
user_guid parameter which will default to previous behaviour when not
specified.

[#176499762](https://www.pivotaltracker.com/story/show/176499762)

### DIFF
--- a/spec/unit/lib/services/service_brokers/v2/http_client_spec.rb
+++ b/spec/unit/lib/services/service_brokers/v2/http_client_spec.rb
@@ -182,9 +182,9 @@ module VCAP::Services::ServiceBrokers::V2
           make_request
 
           expect(a_request(http_method, full_url).
-                 with(basic_auth: basic_auth).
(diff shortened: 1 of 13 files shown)


spec/unit/lib/services/service_brokers/v2/http_client_spec.rb
```diff
@@ -161,6 +161,21 @@ module VCAP::Services::ServiceBrokers::V2
           end
         end
 
+        context 'when user_guid is specified as a parameter' do
```
pivotal-marcela-campo, line 164: Should we also have a test that specifies that the parameter overrides the security-context? I cannot think of a scenario where the two are set and are different from each other tbh, would just be mostly for completeness.

pivotal-marcela-campo (approved):

PR #2078, opened 2021-01-22
v3(tests):  Space-scoped brokers should not be writable in suspended …

pivotal-marcela-campo:
…orgs for space-developers

[#169067752](https://www.pivotaltracker.com/story/show/169067752)

### DIFF
--- a/spec/support/shared_examples/request/suspended_orgs_permissions.rb
+++ b/spec/support/shared_examples/request/suspended_orgs_permissions.rb
@@ -1,0 +1,21 @@
+RSpec.shared_examples 'Permissions when organization is suspended' do
+  before do
+    org.status = 'suspended'
+    org.save
+  end
+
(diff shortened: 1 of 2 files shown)


blgm (approved):

spec/support/shared_examples/request/suspended_orgs_permissions.rb
```diff
@@ -0,0 +1,21 @@
+RSpec.shared_examples 'Permissions when organization is suspended' do
+  before do
+    org.status = 'suspended'
+    org.save
+  end
+
+  it_behaves_like 'permissions for single object endpoint', ALL_PERMISSIONS do
```
blgm, line 7: It's possible to replace `ALL_PERMISSIONS` with something else so as to just test admin and space developer, as some of the other roles are less interesting. However that does add complexity.

Issue #2040, opened 2021-01-08
Issue with unauthenticated rate limits for CF API clients sharing IP addresses


stephanme:
Follow-up of discussion on [capi slack](https://cloudfoundry.slack.com/archives/C07C04W4Q/p1609946417352600).

CF API implements two kinds of [rate limiting](https://docs.cloudfoundry.org/running/setting-rate-limit-cloud-api.html): 
- rate limits for authenticated requests (counted per user)
- rate limits for unauthenticated requests (counted per IP address)

Unauthenticated rate limits based on IP address can become a problem when many CF API clients share the same IP address. This happens e.g. when the CF API clients are CF applications. CF foundations usually have just a very few outgoing IP addresses due to NAT. This means that the unauthenticated rate limit becomes a shared resource for such CF API clients and if one of these clients consumes all unauthenticated requests (e.g. due to bug), other CF API clients sharing the same IPs can get impacted.

One could assume that a well-behaving client never calls the CF API unauthenticated. Unfortunately, this is not the case when using the typical CF API client libs (e.g. https://github.com/cloudfoundry/cf-java-client). These libraries use the access token until the CF API responds with a 401 on an expired token, then refresh the access token and retry the request. Works well ... until unauthenticated rate limiting kicks in and responds 429 instead of 401. 429 is not handled by the CF java client lib.

The current CF API behavior on an expired access token is:
- 401 if the unauthenticated rate limit for the IP is not yet reached, unauthenticated rate limit counter is incremented
- 429 if the unauthenticated rate limit for the IP was already reached

This is not necessarily wrong from CC point of view but it is unexpected from CF API clients/libs that rely on a working access token refresh. As explained above a refresh is not properly implemented in many CF API client libs in case the (shared) unauthenticated rate limit was reached.

Possible solutions:

**CC prefers 401 over 429**

CC responds always 401 on an expired access token so that CF API client libs handle the refresh. The request can still be counted on unauthenticated rate limits.

CC responds 429 on requests without access token or if the access token is invalid for other reasons (e.g. can't be parsed, invalid signature etc.).

This fixes the described issue without having to adapt all CF API client libs and/or apps.

**CF API client libs implement 429 handling**

CF API client libs could implement token refresh on 429 (when hitting the unauthenticated limit only).
This was proposed for cf-java-client in https://github.com/cloudfoundry/cf-java-client/issues/960 nearly 2 years ago but not implemented so far. 

[...]

PR #2079, opened 2021-01-22
ISSUE-2040-favor-401-over-429

svkrieger:

This is a proposal for issue #2040 
This PR changes the behaviour of the "SecurityContextSetter" middleware. This will now check which return code it got back from the next middleware in the call chain. If it gets back a 429 from the "RateLimiter" middleware, it will check if the token is expired, but in general valid. If so, the response is changed from 429 to 401.

The information, whether the token is expired, but in general valid, is retrieved from the SecurityContext.
The SecurityContext was extended by a function which returns true, if the context contains ":expired_token".

The "UaaTokenDecoder" was also changed. It now calls a newly implemented function in the cf_uaa_lib "decode_at_reference_time_exp_warn_only". This function will return the decoded token within the "TokenExpired" exception, if the token is expired, but passed all other checks.
This enables us to execute the remaining checks like the uaa issuer validation.
If all checks in the UaaTokenDecoder passed as well, we throw the catched "TokenExpired" exception again, which will be catched in the "SecurityContextConfigurer", which will store ":expired_token" in the context.
The token is still considered invalid, but the different value in the context enables the "SecurityContextSetter" to distinguish between a bad token (all other cases except expired tokens) and an expired token.

As above mentioned issue describes, the change basically favors 401 over 429 for expired tokens.

This PR depends on the following PR: https://github.com/cloudfoundry/cf-uaa-lib/pull/70

### DIFF
--- a/middleware/security_context_setter.rb
+++ b/middleware/security_context_setter.rb
@@ -16,10 +16,17 @@ module CloudFoundry
           env['cf.user_name'] = VCAP::CloudController::SecurityContext.token['user_name']
         end
 
-        @app.call(env)
+        status, headers, body = @app.call(env)
+
+        # Return a 401 if the token is invalid and if the rate limit is already exceeded
+        if status == 429 && VCAP::CloudController::SecurityContext.invalid_token? && !VCAP::CloudController::SecurityContext.missing_token?
+          return invalid_token!(env, headers)
+        end
+
+        return [status, headers, body]
       rescue VCAP::CloudController::UaaUnavailable => e
         logger.error("Failed communicating with UAA: #{e.message}")
-        [502, { 'Content-Type:' => 'application/json' }, [error_message(env)]]
+        [502, { 'Content-Type:' => 'application/json' }, [error_message(env, 'UaaUnavailable')]]
       end
 
       private
@@ -28,8 +35,8 @@ module CloudFoundry
         VCAP::CloudController::SecurityContext.token['user_id'] || VCAP::CloudController::SecurityContext.token['client_id']
       end
 
-      def error_message(env)
-        api_error = CloudController::Errors::ApiError.new_from_details('UaaUnavailable')
+      def error_message(env, error_name)
+        api_error = CloudController::Errors::ApiError.new_from_details(error_name)
         version = env['PATH_INFO'][0..2]
 
         if version == '/v2'
@@ -37,6 +44,13 @@ module CloudFoundry
         elsif version == '/v3'
           ErrorPresenter.new(api_error, Rails.env.test?, V3ErrorHasher.new(api_error)).to_json
         end
+      end
+
+      def invalid_token!(env, headers)
+        headers['Content-Type']   = 'application/json'
+        message                   = error_message(env, 'InvalidAuthToken')
+        headers['Content-Length'] = message.length.to_s
+        [401, headers, [message]]
       end
 
       def logger
--- a/spec/unit/middleware/security_context_setter_spec.rb
+++ b/spec/unit/middleware/security_context_setter_spec.rb
@@ -53,6 +53,54 @@ module CloudFoundry
           end
         end
 
+        context 'when the rate limiter returns 429' do
+          let(:app) { double(:app, call: [429, { 'X-RateLimit-Remaining' => '0' }, 'a body']) }
+
+          context 'when token is valid' do
+            before do
+              allow(VCAP::CloudController::SecurityContext).to receive(:valid_token?).and_return(true)
+            end
+
+            it 'forwards the response from the rate limiter' do
+              status, headers, body = middleware.call(env)
+              expect(status).to eq(429)
+              expect(headers).to eq({ 'X-RateLimit-Remaining' => '0' })
+              expect(body).to eq('a body')
+            end
+          end
+          context 'when token is invalid' do
+            before do
+              allow(VCAP::CloudController::SecurityContext).to receive(:valid_token?).and_return(false)
+              allow(VCAP::CloudController::SecurityContext).to receive(:missing_token?).and_return(false)
+            end
+
+            it 'changes the response to 401' do
+              status, headers, body = middleware.call(env)
+              expect(status).to eq(401)
+              expect(headers).to eq({ 'Content-Length' => '189', 'Content-Type' => 'application/json', 'X-RateLimit-Remaining' => '0' })
+              json_body = JSON.parse(body.first)
+              expect(json_body).to include(
+                'code' => 1000,
+                'description' => 'Invalid Auth Token',
+                'error_code' => 'CF-InvalidAuthToken',
+              )
+            end
+          end
+          context 'when token is missing' do
+            before do
+              allow(VCAP::CloudController::SecurityContext).to receive(:valid_token?).and_return(false)
+              allow(VCAP::CloudController::SecurityContext).to receive(:missing_token?).and_return(true)
+            end
+
+            it 'forwards the response from the rate limiter' do
+              status, headers, body = middleware.call(env)
+              expect(status).to eq(429)
+              expect(headers).to eq({ 'X-RateLimit-Remaining' => '0' })
+              expect(body).to eq('a body')
+            end
+          end
+        end
+
         context 'when Uaa is unavailable' do
           before do
             allow(VCAP::CloudController::SecurityContext).to receive(:valid_token?).and_raise(VCAP::CloudController::UaaUnavailable)


sethboyles:
I think this approach is reasonable and I would be OK with you all moving forward with this solution.  

When writing tests, can you include one that ensures that when calling an Unauthenticated request with an Expired but Otherwise Valid token, that rate limiting is still being applied?

svkrieger:
Hi @sethboyles,

great! Regarding your specific test case. We saw in the meantime that the current solution unfortunately also changes the response from 429 to 401 for requests to unauthenticated endpoints, which don't skip the rate limit check in case the token is expired but otherwise valid.
This is not a 100% clean solution, but the cases where an unauthenticated endpoint returns a 401 are very rare. This is because most of the endpoints which don't require authentication, skip the rate limiting. Examples for that are the info endpoints. Endpoints which do not require authentication and would return a 401 in case the token is expired, but valid otherwise are marketplace calls. In this case we think it can be even positive to send a 401 because the CF cli will refresh the token and the returned service offering will be the same and not shortened because the user isn't authenticated anymore. The benefit of the current solution is that the rate limiting itself works just as before and only the response, which will be send to the client is changed.
The following table summarises the changes for the different endpoints and requests:

Endpoint   Type | Token   sent? | Token   valid? | Reason | Skip   rate limiting? | Rate   limit exceeded? | Before | Current   solution
-- | -- | -- | -- | -- | -- | -- | --
Unauthenticated | No | - | - | No | No | 200 | 200
Unauthenticated | No | - | - | No | Yes | 429 | 429
Unauthenticated | No | - | - | Yes |   | 200 | 200
Unauthenticated | Yes | No | Bad   token | No | No | 200 | 200
Unauthenticated | Yes | No | Token   expired | No | No | 200 | 200
Unauthenticated | Yes | No | Bad   token | No | Yes | 429 | 429
Unauthenticated | Yes | No | Bad   token | Yes | - | 200 | 200
**Unauthenticated** | **Yes** | **No** | **Token expired** | **No** | **Yes** | **429** | **401 (unwanted)**
Unauthenticated | Yes | No | Token   expired | Yes |   | 200 | 200
Unauthenticated | Yes | Yes | - | No | No | 200 | 200
Unauthenticated | Yes | Yes | - | No | Yes | 429 | 429
Unauthenticated | Yes | Yes | - | Yes |   | 200 | 200
  |   |   |   |   |   |   |  
Authenticated | No | - | - | No | No | 401 | 401
Authenticated | No | - | - | No | Yes | 429 | 429
Authenticated | Yes | No | Bad   token | No | No | 401 | 401
Authenticated | Yes | No | Token   expired | No | No | 401 | 401
Authenticated | Yes | No | Bad   token | No | Yes | 429 | 429
**Authenticated** | **Yes** | **No** | **Token expired** | **No** | **Yes** | **429** | **401 (wanted)**
Authenticated | Yes | Yes | - | No | No | 200 | 200
Authenticated | Yes | Yes | - | No | Yes | 429 | 429

We don't see any chance to distinguish between endpoints which require authentication and which don't in the middlewares, as this decision is made in the super class of the endpoints (ApplicationController) and the basis for this decision is the corresponding action e.g. "index", which is derived by the router after all middlewares were executed. We could hardcode all endpoints which don't require authentication into an array, but that's kinda hacky.
We'll go forward for now with implementing some tests. Let us know if you have other ideas or any concerns.

sethboyles:
@svkrieger thank you for that chart!  It really helps when trying to think about all the different cases here.

Considering we by default have a higher rate limit for authenticated requests ([defined here](https://github.com/cloudfoundry/capi-release/blob/develop/jobs/cloud_controller_ng/spec#L1013)), in the 'unwanted' case, that is probably a good change.  Letting a user know their token is expired would let them refresh their token and potentially get access to a higher rate limit.  

In general, it does seem a little odd to accept invalid tokens (expired or bad or otherwise) for unauthenticated requests. I wonder if it is just simpler to return 401 for invalid tokens wholesale, or at least when rate limiting is exceeded.  So for example, for this case: 


Endpoint   Type | Token   sent? | Token   valid? | Reason | Skip   rate limiting? | Rate   limit exceeded? | Before | Current   solution
-- | -- | -- | -- | -- | -- | -- | --
Unauthenticated | Yes | No | Bad   token | No | Yes | 429 | 429

I think there would be value in returning a 401 here as well because it could also give a user a higher rate limit.

svkrieger:
@sethboyles we discussed internally whether it would make sense to return a 401 generally in case the request was sent with an invalid token to an endpoint which does not require authentication. We don't think that this is a good idea as in general a 401 should be only returned when the requested resource requires authentication.

We also think that returning a 429 in case the rate limiting is exceeded and the token is a bad token (can't be decoded, wrong UAA issuer, not an access token) is preferable. We think we should only return the beneficial 401 to users which at least sent a valid, but expired token. If the token is a bad token, it is likely that the client won't be able to refresh it anyways (because obviously the client does not know which is the correct UAA or tries to construct an access token by itself).

Please let us know if that makes sense to you as well. In the meanwhile we added tests for the new logic. The checks here won't pass successfully though, until the PR in the cf_uaa_lib was merged and made available in a new release and we change the corresponding version in the Gemfile here. We are also open for feedback on the change in the cf_uaa_lib.

sethboyles:
@svkrieger I'm not sure it's worth it to check if a token is otherwise valid.  We might as well just return 401 for all expired tokens.  This would also simplify the solution and allow us to avoid the requirement of updating the uaa lib.

svkrieger:
Okay then I'd suggest we remove the check, whether it is expired as well, as we can't make sure it would be otherwise valid. In this case we can return a 401 for all bad tokens, including expired ones, right?

Then we can just check in the security_context_setter, if we get a 429, whether the token is invalid, valid or nil. If the token is invalid (including expired) we return a 401. If the token is nil or valid we return the 429.

That would result in the following response changes:
Endpoint   Type | Token   sent? | Token   valid? | Reason | Skip   rate limiting? | Rate   limit exceeded? | Before | Current   solution
-- | -- | -- | -- | -- | -- | -- | --
Unauthenticated | Yes | No | Bad   token | No | Yes | 429 | 401   (new)
Authenticated | Yes | No | Bad   token | No | Yes | 429 | 401   (new)

For the first case it could be a bit confusing though. First the client would receive 200s until the rate limit got exceeded and then it will get 401s. Maybe in this case we should also return the rate limit headers with the 401, to make visible, that the rate limit got exceeded? Again the endpoints which will fall into this case are very rare. We can only think of marketplace calls. Other endpoints which do not require authentication, are skipping the rate limit check.

Let us know if we should go for this solution. Then we will change the implementation accordingly.

sethboyles:
I am in favor of this solution.  You could make the argument for returning a 401 even if the rate limit is not exceeded, the endpoint is unauthenticated, but the token is invalid.  I checked a few other APIs and saw this behavior.

We don't have to go that far, but I think this solution is fine.

svkrieger:
@sethboyles the PR is ready for review. We implemented it as discussed above. The tests reflect the behaviour, we wanted to achieve. We also think that always returning a 401 is a bit out of scope here, as it would require code changes in the endpoints too. Let us know if you have any questions about the current implementation.

bruce-ricard:
Hi, Bruce from UAA here.

I would like to talk about this line:

Endpoint Type | Token sent? | Token valid? | Reason | Skip rate limiting? | Rate limit exceeded? | Before | Current solution
-- | -- | -- | -- | -- | -- | -- | --
Authenticated | Yes | Yes | - | No | Yes | 429 | 429


IMHO there is a larger problem here. Why are we validating the token when we know that the user is being rate limited? 
IMHO the first line of the controller should be something like
```
if is_rate_limited() then
  return 429
```

This seems to completely defeat the purpose of rate limiting, if we still do work and maybe let the UAA be DDOSed because we keep validating the tokens, when we know that we are going to return 429 anyways.

Please share your thoughts.

svkrieger:
@bruce-ricard let me answer your questions:
- First of all, this PR only changes the response of the CC from 429 to 401 in case the token is invalid. I'm pretty confident that this can't open any backdoors or enable DoS attacks.
- The token is and always was decoded before the rate limiting middleware. As to what I saw in the CC coding, the CC fetches the public keys from the UAA once and then uses these to decode the tokens locally, by using the cf-uaa-lib and without calling the UAA for every token. The information, whether the token is valid or not, is needed to distinguish between unauthenticated and authenticated rate limits.
- As long as a valid token comes with the request, the rate limiting applies to the specific user and not to the IP. Even if the IP's rate limit got exceeded, the authenticated request coming from the same IP, won't be rejected.

sethboyles:
Hi @bruce-ricard 

As @svkrieger noted above, we check validity of tokens (if present) for rate limiting as users or clients may have a higher rate limit than unauthenticated clients.

See related configs here: https://github.com/cloudfoundry/capi-release/blob/develop/jobs/cloud_controller_ng/spec#L1013

Does that make sense or do you think there are possible issues with this ?  (this logic is already present in CCNG and not introduced by this PR, by the way)

sethboyles:
Merging.  Thanks for submitting this PR!

PR #2085, opened 2021-01-27
v3(services): service broker eager loading

blgm:
[#176660193](https://www.pivotaltracker.com/story/show/176660193)

### DIFF
--- a/app/fetchers/service_broker_list_fetcher.rb
+++ b/app/fetchers/service_broker_list_fetcher.rb
@@ -4,13 +4,13 @@ require 'fetchers/label_selector_query_generator'
 module VCAP::CloudController
   class ServiceBrokerListFetcher < BaseListFetcher
     class << self
-      def fetch(message:, permitted_space_guids: nil)
+      def fetch(message:, permitted_space_guids: nil, eager_loaded_associations: [])
         if permitted_space_guids
           dataset = ServiceBroker.dataset.where(Sequel[:service_brokers][:space_id] => spaces_from(permitted_space_guids))
           return filter(message, dataset)
         end
 
-        dataset = ServiceBroker.dataset
+        dataset = ServiceBroker.dataset.eager(eager_loaded_associations)
         filter(message, dataset)
       end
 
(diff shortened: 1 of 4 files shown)


app/fetchers/service_broker_list_fetcher.rb
```diff
@@ -4,13 +4,13 @@
 module VCAP::CloudController
   class ServiceBrokerListFetcher < BaseListFetcher
     class << self
-      def fetch(message:, permitted_space_guids: nil)
+      def fetch(message:, permitted_space_guids: nil, eager_loaded_associations: [])
         if permitted_space_guids
           dataset = ServiceBroker.dataset.where(Sequel[:service_brokers][:space_id] => spaces_from(permitted_space_guids))
```
pivotal-marcela-campo, line 9: Is there any reason why we are not eager loading here?

pivotal-marcela-campo (changes requested):

app/fetchers/service_broker_list_fetcher.rb
blgm: Well spotted. I didn't spot that it was a branch. Looking at this closer, the `spaces_from()` method is moving a conceptual `join` from the database into Ruby, which is potentially much slower and harder to paginate efficiently, so I have re-written this aspect of the fetcher to work in the same way as other fetchers that we have written.

pivotal-marcela-campo: I am not sure I understand why the rewrite is needed. As far as I can tell the spaces_from() method used to return a query which is nested in the original one, so the final query should have looked like

```Where service_brokers.space_id in ( Select id 
                                                                  from spaces 
                                                                  where guid in [permitted_guid1, permitted_guid_2, ...])```

Which hopefully the db scheduler would notice is always the same subquery and just execute once?

Anyway, looked at the new code and it looks okay to me as well.

**Review by pivotal-marcela-campo — APPROVED** [review]

File: app/fetchers/service_broker_list_fetcher.rb (hunk 1 above)
Comment (blgm) [reply app/fetchers/service_broker_list_fetcher.rb]: @pivotal-marcela-campo you are quite correct. The original query is:
```sql
SELECT * FROM \"service_brokers\" WHERE (\"service_brokers\".\"space_id\" IN (SELECT \"id\" FROM \"spaces\" WHERE (\"guid\" IN ('eda8aec8-a6eb-4cbe-82ee-7fdf5283963c')))) ORDER BY \"service_brokers\".\"id\" ASC LIMIT 50 OFFSET 0
```
The new implementation is more consistent with other fetchers, so I'm tempted to leave it as it is, but the justification relating to performance was not correct.

PR #2093, opened 2021-01-29
v3(services): service credential bindings eager loading

blgm:
[#176660161](https://www.pivotaltracker.com/story/show/176660161)

### DIFF
--- a/app/models/services/service_credential_binding_view.rb
+++ b/app/models/services/service_credential_binding_view.rb
@@ -91,6 +91,7 @@ module VCAP
             'key' => 'VCAP::CloudController::ServiceKey'
           }
 
+        # Custom eager loading: https://github.com/jeremyevans/sequel/blob/master/doc/advanced_associations.rdoc#label-Custom+Eager+Loaders
         many_to_one :service_instance_sti_eager_load,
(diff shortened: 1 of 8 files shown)


app/models/services/service_credential_binding_view.rb
```diff
@@ -85,11 +85,111 @@ module Types
...
+              case scb # keys are joined by ID and bindings by GUID
+              when ServiceKey
+                service_instance_id_to_keys[scb.service_instance_id] ||= []
+                service_instance_id_to_keys[scb.service_instance_id] << scb
+              when ServiceBinding
+                service_instance_guid_to_bindings[scb.service_instance_guid] ||= []
+                service_instance_guid_to_bindings[scb.service_instance_guid] << scb
+              end
+            end
+            ds = ServiceInstance.where(id: service_instance_id_to_keys.keys).or(guid: service_instance_guid_to_bindings.keys)
```
pivotal-marcela-campo, line 112: I am not sure how efficient is this. It seems that Postgresql can do a "bitmap or" of the two indexes, and wouldn't be a big issue. Not sure about other databases. Just something to keep an eye on.

pivotal-marcela-campo (approved):

PR #2096, opened 2021-02-01
V3 consistent service broker permissions 176595049

pivotal-marcela-campo:
[#176595049](https://www.pivotaltracker.com/story/show/176595049)

### DIFF
--- a/app/controllers/v3/service_brokers_controller.rb
+++ b/app/controllers/v3/service_brokers_controller.rb
@@ -75,7 +75,7 @@ class ServiceBrokersController < ApplicationController
 
     if service_broker.space_guid
       space = service_broker.space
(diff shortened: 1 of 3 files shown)


app/controllers/v3/service_brokers_controller.rb
```diff
@@ -75,9 +75,10 @@ def update
 
     if service_broker.space_guid
       space = service_broker.space
-      unprocessable_space! unless space && permission_queryer.can_read_from_space?(space.guid, space.organization_guid)
+      broker_not_found! unless space && permission_queryer.can_read_from_space?(space.guid, space.organization_guid)
```
blgm, line 78: I think scenario 1c says this should be a 422 error, and this would appear to change it to a 404

blgm (changes requested):

blgm (approved):

PR #2101, opened 2021-02-02
v3(services) Introduce last operation for user-provided si

FelisiaM:
[#174695315](https://www.pivotaltracker.com/story/show/174695315)

### DIFF
--- a/app/models/services/user_provided_service_instance.rb
+++ b/app/models/services/user_provided_service_instance.rb
@@ -13,20 +13,6 @@ module VCAP::CloudController
       !(route_service_url.nil? || route_service_url.empty?)
     end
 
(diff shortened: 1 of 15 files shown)


app/models/services/user_provided_service_instance.rb
```diff
@@ -7,13 +7,12 @@ class UserProvidedServiceInstance < ServiceInstance
     strip_attributes :name, :syslog_drain_url, :route_service_url
 
     add_association_dependencies service_bindings: :destroy
+    add_association_dependencies service_instance_operation: :destroy
```
blgm, line 10: Could this be moved to the superclass?

blgm (changes requested):

blgm (approved):
Thanks!

PR #2106, opened 2021-02-03
V3 refactor service instance create

FelisiaM:


app/actions/v3/service_instance_update_managed.rb
```diff
@@ -0,0 +1,362 @@
...
+        @service_instance = instance
+        @message = message
+        @user_audit_info = user_audit_info
+        @audit_hash = audit_hash
+      end
+
+      def precursor
+        raise_if_cannot_proceed!
+
+        unless is_deleting?(service_instance) || !only_metadata?
```
blgm, line 29: This method is hard to reason about. It takes a lot of effort to work out what `unless !only_metadata?` means, and why the metadata could be updated in more than one place.

app/jobs/v3/create_service_instance_job.rb
```diff
@@ -23,19 +33,129 @@ def operation_type
...
+
+        raise_if_other_operations_in_progress!
+
+        compute_maximum_duration
+
+        begin
+          if @first_time
+            @first_time = false
+            action.provision(service_instance, parameters: @arbitrary_parameters, accepts_incomplete: true)
+            compatibility_checks # TODO: get the warning from the action?
```
blgm, line 63: What's the plan for this TODO?

blgm (changes requested):
This is a great refactor. I think the update "precursor" method needs some simplification, probably because it's trying to do too many things in a single method. Some potential alternative methods might be:

- `preflight()` is this update even valid?
- `needs_broker_update?` - is the update sync or async
- `perform_sync_update()` - do the sync updates

app/actions/v3/service_instance_create_managed.rb
```diff
@@ -0,0 +1,182 @@
...
+        client = VCAP::Services::ServiceClientProvider.provide(instance: instance)
+        details = client.provision(
+          instance,
+          arbitrary_parameters: parameters,
+          accepts_incomplete: accepts_incomplete,
+          maintenance_info: instance.service_plan.maintenance_info,
+          user_guid: @user_audit_info.user_guid
+        )
+
+        if details[:last_operation][:state] == 'in progress' && details[:last_operation][:type] == 'create'
```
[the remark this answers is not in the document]
FelisiaM: come to think of it just checking for the state should be enough.

app/actions/v3/service_instance_create_managed.rb
```diff
@@ -0,0 +1,182 @@
...
+        save_failed_state(instance, e)
+
+        raise e
+      end
+
+      def poll(instance)
+        client = VCAP::Services::ServiceClientProvider.provide(instance: instance)
+        details = client.fetch_service_instance_last_operation(instance, user_guid: @user_audit_info.user_guid)
+
+        case details[:last_operation][:state]
```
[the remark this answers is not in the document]
FelisiaM: that is handled by the job for all operations

app/actions/v3/service_instance_create_managed.rb
```diff
@@ -0,0 +1,182 @@
...
+          instance.save_with_new_operation(
+            broker_response[:instance] || {},
+            broker_response[:last_operation] || {}
+          )
+        end
+
+        event_repository.record_service_instance_event(:create, instance, @audit_hash)
+      end
+
+      def save_incomplete_instance(instance, broker_response)
```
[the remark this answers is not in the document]
FelisiaM: Updated to use similar methods

app/jobs/v3/create_service_instance_job.rb
FelisiaM: Removed it in the benefit of completing this work.

app/actions/v3/service_instance_update_managed.rb
FelisiaM: Updated the if statement to `if update_metadata_only?`. Hope this makes it more clear

app/controllers/v3/service_instances_controller.rb
```diff
@@ -260,21 +269,24 @@ def update_user_provided(service_instance)
...
-    end
-
-    service_instance, job = ServiceInstanceUpdateManaged.new(service_event_repository).update(service_instance, message)
-
-    if job.nil?
-      render status: :ok, json: Presenters::V3::ServiceInstancePresenter.new(service_instance)
+    raise_if_invalid_service_plan!(service_instance, message)
+
+    action = V3::ServiceInstanceUpdateManaged.new(service_instance, message, user_audit_info, message.audit_hash)
+    service_instance, continue_async = action.precursor
```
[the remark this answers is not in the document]
FelisiaM: Updated this - went with the first option.

app/actions/v3/service_instance_update_managed.rb
```diff
@@ -23,10 +23,16 @@ def initialize(instance, message, user_audit_info, audit_hash)
...
+        raise_if_invalid_update!
+        raise_if_renaming_shared_service_instance!
+        raise_if_invalid_plan_change!
+        raise_if_invalid_maintenance_info_change!
+        raise_if_cannot_update!
+      end
 
-        unless is_deleting?(service_instance) || !only_metadata?
+      def try_update_sync
+        if update_metadata_only?
```
blgm, line 35: This is definitely improved. There are some questions that occur to me that are hard to understand, but perhaps could be made clearer with better names:
- if a service instance is being deleted, why is it that I can't change the metadata, but I can change the metadata and the name together as long as `allow_context_updates` is false?
- why is the audit event reported for the original service instance and not the updated one? But when only metadata is changed, it's reported on the updated service instance?

I'm sure there are good reasons, it's just that for someone unfamiliar with the code, it's not obvious. But perhaps the thing to do is merge this and improve it later.

blgm (approved):

app/actions/v3/service_instance_update_managed.rb
FelisiaM: Both questions that I have had to think about in the code.

> If a service instance is being deleted, why is it that I can't change the metadata, but I can change the metadata and the name together as long as allow_context_updates is false?

I think there is some more subtlety here that was not captured either in the previous implementation and neither here. When changing metadata we made the decision to allow it only while create and update is in progress for all resources. In that sense it's consistent. I guess the idea is that while delete is happening it might be more confusing to accept any other operation. 
Any other operation makes the state check first and will fail if another operation is in progress - delete in this case. So name and metadata update that could happen later doesn't really happen. 
```
 lock = UpdaterLock.new(service_instance)
 lock.lock!
```
will raise when other operations are in progress.

> why is the audit event reported for the original service instance and not the updated one? But when only metadata is changed, it's reported on the updated service instance?

My understanding is that the event log does not include metadata so it is not relevant which instance is used. However, I will change it to be the updated one just for consistency.

app/actions/v3/service_instance_update_managed.rb
```diff
@@ -0,0 +1,372 @@
...
+            broker_response[:last_operation] || {}
+          )
+        end
+
+        event_repository.record_service_instance_event(:start_update, instance, @audit_hash)
+      end
+
+      def save_failed_state(instance, e)
+        instance.save_with_new_operation(
+          {},
```
pivotal-marcela-campo, line 160: Not sure I follow why sometimes we lock last_operation and sometimes we don't. In particular we used to lock when we saved the failed state. So maybe it's worth reviewing it.

pivotal-marcela-campo (approved):

PR #2108, opened 2021-02-04
services: validate binding credentials

blgm:
OSBAPI requires binding credentials to be a JSON object. When a binding
is retrieved this is validated, and this adds validation when the
credentials are returned for synchronous binding response.

[#174599297](https://www.pivotaltracker.com/story/show/174599297)


FelisiaM (approved):
🔑   ✅

blgm:
It's worth noting that if the credentials are not an object then the `cf service-key` fails later, so this is about catching errors earlier.

PR #2083, opened 2021-01-26
Eirini LRP instances client can handle processes with 0 instances

tcdowney:
https://github.com/cloudfoundry/cloud_controller_ng/pull/1934 updated this code to handle `STOPPED` processes, but did not account for `STARTED` processes that had 0 desired running instances. Processes with 0 instances do not have a
StatefulSet backing them on Kubernetes which caused the instance stats check to raise an exception.

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats) -- nope 🙂

cc @Birdrock @davewalter @acosta11

PR #1934, opened 2020-10-29
Prevent unnecessary retries in the LRP instances client

cwlbraa:
Prior to this change, the LRP instances client would retry repeatedly if you asked for instances of STOPPED processes via the various `/stats` endpoints. Instead, this change confirms that Eirini is not running stopped apps. In addition, this change will retry any non-404 error. We think this might help with reliability in large clusters.

tracker story: https://www.pivotaltracker.com/story/show/175381591

vmware-internal report of the bug: https://vmware.slack.com/archives/C01BQNTR9RT/p1603226144071200

PR #2123, opened 2021-02-10
Re-introduce prevention of unnecessary retrying of fetching stats for stopped OPI instances

jspawar:

Revert the changes we made to prevent unnecessary retrying of fetching metrics for stopped/0 instances apps and resolve the null-pointer dereferencing error those changes inadvertently introduced

Should mitigate slow response times for the `cf apps` command when there are more than 25+ apps pushed to a standard cf-for-k8s installation

Reverts the reverting of the following:
- #1934
- #2083

### DIFF
--- a/lib/cloud_controller/opi/instances_client.rb
+++ b/lib/cloud_controller/opi/instances_client.rb
@@ -38,7 +38,11 @@ module OPI
     end
 
     def lrp_instances(process)
-      return confirm_not_running(process) if process_has_no_desired_instances?(process)
+      if process_has_no_desired_instances?(process)
+        confirm_not_running(process)
+        # return zero instances
+        return []
+      end
 
       parsed_response = get_instances(process)
       parsed_response['instances'].map do |instance|
@@ -60,9 +64,12 @@ module OPI
 
     def confirm_not_running(process)
       parsed_response = JSON.parse(client.get(instances_path(process)).body)
-      raise Error.new("expected no instances for stopped process: #{parsed_response['error']}") unless parsed_response['error'].include?('not found')
+      possible_error = parsed_response['error']
 
-      []
+      # we are _only_ ok with receiving no errors or a 404-esque error
+      if !possible_error.blank? && !possible_error.include?('not found')
+        raise Error.new("expected no instances for stopped process: #{possible_error}")
+      end
     end
 
     def get_instances(process)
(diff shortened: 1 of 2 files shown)


lib/cloud_controller/opi/instances_client.rb
```diff
@@ -59,11 +58,52 @@ def desired_lrp_instance(process)
...
 
-    def handle_error(response_body)
-      error = response_body['error']
-      return unless error
+    def process_has_no_desired_instances?(process)
+      process.stopped? || process.instances == 0
+    end
+
+    def confirm_not_running(process)
+      parsed_response = JSON.parse(client.get(instances_path(process)).body)
```
tcdowney, line 66: What if the response is not JSON parseable? Like if Eirini is not reach able and you just get back like a `"no healthy upstream"` error from its Envoy sidecar?

tcdowney: I'm wondering if we should have the `Retryable.retryable` block either in here to retry those cases or at a higher level up in the `lrp_instances` method and have both this and the `get_instances` method in its block.

jspawar: Oh lol ya, should prob actually handle errors here. The 1-2 levels above which are calling this are handling generic errors being raised, so they won't be lost entirely at least. Wrapping them to make them more clear would probably be good though.

If we put it in a retryable block though, the PR to limit retries which prompted this is kind of moot isn't it?

tcdowney: The intent of the original PR was to not retry only on "not found"/404 style errors. It was because if your process was `STOPPED`/had zero instances this call would always retry five times and take five seconds to respond for every stopped app.

I think it was a mistake in the original PR that it didn't retry other classes of errors in the "no running instances" case.

jspawar: Ah ok, thanks for clarifying, that makes sense. Will adjust this accordingly then

jspawar: @tcdowney just pushed changes to match the original intent, seem fine?

tcdowney: yes, I think this looks good 👍 

thanks! 🙂

tcdowney (approved):

lib/cloud_controller/opi/instances_client.rb
jspawar: :tq:

jspawar:
review complete, merging and closing

PR #2131, opened 2021-02-19
services: allow longer service instance names

### LINKED ISSUE #2111: Creation of service instance with name more than 50 characters fails
## Issue

Service instance creation fails when I tried to create the instance with name more than 50 characters. 

## Context

When I tried to create a service instance of name more than 50 characters, it failed with error: **You have requested an invalid service instance name. Names are limited to 50 characters.**
What is the reason behind this limitation? Is it possible to bypass this length restriction somehow as it breaks our live scenario? Also, I observed that the service key name has no length restriction, then why only service name?

blgm:
Issue #2111 documents a problem with service instance names being
limited to 50 characters and a case to increase that limit.

The original reason for this limit was not documented when it was
introduced in 2013. The best working theory is that there was a service
broker used in Pivotal Web Services (PWS) that was not compatible with
names longer than 50 characters. That is no longer the case. But there
is a risk that this change could cause issues for some service brokers.
We note that OSBAPI does not specify a limit.

The limit has been increased to 255 characters so that it matches the
limit for service binding names.

[#176818363](https://www.pivotaltracker.com/story/show/176818363)

Resolves #2111

### DIFF
--- a/app/models/services/service_instance.rb
+++ b/app/models/services/service_instance.rb
@@ -115,7 +115,7 @@ module VCAP::CloudController
       validates_presence :name
       validates_presence :space
       validates_unique :name, where: name_clashes
-      validates_max_length 50, :name
+      validates_max_length 255, :name
       validates_max_length 10_000, :syslog_drain_url, allow_nil: true
       validate_tags_length
     end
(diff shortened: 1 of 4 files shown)


FelisiaM (approved):
👍

Issue #2010, opened 2020-12-17
Race condition when a buildpack is deleted


jochenehret:
## Issue

Race condition when a buildpack is being deleted.

## Context

When you delete a buildpack, the blob is deleted first and then (to our understanding) the record in the database:
https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/actions/buildpack_delete.rb

If an app is pushed at the same time and uses buildpack auto-detection, the push can fail with:
```
Staging app and tracing logs...
Staging error: Missing blob for <buildpack being deleted>. Specify a different buildpack with the -b flag or contact your operator.
```
Note that the app uses a different buildpack. Workaround is to specify a fixed buildpack so that auto-detection does not try to use the buildpack being deleted.

## Steps to Reproduce

Could be reproduced by inserting a wait/sleep after blobstore deletion and before database deletion:
https://github.com/cloudfoundry/cloud_controller_ng/blob/76cb268b88909a0e07e60c3c3b82f1ce52fbe1be/app/actions/buildpack_delete.rb#L12

## Expected result

The push should not fail.

## Current result

The push can fail with the error shown above.

## Possible Fix

The database record should be deleted first so that the buildpack is not referenced in the auto-detect process. The buildpack blob should be deleted afterwards.

philippthun:
Fixed with #2133.

PR #2133, opened 2021-02-24
Fix buildpack delete race condition

jochenehret:
Fix for https://github.com/cloudfoundry/cloud_controller_ng/issues/2010

This PR fixes the race condition explained in https://github.com/cloudfoundry/cloud_controller_ng/issues/2010

If a CF buildpack is deleted the blobstore object is deleted before the database record. In scenarios with heavy load and slower procession of delayed background jobs, it can happen that a simultaneously executed "cf push" fails with:
```
Staging app and tracing logs...
Staging error: Missing blob for <buildpack name>. Specify a different buildpack with the -b flag or contact your operator.
```
The fix first deletes the database record and cleans up the blobstore afterwards. Should the blobstore deletion fail, the OrphanedBlobsCleanup class would clean up the orphaned buildpack blob.

### DIFF
--- a/app/actions/buildpack_delete.rb
+++ b/app/actions/buildpack_delete.rb
@@ -4,13 +4,11 @@ module VCAP::CloudController
       buildpacks.each do |buildpack|
         Buildpack.db.transaction do
           Locking[name: 'buildpacks'].lock!
-
-          if buildpack.key
-            blobstore_delete = Jobs::Runtime::BlobstoreDelete.new(buildpack.key, :buildpack_blobstore)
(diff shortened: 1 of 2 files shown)


JenGoldstrich:
Hey @jochenehret

Thank you so much for this PR, @belinda-liu and I reviewed it and think it looks great, we also pulled it in locally and run `bundle exec rake` to confirm it will work for us locally

Good callout @FloThinksPi, based on investigating Belinda and I saw that the OrphanedBlobCleanup job runs every day https://docs.cloudfoundry.org/concepts/architecture/cloud-controller.html#automatic-clean which makes us confident in this change.

Merging this PR now.

Thanks again for your contribution!
@belinda-liu and Jenna

PR #2149, opened 2021-03-10
v3(services) Expose parameters in schema object

FelisiaM:
Keeping consistency with how we show empty objects.

[#177270692](https://www.pivotaltracker.com/story/show/177270692)

### DIFF
--- a/app/presenters/v3/service_plan_presenter.rb
+++ b/app/presenters/v3/service_plan_presenter.rb
@@ -90,7 +90,7 @@ module VCAP::CloudController
         end
 
         def parse_schema(schema)
-          return {} unless schema
+          return { parameters: {} } unless schema
 
           { parameters: JSON.parse(schema) }
         rescue JSON::ParserError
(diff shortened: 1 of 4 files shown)


app/presenters/v3/service_plan_presenter.rb
```diff
@@ -90,7 +90,7 @@ def maintenance_info
         end
 
         def parse_schema(schema)
-          return {} unless schema
+          return { parameters: {} } unless schema
 
           { parameters: JSON.parse(schema) }
         rescue JSON::ParserError
```
blgm, line 96: What about if there's a parse error?

blgm (changes requested):

app/presenters/v3/service_plan_presenter.rb
FelisiaM: Moved the "parameter" property to the calling code for `parse_schema`.

blgm (approved):

PR #2153, opened 2021-03-15
Added SI related info for orphan mitigation

FelisiaM:
[#176053241](https://www.pivotaltracker.com/story/show/176053241)


JenGoldstrich:
Hey @FelisiaM 

This PR looks good to me, since this change is only docs I think its worth ignoring the test failure since it seems like an unrelated flake,

Thanks, merging this!
Jenna

PR #2154, opened 2021-03-15
Sending app annotation to service broker on bind request

FelisiaM:
As defined in [OSBAPI 2.16](https://github.com/openservicebrokerapi/servicebroker/blob/v2.16/profile.md#cloud-foundry-bind-resource-object)

[#174946388](https://www.pivotaltracker.com/story/show/174946388)

### DIFF
--- a/app/models/services/service_binding.rb
+++ b/app/models/services/service_binding.rb
@@ -1,4 +1,5 @@
 require 'models/helpers/process_types'
+require 'presenters/mixins/metadata_presentation_helpers'
 
 module VCAP::CloudController
   class ServiceBinding < Sequel::Model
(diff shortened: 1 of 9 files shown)


app/models/services/service_binding.rb
```diff
@@ -3,6 +3,7 @@
 module VCAP::CloudController
   class ServiceBinding < Sequel::Model
     include Serializer
+    include VCAP::CloudController::Presenters::Mixins::MetadataPresentationHelpers
```
FelisiaM, line 6: Already thinking this is not a great idea. 
Might move the `required_parameters` to the client since it only relates to OSBAPI interface

blgm: Agreed. It has the potential to become a circular dependency, as the presenters depend on the models; do we don't really want the models to depend on the presenters...

blgm (approved):

Issue #2161, opened 2021-03-24
Service binding parameters in manifest are not passed to the service broker when cf push


svkrieger:
## Issue

When parameters are added to a service binding in the manifest and cf push is executed, the service broker won't receive the provided parameters.

## Context

The manifest could look like this:
```
applications:
- buildpacks:
  - binary_buildpack
  disk_quota: 128M
  instances: 1
  memory: 128M
  name: my-app
  services:
  - name: myservice
    parameters:
      myparameter: "myvalue"
      mysecondparameter: "mysecondvalue"
```
Setup:
CF CLI version: 7.2.0+be4a5ce2b.2020-12-10
CC API version: 3.98.0 (main branch, but does occur in previous releases too)

## Steps to Reproduce

Put some parameters into the manifest for a service binding and execute `cf push`.
Then check whether the service broker received a request containing the provided parameters.

## Expected result

The service broker receives the provided parameters with the service binding request.
The following shows how a request, which arrives at the service broker should look like:
```
Request: PUT /v2/service_instances/0c028742-0d9f-4618-9f1c-e90847343d0f/service_bindings/c381bc54-9b0a-49df-a006-22f515799f80 HTTP/1.1
Host: service-broker:3000
Accept: application/json
Authorization: Basic YnJva2VyLXVzZXI6YnJva2VyLXBhc3N3b3Jk
Content-Length: 739
Content-Type: application/json
Date: Wed, 24 Mar 2021 10:40:12 GMT
User-Agent: HTTPClient/1.0 (2.8.3, ruby 2.7.2 (2020-10-01))
X-Api-Info-Location: localhost/v2/info
X-Broker-Api-Originating-Identity: cloudfoundry ewogICJ1c2VyX2lkIjogImRmYWZmZjczLTYyZDAtNDVhMi1hNWFjLTUyYTJjZTA4ZWJhZCIKfQ==
X-Broker-Api-Request-Identity: a621a2cd-a8bd-40e3-8ba9-16de18c30932
X-Broker-Api-Version: 2.15
X-Vcap-Request-Id: 00943c2d-db21-4c14-b396-cc08dfc1f4d0

{
  "service_id": "1234-5678-8765-4321-service-my-service",
  "plan_id": "1234-5678-8765-4321-plan-unlimited",
  "app_guid": "ecde9a4f-7e68-4e88-8cb1-7951f5648b7a",
  "bind_resource": {
[...]

PR #2163, opened 2021-03-24
ISSUE-2161 Pass service binding parameters to service broker

svkrieger:

Pass service binding parameters, which were provided in an app manifest, to the service broker.
This PR is a suggestion for solving #2161 

### DIFF
--- a/app/actions/app_apply_manifest.rb
+++ b/app/actions/app_apply_manifest.rb
@@ -156,7 +156,7 @@ module VCAP::CloudController
             message: binding_message)
 
           begin
-            result = action.bind(binding)
+            result = action.bind(binding, parameters: binding_message.parameters)
             if result[:async]
               raise ServiceBrokerRespondedAsyncWhenNotAllowed
             end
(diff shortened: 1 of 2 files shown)


JenGoldstrich:
Hey @svkrieger, this PR looks good to us, one thing @tstannard and I think you could maybe add explicit unit tests for when you pass in service binding parameters, and when you don't.  Otherwise this seems reasonable to us, will ping folks from the services team on slack for them to also scan through this to double check that.

Thanks,
Jenna & Teal

app/actions/app_apply_manifest.rb
```diff
@@ -156,7 +156,11 @@ def create_service_bindings(manifest_service_bindings_message, app)
             message: binding_message)
 
           begin
-            result = action.bind(binding)
+            result = if binding_message.parameters
+                       action.bind(binding, parameters: binding_message.parameters)
+                     else
+                       action.bind(binding)
+                     end
```
FelisiaM, line 163: The if statement is needed here. `binding_message.parameter` will be set to an empty object if not set in the manifest and `action.bind` can handle that.

svkrieger: Good point!

svkrieger:
@JenGoldstrich I can't really come up with a meaningful explicit test for the arbitrary parameters, without adding two contexts and duplicating some tests, as it is tested implicitly at the moment. And I don't think it would add that much confidence to be worth it. But let me know if you feel different. Or feel free to add one yourself, if you have one in mind.

JenGoldstrich:
Merging, see further context: https://cloudfoundry.slack.com/archives/C07C04W4Q/p1616585842025800

Thanks again @svkrieger !!!

PR #2167, opened 2021-03-25
Set additional header with user guid

johha:
  Set user guid as additional header. This can be used for example in the nginx access logs.
  There will be a related capi PR which ensures that the guid is not forwarded to the client/user

  * https://github.com/cloudfoundry/capi-release/pull/191 

### DIFF
--- a/middleware/security_context_setter.rb
+++ b/middleware/security_context_setter.rb
@@ -23,6 +23,8 @@ module CloudFoundry
           return invalid_token!(env, headers)
         end
 
+        headers['X-USER-GUID'] = env['cf.user_guid'] if env['cf.user_guid']
+
         return [status, headers, body]
       rescue VCAP::CloudController::UaaUnavailable => e
         logger.error("Failed communicating with UAA: #{e.message}")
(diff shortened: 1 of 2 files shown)


sweinstein22:
Hi @johha ,

Thanks so much for the PR! One minor question, it seems like we've eased an assertion in moving from `eq` to an `include` in the ` it 'forwards the response from the rate limiter'` test. It feels like it would be beneficial to keep at least one equality test, to guard against any potential accidental header inclusions in the future. What would you think about either moving back to an `eq` and including the `X-USER-GUID` header in that test, or changing the `it 'sets the X-USER-GUID header'` assertion to say `it 'adds the X-USER-GUID header'` and moving to an `eq` in that case?

Thanks,
Sarah & @MarcPaquette

johha:
Hi @sweinstein22 @MarcPaquette 

thanks for your suggestions. I adjusted the tests but now the CI jobs are failing. Locally all tests are green. Could you have a look?

Thanks,
Johannes

sethboyles:
The failing tests are unrelated flakes, so I am going to merge this in.  Thanks for the PR!

PR #2169, opened 2021-03-26
Introduce LRP Custom Resources for Apps

mnitchev:
# A short explanation of the proposed change:

This PR introduces a flag to use the LRP CR for Eirini and implements a client that creates an LRP CR in Kubernetes. This PR implements **only** the `create` operation and uses the REST API for everything else. The flag should be `false` by default, because some things like scaling will be broken until we implement the the whole client, hence the `experimental` prefix to the flag. Our intention is to avoid a big PR and merge smaller chunks when we're ready with them. 

# An explanation of the use cases your change solves

Eventually the LRP custom resource will replace the REST API and it will be deprecated in Eirini.

# Link to Story in the Eirini Tracker
https://www.pivotaltracker.com/story/show/176919034

# Checklist

cc @cloudfoundry/eirini


lib/kubernetes/eirini_client.rb
```diff
@@ -0,0 +1,27 @@
+require 'steno'
+require 'kubernetes/kube_client_builder'
+require 'cloud_controller/errors/api_error'
+
+module Kubernetes
```
jspawar, line 5: Would be preferable if we put these methods instead in `Kubernetes::ApiClient` for consistency (all of our other Kubernetes interactions currently go through that class)

Would be happy to discuss refactoring that after though if we'd like.

jspawar, line 5: Would be preferable if we put these methods instead in `Kubernetes::ApiClient` for consistency (all of our other Kubernetes interactions currently go through that class)

jspawar:
Noticed that the box indicating if y'all ran CATS wasn't checked; wondering then where and how you tested this beyond the unit test level?

lib/cloud_controller/opi/apps_client.rb
```diff
@@ -8,252 +8,107 @@
...
+            cpuWeight: cpu_weight,
+            diskMB: process.disk_quota,
+            health: {
+              type: process.health_check_type,
+              timeoutMs: timeout_ms
+            },
+            lastUpdated: process.updated_at.to_f.to_s,
+            volumeMounts: generate_volume_mounts(process),
+            ports: process.open_ports,
+            appRoutes: OPI.routes(process),
```
tcdowney, line 80: Does this stay in sync as the user maps and unmaps routes for a running LRP? I believe the same code that does it for Diego LRPs should still be doing it in this case, but am wondering if you've tested this.

lib/kubernetes/eirini_client.rb
mnitchev: Alright, moved to `Kubernetes::ApiClient`

lib/cloud_controller/opi/apps_client.rb
mnitchev: It wouldn't update, since that requires the update to be implemented with CRs, which we'll do in a future PR. The `OPI.routes` function is just extracted in the OPI module, but still calls `VCAP::CloudController::Diego::Protocol::RoutingInfo.new(process).routing_info` to get the routes

mnitchev:
> @mnitchev Thanks for the PR! Noticed that the box indicating if y'all ran CATS wasn't checked; wondering then where and how you tested this beyond the unit test level?

Hello @jspawar we've just added some integration tests in `spec/integration` for the eirini kube client, which we run in our CI. These tests will be expanded as we complete more stories (see [this epic in our tracker](https://www.pivotaltracker.com/epic/show/4739977)). 

Running CATs with this feature turned on wouldn't be possible with _just_ this PR. We'd need to implement the whole client using CRs before we could do that. We decided to introduce the change in small instalments (just the creation now), because we think it would be easier to review for you and easier for us to accept.

Like I said in the description, this feature is behind an "off by default" flag and shouldn't interfere with the regular REST API flow, but we'd also understand if you don't want to merge the PR in by itself.

Birdrock:
Are we expecting the `spec/integration/eirini_k8s_client_spec.rb` tests to be pending? I suspect that Github actions needs to have the env var (`CF_RUN_EIRINI_SPECS`) set to run them.

spec/integration/eirini_k8s_client_spec.rb
```diff
@@ -0,0 +1,108 @@
+require 'spec_helper'
+
+skip_eirini_tests = ENV['CF_RUN_EIRINI_SPECS'] != 'true'
```
Birdrock, line 3: These tests don't appear to run in automation. If we add this var, they will fail without a kubeconfig. How are you testing this? We'd like to ensure test coverage, but it may not be appropriate to run it on the CC baseline.

Birdrock (commented):
Conditionally looks pretty good. I'd like to ensure we have a testing strategy in place before we merge this, however.

spec/integration/eirini_k8s_client_spec.rb
heycait: We will leave these tests turned off. We would recommend adding acceptance tests to https://github.com/cloudfoundry/capi-bara-tests to validate the CRDs are created and are correct. See [an example here](https://github.com/cloudfoundry/capi-bara-tests/blob/main/baras/route_crds.go). If these tests still provide value after adding to capi-bara-tests, we think it may be better to turn them into unit tests and mock out the actual k8s requests using webmock.

heycait:
For transparency, we will merge these PRs as they come in.

jspawar:
@cloudfoundry/eirini changes here remain fine; however, we realized that we're missing necessary RBAC rules on the `cf-for-k8s`/`capi-k8s-release` side which actually allows the various CAPI Deployments to create LRPs. Was wondering if y'all could please provide changes along with your next PR to introduce whatever requisite RBAC is needed?

Thank you!

PR #2195, opened 2021-04-14
Strictly validate page and per_page pagination options

### LINKED ISSUE #2194: Paginated v3 endpoints do not properly validate that the page and per_page query parameters are integers
## Issue
Paginated `v3` endpoints do not properly validate that the `page` and `per_page` query parameters are integers

## Context
When querying a paginated `v3` endpoint, the page parameter is only loosely validated. Values such as `page=1.5` or `page=1.abcd` are accepted (and converted to integer 1)

## Steps to Reproduce
`cf curl /v3/organizations?page=1.abcd`

## Expected result
400 Bad Request with useful error message stating that `page` should be a positive integer

## Current result
The first page of results is returned

## Possible Fix
Strictly validate the `page` and `per_page` query parameters

andy-paine:
Existing validation just compared the value of the input `.to_i` against
the limits. This method means that floats and strings such as `1.abcd`
get treated as valid integers and so are accepted by the API. By using
the Rails validation methods we can get stricter controls over what
is allowable and save writing custom validation functions

Fixes #2194 

* Use Rails builtin validation for v3 pagination query params

An explanation of the use cases your change solves
* Users submitting non-integer `page` parameters should get 400 errors rather than fuzzy integer casts

### DIFF
--- a/app/messages/list_message.rb
+++ b/app/messages/list_message.rb
@@ -52,6 +52,7 @@ module VCAP::CloudController
 
         validate_page(page, record) if !page.nil?
         validate_per_page(per_page, record) if !per_page.nil?
+        validate_maximum_result(page, per_page, record) if !page.nil? || !per_page.nil?
       end
 
       def validate_page(value, record)
@@ -63,6 +64,12 @@ module VCAP::CloudController
           record.errors.add(:per_page, 'must be between 1 and 5000') unless value.to_i <= 5000
         else
           record.errors.add(:per_page, 'must be a positive integer')
+        end
+      end
+
+      def validate_maximum_result(page, per_page, record)
+        if (page || PaginationOptions::PAGE_DEFAULT).to_i * (per_page || PaginationOptions::PER_PAGE_DEFAULT).to_i > 9223372036854775807
+          record.errors.add(:maximum_result, '(page * per_page) must be less than 9223372036854775807')
         end
       end
     end
--- a/spec/unit/messages/list_message_spec.rb
+++ b/spec/unit/messages/list_message_spec.rb
@@ -65,6 +65,35 @@ module VCAP::CloudController
       end
     end
 
+    describe 'maximum_result' do
+      it 'is valid if page and per_page is nil' do
+        message = ListMessage.from_params({ page: nil, per_page: nil }, [])
+        expect(message).to be_valid
+      end
+
+      it 'is valid when using default page' do
+        message = ListMessage.from_params({ page: nil, per_page: 5000 }, [])
+        expect(message).to be_valid
+      end
+
+      it 'is valid when using default per_page' do
+        message = ListMessage.from_params({ page: 20, per_page: nil }, [])
+        expect(message).to be_valid
+      end
+
+      it 'is invalid when page * default per_page exceeds DB limits' do
+        message = ListMessage.from_params({ page: 184467440737095517, per_page: nil }, [])
+        expect(message).to be_invalid
+        expect(message.errors[:maximum_result]).to include('(page * per_page) must be less than 9223372036854775807')
+      end
+
+      it 'is invalid when page * per_page exceeds DB limits' do
+        message = ListMessage.from_params({ page: 9223372036854775807, per_page: 2 }, [])
+        expect(message).to be_invalid
+        expect(message.errors[:maximum_result]).to include('(page * per_page) must be less than 9223372036854775807')
+      end
+    end
+
     describe 'order validations' do
       context 'when order_by is present' do
         it 'validates when order_by is `created_at`' do


lib/cloud_controller/paging/pagination_options.rb
```diff
@@ -7,29 +7,49 @@ class PaginationOptions
...
-    attr_writer :order_by, :order_direction, :default_order_by
-    attr_accessor :page, :per_page
+    attr_writer :order_by, :order_direction, :default_order_by, :page, :per_page
+
+    validates_numericality_of :page_string,
+      allow_nil: true,
+      only_integer: true,
+      greater_than_or_equal_to: 1,
+      message: 'must be a positive integer'
+    validates_numericality_of :per_page_string,
```
sethboyles, line 20: why validate on `per_page_string` rather than `per_page`?

app/messages/list_message.rb
```diff
@@ -25,6 +27,7 @@ def initialize(params={})
       params = params.symbolize_keys
       @pagination_params = params.slice(*ALLOWED_PAGINATION_KEYS)
       @pagination_options = PaginationOptions.from_params(params)
+      invalid_param!(@pagination_options.errors.full_messages) unless @pagination_options.valid?
```
sethboyles, line 30: A Message class shouldn't need to require a controller or have access to the `invalid_param!` method.  Generally, messages should be informative while the controller actually decides what to do with any errors reported by the message.  Please see https://github.com/cloudfoundry/cloud_controller_ng/wiki/Notes-on-V3-Architecture#controllers for more details

spec/unit/lib/cloud_controller/paging/pagination_options_spec.rb
```diff
@@ -115,31 +115,66 @@ module VCAP::CloudController
...
+        expect(pagination_options.per_page).to eq(per_page.to_i)
         expect(pagination_options.order_by).to eq(order_by)
         expect(pagination_options.order_direction).to eq(order_direction)
       end
     end
 
     describe 'validations' do
       context 'when page is not an integer' do
-        let(:params) { { page: 3.5 } }
+        let(:params) { { page: '3.5' } }
```
sethboyles, line 127: what's the behavior for a case like `page: '3.0'` ?

sethboyles:
Hi @andy-paine , 

Thank you for the PR!

It's a bit unfortunate this behavior this PR fixes has existed for so long and affects so much of the codebase.  I am wondering if this might be a breaking change for any existing clients which have, for better or worse, relied on the loose validation we have had for pagination options.  

Can you explain a little bit as to why you would like to see this fixed?

andy-paine:
@sethboyles Thanks for your comment on the PR.

Initially we wanted to change this because we see 500 Errors when `page multiplied by `per_page` is exceeding the database bigint limit. Just on coincidence we also experienced that the unittests of the page parameter validation is reversed. Instead of testing that no string can be passed it tested that a string can be passed.
So we also tried to make the parameter validation follow the documented behaviour.
Understandably a user that supplies a float like `3.0` may be hit by the change.
However both parameters are explicitly documented as an integer https://v3-apidocs.cloudfoundry.org/version/3.99.0/#get-an-app.

So we can now either fix both things (validate parameters as documented and enforce maximum)
or we can just add the validation for the maximum value and keep strings allowed.

It would feel more correct if a user is really just able to pass an integer instead of floats and strings.

lib/cloud_controller/paging/pagination_options.rb
andy-paine: The validator tests a function however since we call to_i we need to validate the value of page before it gets casted to an integer.

app/messages/list_message.rb
andy-paine: That's really useful to know, thanks for the link. Looks like we can just add a validator to the list message class that will call the pagination options validate method and add in any errors found so we don't need to use the `invalid_param` method and it is up to the calling controller to throw the correct exception 👍

spec/unit/lib/cloud_controller/paging/pagination_options_spec.rb
andy-paine: Looks like this is invalid too which is a bit unfortunate but seems to be the behaviour of the `only_integer` flag to the validator

MarcPaquette:
Hi @andy-paine,

We took a look at this PR, and there's quite a bit of goodness here.  We see  value in the limiting of the page size and feel it adds to the stability of the CC overall.  The input sanitation is a bit concerning for us as our users have reacted in a negative way to breaking changes to existing behavior without ample notice. 

With all that being said, would it be possible to scale back this code to just include the page limits, and omit the input validation till another time when such changes are necessary to migitage a specific pain point for users?

Thanks,
@sweinstein22  & @MarcPaquette

andy-paine:
Hi @MarcPaquette - thanks for taking a look! We are really just looking to remove the 500 errors so we can drop the strict validation if needed. Rather than reinvent the wheel of validation, I've just added a new function and some tests to the existing validation to check that the page sizes aren't enormous. This should stop the DB getting queried with an `OFFSET` that it can't handle which should hopefully remove the 500 error.

Issue #1591, opened 2020-04-07
"cf service" slow when user is member of many orgs


stephanme:
## Issue

"cf service" slow when user is member of many orgs.

Users complain about really slow "cf service <myservice>" calls. Execution takes sometimes >5min.
Execution time seems to be related to the number of orgs that the user is member of. We have users with access to several thousand orgs.

## Context

CF 12.36
CF API 2.146.0
cf version 6.50.0+4f0c3a2ce.2020-03-03

Our CF foundation is really large: ~800 cells, 90 CC, 37k app instances.
CC VMs are well loaded but not overloaded according to our monitoring.
CC DB: AWS RDS db.m5.12xlarge, 20..40% CPU

## Steps to Reproduce

This is not exactly 'steps to reproduce' but more an analysis of a 'cf service postgres01' call.
The user in this example has access to 35 orgs.

## Expected result

Execution is fast (a few seconds).

## Current result

Execution took 52s.

I traced the CF API calls and the involved DB queries:

GET /v2/spaces/b79d0289-75dc-47bd-9460-7283508db450/service_instances?q=name%3Apostgres01&return_user_provided_service_instances=true HTTP/1.1
X-Vcap-Request-Id: 34b1ecb0-f22d-4ca8-473f-012d93d29f33::20e6abf8-4e51-4edc-8756-1c2e813812c9
- 13s
- results in 1 service instance (d1703063-df67-4952-ba09-8f615db92771)
- 51 DB queries
- 35x SELECT * FROM "service_plan_visibilities" WHERE ("service_plan_visibilities"."organization_id" = ?)
  - most time spent here (9s)
  - singly query took ~100ms

GET /v2/service_instances/d1703063-df67-4952-ba09-8f615db92771/shared_to HTTP/1.1
X-Vcap-Request-Id: 76563e6b-687b-4157-6671-17f584110e51::8af040ac-0009-441b-89c6-fc7b0a7b9102
- 2s
- 11 DB queries

GET /v2/service_plans/eefb5965-d893-4377-805d-876898887a28 HTTP/1.1
X-Vcap-Request-Id: 22465555-4ded-4f7b-4cb9-ad5ab695b931::a77c1b10-f7a1-4dec-b79a-9b3f3a79560f
- 15s
- 1178 DB queries
- 1117x SELECT * FROM "service_plans" WHERE "id" = ?
  - took ~4s
  - single queries are fast
  - no delay between queries (= processing CC was fast)
[...]

PR #2215, opened 2021-04-21
Speed up v3 service plans by removing eager loading of visibilities

### LINKED ISSUE #2213: /v3/service_plans endpoint is slow in deployments with large numbers of service plan visibilities
## Issue
The performance of the `/v3/service_plans` in foundations with large numbers of organizations (and thus large numbers of service plan visibilities) is significantly slower than the `/v2/service_plans` endpoint

## Context
(This is a continuation of parts of #1591, using the numbers in [this comment](https://github.com/cloudfoundry/cloud_controller_ng/issues/1591#issuecomment-733059916))
For a large CF deployment with many orgs and brokers, the `/v3/service_plans` endpoint is up to 25x slower than the `/v2/service_plans` endpoint. This means that for a user using the CLI v6, `cf marketplace` takes ~4 seconds but a CLI v7 (thus /v3 user) `cf marketplace` takes ~90 seconds. 

## Steps to Reproduce
* Seed the DB with the orgs, spaces, service brokers, services, service plans and service visibility numbers from [this comment](https://github.com/cloudfoundry/cloud_controller_ng/issues/1591#issuecomment-733059916)
* Run `cf curl /v2/service_plans` and time it
* Run `cf curl /v3/service_plans` and time it
* (Bonus) Run `cf curl /v2/service_plans?per_page=500` and time it
* (Bonus) Run `cf curl /v3/service_plans?per_page=500` and time it

## Expected result
`/v3` timings are at least comparable to `/v2` timings
`/v3` timings should not increase significantly when querying more results per page (`/v2` timings don't)

## Current result
`/v3` timings are 10-100x slower than `/v2`
`/v3` timings scale roughly linearly with `per_page` size

## Possible Fix
We've done a lot of investigation into this, full context will be in a PR to follow. The basic explanation is that the service plan visibilities table is so large since it is `~ NUM_ORGS * NUM_SERVICE_PLANS` that eagerly fetching results from it is extremely slow

andy-paine:
## Overview
Remove eager loading of service plan visibilities when fetching the list of service plans

> Note: there are 2 commits here - the 1st one just adds a helpful rake task for populating the DB and isn't critical to the behaviour, we just included it to help test locally/test in the future. If whoever reviews this would rather just take the functional second commit please let us know

Users trying to query `cf marketplace` against deployments with large numbers of organisations and a reasonably large number of brokers/services/plans which thus have an enormous number of service plan visibilities

Fixes #2213 

## More info
After a bunch of investigation we found that the service plan visibilities table was causing issues when eagerly loaded into the service plans endpoint, particularly when listing service plans (as `cf marketplace` does). Whilst the query that the ORM executes (`SELECT * FROM service_plan_visibilities WHERE service_plan_id IN (<long list of IDs>)`) isn't that expensive for a DB, it returns >10k rows in many cases and the process of parsing that to Ruby objects and linking it to the correct plan does seem to be pretty slow.

The only usage of this data in this endpoint appears to be [checking whether there are any visibilities](https://github.com/cloudfoundry/cloud_controller_ng/blob/9c796d83da5723d576e8430fa30186c063e6fdcb/app/models/services/service_plan.rb#L167-L175) for this plan and only in the case where the plan is not public or not space-scoped. By _not_ eagerly loading this table, the ORM instead issues a `SELECT 1 ... WHERE service_plan_id = '1234'` for each plan in the list but as this a) is only called if it is not public or space scoped and b) returns much less data it is far faster than the eager loading still. _Potentially_, this may remain true in environments with **significant** DB latency as more, smaller requests are needed but as the numbers below will say it would have to be **really** bad to be worse than it currently is.

## Tests
Setting up a database as per [this comment](https://github.com/cloudfoundry/cloud_controller_ng/issues/1591#issuecomment-733059916):
* 22,000 orgs
* 66,000 spaces
* 300 service brokers
* 600 services
* 1200 service plans
* 660,000 service plan visibilities (all plans have worst-case scenario of being organization visible)

We got the following timings for querying the `service_plan` endpoints before and after this change:
* `/v2/service_plans?results-per-page=50`  - 0.12s
* `/v2/service_plans?results-per-page=100` - 0.17s
Before change:
* `/v3/service_plans?per_page=50` - 1.24s
* `/v3/service_plans?per_page=100` - 2.21s
* `/v3/service_plans?per_page=5000` - 30.20s
After change:
* `/v3/service_plans?per_page=50` - 0.16s
* `/v3/service_plans?per_page=100` - 0.22s
* `/v3/service_plans?per_page=5000` - 1.72s

Unfortunately this hasn't actually done much for CLI users yet as there seems to be something related to the `space_guids` filter (it hasn't made it worse) but at the very least this clearly makes the basic endpoint much faster for other users (e.g. Stratos, AppsMan)

## Checklist


sethboyles:
Interesting!  Thanks for the detailed investigation.  

If the including the n+1 ends up being slow for some cases, we could consider creating a new association on the ServicePlan model `:organization, through: :service_plan_visibilities`, which would allow us to check the same information but reduce the scope from Org count * Service Plan count to just Org count.  However, given that this query isn't needed for every ServicePlan, it still might not be worth it.

andy-paine:
After doing some more profiling with `rbspy`, it looks like the majority of the time is spent parsing the `created_at` and `updated_at` columns from the `service_plan_visibilities` table. If we don't want to remove eager loading entirely I think we could probably just add a [:select](https://sequel.jeremyevans.net/rdoc/files/doc/association_basics_rdoc.html#label-3Aselect) to the visibilities association in service plans.

I tested this locally and it is not as big an improvement as removing the eager loading entirely but it is still ~6x faster I think (early numbers). That would suggest that even the worst case "I need to run 5000 `SELECT 1` queries" is better but I think it is interesting to note that most of the processing time when parsing that table is dealing with the datetimes.

sethboyles:
Closing this as we cherry-picked the relevant commit: https://github.com/cloudfoundry/cloud_controller_ng/commit/2e363c69ecaa410ae713ca6a042928ef84ac05c6

We tested this out on one of our environments and everything looked good.  FWIW we did quickly try a version of my above suggestion of eager-loading a newly defined `organizations` association and did not see a performance improvement.  But I think we discovered why there was still an `n+1`, even after eager loading `service_plan_visibilities`--we believe it is because this method: 

https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/models/services/service_plan.rb#L172

calls `service_plan_visibilities_dataset` and not `service_plan_visibilities`.

as a side note there is probably a minor perf improvement to call `service_plan_visibilities_dataset.count > 0` instead of `empty?`, since `empty?` still loads the ruby models into memory and `count` executes a SQL `COUNT` statement.   

Again, thanks for this investigation and PR!

Issue #2227, opened 2021-04-22
Consider returning empty usage data for /v3/processes/:guid/stats


tcdowney:
Thanks for submitting an issue to `cloud_controller_ng`. We are always trying to improve! To help us, please fill out the following template.

## Issue

When `log-cache` (or `metric-proxy` on cf-for-k8s) is unavailable or returns a bad response for a particular process the entire `/v3/processes/:guid/stats` endpoint will exit with an error. 

## Context

This behavior is unfortunate for clients who do not care about process metrics and instead are using the endpoint for other information (e.g. is my process running? what ports is my process listening on?). Ideally the endpoint could continue to respond with the information it _is_ able to provide when the platform is degraded.

## Steps to Reproduce

On cf-for-k8s you can easily reproduce this behavior by running `cf restart` on an app in a loops. It fails roughly every 10 restarts since `metric-proxy` has difficulty fetching metrics from the Kubernetes API while pod containers are being churned.

More easily you can reproduce this by explicitly raising an error in the log-cache client in this method:
https://github.com/cloudfoundry/cloud_controller_ng/blob/637fe7d6449a593e89fc344d0745aadfad221f5c/lib/logcache/client.rb#L32

## Expected result

**Request:**
```
curl "https://api.example.org/v3/processes/[guid]/stats" \
  -X GET \
  -H "Authorization: bearer [token]"
```

**Response:**
```json
HTTP/1.1 200 OK
Content-Type: application/json

{
  "resources": [
    {
      "type": "web",
      "index": 0,
      "state": "RUNNING",
      "usage": { },
      "host": "10.244.16.10",
      "instance_ports": [
        {
          "external": 64546,
          "internal": 8080,
          "external_tls_proxy_port": 61002,
          "internal_tls_proxy_port": 61003
        }
      ],
      "uptime": 9042,
      "mem_quota": 268435456,
      "disk_quota": 1073741824,
      "fds_quota": 16384,
      "isolation_segment": "example_iso_segment",
      "details": null
    }
  ]
}
```

or

```json
HTTP/1.1 200 OK
Content-Type: application/json

{
  "resources": [
    {
      "type": "web",
      "index": 0,
      "state": "RUNNING",
      "usage": {
        "time": "2016-03-23T23:17:30.476314154Z",
        "cpu": null,
        "mem": null,
        "disk": null
      },
      "host": "10.244.16.10",
      "instance_ports": [
        {
          "external": 64546,
[...]

PR #2250, opened 2021-04-28
Support partial Process Stats responses when the stats service (log-cache or metric-proxy) is unavailable

tcdowney:
This change rescues errors from log-cache and metric-proxy and passes them up as warnings to callers instead of exceptions so that the stats endpoint can continue to function on a best-effort basis. On /v3/processes/:guid/stats these warnings are included in a new "warnings" key on the response, on v2 they are ignored. Alternatively/additionally we could add these to the `'X-Cf-Warnings'` header as @sethboyles [mentioned on the issue](https://github.com/cloudfoundry/cloud_controller_ng/issues/2227#issuecomment-828039303).

The related issue has more information: https://github.com/cloudfoundry/cloud_controller_ng/issues/2227

### DIFF
--- a/app/controllers/runtime/stats_controller.rb
+++ b/app/controllers/runtime/stats_controller.rb
@@ -22,7 +22,12 @@ module VCAP::CloudController
       end
 
       begin
-        stats, _ = instances_reporters.stats_for_app(process)
+        stats, warnings = instances_reporters.stats_for_app(process)
+
+        warnings.each do |warning_message|
+          add_warning(warning_message)
+        end
+
         stats.each_value do |stats_hash|
           if stats_hash[:stats]
             stats_hash[:stats].delete_if { |key, _| key == :net_info }
--- a/app/controllers/runtime/stats_controller.rb
+++ b/app/controllers/runtime/stats_controller.rb
@@ -22,7 +22,7 @@ module VCAP::CloudController
       end
 
       begin
-        stats = instances_reporters.stats_for_app(process)
+        stats, _ = instances_reporters.stats_for_app(process)
         stats.each_value do |stats_hash|
           if stats_hash[:stats]
             stats_hash[:stats].delete_if { |key, _| key == :net_info }
--- a/docs/v3/source/includes/resources/processes/_stats_object.md.erb
+++ b/docs/v3/source/includes/resources/processes/_stats_object.md.erb
@@ -12,6 +12,7 @@ Name | Type | Description
 **type** | _string_ | Process type; a unique identifier for processes belonging to an app
 **index** | _integer_ | The zero-based index of running instances
 **state** | _string_ | The state of the instance; valid values are `RUNNING`, `CRASHED`, `STARTING`, `DOWN`
+**usage** | _object_ | Object containing actual usage data for the instance; the value is `{}` when usage data is unavailable
 **usage.time** | _[timestamp](#timestamps)_ | The time when the usage was requested
 **usage.cpu** | _number_ | The current cpu usage of the instance
 **usage.mem** | _integer_ | The current memory usage of the instance
@@ -19,8 +20,8 @@ Name | Type | Description
 **host** | _string_ | The host the instance is running on
 **instance_ports** | _object_ | JSON array of port mappings between the network-exposed port used to communicate with the app (`external`) and port opened to the running process that it can listen on (`internal`)
 **uptime** | _integer_ | The uptime in seconds for the instance
-**mem_quota** | _integer_ | The current maximum memory allocated for the instance
-**disk_quota** | _integer_ | The current maximum disk allocated for the instance
+**mem_quota** | _integer_ | The current maximum memory allocated for the instance; the value is `null` when memory quota data is unavailable
+**disk_quota** | _integer_ | The current maximum disk allocated for the instance; the value is `null` when disk quota data is unavailable
 **fds_quota** | _integer_ | The maximum file descriptors the instance is allowed to use
 **isolation_segment** | _string_ | The current isolation segment that the instance is running on; the value is `null` when the instance is not placed on a particular isolation segment
 **details** | _string_ | Information about errors placing the instance; the value is `null`  if there are no placement errors
--- a/lib/cloud_controller/diego/reporters/instances_stats_reporter.rb
+++ b/lib/cloud_controller/diego/reporters/instances_stats_reporter.rb
@@ -20,22 +20,9 @@ module VCAP::CloudController
         logger.debug('stats_for_app.fetching_container_metrics', process_guid: process.guid)
         desired_lrp = bbs_instances_client.desired_lrp_instance(process)
 
-        log_cache_data = envelopes(desired_lrp, process)
-        stats = log_cache_data.
-                map { |e|
-          [
-            e.containerMetric.instanceIndex,
-            converted_container_metrics(e.containerMetric, formatted_current_time),
-          ]
-        }.to_h
-
-        quota_stats = log_cache_data.
-                      map { |e|
-                        [
-                          e.containerMetric.instanceIndex,
-                          e.containerMetric,
-                        ]
-                      }.to_h
+        log_cache_data, log_cache_errors = envelopes(desired_lrp, process)
+        stats = formatted_process_stats(log_cache_data, formatted_current_time)
+        quota_stats = formatted_quota_stats(log_cache_data)
 
         bbs_instances_client.lrp_instances(process).each do |actual_lrp|
           index = actual_lrp.actual_lrp_key.index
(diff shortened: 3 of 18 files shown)


lib/cloud_controller/diego/reporters/instances_stats_reporter.rb
```diff
@@ -92,15 +121,16 @@ def envelopes(desired_lrp, process)
...
 
-        @logstats_client.container_metrics(
+        return @logstats_client.container_metrics(
           source_guid: source_guid,
           auth_token: VCAP::CloudController::SecurityContext.auth_token,
           logcache_filter: filter
-        )
+        ), nil
+      rescue GRPC::BadStatus, CloudController::Errors::ApiError => e
+        logger.error('stats_for_app.error', error: e.message, backtrace: e.backtrace.join($INPUT_RECORD_SEPARATOR))
```
sethboyles, line 130: what does this global variable do?

sethboyles:
so far looks good to me.  Is the non-idiomatic part of this you mentioned the multiple return values?  If so I don't have any qualms with it.

lib/cloud_controller/diego/reporters/instances_stats_reporter.rb
tcdowney: It just joins them with the system's concept of what the newline separator should be. I think it's always `"\n"` in our case and it's probably better if I just explicitly declare that.

tcdowney:
> so far looks good to me. Is the non-idiomatic part of this you mentioned the multiple return values? If so I don't have any qualms with it.

Yeah, that was mostly it. I don't think we've done much of it in the codebase so wanted opinions. I think it's better than trying to juggle the exceptions all the way up to the controller in this case, though.

tcdowney:
@sethboyles do you think we should go with the warning header instead of the `"warnings"` field in the response? I could see that being a better UX out of the box since the CLI already supports it.

sethboyles:
I think for now that might be the better option.  I am uncertain about the `"warnings"` property as it has only been used for jobs in the past, and I am wary of introducing it to other parts of the API without giving it enough thought or discussion. 
 Clients probably don't have existing ways of consuming non-job related warnings, and I'm unsure its a pattern we want to adopt generally as we probably don't want to allow partial successes/failures when we can avoid it.  

We are going to discuss warnings in general as a team tomorrow, but I don't want to block this PR--how about we go with the header warning for now, and if we decide later on to use the JSON field, we can put it in our backlog?

tcdowney:
Sounds good. I'll make that change. 👍

tcdowney:
## Validating Steps
Here's how I tested this out manually.

### CF for VMs
1.  `cf push dora` and observe the metrics work
2. `cf restart dora` and observe the metrics work
3. `bosh ssh doppler`
4. `monit stop log-cache`
5. `cf restart dora` and observe that the workflow succeeds even though metrics are missing. My output looked like this:

```console
$ cf restart dora
Restarting app dora in org org / space space as admin...

Stopping app...

Waiting for app to start...

Instances starting...
Instances starting...
Instances starting...
Instances starting...

Stats server temporarily unavailable.
Stats server temporarily unavailable.
Stats server temporarily unavailable.
Stats server temporarily unavailable.
Stats server temporarily unavailable.
No running instances found for process guid 63461f63-0ab9-45c0-bc16-482af9179d3c-7219907a-a412-45cd-a4f8-8fe6bd6477f6
name:              dora
requested state:   started
routes:            dora.diamond-legend.capi.land
last uploaded:     Thu 29 Apr 09:20:20 PDT 2021
stack:             cflinuxfs3
buildpacks:
	name             version   detect output   buildpack name
	ruby_buildpack   1.8.37    ruby            ruby

type:           web
sidecars:
instances:      1/1
memory usage:   256M
     state     since                  cpu    memory   disk     details
#0   running   2021-04-29T16:22:54Z   0.0%   0 of 0   0 of 0

type:           worker
sidecars:
instances:      0/0
memory usage:   256M
There are no running instances of this process.
```

Directly hitting the endpoint returns:
```console
cf curl /v3/processes/8f79dd94-2f93-427c-b66c-9290b81db79d/stats
{
   "resources": [
      {
         "type": "web",
         "index": 0,
         "state": "RUNNING",
         "host": "10.244.0.138",
         "uptime": 258,
         "mem_quota": null,
         "disk_quota": null,
         "fds_quota": 16384,
         "isolation_segment": null,
         "details": null,
         "instance_ports": [
            {
               "external": 61004,
               "internal": 8080,
               "external_tls_proxy_port": 61006,
               "internal_tls_proxy_port": 61001
            },
            {
               "external": 61005,
               "internal": 2222,
               "external_tls_proxy_port": 61007,
               "internal_tls_proxy_port": 61002
            }
         ],
         "usage": {}
      }
   ]
}
```

`cf curl` also prints out the `Stats server temporarily unavailable.` warning after it finishes.

Even though the API is returning `null`/empty values,  CLI is reporting 0s for the metrics that is the default numeric value in Go. Since we're returning empty values on the API side, the CLI could be enhanced to display something more meaningful if it wanted to, but I think this is fine for now.

### CF for Kubernetes
On Kubernetes the CLI/API experience is the same, just the repro steps are a little different.
1.  `cf push dora` and observe the metrics work (on cf-for-k8s I always see the quotas missing even without this change)
2. `cf restart dora` and observe the metrics work (on cf-for-k8s I always see the quotas missing even without this change)
3. `kubectl -n cf-system scale deployment metric-proxy --replicas=0`
4. `cf restart dora` and observe that the workflow succeeds even though metrics are missing.

lib/cloud_controller/diego/reporters/instances_stats_reporter.rb
tcdowney: Made the change:
https://github.com/cloudfoundry/cloud_controller_ng/pull/2250/files#diff-ad5dff07755b8d42aeadbf674ff7947d38ac428c9e7e52e415a6194d07ba202eR130

tcdowney:
@sethboyles I rebased the original commit to use the `'X-Cf-Warnings'` header instead of adding `"warnings"` to the response. Also added a commit to remove a now unused config flag and some light-weight documentation to point out these fields can now be empty or `null` (the `mem_quota` and `disk_quota` ones could have been `null` before this change fwiw).

I'm gonna take this PR out of "draft" state now, I think it's ready. 🙂

app/controllers/runtime/stats_controller.rb
```diff
@@ -22,7 +22,7 @@ def stats(guid, opts={})
       end
 
       begin
-        stats = instances_reporters.stats_for_app(process)
+        stats, _ = instances_reporters.stats_for_app(process)
```
jspawar, line 25: Why do we ignore warnings in v2?

jspawar (approved):
Looks good, I don't quite understand why we ignore the errors in v2, although I'm guessing this is because v2 endpoints should not be modified and they don't support that warnings header?

app/controllers/runtime/stats_controller.rb
tcdowney: > I don't quite understand why we ignore the errors in v2, although I'm guessing this is because v2 endpoints should not be modified and they don't support that warnings header?

Yeah, I didn't want to change v2 too much. It does have code in place that would support adding the header, though: https://github.com/cloudfoundry/cloud_controller_ng/blob/4a23a419862ef9b7272f411a4435e33fc7305f30/app/controllers/base/base_controller.rb#L124

sethboyles: Sounds good to me!

docs/v3/source/includes/resources/processes/_stats_object.md.erb
```diff
@@ -12,15 +12,16 @@ Name | Type | Description
...
 **usage.cpu** | _number_ | The current cpu usage of the instance
 **usage.mem** | _integer_ | The current memory usage of the instance
 **usage.disk** | _integer_ | The current disk usage of the instance
 **host** | _string_ | The host the instance is running on
 **instance_ports** | _object_ | JSON array of port mappings between the network-exposed port used to communicate with the app (`external`) and port opened to the running process that it can listen on (`internal`)
 **uptime** | _integer_ | The uptime in seconds for the instance
-**mem_quota** | _integer_ | The current maximum memory allocated for the instance
-**disk_quota** | _integer_ | The current maximum disk allocated for the instance
+**mem_quota** | _integer_ | The current maximum memory allocated for the instance; the value is `null` when memory quota data is unavailable
+**disk_quota** | _integer_ | The current maximum disk allocated for the instance; the value is `null` when disk quota data is unavailable
```
sethboyles, line 24: 🙏  thank you for updating the docs!

sethboyles:
LGTM!

```
± sb |handle-log-cache-unavailability → origin {2} ?:1 ✗| → cf curl v3/apps/$(cf app dora --guid)/processes/web/stats
{
   "resources": [
      {
         "type": "web",
         "index": 0,
         "state": "RUNNING",
         "host": "10.244.0.138",
         "uptime": 1525,
         "mem_quota": null,
         "disk_quota": null,
         "fds_quota": 16384,
         "isolation_segment": null,
         "details": null,
         "instance_ports": [
            {
               "external": 61028,
               "internal": 8080,
               "external_tls_proxy_port": 61030,
               "internal_tls_proxy_port": 61001
            },
            {
               "external": 61029,
               "internal": 2222,
               "external_tls_proxy_port": 61031,
               "internal_tls_proxy_port": 61002
            }
         ],
         "usage": {}
      }
   ]
}
Stats server temporarily unavailable.
```

@tcdowney Let me know if/when you add the warnings to v2, then I'll go ahead and merge this in.  Thank you for the PR!

tcdowney:
```console
19:04 $ cf curl /v2/apps/$(cf app dora --guid)/stats
{
   "0": {
      "state": "RUNNING",
      "isolation_segment": null,
      "stats": {
         "name": "dora",
         "uris": [
            "dora.diamond-legend.capi.land"
         ],
         "host": "10.244.0.138",
         "port": 61004,
         "uptime": 62,
         "fds_quota": 16384,
         "mem_quota": null,
         "disk_quota": null,
         "usage": {}
      }
   }
}
Stats server temporarily unavailable.
```

Done! 👍

app/controllers/runtime/stats_controller.rb
tcdowney: 932e8ecac467ae656432a38f49a68b240c380a74

Issue #2210, opened 2021-04-20
Create Space Application Supporter role in ccdb and allow role to be assigned to a user


MerricdeLauney:
## Issue

Create Space Application Supporter role in ccdb and allow role to be assigned to a user.

## Context 
We are introducing a [new role](https://github.com/cloudfoundry/cfar-proposals/issues/22) and we want to make sure it has the right access.

## Expected result

**As** admin or space manager
**I want** to assign the role Space Application Supporter to a user in my space
**So that** the user and space guid association is stored in a ccdb table titled to explain the relationship.

### Acceptance Criteria

```gherkin
Scenario:
Given I have a user with guid "user_guid" who has the role `organization_user` in an org containing a space with the guid "space_guid"
When I run 
cf curl v3/roles  -d '{
      "type": "space_application_supporter",
      "relationships": {
        "user": {
          "data": {
            "guid": "user-guid"
          }
        },
        "space": {
          "data": {
            "guid": "space-guid"
          }
        }
      }
    }'
Then I get a 201 response with a role returned and can see a table with a relevant title has a row populated with the user and space guid.
```

**Dev Notes:**
This should require a migration. Be careful naming the new table and try to match existing naming conventions very closely

### Documentation
When I browse to the [valid role types](https://v3-apidocs.cloudfoundry.org/version/3.99.0/index.html#valid-role-types) in v3 docs I see the new role type listed with an explanation that it is not to be used and is not fully implemented.

PR #2251, opened 2021-04-29
Space application supporter role

MerricdeLauney:
Implements requested feature from https://github.com/cloudfoundry/cloud_controller_ng/issues/2210


monamohebbi (commented):
An interesting side effect of this commit is deleting users will now fail with an unknown error because we haven't implemented a delete action for the space_application role:

`Sequel::ForeignKeyConstraintViolation: Mysql2::Error: Cannot delete or update a parent row: a foreign key constraint fails `

Is it alright to leave things in a semi broken state given we have the disable flag and properly documented that the role is under active development?

cc @MerricdeLauney @weymanf @mkocher @sweinstein22 @sethboyles

app/models.rb
```diff
@@ -19,7 +19,9 @@
 require 'models/runtime/app_usage_event'
 require 'models/runtime/auto_detection_buildpack'
 require 'models/runtime/app_event'
+require 'models/runtime/helpers/organization_role_mixin'
```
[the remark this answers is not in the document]
monamohebbi: Given inheritance isn't an option in the models where this mixin is being used, a mixin seems like the best solution for a lot of repeated code.

mkocher:
@elenasharma we fixed the deletion issue, so this should be set to go. Please take a look. /cc @monamohebbi

app/models/runtime/helpers/space_role_mixin.rb
```diff
@@ -0,0 +1,22 @@
+module VCAP::CloudController
+  module SpaceRoleMixin
+    def self.included(included_class)
+      included_class.class_eval <<-RUBY, __FILE__, __LINE__ + 1
```
sethboyles, line 4: Is there a reason to use this syntax rather than a do...end block?

sethboyles:
@mkocher tests are failing 👀

app/models/runtime/space_application_supporter.rb
```diff
@@ -0,0 +1,9 @@
+module VCAP::CloudController
+  class SpaceApplicationSupporter < Sequel::Model(:spaces_application_supporters)
+    include SpaceRoleMixin
+
+    def type
+      @type ||= RoleTypes::SPACE_APPLICATION_SUPPORTER
```
sethboyles, line 6: I know this was pattern matched from existing roles, but I don't think there is a benefit to saving a constant to an instance variable like this, is there?

weymanf:
There was some funky stuff going on with this branch that was missing a bunch of code but was sitting on another space supporter branch. Just cherry picked those over and also rebased onto main.

app/models/runtime/helpers/space_role_mixin.rb
weymanf: We had to do this to for the special rails associations methods.
`many_to_one` don't translate the same as normal functions because they are rails magic functions that will eventually inject association methods.

Unless you are saying there is another special thing to inject these rail methods into our mixin?

weymanf: We could do it this way but I have no idea what extra stuff we get by adding ActiveSupport::Concern will give us.
```
module VCAP::CloudController
    module SpaceRoleMixin
        extend ActiveSupport::Concern

        included do
          many_to_one :user
          many_to_one :space

          def_column_alias :guid, :role_guid
        end
...
```

sethboyles (approved):

app/models/runtime/helpers/space_role_mixin.rb
sethboyles: We discussed this in the sync, but adding here for posterity:

`class_eval` can either take a string or block.  So this could be rewritten like so: 

```
included_class.class_eval do
     many_to_one :user
     many_to_one :organization
     def_column_alias :guid, :role_guid
end
```

However, I am not sure if you can still pass in the file/line number for the backtrace with the block method and how that behavior will be different.

monamohebbi (approved):

PR #2258, opened 2021-05-05
Do not set VCAP_APPLICATION env var during kpack staging

kieron-dev:

We have noticed that during kpack staging, VCAP_APPLICATION is set to a Hash#to_s string. It should be a json string, although in diego staging, VCAP_APPLICATION is not set at all.

This PR excludes VCAP_APPLICATION from the environment in the same way that VCAP_SERVICES is already excluded.

When deploying postfacto, for example, the rails-assets precompilation triggers `bundle exec rails assets:precompile` which ends up trying to [deserialise VCAP_APPLICATION from json](https://github.com/pivotal/postfacto/blob/89d4d5e07b9399e99fa963b6020547fba04d358b/api/lib/configurations/action_cable_host_provider.rb#L37). This fails as it gets the ruby hash string. In CF on VMs, VCAP_APPLICATION is not set and the staging succeeds.

### DIFF
--- a/lib/cloud_controller/kpack/stager.rb
+++ b/lib/cloud_controller/kpack/stager.rb
@@ -101,7 +101,7 @@ module Kpack
 
     def get_environment_variables(staging_details)
       staging_details.environment_variables.
-        except('VCAP_SERVICES').
+        except('VCAP_SERVICES', 'VCAP_APPLICATION').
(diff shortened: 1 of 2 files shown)


acosta11:
LGTM, thanks for the PR!

Thanks,
Andrew and @jspawar

PR #2288, opened 2021-05-20
Allow application supporters to use route endpoints

### LINKED ISSUE #2229: Allow space application supporter to access specific route endpoints.
## Issue

Allow space application supporter to access specific **route** endpoints. 

## Context

## Expected result

POST /v3/routes
GET /v3/routes
DELETE /v3/routes/:guid
GET /v3/domains/:guid/route_reservations
GET /v3/routes/:guid/destinations
GET /v3/apps/:guid/routes
POST /v3/routes/:guid/destinations
PATCH /v3/routes/:guid/destinations
DELETE /v3/routes/:guid/destinations/:destination_guid
DELETE /v3/spaces/:guid/routes?unmapped=true

## Acceptance
A space application supporter would see the same info as a space developer assigned to the same space for these and only these routes endpoints.

## Documentation

andy-paine:
The readable_org_guids_for_domains function takes a flag for now about
whether to include application supporters. All usages of this function
will eventually set this flag to true at which point it can be removed.

As some controllers start using the untrusted_* functions, the
user_helpers spec file needs to return the correct value for whichever
of these functions has been called. The tests that use this helper do
not need to know which of the two functions was called, only that the
permission will or will not be granted.

Added a has_application_supporter? public function to the space object
so that we can use untrusted_can_read_route? as the can_read_route?
function is used by v2 endpoints which we don't want to grant access to
for application supporters

Co-authored-by: Philipp Thun <philipp.thun@sap.com>

Closes #2229 

### DIFF
--- a/app/controllers/v3/domains_controller.rb
+++ b/app/controllers/v3/domains_controller.rb
@@ -56,7 +56,7 @@ class DomainsController < ApplicationController
     message = DomainShowMessage.new({ guid: hashed_params['guid'] })
     unprocessable!(message.errors.full_messages) unless message.valid?
 
-    domain = find_domain(message)
+    domain = find_domain(message, include_application_supporters: true)
     domain_not_found! unless domain
 
     check_route_params = to_route_list_params(query_params, domain)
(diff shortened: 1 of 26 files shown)


monamohebbi (approved):
LGTM

We have deployed changes to a test environment and ran through the basic commands with a [script](https://github.com/monamohebbi/RBAC_acceptance/blob/main/README.md
) that can be modified for any local environment to test changes as well. Now just waiting on the official maintainers to review and merge this as well.

sweinstein22 (approved):
Looks good, thanks for the PR!

PR #2293, opened 2021-05-25
Log service broker requests at info level

### LINKED ISSUE #2292: Service Broker HTTP requests are only logged at debug level
## Issue
Interactions with service brokers are only logged at debug levels

## Context
When a service broker is responding incorrectly, it is useful to be able to tell where in the Cloud Controller the requests to it originated from. This would help platform operators understand why a service or binding is in a certain state and the overall error rates of various brokers.

As service brokers represent fairly complex external systems that the platform operators may not have access to, being able to report what the Cloud Controller was trying to do is helpful when diagnosing integration issues with brokers. Debug logging is unfortunately far too noisy to be enabled the whole time so we would prefer this to be logged at either a different level or be able to enable debug logging specifically for service brokers.

## Steps to Reproduce
Enable `info` logging in CC - see no logs about service broker interactions
Enable `debug` logging in CC - see logs about requests and responses with service brokers

## Possible Fix
Raising this issue mostly to discuss approaches. As I see it there are a few different options we might take:
1. Change existing service broker client `debug` logging to `info` (affects all users)
2. Add additional (less rich) `info` level logging to the service broker client (affects all users)
3. Add a config option to include `debug` level logs specifically for the service broker client
4. Add a config option to log additional (less rich) `info` level logging in service broker client

Happy to contribute but would be good to get a steer on best direction given there is existing logging behaviour for the service broker client

andy-paine:
Service brokers represent complex external systems that can easily fail
so adding some logging at info level should allow operators to debug
failing brokers more easily. Both request and response log lines include
a request_guid (which is the VCAP request ID) from the Steno context for
tracking requests from both the CC and broker sides

Resolves #2292 

### DIFF
--- a/lib/services/service_brokers/v2/http_client.rb
+++ b/lib/services/service_brokers/v2/http_client.rb
@@ -69,9 +69,11 @@ module VCAP::Services
 
         headers = client.default_header.merge(opts[:header])
 
+        logger.info 'request', method: method, url: uri.to_s
         logger.debug "Sending #{method} to #{uri}, BODY: #{body.inspect}, HEADERS: #{headers}"
 
         response = client.request(method, uri, opts)
+        logger.info 'response', status: response.code
         logger.debug "Response from request to #{uri}: STATUS #{response.code}, BODY: #{redact_credentials(response)}, HEADERS: #{response.headers.inspect}"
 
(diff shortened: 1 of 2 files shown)


JenGoldstrich:
Looks good to me, thanks Andy!

Merging this now

Issue #2223, opened 2021-04-21
Allow space application supporter to access specific organizations endpoints.


ctlong:
## Issue

Allow space application supporter to access specific **organizations** endpoints. 

## Context

We are introducing a [new role](https://github.com/cloudfoundry/cfar-proposals/issues/22) and we want to make sure it has the right access.

## Expected result
A space application support should be able access the following endpoints:

GET /v3/organizations/:guid
GET /v3/organizations

## Acceptance
A space application supporter would see the same info as a space developer assigned to the same space for these and only these organizations endpoints.

sweinstein22:
`GET /v3/organizations/:guid/domains/default` and `GET /v3/orgaizations/:guid/usage_summary` have both been addressed as part of PR https://github.com/cloudfoundry/cloud_controller_ng/pull/2300

philippthun:
Is someone working on this one? I'm asking, as I cannot assign issues to myself and can only indicate that I am working on something by moving it to "In progress"...

sweinstein22:
Thanks for checking in @philippthun , this one has been spoken for - Elena is one of our engineering managers and anything she moves from "To do" into "In progress" is going to be slated for someone at VMware to eventually pick up. Would it be helpful for us to assign the issue to someone on our team just so you know it's been spoken for?

philippthun:
@sweinstein22 Ok, that's fine. No need to change anything on your side. Will will just don't touch anything in "In progress"...

ctlong:
It appears that
```
GET /v3/organization_quotas/:guid
GET /v3/organization_quotas
```
were resolved in #2387.

So the only endpoints that @moleske and I believe we need to update are
```
GET /v3/organizations/:guid
GET /v3/organizations
```
rather than
```
GET /v3/organizations/:guid
GET /v3/organizations
GET /v3/organizations/:guid/domains/default
GET /v3/organizations/:guid/usage_summary
GET /v3/organization_quotas/:guid
GET /v3/organization_quotas
```

We are going to update the issue description to reflect that.

PR #2300, opened 2021-05-27
Allow space application supporter to access specific domain endpoints

### LINKED ISSUE #2219: Allow space application supporter to access specific domain endpoints.
## Issue

Allow space application supporter to access specific **domain** endpoints. 

## Context

## Expected result

GET /v3/domains/:guid
GET /v3/domains
GET /v3/organizations/:guid/domains

## Acceptance
A space application supporter would see the same info as a space developer assigned to the same space for these and only these domain endpoints.

## Documentation

philippthun:
Closes #2219

### DIFF
--- a/app/controllers/v3/apps_controller.rb
+++ b/app/controllers/v3/apps_controller.rb
@@ -87,7 +87,7 @@ eager_loaded_associations: Presenters::V3::AppPresenter.associated_resources)
     unprocessable!(message.errors.full_messages) unless message.valid?
 
     space = Space.where(guid: message.space_guid).first
-    unprocessable_space! unless space && permission_queryer.can_read_from_space?(space.guid, space.organization_guid)
+    unprocessable_space! unless space && permission_queryer.untrusted_can_read_from_space?(space.guid, space.organization_guid)
     unauthorized! unless permission_queryer.can_write_to_space?(space.guid)
     # TODO: only fail if also not `kpack` app lifecycle
     if message.lifecycle_type == VCAP::CloudController::PackageModel::DOCKER_TYPE
@@ -244,7 +244,7 @@ eager_loaded_associations: Presenters::V3::AppPresenter.associated_resources)
 
     FeatureFlag.raise_unless_enabled!(:env_var_visibility)
 
-    app_not_found! unless app && permission_queryer.can_read_from_space?(space.guid, org.guid)
+    app_not_found! unless app && permission_queryer.untrusted_can_read_from_space?(space.guid, org.guid)
     unauthorized! unless permission_queryer.can_read_secrets_in_space?(space.guid, org.guid)
 
     FeatureFlag.raise_unless_enabled!(:space_developer_env_var_visibility)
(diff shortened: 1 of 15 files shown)


sweinstein22:
## Acceptance

Confirmed user only has space_application_supporter and organization_user roles:

```
$ cf curl /v3/roles?user_guids="d8685663-a98a-4c4d-86b0-4523dd45a8b1"
{
  ...
  "resources": [
    {
      "guid": "29ab067c-7205-4fe4-be13-bd63009fcc2a",
      "relationships": {
        "user": {
          "data": {
            "guid": "d8685663-a98a-4c4d-86b0-4523dd45a8b1"
          }
        },
      ...
    },
    {
      "guid": "f079f79a-81d9-4ec1-8229-051ddafc3c72",
      "type": "organization_user",
      "relationships": {
        "user": {
          "data": {
            "guid": "d8685663-a98a-4c4d-86b0-4523dd45a8b1"
          }
        },
      ...
    }
  ]
}
```

Ran `cf curl` against `/v3/domains/:guid`, `/v3/domains`, and `/v3/organizations/:guid/domains` as user and as a space developer, piped results to files

Confirmed that output of calls matched by checking for empty diff:
```
$ diff space_dev_domains space_app_supporter_domains

$ diff space_dev_domains_guid space_app_supporter_domains_guid

$ diff space_dev_org_guid_domains space_app_supporter_org_guid_domains
```

### Code Review:
Checked that `readable_org_guids_for_domains` is only being used in cases where the space_application_supporter role has been introduced, allowing us to remove the conditional statement
 - In doing so we're also changing permissions for the `GET /v3/organizations/:guid/domains/default` endpoint.
    - It looks like this was originally intended to be included in [this github issue](https://github.com/cloudfoundry/cloud_controller_ng/issues/2223), but given the work is already done in this PR I think it makes sense to include it here.

    - As the code change has already been implemented, the output between a space_application_supporter and a space_developer are already the same:
    ```
    $ diff space_dev_org_guid_domains_default space_app_supporter_org_guid_domains_default
   
     ```
    - The request spec for this endpoint has also been adjusted accordingly, but it looks like the documentation for the endpoint has not been updated.

Examined the request spec changes, looks like we've a specs around the `GET /v3/organizations/:guid/usage_summary` endpoint indicating they should receive a `200`, which is not included in this this issue.
  - Confirmed the output between a space_application_supporter and a space_developer are the same:
    ```
    diff space_dev_org_usage_summary space_app_supporter_org_usage_summary
    
    ```
  - Checked the docs and it's indicated as visible to All Roles, no further changes required
  - Similar to the default domain endpoint, it looks like it was slated to be addressed in [this github issue](https://github.com/cloudfoundry/cloud_controller_ng/issues/2223). I'll make note in that issue to indicate that the two endpoints have been addressed in this PR as the request spec has already been changed.

Checked that `find_domain` is only being used in cases where the space_application_supporter role has been introduced, allowing us to remove the conditional statement
- While `find_domain` is used in other endpoints that the space_application_supporter won't have access to, they're all guarded by checks for write permissions

sweinstein22 (changes requested):
### Changes Requested:
- Add documentation to `GET /v3/organizations/:guid/domains/default` indicating that the Space Application Supporter role has access to the endpoint

philippthun:
@sweinstein22 Thanks for reviewing the PR. I've added the missing documentation.

sweinstein22 (approved):
Perfect, thanks @philippthun !

PR #2317, opened 2021-06-07
Ensure that space application supporter can access user GET endpoints

### LINKED ISSUE #2236: Allow space application supporter to access specific user endpoints.
## Issue

Allow space application supporter to access specific **user** endpoints. 

## Context

## Expected result

GET /v3/users/:guid
GET /v3/users

## Acceptance
A space application supporter would see the same info as a space developer assigned to the same space for these and only these user endpoints.

## Documentation

philippthun:
- Add `space_application_supporter` to permission tests.
- Remove `set_current_user_as_role` from "when the user is logged in" context; for this test only the scope (`cloud_controller.read`) is important.

Closes #2236.

### DIFF
--- a/docs/v3/source/includes/resources/users/_get.md.erb
+++ b/docs/v3/source/includes/resources/users/_get.md.erb
@@ -36,5 +36,6 @@ Org Manager | Can only view users affiliated with their org
 Space Auditor | Can only view users affiliated with their org
 Space Developer | Can only view users affiliated with their org
 Space Manager | Can only view users affiliated with their org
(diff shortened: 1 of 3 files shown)


sweinstein22 (approved):
Ran `cf curl /v3/users` and `cf curl /v3/users/:guid` against an environment with latest CAPI to confirm that implementation changes were not required. Confirmed that output as a space dev and space app supporter were the same, and manually confirmed that output looked correct
```
$ diff space-dev-users space-app-supporter-users

$ diff space-dev-user-guid space-app-supporter-user-guid

```

Code changes and docs changes look good, thanks for the PR!

PR #2324, opened 2021-06-09
Allow Space Application Supporter access to specific Audit Events endpoints

### LINKED ISSUE #2216: Allow space application supporter to access specific audit event endpoints.
## Issue

Allow space application supporter to access specific **audit event** endpoints. 

## Context

## Expected result

GET /v3/audit_events/:guid
GET /v3/audit_events 

## Acceptance
A space application supporter would see the same info as a space developer assigned to the same space for these and only these audit endpoints.

## Documentation

ctlong:
Also refactored v3 Events permissions checks to be in the permissions queryer because we believe that to be the desired pattern + it seems easier to follow. We also updated the documentation.

Closes #2216

### DIFF
--- a/app/controllers/v3/events_controller.rb
+++ b/app/controllers/v3/events_controller.rb
@@ -7,7 +7,7 @@ class EventsController < ApplicationController
     message = EventsListMessage.from_params(query_params)
     invalid_param!(message.errors.full_messages) unless message.valid?
 
-    dataset = EventListFetcher.fetch_all(message, readable_event_dataset)
+    dataset = EventListFetcher.fetch_all(message, permission_queryer.readable_event_dataset)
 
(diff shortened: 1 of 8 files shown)


monamohebbi (approved):
LGTM. Deployed changes to environment candle-chopper. Created users louie and dev with roles space application supporter and space developer respectively. Both have password 'password'

sweinstein22 (approved):
Did a quick check to make sure the output matched up between space devs and space application supporters for the two endpoints, `GET /v3/audit_events/:guid` and `GET /v3/audit_events`. Confirmed that we now have permissions tests for both endpoints, and the refactor looks good!

Checked for the docs changes, everything looks good! Thanks for catching the stray alphabetization ordering

PR #2325, opened 2021-06-10
Space application supporter can access environment variable groups

### LINKED ISSUE #2221: Allow space application supporter to access specific environment variable group endpoints.
## Issue

Allow space application supporter to access specific **environment variable group** endpoints. 

## Context

## Expected result

GET /v3/environment_variable_groups/:name

## Acceptance
A space application supporter would see the same info as a space developer assigned to the same space for these and only these environment variable group endpoints.

## Documentation

will-gant:
* Add unit tests to prove that all roles can _get_ access environment variable groups and that only the admin role can _update_ environment variable groups (as documented [here](https://v3-apidocs.cloudfoundry.org/version/3.101.0/index.html#get-an-environment-variable-group) and [here](https://v3-apidocs.cloudfoundry.org/version/3.101.0/index.html#update-environment-variable-group))

Closes #2221

### DIFF
--- a/spec/request/environment_variable_groups_spec.rb
+++ b/spec/request/environment_variable_groups_spec.rb
@@ -72,6 +72,19 @@ RSpec.describe 'Environment group variables' do
         expect(last_response.status).to eq(401)
       end
     end
+
+    context 'permissions' do
+      let(:org) { VCAP::CloudController::Organization.make }
+      let(:space) { VCAP::CloudController::Space.make(organization: org) }
+      let(:api_call) { lambda { |user_headers| get '/v3/environment_variable_groups/running', nil, user_headers } }
+      let(:expected_codes_and_responses) { Hash.new(code: 200).freeze }
+
+      before do
+        space.organization.add_user(user)
+      end
+
+      it_behaves_like 'permissions for single object endpoint', ALL_PERMISSIONS + ['space_application_supporter']
+    end
   end
 
   describe 'PATCH /v3/environment_variable_groups/:name' do
@@ -239,6 +252,28 @@ RSpec.describe 'Environment group variables' do
 
         expect(last_response.status).to eq(401)
       end
+    end
+
+    context 'permissions' do
+      let(:org) { VCAP::CloudController::Organization.make }
+      let(:space) { VCAP::CloudController::Space.make(organization: org) }
+      let(:api_call) { lambda { |user_headers| patch '/v3/environment_variable_groups/running', params.to_json, user_headers } }
+      let(:expected_codes_and_responses) do
+        h = Hash.new(
+          code: 403
+        )
+        h['admin'] = {
+          code: 200
+        }
+
+        h.freeze
+      end
+
+      before do
+        space.organization.add_user(user)
+      end
+
+      it_behaves_like 'permissions for single object endpoint', ALL_PERMISSIONS + ['space_application_supporter']
     end
   end
 end
--- a/spec/request/environment_variable_groups_spec.rb
+++ b/spec/request/environment_variable_groups_spec.rb
@@ -73,7 +73,7 @@ RSpec.describe 'Environment group variables' do
       end
     end
 
-    context 'permissions' do
+    context 'when the user is logged in' do
       let(:org) { VCAP::CloudController::Organization.make }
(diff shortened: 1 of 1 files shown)


sweinstein22:
## Acceptance

### Setup

Created environment variables for `staging` and `running` groups as `admin`:
```
$ cf curl -X PATCH /v3/environment_variable_groups/staging -d '{"var":{"test_staging":"variable_staging"}}'
$ cf curl -X PATCH /v3/environment_variable_groups/running -d '{"var":{"test_running":"variable_running"}}'
```

Confirmed that `admin` could view both groups
```
$ cf curl /v3/environment_variable_groups/staging
{
  "name": "staging",
  "var": {
    "test_staging": "variable_staging"
  },
  ...
}

$ cf curl /v3/environment_variable_groups/running
{
  "name": "running",
  "var": {
    "test": "variable"
  },
  ...
}
```

Confirmed user only has space_application_supporter and organization_user roles:

```
$ cf curl /v3/roles?user_guids="0776980f-9ea5-499a-9a23-ecc09d0d5b39"
{
  ...
  "resources": [
    {
      "guid": "7555955f-5baf-4700-93e1-3952ef664a1f",
      "type": "organization_user",
      "relationships": {
        "user": {
          "data": {
            "guid": "0776980f-9ea5-499a-9a23-ecc09d0d5b39"
          }
        },
      ...
    },
    {
      "guid": "e158488c-2b95-4806-a6f1-68df9dde8fdc",
      "relationships": {
        "user": {
          "data": {
            "guid": "0776980f-9ea5-499a-9a23-ecc09d0d5b39"
          }
        },
       ...
  ]
}
```

### Behavior Check

Ran `cf curl` against `/v3/environment_variable_groups/running`, `/v3/environment_variable_groups/staging`, and `/v3/environment_variable_groups/bogus` as both space application supporter and space developer, confirmed that there was no diff and manually confirmed that output looked correct:
```
$ diff space-app-supporter-running space-dev-running

$ diff space-app-supporter-staging space-dev-running

$ diff space-app-supporter-bogus space-dev-bogus

```

Confirmed that space application supporter does not have permission to hit `PATCH` endpoint:
```
$ cf curl -X PATCH /v3/environment_variable_groups/staging -d '{"var":{"test_staging":"variable_staging"}}'
{
  "errors": [
    {
      "detail": "You are not authorized to perform the requested action",
      "title": "CF-NotAuthorized",
      "code": 10003
    }
  ]
}
```

### Code Review

This change did not require any changes to the implementation. Some adjustments to the request spec changes would bring the spec more inline with the current patterns of the file.

### Docs

No changes required, as the docs already indicate that this endpoint is available to all roles.

sweinstein22 (changes requested):
### Requested changes to `environment_variable_groups_spec.rb`

#### `describe 'GET /v3/environment_variable_groups/:name'`

Minor detail but it would be more inline with [current patterns of the file](https://github.com/cloudfoundry/cloud_controller_ng/blob/72f7115f3fcea74625c414527b287a4668e11755/spec/request/environment_variable_groups_spec.rb#L172) to call the new context 'when the user is logged in'

#### `describe 'PATCH /v3/environment_variable_groups/:name'`

The added changes to the spec are on the right track, but the organization of the tests is a bit disjointed with these changes. There's a [section that was already doing the shared examples testing for `staging`](https://github.com/cloudfoundry/cloud_controller_ng/blob/72f7115f3fcea74625c414527b287a4668e11755/spec/request/environment_variable_groups_spec.rb#L200), which haven't been updated to include the new role. There's also [a section dedicated to the `running` group](https://github.com/cloudfoundry/cloud_controller_ng/blob/72f7115f3fcea74625c414527b287a4668e11755/spec/request/environment_variable_groups_spec.rb#L101-L123), which the newly introduced permissions check is not part of.

As the pattern in this describe block is to do the majority of the tests for just one of the two cases (`running` or `staging`) because they run through the same controller flow, it would be great to adjust the tests so that we update the existing permission test under the `staging` context, and scrap the one testing against `running`.

will-gant:
Many thanks for the thorough review and apologies for the sloppiness on my side. I think/hope I've now implemented each of the improvements you've suggested.

sweinstein22:
No worries, thanks for making the adjustments, looks good!

PR #2338, opened 2021-06-15
Log when DB queries return large numbers of rows

andy-paine:
Add some extra logging for when DB queries return a large number of rows including a configuration flag which can both toggle the feature off as well as set a threshold for when to log.

We've seen [some DB queries](https://github.com/cloudfoundry/cloud_controller_ng/pull/2215) return large numbers of rows which CC then has to parse/marshal which can be particularly slow. By adding some additional logging when queries return lots of data we can hopefully identify specific requests that cause large numbers or rows to be handled.

We've also looked at using pg_stat_statments but by using the existing Steno logger in CC we can identify a specific request as our users submit error reports about slow endpoints including the VCAP request ID.

Slow service plan visibilities caused by large numbers of returned rows - https://github.com/cloudfoundry/cloud_controller_ng/pull/2215

## Checks

### DIFF
--- a/lib/cloud_controller/config_schemas/base/api_schema.rb
+++ b/lib/cloud_controller/config_schemas/base/api_schema.rb
@@ -88,6 +88,7 @@ module VCAP::CloudController
               pool_timeout: Integer, # timeout before raising an error when connection can't be established to the db
               log_level: String, # debug, info, etc.
               log_db_queries: bool,
+              optional(:query_size_log_threshold) => Integer,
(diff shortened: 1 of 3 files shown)


sethboyles:
@andy-paine thanks for the PR.  Is there an associated capi-release PR to allow configuration of the `query_size_log_threshold` field?

andy-paine:
Rebased this branch and added a CAPI PR https://github.com/cloudfoundry/capi-release/pull/200 now 🙂

PR #2340, opened 2021-06-16
Correct documentation to properly reflect allowed roles.

MerricdeLauney:
We noticed a noticed a disparity in the documented behavior vs our code.
It seems like this story (https://www.pivotaltracker.com/story/show/162360247)
implies that the code was correct in only
allowing the space developer and admin to use PATCH /v3/droplets/guid.

Co-authored-by: Galen Hammond <galenh@vmware.com>


sweinstein22:
Thanks for catching that discrepancy! @galenhammond could you click through the link provided by EasyCLA so that you're authorized before I merge in this PR?

PR #2349, opened 2021-06-23
Allow space application supporter to access specific service plan endpoints

svkrieger:
A space application support can access the following endpoints:

GET /v3/service_plans
GET /v3/service_plans/:guid/visibility
GET /v3/service_plans/:guid

Implements #2237

Remark: I'd be glad for feedback on whether I put the role into the right places and whether I added appropriate methods in the correct places. The changes should grant the new role the right permissions and the changes do not interfere with other endpoints which might use methods which were changed, because all unit tests are green. But I don't know whether the coding I added or adjusted fits into the overall picture.

### DIFF
--- a/app/controllers/v3/mixins/service_permissions.rb
+++ b/app/controllers/v3/mixins/service_permissions.rb
@@ -26,7 +26,7 @@ module VCAP::CloudController
     end
 
     def visible_space_scoped?(space)
-      current_user && space && (space.has_member?(current_user) || space.has_application_supporter?(current_user))
+      current_user && space && (space.has_member?(current_user) || space.has_supporter?(current_user))
     end
 
     def visible_to_current_user?(service: nil, plan: nil)
--- a/app/controllers/v3/mixins/service_permissions.rb
+++ b/app/controllers/v3/mixins/service_permissions.rb
@@ -26,7 +26,7 @@ module VCAP::CloudController
     end
 
     def visible_space_scoped?(space)
-      current_user && space && space.has_member?(current_user)
+      current_user && space && (space.has_member?(current_user) || space.has_application_supporter?(current_user))
     end
 
     def visible_to_current_user?(service: nil, plan: nil)
--- a/app/controllers/v3/service_plans_controller.rb
+++ b/app/controllers/v3/service_plans_controller.rb
@@ -55,7 +55,7 @@ class ServicePlansController < ApplicationController
                   message,
                   eager_loaded_associations: Presenters::V3::ServicePlanPresenter.associated_resources,
                   readable_org_guids: permission_queryer.readable_org_guids,
-                  readable_space_guids: permission_queryer.readable_application_supporter_space_scoped_space_guids,
+                  readable_space_guids: permission_queryer.readable_space_supporter_space_scoped_space_guids,
                 )
               end
 
--- a/app/controllers/v3/service_plans_controller.rb
+++ b/app/controllers/v3/service_plans_controller.rb
@@ -55,7 +55,7 @@ class ServicePlansController < ApplicationController
                   message,
                   eager_loaded_associations: Presenters::V3::ServicePlanPresenter.associated_resources,
                   readable_org_guids: permission_queryer.readable_org_guids,
-                  readable_space_guids: permission_queryer.readable_space_scoped_space_guids,
+                  readable_space_guids: permission_queryer.readable_application_supporter_space_scoped_space_guids,
                 )
               end
 
--- a/lib/cloud_controller/permissions.rb
+++ b/lib/cloud_controller/permissions.rb
@@ -234,6 +234,14 @@ class VCAP::CloudController::Permissions
     end
   end
 
+  def readable_application_supporter_space_scoped_space_guids
+    if can_read_globally?
+      VCAP::CloudController::Space.select(:guid).all.map(&:guid)
+    else
+      membership.space_guids_for_roles(SPACE_ROLES_INCLUDING_APPLICATION_SUPPORTERS)
+    end
+  end
+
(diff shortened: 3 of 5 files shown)


mkocher:
chatted with @sweinstein22. Let's 1) add a guid check to the shared behavior and 2) rename the "as a space developer" context to something like "as a logged in user" or "as a permitted user".

monamohebbi (changes requested):
Code review looks good to me except for one thing to call out - I noticed you have created a new `COMPLETE_PERMISSIONS` in the shared_example spec instead of following the `+ space_application_supporter`. Our plan is to do a larger refactor at the completion of the entire track where we remove all the `+` and add space_application_supporter to the offical `ALL_PERMISSIONS` so that when run the entire test suite we can be sure all examples work with the set of all permissions. I think changing the pattern halfway through might make us a little error prone later on, and I wouldn't want to end up accidentally keeping `ALL_PERMISSIONS` and `COMPLETE_PERMISSIONS` since that could be a confusing pattern .Could you refactor it to match the existing pattern?

svkrieger:
> Code review looks good to me except for one thing to call out - I noticed you have created a new `COMPLETE_PERMISSIONS` in the shared_example spec instead of following the `+ space_application_supporter`. Our plan is to do a larger refactor at the completion of the entire track where we remove all the `+` and add space_application_supporter to the offical `ALL_PERMISSIONS` so that when run the entire test suite we can be sure all examples work with the set of all permissions. I think changing the pattern halfway through might make us a little error prone later on, and I wouldn't want to end up accidentally keeping `ALL_PERMISSIONS` and `COMPLETE_PERMISSIONS` since that could be a confusing pattern .Could you refactor it to match the existing pattern?

Oh I see. Yeah that makes sense. Just thought I reduce code duplication a bit. I refactored that to match the other PRs.

svkrieger:
> chatted with @sweinstein22. Let's 1) add a guid check to the shared behavior and 2) rename the "as a space developer" context to something like "as a logged in user" or "as a permitted user".

I don't really know which code pieces you are referring to. Could it be that you are talking about another PR? I found the context "as a space developer" in this PRs changed files where I also commented on: https://github.com/cloudfoundry/cloud_controller_ng/pull/2323
If I'm wrong and you are talking about this one here, could you point me to the code which should be changed? Thanks!

sweinstein22:
Hi @svkrieger , you're right that comment was in relation to a different PR, hopefully it didn't cause you too much confusion. Thanks for checking in!

sweinstein22 (commented):
## Acceptance

### Setup

Pushed [overview-broker](https://github.com/cloudfoundry/overview-broker) and enabled plans globally

### Behavior Check

Mostly leaving this to Mona, just ran through a quick curl against the endpoints, looks reasonable to me.

### Docs Check

All three endpoints already indicate that they can be accessed by All Roles, no changes required

sweinstein22 (changes requested):
### Code Check

Implementation looks good to me , in request spec confirmed that shared examples are being used, that response codes and bodies are being checked.

Looks like we're still using `COMPLETE_PERMISSIONS` in ` spec/request/service_plans_spec.rb`, if we could swap that one back to `ALL_PERMISSIONS` just to make it easier for us to refactor everything once the role has been completely implemented that would be great. Other than that, this looks good to go, thanks for the PR!

sweinstein22 (approved):
Looks great, thanks for the PR!

One minor detail, @iaftab-alam could you sign the EasyCLA cert? Thanks!

iaftab-alam:
@sweinstein22 already requested for the EasyCLA cert.

iaftab-alam:
@sweinstein22 got approval for  EasyCLA cert. Can we re-run the failing validation job again?

iaftab-alam:
Hey @monamohebbi, can you please approve the changes.

monamohebbi:
## Acceptance

### Behavior Check

Setup: created 4 service brokers each assigned a plan with a different visibility type from set admin, org, space, and public. 

#### `GET` Endpoints
Ran `cf curl` against `/v3/service_plans/:guid`, `/v3/service_plans` as a space supporter in the targeted org and space. Manually confirmed output looked correct, and checked for diffs:
```
$ diff dev supporter

$ diff show_dev show_supporter

```

#### `GET` Endpoints when assigned a role in a separate space in the same org
Confirmed that output looked correct manually. Checked that diff only included timestamps, guids, and user details under `created_by` section:
```
$ diff otherspace_dev otherspace_supporter

$ diff otherspace_show_dev otherspace_show_supporter

```

#### `GET` Endpoints when assigned a role in a separate org
Confirmed that output looked correct manually. Checked that diff only included timestamps, guids, and user details under `created_by` section:
```
$ diff otherorg_dev otherorg_supporter

$ diff otherorg_show_dev otherorg_show_supporter

```


### Code Check
Unfortunately, due to a series of vacations and other requests I've had to deprioritize acceptance for a while. Since I'm just getting around to this now, the rebase didn't fix the naming conflicts in the newly added functions in the queryer/permissions/controller. If you could just take the word application out of the new functions and the respective calls I think you should be good to go; that is how I tested things off this branch locally.

svkrieger:
Hi @monamohebbi,

I did the changes. It was kinda complex though, because of the perm deletion. I'd suggest repeating the acceptance tests with the current state.

monamohebbi:
Sorry about that! Thanks for the quick updates. I'll run through them and do unit testing again.

monamohebbi:
Double checked the endpoints using the same work flow and got the same results. This looks ready to be merged. Thanks!

monamohebbi (approved):
Double checked the endpoints using the same work flow and got the same results. This looks ready to be merged. Thanks!

PR #2350, opened 2021-06-23
Space application supporter revisions

### LINKED ISSUE #2232: Allow space application supporter to access specific revision endpoints.
## Issue

Allow space application supporter to access specific **revision** endpoints. 

## Context

## Expected result

GET /v3/apps/:guid/revisions
GET /v3/apps/:guid/revisions/deployed
GET /v3/revisions/:guid

## Acceptance
A space application supporter would see the same info as a space developer assigned to the same space for these and only these revision endpoints.

## Documentation

will-gant:
Closes #2232 

### DIFF
--- a/app/controllers/v3/app_revisions_controller.rb
+++ b/app/controllers/v3/app_revisions_controller.rb
@@ -14,7 +14,7 @@ class AppRevisionsController < ApplicationController
     invalid_param!(message.errors.full_messages) unless message.valid?
 
     app, space, org = AppFetcher.new.fetch(hashed_params[:guid])
-    app_not_found! unless app && permission_queryer.can_read_from_space?(space.guid, org.guid)
+    app_not_found! unless app && permission_queryer.untrusted_can_read_from_space?(space.guid, org.guid)
 
     dataset = AppRevisionsListFetcher.fetch(app, message)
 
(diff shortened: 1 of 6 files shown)


sweinstein22 (commented):
## Acceptance

### Behavior Check

Ran `cf curl` against `GET /v3/apps/:guid/revisions`, `/v3/apps/:guid/revisions/deployed`, `/v3/revisions/:guid` as both a space app supporter and space developer, confirmed that output looked correct manually and that there was no diff in output:
```
$ diff space-app-supporter-app-guid-revisions space-dev-app-guid-revisions

$ diff space-app-supporter-app-guid-revisions-deployed space-dev-app-guid-revisions-deployed

$ diff space-app-supporter-revisions-guid space-dev-revisions-guid

```

Also confirmed that curing `/v3/revisions/:guid/environment_variables` returns a `403` as a space app supporter

### Code Check

Looks good - implementation looks good and the changes leverage the shared examples where possible. There's a minor thing I'd like to pick at, going to leave it as a separate comment.

### Docs Check

Confirmed that new role has been added to all three endpoints with Experimental indicator, and that docs are rendering correctly

sweinstein22 (changes requested):
Super minor, but we've been aiming to be consistent with the way we've been adding the permissions to the request spec shared examples, specifically using the `ALL_PERMISSIONS + ['space_application_supporter']` structure. I know it's a little counter to what we would normally be aiming to do as far as code duplication, but it'll make it easier for us to go through and refactor `'space_application_supporter'` into the default set for `ALL_PERMISSIONS` down the line if things are consistent among the changes we're making now. If you wouldn't mind swapping that out everything else looks good to go! Thanks for the PR!

will-gant:
@sweinstein22 Many thanks for the feedback. You're right - it's going to be easier to work on in future if we keep things like that consistent across these new files. I've implemented the change you requested. Let me know if there's anything else.

sweinstein22 (approved):
Looks great, thanks so much for the PR!

PR #2351, opened 2021-06-24
Support expiring of DB connections

andy-paine:
Support DB connections being expired after a certain time period (with a random delay)

We have long running CC processes and connections appear to be consuming and not freeing a lot of Postgres resources. We are hoping that by expiring connections after a certain (configurable) time limit, we can free up these resources before we see more damaging effects on the DB

It uses a builtin extension to Sequel which should hopefully be well understood and tested and just exposes some configuration parameters. Additionally, the extension should only end up being enabled for users who opt-in by specifying the timeout and optionally a random delay


sweinstein22:
Looks good to me, thanks for the PR!

PR #2353, opened 2021-06-24
Replace readable_org_guids_for_domains with query

andy-paine:
Replace the `readable_org_guids_for_domains` function with one that returns a SQL subquery as a filter rather than an actual Ruby array.

We have seen extremely large query strings when an admin or a power-user in many orgs requests domains as the entire list of (sometimes >100k) guids is turned into a giant `WHERE IN ('abc', 'def' etc.)` SQL query.
This is expensive for Ruby to turn into the query, expensive to send over the network and expensive for the DB to process.

Databases are very very good at working out the most efficient way to run a query so even though the subqueries produced by this look pretty complex (due to each role being a separate table and space roles needing a join to the orgs table), the actual runtime is much better this way.

https://github.com/cloudfoundry/cloud_controller_ng/pull/2255 - similar problem of huge query strings that may actually be better addressed by this approach

## Comparison with performance tests
Using the tests at https://github.com/cloudfoundry-incubator/cf-performance-tests I ran (locally) a comparison of the /v3/domains endpoint with:
* 50,000 orgs
* 100,000 spaces
* 20,000 domains

The results seem pretty clear cut:
BEFORE
AFTER

### DIFF
--- a/app/controllers/v3/organizations_controller.rb
+++ b/app/controllers/v3/organizations_controller.rb
@@ -142,7 +142,7 @@ class OrganizationsV3Controller < ApplicationController
     message = DomainsListMessage.from_params(query_params.except(:guid))
     invalid_param!(message.errors.full_messages) unless message.valid?
 
-    domains = DomainFetcher.fetch(message, domain_readable_org_guids([org.guid]))
+    domains = DomainFetcher.fetch(message, permission_queryer.readable_org_guids_for_domains.where(guid: org.guid))
 
     render status: :ok, json: Presenters::V3::PaginatedListPresenter.new(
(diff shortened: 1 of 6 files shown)


sethboyles:
Hi @andy-paine, thanks for the PR!  Looking forward to merging this.

Unfortunately I think the merge of https://github.com/cloudfoundry/cloud_controller_ng/pull/2405 caused most of the merge conflicts with this PR.  I think most of those files were just deleted, though, I hope the resolution of those conflicts are relatively straightforward...

andy-paine:
Thanks for the heads up @sethboyles. I've rebase this onto `main` now so should be good to go

Issue #2306, opened 2021-06-01
Cloud Controller API does not validate if 2 services with the same name are shared in the same spaces


sweinstein22:
## Issue

The cloud controller does not enforce the name uniqueness for shared services in a spaces. It creates a situation where you can have multiple SI with the same name in a space. 

## Steps to Reproduce

Use the [overview-broker service](https://github.com/cloudfoundry/overview-broker) to deploy a service broker:
```
$ git clone git@github.com:cloudfoundry/overview-broker.git
$ cf push overview-broker
$ cf create-service-broker overview-broker admin password <app_url>
$ cf enable-service-access overview-service
```

Create service instances in two different spaces, then share them to a third:
```
$ cf enable-feature-flag service_instance_sharing
$ cf create-org source
$ cf create-space -o source space-1
$ cf create-space -o source space-2
$ cf create-org target
$ cf create-space -o target target
$ cf target -o source -s space-1
$ cf create-service overview-service small si -c '{"foo": "bar"}'
$ cf target -o source -s space-2
$ cf create-service overview-service small si -c '{"foo": "bar"}'
$ cf target -o source -s space-1
$ cf share-service si -s target -o target
$ cf target -o source -s space-2
$ cf share-service si -s target -o target
```

See that there are two service instances with the same name in the `target` org/space:
```
$ cf target -o target -s target
$ cf services
```
Output:
```
name   offering              plan    bound apps   last operation     broker            upgrade available 
si     overview-service   small                   create succeeded   overview-broker   no 
si     overview-service   small                   create succeeded   overview-broker   no
```

## Expected result

The second `share-service` command should fail with a `422 Unprocessable Entity`

## Current result

Currently it succeeds, allowing users to end up with two service instances that have the same name in a single space.

This validation is already in place when trying to create a new service instance in a space:
```
$ cf target -o target -s target
$ cf create-service overview-broker small si -c '{"foo": "bar"}'
The service instance name is taken: si.
FAILED
```

## Possible Fix

Validate service instance name uniqueness when sharing services

sweinstein22:
@tstannard and I have done some preliminary exploration and determined that we are not checking for name uniqueness among service instances that have been shared into the space. We're only [checking against service instances that originated in the space.](https://github.com/cloudfoundry/cloud_controller_ng/blob/05a2399af15f9517d815ba3979c3e9fddd9d47eb/app/actions/service_instance_share.rb#L56-L59)

[...]

PR #2362, opened 2021-07-01
Service name validation

bepotts:
This PR addresses issue #2306


sweinstein22 (commented):
Started by reproing the issue on a boshlite using the steps outlined in the issue. Minor thing worth noting for reproduction steps, when using the latest nodejs buildpack the `overview-broker` needs to be adjusted so that the `package.lock` file indicates the node version is at least `12.x.x`, as node 10 is no longer in support.
```
$ cf target -o target -s target
...
org:            target
space:          target
$ cf services
Getting service instances in org target / space target as admin...

name   offering           plan    bound apps   last operation     broker            upgrade available
si     overview-service   small                create succeeded   overview-broker   no
si     overview-service   small                create succeeded   overview-broker   no

```

Unshared both service instances:
```
$ cf services
Getting service instances in org target / space target as admin...

No service instances found.
```

Then ran `deploy_only_new_capi` and attempted to reproduce the issue again. It looks like as part of the changes introduced we've lost the ability to share the first service instance into the space:

```
$ cf services
Getting service instances in org target / space target as admin...

No service instances found.

$ cf target -o source -s space-1
...
org:            source
space:          space-1

Unexpected Response
Response Code: 500
Request ID:    25747d10-4723-45ca-5d6d-f0c32d3eae18::db8c01ca-9843-4fb6-bf9c-cb668c686469
Code: 0, Title: , Detail: {
  "errors": [
    {
      "title": "UnknownError",
      "detail": "An unknown error occurred.",
      "code": 10001
    }
  ]
}
FAILED
```

I haven't looked at the code yet to determine what would have caused this error, I'll plan to get around to it sometime this afternoon. Just wanted to get the feedback out there in case you'd like to dig into it while I'm offline for lunch @bepotts !

sweinstein22:
Did some more poking around after a chat with @bepotts , things that might be notable:

- I poked around the code on the api vm and realized I documented the above after adding a `!` after `validate_uniqueness_of_service_names` because initially it didn't seem like there was any change after deploying the code. (final result: `validate_uniqueness_of_service_names!(service_instance, target_spaces)`), which is in line with the syntax of the other validation functions being called. With that `!`, all service instance sharing errors out as mentioned above:
  ```
  Unexpected Response
  Response Code: 500
  Request ID:    64ef7073-2d92-4073-5cc5-59f9e2b4200f::96e01e7a-af95-4c98-9dca-ab409a354b91
  Code: 0, Title: , Detail: {
    "errors": [
      {
        "title": "UnknownError",
        "detail": "An unknown error occurred.",
        "code": 10001
      }
    ]
  }
  FAILED
  ```
- Without the `!` (as the code currently stands) I was actually still able to share both service instances into the target space, which means the bug is still present but it's maybe not failing out quite as mysteriously as I originally thought

I confirmed that it's hitting `POST /v3/service_instances/:guid/relationships/shared_spaces` by running the CLI command with a `-v` flag.

Beyond the functionality, some things that might be notable within the code:
- it looks like some comments got left behind (a `TODO` in particular, but there are a few scattered around)
- The name `validate_uniqueness_of_service_names` is a bit misleading, as what we're actually trying to check is the uniqueness of the name in context of service instances that have been shared into the space. The [`validate_name_uniqueness` call](https://github.com/cloudfoundry/cloud_controller_ng/blob/5787b935f80f4bdc48b236cca68d0595928dc944/app/actions/service_instance_share.rb#L55-L60) is actually checking against the service instances that were created in the space. Generally interacting with services is the worst from a naming perspective, but wanted to call that out since we're testing two separate things.
- I'm noticing we've also got two different space loops going, [one that has validations in it](https://github.com/cloudfoundry/cloud_controller_ng/pull/2362/files#diff-29dfbac596bd545b15934fc393c342ce39c592b4988a0889e3e249575d8d5194L35-L38) and [one that's part of the newly introduced function](https://github.com/cloudfoundry/cloud_controller_ng/pull/2362/files#diff-29dfbac596bd545b15934fc393c342ce39c592b4988a0889e3e249575d8d5194R61-R63). Just wondering whether it would make sense to collapse it down and model the new check on `validate_name_uniqueness`

I know you said you had some ideas about where to go next so feel free to take as many or as few of these thoughts as you'd like, just wanted to do some more digging before I head out on vacation!

PR #2384, opened 2021-07-09
Create new table to store newly created space_supporters in an appropriately named table

monamohebbi:
We have created a migration to create a new table with a name that matches the new role title. We also introduced a time bomb test that will remind us that an unused table exists in 6 months so we can consider removing it after we have give users enough time to upgrade their capi-release versions.

This solves the problem of having a table name that is inconsistent with the role name.

Follow up to:
https://github.com/cloudfoundry/cloud_controller_ng/pull/2375


sweinstein22:
For the most part this PR looks good to me, thanks so much for taking the time to write up an ADR! Unfortunately the unit tests are failing. I've tried rerunning them, and running them locally after rebasing off of `main`, neither of which seemed to solve the issues. cc: @monamohebbi @galenhammond

sweinstein22:
I've opened https://github.com/cloudfoundry/cloud_controller_ng/pull/2446 in lieu of this PR as the rebase is giving us issues with EasyCLA and `capi-bot`. 

I double checked that the diffs are identical between the two PRs, the only difference is that this PR had [a comment removal](https://github.com/cloudfoundry/cloud_controller_ng/pull/2384/files#diff-5ed123e8b5e7f61aecf929a278f53c219544d8c6aa80a65f42dac5fa9b63654bL4-L5) which we do actually want to keep.

PR #2405, opened 2021-07-14
Removing perm

mkocher:
Perm was a valiant effort, and many good developers gave it their all. It was
time to admit that it was dead weight we were hauling along with us.

My main motivation was to get the permission querying code to be
something our vim setup could navigate through when going to the
definitions of functions named the same for db permissions and
perm. Once I got started I couldn't stop myself.

Assorted Notes:
- removed all traces of perm I could find. There are some instance of
"perm" in the codebase that have nothing to do with externalizing permissions.
- Made all perm config optional, as it's no longer used for anything.

Authored-by: Matthew Kocher <mkocher@pivotal.io>

There are no short explanations of what the perm project was. The proposed change here is to remove all traces of it from Cloud Controller. 

As a developer, I want to go to the definition of redable_space_guids and see the code that is in use in 100% cloud controller instances.

As an operator I don't want to specify in my config file that I'm not using perm. Since I'm not. 

I probably should do this one, but I haven't yet. Would like to get some feedback on this PR before I put in the effort.


lib/cloud_controller/config_schemas/base/api_schema.rb
```diff
@@ -311,8 +311,11 @@ class ApiSchema < VCAP::Config
               optional(:diego_sync) => { timeout_in_seconds: Integer },
             },
 
-            perm: {
-              enabled: bool,
+            # perm settings no longer have any effect but are preserved here
+            # for the time being to avoid breaking users as the perm
+            # setting was once required.
+            optional(:perm) => {
+              optional(:enabled) => bool,
```
ctlong, line 318: @mkocher so passing in `perm` as a config option will break the api_schema if `perm` isn't at least defined as optional here?

Am I right in thinking this config option ultimately comes from the bosh release? If so, why remove the `perm` property in `config/cloud-controller.yaml` but not here.

ctlong:
I think this is a really smart PR

lib/cloud_controller/config_schemas/base/api_schema.rb
mkocher: Yes, you're right in thinking passing in perm config will break things if it's not defined as optional.

This usually comes from the bosh release and we could delete it there at the same time, but I'd prefer to not tie this change to an exact version of capi-release. I'm also suspecting that the various k8s cf projects might not use the bosh config.

I removed it from config/cloud-controller.yaml as I think of that as the canonical sample config for cloud controller, so it seems right to update it to not have any perm settings as that's what we'd hope all users will do eventually.

sweinstein22 (approved):
This all looks good to me, pretty straightforward deletions for the most part with some swaps `Permissions::Queryer` => `VCAP::CloudController::Permissions` thrown into the mix. Thanks for taking the time to dig all this out!

PR #2406, opened 2021-07-14
Allow space supporter to access specific isolation segment endpoints

### LINKED ISSUE #2222: Allow space application supporter to access specific isolation segments endpoints.
## Issue

Allow space application supporter to access specific **isolation segments** endpoints. 

## Context

## Expected result
A space supporter should be able access the following endpoints:

GET /v3/isolation_segments/:guid
GET /v3/isolation_segments
GET /v3/isolation_segments/:guid/relationships/organizations
GET /v3/isolation_segments/:guid/relationships/spaces

A space supporter should get a 403 when attempting to curl the following endpoint: 

PATCH /v3/spaces/:guid/relationships/isolation_segment

## Acceptance
A space application supporter would see the same info as a space developer assigned to the same space for these and only these isolation segments endpoints.

## Documentation

galenhammond:
Closes #2222  

### DIFF
--- a/app/controllers/v3/spaces_controller.rb
+++ b/app/controllers/v3/spaces_controller.rb
@@ -149,7 +149,7 @@ class SpacesV3Controller < ApplicationController
     space_not_found! unless space
     org = space.organization
     org_not_found! unless org
-    space_not_found! unless permission_queryer.can_read_from_space?(space.guid, org.guid)
+    space_not_found! unless permission_queryer.untrusted_can_read_from_space?(space.guid, org.guid)
     unauthorized! unless roles.admin? || space.organization.managers.include?(current_user)
 
     message = SpaceUpdateIsolationSegmentMessage.new(hashed_params[:body])
--- a/docs/v3/source/includes/resources/isolation_segments/_get.md.erb
+++ b/docs/v3/source/includes/resources/isolation_segments/_get.md.erb
@@ -27,10 +27,5 @@ Content-Type: application/json
 #### Permitted roles
  |
 --- | ---
-Admin |
-Admin Read-Only |
-Global Auditor |
-Org Manager |
-Space Auditor |
-Space Developer |
-Space Manager |
+All Roles |
+
--- a/docs/v3/source/includes/resources/isolation_segments/_list_spaces.md.erb
+++ b/docs/v3/source/includes/resources/isolation_segments/_list_spaces.md.erb
@@ -29,10 +29,4 @@ This endpoint lists the spaces to which the isolation segment is assigned.  For 
 #### Permitted roles
  |
 --- | ---
-Admin |
-Admin Read-Only |
-Global Auditor |
-Org Manager |
-Space Auditor |
-Space Developer |
-Space Manager |
+All Roles |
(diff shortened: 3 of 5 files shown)


sweinstein22 (commented):
## Acceptance

### Behavior Check

Ran `cf curl` against `v3/isolation_segments/:guid`, `v3/isolation_segments`, `v3/isolation_segments/:guid/relationships/organizations` as a space supporter and a space developer. Confirmed output looked correct manually and checked for diffs:
```
$ diff space-supporter-isolation-segments space-dev-isolation-segments

$ diff space-supporter-isolation-segments-guid space-dev-isolation-segments-guid

$ diff space-supporter-isolation-segments-relationship-orgs space-dev-isolation-segments-relationship-orgs

```

Confirmed that `PATCH /v3/spaces/:guid/relationships/isolation_segment` returned a `403` as a space supporter:
```
cf curl -X PATCH /v3/isolation_segments/5792dac1-31da-407b-98fa-fc4a63d8e618 -v
REQUEST: [2021-07-14T15:11:33-07:00]
PATCH /v3/isolation_segments/5792dac1-31da-407b-98fa-fc4a63d8e618 HTTP/1.1
...

RESPONSE: [2021-07-14T15:11:33-07:00]
HTTP/1.1 403 Forbidden
...
```

### Docs Check

All the associated endpoints now indicate that all roles should have access, and the docs render as expected

sweinstein22 (changes requested):
### Behavior Check

Ran `cf curl` against `v3/isolation_segments/:guid/relationships/spaces`, looks like we're not getting the correct return value:

```
$ diff space-supporter-isolation-segments-relationship-spaces space-dev-isolation-segments-relationship-spaces
3c3,5
<
---
>     {
>       "guid": "a82da199-03e7-4607-9f8b-36a16fb782ea"
>     }
```

### Code Check

Looks like we're not checking against a guid in the request spec shared examples, which might be why we didn't catch that `v3/isolation_segments/:guid/relationships/spaces` isn't returning the correct output. If we could add a response body check, or at least a guid check, that would be great!

galenhammond:
@sweinstein22,

We have pushed an update addressing your previous comments to the PR branch. Let us know if there is anything else that it needs!

Thanks,
@moleske, @mkocher, Galen

sweinstein22:
Thanks @moleske , @mkocher, @galenhammond ! Looks like the unit tests are passing with some iso segment related errors, do you all mind taking a look?

mkocher:
This should be good to go now, we encountered a random array ordering issue so changed tactics a bit.

sweinstein22 (changes requested):
### Behavior Check

Ran cf curl against `v3/isolation_segments/:guid/relationships/spaces` as both a space supporter and space developer, confirmed that there was no diff and manually checked that output looked correct:
```
$ diff space-supporter-iso-segments-for-space space-dev-iso-segments-for-space

```

### Code Check

Thanks for updating to check against the response body for list endpoints, it looks like we missed `GET /v3/isolation_segments`. Just want to make sure we're getting content, not just a `200`. If we could make that one adjustment this PR should be good to go!

sweinstein22 (commented):
Changes look good, thanks for making the docs more robust as well!

sweinstein22 (approved):

PR #2413, opened 2021-07-16
Retry polling on broker errors in V3

### LINKED ISSUE #2354: Polling on last operation fails in v3 on http request errors
## Issue

After an instance creation/update/deletion the cloud controller polls for the last operation. If the polling request fails for connection issues the creation/update/deletion fails, even if everything is fine on service broker side. It seems that this behavior changed between v2 and v3.

## Context

When I look at the source code it seems that the v2 last operation polling handles http request errors:

https://github.com/cloudfoundry/cloud_controller_ng/blob/4a23a419862ef9b7272f411a4435e33fc7305f30/app/jobs/v2/services/service_instance_state_fetch.rb#L32

while the v3 last operation polling sets the instance to failed on every error that occurs:

https://github.com/cloudfoundry/cloud_controller_ng/blob/4a23a419862ef9b7272f411a4435e33fc7305f30/app/actions/v3/service_instance_create_managed.rb#L94-L97

At least in my understanding this is a regression and the cloud controller should continue polling at least for status code 502, 503, and tcp connection errors (or just for all http errors).

## Steps to Reproduce

1. Create and register a service broker
2. Create a service instance
3. Let the service broker fail one of the last operation requests on tcp/http layer (while all other requests are fine)

## Expected result

I would expect that the instance state stays "in progress"

## Current result

The instance state switches to failed

## Possible Fix

Catch and handle http request errors during last operation polling and just continue the polling

iaftab-alam:

By retrying HTTP errors in the V3 action CC can be more resilient against failing service brokers.

In V2 service broker HTTP errors would be retried until successful. In V3 a service broker HTTP error would immediately mark the instance as failed.

Fixes #2354

### DIFF
--- a/app/actions/v3/service_instance_create_managed.rb
+++ b/app/actions/v3/service_instance_create_managed.rb
@@ -74,8 +74,14 @@ module VCAP::CloudController
       end
 
       def poll(instance)
+        logger = Steno.logger('cc.action.service_instance_create_managed')
         client = VCAP::Services::ServiceClientProvider.provide(instance: instance)
-        last_operation_result = client.fetch_service_instance_last_operation(instance, user_guid: @user_audit_info.user_guid)
+        begin
+          last_operation_result = client.fetch_service_instance_last_operation(instance, user_guid: @user_audit_info.user_guid)
+        rescue HttpRequestError, HttpResponseError, Sequel::Error => e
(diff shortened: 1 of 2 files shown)


sethboyles:
LGTM! Thanks for the PR!

PR #2426, opened 2021-07-19
Allows space supporter to access GET v3/service_route_bindings

### LINKED ISSUE #2422: Allow space application supporter to list service route bindings
## Issue

Allow space application supporter to list **service route bindings**. 

## Context

## Expected result

GET /v3/service_route_bindings

## Acceptance
A space application supporter would see the same info as a space developer assigned to the same space for these and only these service route binding endpoints.

## Documentation

galenhammond:
Allows space supporter access to service route binding list endpoint

Closes #2422

### DIFF
--- a/app/controllers/v3/service_route_bindings_controller.rb
+++ b/app/controllers/v3/service_route_bindings_controller.rb
@@ -182,7 +182,7 @@ class ServiceRouteBindingsController < ApplicationController
   end
 
   def space_guids
-    permission_queryer.readable_space_guids
+    permission_queryer.readable_supporter_space_guids
   end
 
   def parse_create_request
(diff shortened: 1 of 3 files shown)


sweinstein22 (commented):
## Acceptance

### Behavior Check

Ran `cf curl` against `v3/service_route_bindings` as both a space supporter and space developer. Manually confirmed output looks correct and checked for a diff:

```
$ diff space-supporter-service-route-bindings space-dev-service-route-bindings

```

### Code Check

Looks good, pretty minimal changes that use the established patterns for this work

sweinstein22 (changes requested):
### Docs Check

Looks like we missed the "Experimental" note on the docs, but  otherwise this looks good to go!

sweinstein22 (approved):
Looks good, thanks for adjusting the docs and for the PR!

Issue #2410, opened 2021-07-15
Allow space supporter to access any endpoints we might have missed


monamohebbi:
Issue
Allow space supporter to access specific app environment variable endpoints.

Context
We are introducing a new role and we want to make sure it has the right access.

Expected result
A space supporter should be able access the following endpoints:

GET /v3/apps/:guid/ssh_enabled
GET v3/stacks/:guid/apps

Acceptance
A space application supporter would see the same info as a space developer assigned to the same space for these and only these endpoints.

Documentation
When I browse to any of these endpoints on v3 docs I can see the Space Supporter role in the list of permitted roles with an indication that this role is not fully implemented and the permissions will be changing, and that this role will only be able to see a redacted view.

mariash:
MIssed endpoints:

| CLI Command | Endpoint | Handler |
| -------------- | --------- | -------- |
| services | /v3/service_instances | service_instances_v3#index |

galenhammond:
A few more endpoints that were missed:

GET /v3/apps/:guid/environment_variables
PATCH /v3/apps/:guid/environment_variables
PATCH /v3/deployments/:guid 
PATCH /v3/environment_variable_groups/:name

I would also be interested in reviewing the following endpoint specs:

GET v3/buildpacks -> not sure about permissions test setup as user is defined as a member of space by default/using single object endpoint test
GET v3/deployments -> no permissions test

monamohebbi:
Thank you for looking over these endpoints! 

Due to some CI issues some of these endpoints are already complete and just waiting on being merged in, some are still waiting on acceptance testing, and some are currently in flight:

I believe these: 
https://github.com/cloudfoundry/cloud_controller_ng/pull/2388
https://github.com/cloudfoundry/cloud_controller_ng/pull/2378
https://github.com/cloudfoundry/cloud_controller_ng/issues/2241
should cover environment variable and services endpoints you all had listed.

As for `PATCH /v3/deployments/:guid` I believe we are blocking `PATCH` updates that only allow for metadata updates for now due to a lack of user data on metadata. I can update the spreadsheet accordingly.

I'll look into backfilling the tests!

PR #2431, opened 2021-07-21
Allow space supporter to get ssh_enabled app feature and to access apps in the stack

mariash:
Allow space supporter to get ssh_enabled app feature and to access apps in the stack.

Endpoints:
GET /v3/apps/:guid/ssh_enabled
GET v3/stacks/:guid/apps

Issue #2410

### DIFF
--- a/app/controllers/v3/app_features_controller.rb
+++ b/app/controllers/v3/app_features_controller.rb
@@ -57,7 +57,7 @@ class AppFeaturesController < ApplicationController
   def ssh_enabled
     app, space, org = AppFetcher.new.fetch(hashed_params[:guid])
 
-    app_not_found! unless app && permission_queryer.can_read_from_space?(space.guid, org.guid)
+    app_not_found! unless app && permission_queryer.untrusted_can_read_from_space?(space.guid, org.guid)
 
     render status: :ok, json: Presenters::V3::AppSshStatusPresenter.new(app, Config.config.get(:allow_app_ssh_access))
   end
--- a/app/controllers/v3/stacks_controller.rb
+++ b/app/controllers/v3/stacks_controller.rb
@@ -67,7 +67,7 @@ class StacksController < ApplicationController
     dataset = if permission_queryer.can_read_globally?
                 AppListFetcher.fetch_all(message)
               else
-                AppListFetcher.fetch(message, permission_queryer.readable_space_guids)
+                AppListFetcher.fetch(message, permission_queryer.readable_supporter_space_guids)
               end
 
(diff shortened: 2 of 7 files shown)


monamohebbi:
### Acceptance
#### Setup 
Pushed app dora from cf-acceptance-tests repo as admin.

#### Behavior check
Running `cf curl v3/stacks/:guid/apps` as supporter user and developer user returns the same result. 
Running `cf curl v3/apps/$(cf app dora --guid)/ssh_enabled` as supporter user and developer user returns the same result. 

Additionally I tested that a stack associated with an app dora2 in a different space would not show dora2's info via 
```
 cf curl v3/stacks/f4492bdc-2e16-43d9-a6f7-fb867035d6d3/apps
{
   "pagination": {
      "total_results": 0,
      "total_pages": 1,
      "first": {
         "href": "https://api.stream-dart.capi.land/v3/stacks/f4492bdc-2e16-43d9-a6f7-fb867035d6d3/apps?page=1&per_page=50"
      },
      "last": {
         "href": "https://api.stream-dart.capi.land/v3/stacks/f4492bdc-2e16-43d9-a6f7-fb867035d6d3/apps?page=1&per_page=50"
      },
      "next": null,
      "previous": null
   },
   "resources": []
}
```
LGTM!

sweinstein22 (changes requested):
### Code Check

Mostly looks good, a minor detail - we've had [conversations previously](https://github.com/cloudfoundry/cloud_controller_ng/pull/2343#pullrequestreview-694404214) about avoiding using a specific role name when testing things that are separate from explicit permissions tests. To that end, if we could change the ['as a space developer' context](https://github.com/cloudfoundry/cloud_controller_ng/pull/2431/files#diff-736eb52ed00531b22ccd8cfed45e6f1367debef2ac98c83834e23679d69032c0R306) to 'as a permitted user' that would be great. Generally I'm trying to avoid creating a pattern where folks may feel like they need to copy the whole context when introducing a new role, unless there's an outlier case.

### Docs Check

For `GET /v3/apps/:guid/ssh_enabled` it looks like we missed the 'Experimental' note, but otherwise looks good!

sweinstein22 (approved):
Made one minor change, we needed to include the new stack endpoint in the docs index so it appears in the sidebar, but otherwise looks good. I'm going to let the checks run but otherwise this should be good to merge in!

moleske:
> 
> 
> Made one minor change, we needed to include the new stack endpoint in the docs index so it appears in the sidebar, but otherwise looks good. I'm going to let the checks run but otherwise this should be good to merge in!

D'oh! Thanks for catching that!

PR #2438, opened 2021-07-23
A service instance can no longer shared if a service instance's name is already in use in the target space.

bepotts:
This issue resolves GitHub issue #2306 . 

A service instance can no longer shared if a service instance's name is already in use in the target space.

### DIFF
--- a/app/actions/service_instance_share.rb
+++ b/app/actions/service_instance_share.rb
@@ -57,6 +57,11 @@ module VCAP::CloudController
         error_msg = "A service instance called #{service_instance.name} already exists in #{space.name}."
         error!(error_msg)
       end
+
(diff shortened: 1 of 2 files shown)


sweinstein22 (approved):
## Acceptance

Confirmed that a service instance named `si` was already shared with org `target` space `target`:
```
$ cf services
Getting service instances in org target / space target as admin...

name   offering           plan    bound apps   last operation     broker            upgrade available
si     overview-service   small                create succeeded   overview-broker   no
```
Tried sharing service `si` from org `source` space `space-2`:
```
$ cf target -o source -s space-2
API endpoint:   https://api.clover-horse.capi.land
API version:    3.102.0
user:           admin
org:            source
space:          space-2
A service instance called si has already been shared with target.
FAILED
```

Looks good to me! Snyk is failing, not sure why, but other than that this should be good to merge in.

PR #2439, opened 2021-07-23
Add route destinations protocols

MerricdeLauney:

Add protocols to route_destinations

We are not using this field yet. next we will use this to decorate the
routing information of the LRPs. This change will be added to this PR.

### DIFF
--- a/app/actions/update_route_destinations.rb
+++ b/app/actions/update_route_destinations.rb
@@ -132,10 +132,7 @@ module VCAP::CloudController
       def compare_destination_hashes(new_route_mappings, existing_route_mappings)
         new_route_mappings.reject do |new|
           matching_route_mapping = existing_route_mappings.find do |existing|
-            new[:app_guid] == existing[:app_guid] &&
-              new[:process_type] == existing[:process_type] &&
-              new[:route_guid] == existing[:route_guid] &&
-              new[:app_port] == existing[:app_port]
+            new.slice(:app_guid, :process_type, :route_guid, :app_port) == existing.slice(:app_guid, :process_type, :route_guid, :app_port)
           end
 
           if matching_route_mapping && matching_route_mapping[:protocol] != new[:protocol]
--- a/app/actions/update_route_destinations.rb
+++ b/app/actions/update_route_destinations.rb
@@ -21,7 +21,7 @@ module VCAP::CloudController
           raise Error.new('Destinations cannot be inserted when there are weighted destinations already configured.')
         end
 
-        to_add = new_route_mappings - existing_route_mappings
+        to_add = compare_destination_hashes(new_route_mappings, existing_route_mappings)
 
         update(route, to_add, [], user_audit_info, manifest_triggered)
       end
@@ -57,7 +57,8 @@ module VCAP::CloudController
           processes_to_ports_map = {}
 
           to_delete.each do |rm|
-            route_mapping = RouteMappingModel.find(rm)
+            rm.delete(:protocol)
+            route_mapping = RouteMappingModel[rm]
             route_mapping.destroy
 
             Copilot::Adapter.unmap_route(route_mapping)
@@ -76,6 +77,8 @@ module VCAP::CloudController
           end
 
           to_add.each do |rm|
+            validate_protocol_matches(rm)
+
             route_mapping = RouteMappingModel.new(rm)
             route_mapping.save
 
@@ -126,6 +129,23 @@ module VCAP::CloudController
         raise Error.new(e.message)
       end
 
+      def compare_destination_hashes(new_route_mappings, existing_route_mappings)
+        new_route_mappings.reject do |new|
+          matching_route_mapping = existing_route_mappings.find do |existing|
+            new[:app_guid] == existing[:app_guid] &&
+              new[:process_type] == existing[:process_type] &&
+              new[:route_guid] == existing[:route_guid] &&
+              new[:app_port] == existing[:app_port]
+          end
+
+          if matching_route_mapping && matching_route_mapping[:protocol] != new[:protocol]
+            raise Error.new('Destination exists with conflicting protocol')
+          end
+
+          matching_route_mapping
+        end
+      end
+
       def add_route(destinations, route)
         destinations.map do |dst|
           dst.merge({ route: route, route_guid: route.guid })
@@ -163,7 +183,8 @@ module VCAP::CloudController
           process_type: destination.process_type,
           app_port: destination.app_port,
           route: route,
-          weight: destination.weight
+          weight: destination.weight,
+          protocol: destination.protocol_without_defaults
         }
       end
 
@@ -173,6 +194,21 @@ module VCAP::CloudController
 
       def validate_unique!(new_route_mappings)
         raise DuplicateDestinationError.new('Destinations cannot contain duplicate entries') if new_route_mappings.any? { |rm| new_route_mappings.count(rm) > 1 }
+      end
+
+      def validate_protocol_matches(route_destination)
+        destination_protocol = route_destination[:protocol]
+        return unless destination_protocol
+
+        route_protocol = route_destination[:route].protocol
+        dest_to_route_map = { 'tcp' => 'tcp', 'http1' => 'http', 'http2' => 'http' }
+
+        raise protocol_mismatch_error(route_protocol, destination_protocol) if route_protocol != dest_to_route_map[destination_protocol]
+      end
+
+      def protocol_mismatch_error(route_protocol, destination_protocol)
+        valid_options = { 'tcp' => 'tcp', 'http' => 'http1, http2' }
+        Error.new("Cannot use '#{destination_protocol}' protocol for #{route_protocol} routes; valid options are: [#{valid_options[route_protocol]}].")
       end
 
       def validate_max_destinations!(existing_route_mappings, new_route_mappings)
--- a/app/actions/update_route_destinations.rb
+++ b/app/actions/update_route_destinations.rb
@@ -12,6 +12,7 @@ module VCAP::CloudController
       def add(new_route_mappings, route, apps_hash, user_audit_info, manifest_triggered: false)
         existing_route_mappings = route_to_mapping_hashes(route)
         new_route_mappings = update_port(new_route_mappings, apps_hash)
+        new_route_mappings = update_protocol(new_route_mappings, route)
         new_route_mappings = add_route(new_route_mappings, route)
 
         validate_max_destinations!(existing_route_mappings, new_route_mappings)
@@ -29,6 +30,7 @@ module VCAP::CloudController
       def replace(new_route_mappings, route, apps_hash, user_audit_info, manifest_triggered: false)
         existing_route_mappings = route_to_mapping_hashes(route)
         new_route_mappings = update_port(new_route_mappings, apps_hash)
+        new_route_mappings = update_protocol(new_route_mappings, route)
         new_route_mappings = add_route(new_route_mappings, route)
 
         validate_max_destinations!([], new_route_mappings)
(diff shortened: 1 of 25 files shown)


mkocher:
Note to future self: not completely happy with the destination.protocol_without_defaults in actions/update_route_destinations.rb. Consider converting the hashes to objects?

ctlong:
cloudfoundry/routing-release#200

docs/v3/source/includes/resources/routes/_destination_object.md.erb
```diff
@@ -25,3 +25,4 @@ If a destination does not specify a port, the default port depends on the app li
 | **app.process.type**        | _string_                 | Type of the process belonging to the app to route traffic to                                                                      |
 | **port**                    | _integer_                | Port on the destination process to route traffic to
 | **weight** (*experimental*) | _integer_ or `null`      | Percentage of traffic which will be routed to this destination; _this feature is only available via the BETA Service Mesh routing plane_ |
+| **protocol**                | _string_                 | Protocol to use, valid protocols are `http1`, `http2`, or `tcp`; _this feature is only available if you turn it on in routing release_ |
```
ctlong, line 28: Should we declare this as experimental?

Gerg: Yes

app/actions/update_route_destinations.rb
```diff
@@ -126,6 +129,23 @@ def update_processes(processes_to_ports_map)
...
+        new_route_mappings.reject do |new|
+          matching_route_mapping = existing_route_mappings.find do |existing|
+            new[:app_guid] == existing[:app_guid] &&
+              new[:process_type] == existing[:process_type] &&
+              new[:route_guid] == existing[:route_guid] &&
+              new[:app_port] == existing[:app_port]
+          end
+
+          if matching_route_mapping && matching_route_mapping[:protocol] != new[:protocol]
+            raise Error.new('Destination exists with conflicting protocol')
```
Gerg, line 142: Will users be able to hit this error when using the replace (`PATCH`) destinations endpoint?

If so, I don't think we want to error if the user is trying to replace all destinations, and one of them has a conflicting protocol with an existing destination.

MerricdeLauney: Currently, this method is only being used in the add (POST) cast so i think it's safe. However the name is misleading since it makes it seem more benign than it actually is. I could see someone trying to refactor this method into the update case and causing bugs down the line.

spec/request/route_destinations_spec.rb
```diff
@@ -590,6 +596,72 @@
...
+          route: route,
+          process_type: 'assistant',
+          protocol: existing_protocol
+        )
+      end
+      context 'http1/http2' do
+        [
+          ['http2', 'http1', 'http1'],
+          ['http2', 'http2', 'http2'],
+          ['http2', nil,     'http1'],
```
sethboyles, line 627: can you help me understand this case?  Why would `nil` change from `http2` to `http1`?  Would this happen if `protocol` was completely elided or only if it is explicitly set to `null`?

docs/v3/source/includes/resources/routes/_destination_object.md.erb
```diff
@@ -18,10 +18,11 @@ Note that when using a custom port, the app process must be listening on the spe
...
+| **guid**                      | _uuid_                   | Unique identifier for the destination                                                                                                    |
+| **app.guid**                  | _uuid_                   | Unique identifier for the app to route traffic to                                                                               |
+| **app.process.type**          | _string_                 | Type of the process belonging to the app to route traffic to                                                                      |
+| **port**                      | _integer_                | Port on the destination process to route traffic to
+| **weight** (*experimental*)   | _integer_ or `null`      | Percentage of traffic which will be routed to this destination; _this feature is only available via the BETA Service Mesh routing plane_ |
+| **protocol** (*experimental*) | _string_                 | Protocol to use, valid protocols are `http1`, `http2`, or `tcp`; _this feature is only available if you turn it on in routing release_ |
```
sethboyles, line 28: based on the tests below, `null` is a valid value, too--or am I missing something?  Should that behavior be clarified here?

app/actions/update_route_destinations.rb
```diff
@@ -126,6 +129,23 @@ def update_processes(processes_to_ports_map)
         raise Error.new(e.message)
       end
 
+      def compare_destination_hashes(new_route_mappings, existing_route_mappings)
+        new_route_mappings.reject do |new|
+          matching_route_mapping = existing_route_mappings.find do |existing|
+            new[:app_guid] == existing[:app_guid] &&
+              new[:process_type] == existing[:process_type] &&
+              new[:route_guid] == existing[:route_guid] &&
+              new[:app_port] == existing[:app_port]
```
sethboyles, line 138: if you like, can one or two line this:

```
new.slice(:app_guid, :process_type, :route_guid, :app_port) == existing.slice(:app_guid, :process_type, :route_guid, :app_port)
```

or

```
attrs = [:app_guid, :process_type, :route_guid, :app_port]
new.slice(attrs) == existing.slice(attrs)
```

sethboyles: or maybe use https://apidock.com/rails/v5.2.3/Hash/except if we want to compare everything but `protocol`?  that might clearer in intent

sethboyles:
Can you help me understand the complexity around defaults/nil?  Curious to hear why we didn't just make http1 an explicit default in the db.  At a quick glance, it feels to me that would clean up a lot of the alias method chains and default logic.

sethboyles:
All the commits in this PR point to a private Tracker.  Is there a publicly viewable version of the story?  It would be helpful to create a corresponding GitHub issue (if one doesn't exist), and then clean up the commits and add the GitHub issue ID instead of the private tracker story ID

Gerg:
@sethboyles On the OSS side, we have been tracking this work in cloudfoundry/routing-release#200

You can see the written proposal in https://github.com/cloudfoundry/routing-release/issues/200#issuecomment-876780684

mkocher:
We didn't make http1 the explicit default in the DB since there are TCP routes in the same table that wouldn't make sense to default to http1. So in order to save http1 as a string in the db we would need to write a migration that joined in the route & domain and set the value for each row appropriately.

It's worth spending some time investigating the above alternative or something else entirely. The "we'll just default it in code" approach started out meh and got worse from there as the edge cases piled up.

app/models/runtime/route.rb
```diff
@@ -183,6 +183,10 @@ def tcp?
       domain.shared? && domain.tcp? && port.present? && port > 0
     end
 
+    def protocol
+      self.domain.protocols.first
```
sethboyles, line 187: How does the route's protocol interplay with a destination's protocol?  It might be helpful to clarify the new behavior in the [Route Object's documentation](https://v3-apidocs.cloudfoundry.org/version/3.102.0/index.html#the-route-object) since the description seems to be out of date: 

```
The protocol supported by the route, based on the route’s domain configuration. Valid protocols are http and tcp
```

sethboyles:
@mkocher I am trying to familiarize myself with the intricacies of how domain protocols work, let me know if this matches your understanding.

As far as I can tell, despite domains having a `supported_protocols` attribute, which ostensibly is a list of supported protocols, in actuality you cannot have a domain support both `http` and `tcp` at the same time:

```
TCP domains are defined as being associated with the TCP Router Group. The TCP Router Group defines the range of ports available to be reserved with TCP Routes. Currently, only Shared Domains can be TCP
```

https://docs.cloudfoundry.org/devguide/deploy-apps/routes-domains.html#http-vs-tcp-shared-domains

The fact that we don't support creating custom routing groups, and domains can only change their protocol by being assigned to the one default TCP router group, means that we can rely on domains only having one protocol in the `supported_protocols` array, right? 

I guess I should have just looked at the code first: 

https://github.com/cloudfoundry/cloud_controller_ng/blob/a698d407d9f11263152cfdc4317f4786567bb16f/app/models/runtime/shared_domain.rb#L64-L68

we only ever return one element in the supported protocols array

So a migration using joins like you stated should be possible.  It seems like a straightforward enough join to me that it is worth investigating.  Something like update all route_mappings where domains have a tcp router-group?  I think by doing the migration upfront it probably would help clean up the code.  

I guess the question is if that migration is will be performant enough for large foundations?

app/models/runtime/route_mapping_model.rb
```diff
@@ -17,6 +17,24 @@ class RouteMappingModel < Sequel::Model(:route_mappings)
     one_to_many :processes, class: 'VCAP::CloudController::ProcessModel',
       primary_key: [:app_guid, :process_type], key: [:app_guid, :type]
 
+    def default_protocol_mapping
+      @default_protocol_mapping ||= { 'tcp' => 'tcp', 'http' => 'http1', nil: nil }
```
sethboyles, line 21: `nil:` is a symbol, should this be `nil => nil`

sethboyles: or just removed entirely?

app/actions/update_route_destinations.rb
Gerg: While reviewing with Seth, we were also wondering if this error might be confusing in cases where the user hasn't specified a protocol in their request. This could then default to `http1` and conflict with an existing `http2` destination.

ctlong:
After some investigation we concluded that it makes sense to stick with the "default it in code" approach for RouteMappings protocols because writing a migration to set each row of the RouteMappings table to have the correct protocol proved very tricky.

Part of the calculation to check if a RouteMapping protocol is `route_mapping.route.domain.router_group.type == 'tcp'`. However, RoutingGroups are not contained in the CC DB, and therefore any migration to set RouteMapping protocols to contain the correct value would need to make external calls to the Routing API.

cc @moleske @sethboyles

app/models/runtime/route.rb
ctlong: That description looks correct to me. We expanded upon the connection between route and route destination protocols in the route destination documentation instead of in the route documentation.

app/actions/update_route_destinations.rb
ctlong: > we were also wondering if this error might be confusing in cases where the user hasn't specified a protocol in their request. This could then default to http1 and conflict with an existing http2 destination.

Currently, an unspecified protocol in the request does default to `http1`, and that could conflict with an existing `http2` destination. @sethboyles is the ask to change the behavior in this case to not error and not update the protocol, or is it simply to change the error message?

sethboyles: Hm, this is a tough one.  I think if we are keeping the null/default behavior, and telling api consumers it behaves like this: 

https://github.com/cloudfoundry/cloud_controller_ng/pull/2439/files#diff-ad56f3840f191aea04f2453955cda4fb69cd82a80272a01e17442a73fd40d239R28

then maybe we have to assume intentionality on the client's part and stick to the contract we state in the documentation.  That is, that `null` is an attempt to set a route destination's protocol to `http`.  It could otherwise lead to a surprising scenario wherein a user **thought** they set a destination to `http`, only for the API to silently ignore that 'insert', and who knows how long it would take to discover that their destination, which should be `http1`, is set to `http2`.

By erroring, we (perhaps annoyingly) alert the user that the route destination already exists with a different protocol.  But an immediate surprise is better than one that goes undiscovered, I think.  

However the current error message won't be very helpful if the client is attempting to insert several destinations, even for cases outside this particular null/default issue.  I think in the error message should include all identifying attributes of the offending destination(s), as well as the protocol (including what it was attempted to be set to and what the existing protocol is).

app/models/runtime/route.rb
sethboyles: Makes sense.  Before our sync last week, I was expecting `domain.protocol` to support `tcp`, `http1`, or `http2`, but understand now that it is only on the route destination where the distinction between `http1` and `http2` is made.

ctlong:
Is this all work we should be feature flagging?

cc @mariash @sethboyles

app/actions/update_route_destinations.rb
ctlong: We updated the error message to include more information: `Destination ROUTE_URI for app APP_GUID with process PROCESS_TYPE exists with conflicting protocol EXISTING_PROTOCOL, can't create destination with protocol NEW_PROTOCOL`

lib/cloud_controller/diego/protocol/routing_info.rb
```diff
@@ -56,18 +24,52 @@ def routing_info
...
 
-        def route_id_app_ports_map
-          process.route_mappings(reload: true).each_with_object({}) do |route_mapping, route_app_port_map|
-            route_app_port_map[route_mapping.route_guid] ||= []
-            route_app_port_map[route_mapping.route_guid].push(get_port_to_use(process, route_mapping))
+        def http_info
+          relevant_http_routes = process.routes.reject do |route|
+            route.internal? ||
+              (route.tcp? &&
+                route.domain.router_group.present?)
```
sethboyles, line 34: correct me if I am wrong, but routes can only be `tcp` if there is a router_group present, right?  I **think** the `route.domain.router_group.present?` check is superfluous

https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/models/runtime/shared_domain.rb#L74

sethboyles (approved):
LGTM!

PR #2443, opened 2021-07-26
Wrap all errors in delete_non_transactional_subresources

### LINKED ISSUE #2445: Follow-up for #2147 - Real error is hidden behind an UnknownError
## Issue

This is a follow-up for issue #2147. We are still seeing jobs failing with an `UnknownError` although the `delayed_jobs` database table contains the detailed error.

## Context

With PR #2148, it was ensured that compound errors are saved correctly to the database. But for some reason only specific errors (i.e. `AsyncBindingDeletionsTriggered`, see https://github.com/cloudfoundry/cloud_controller_ng/blob/92e3ae6fecbaaa286d359bdcf733c19f4533510f/app/actions/app_delete.rb#L107) are returned from the `delete_non_transactional_subresources` method and then raised as a `SubResourceError` (see https://github.com/cloudfoundry/cloud_controller_ng/blob/92e3ae6fecbaaa286d359bdcf733c19f4533510f/app/actions/app_delete.rb#L44) which will later on be translated into a `CompoundError`.

All other errors are raised as they are in `delete_non_transactional_subresources` (only the first one, see https://github.com/cloudfoundry/cloud_controller_ng/blob/92e3ae6fecbaaa286d359bdcf733c19f4533510f/app/actions/app_delete.rb#L108).

If - for example - the service binding deletion fails due to a service broker error (e.g. https://github.com/cloudfoundry/cloud_controller_ng/blob/92e3ae6fecbaaa286d359bdcf733c19f4533510f/app/actions/v3/service_binding_delete.rb#L100), the `UnprocessableDelete` error is raised as is. This results in an `UnknownError` reported to the user.

## Expected result

The expected result is a clear error message that contains the root cause of the error instead of `UnknownError`.

## Possible Fix

A possible fix (as proposed in PR #2443) would be to wrap all errors in the `delete_non_transactional_subresources` method in a `SubResourceError`.

philippthun:
Fixes #2445.

It is unclear to me, why `delete_non_transactional_subresources` differentiates between `AsyncBindingDeletionsTriggered` and all other errors. In the end, all errors are raised, the difference is: first error raised as is vs. all errors wrapped into a `SubResourceError`.

So the idea of this PR is to always raise a `SubResourceError` containing all errors returned from `delete_bindings`. This ensures proper logging and saving to the database when executed as a job.

### DIFF
--- a/app/actions/app_delete.rb
+++ b/app/actions/app_delete.rb
@@ -38,11 +38,7 @@ module VCAP::CloudController
       apps.each do |app|
         logger.info("Deleting app: #{app.guid}")
 
-        errs = delete_non_transactional_subresources(app)
-
-        if errs&.any?
-          raise SubResourceError.new(errs)
-        end
+        delete_non_transactional_subresources(app)
 
         app.db.transaction do
           app.lock!
@@ -104,10 +100,7 @@ module VCAP::CloudController
 
     def delete_non_transactional_subresources(app)
       errors = delete_bindings(app.service_bindings, user_audit_info: @user_audit_info)
-      non_async_errors = errors.reject { |e| e.is_a?(AsyncBindingDeletionsTriggered) }
-      raise non_async_errors.first unless non_async_errors.empty?
-
-      errors
+      raise SubResourceError.new(errors) if errors.any?
     end
 
     def stagers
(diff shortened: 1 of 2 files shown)


sethboyles:
Thanks for the PR!

PR #2452, opened 2021-08-03
Tokens slow uploads

will-gant:

Uses the upload start time instead of the current time when calculating authorization token validity for POST requests to `/v3/packages/:guid/upload`.

Currently users whose token is still valid at the time a package upload starts, but whose token expires before that upload finishes, are presented with a CF-InvalidAuthToken error. Although the option to set a grace period already exists, this is only useful to users who are able to easily change platform-wide configuration, and who can predict the time their uploads will need.

There will be a link to a PR to capi-release that I intend to make immediately after this one...


lib/cloud_controller/uaa/uaa_token_decoder.rb
```diff
@@ -7,12 +7,13 @@ class BadToken < StandardError
 
     attr_reader :config
 
-    def initialize(uaa_config, grace_period_in_seconds=0)
+    def initialize(uaa_config, grace_period_in_seconds=0, alternate_reference_time=nil)
```
philippthun, line 10: What about adding a constraint that only one of the properties `grace_period_in_seconds` and `alternate_reference_time` can be set? I don't think that it makes much sense to have an alternate reference time and a grace period. So if both are provided (i.e. != 0 and != nil), an exception could be raised.

middleware/security_context_setter.rb
```diff
@@ -1,16 +1,29 @@
...
         header_token = env['HTTP_AUTHORIZATION']
+        request_path = env['REQUEST_PATH']
 
         @security_context_configurer.configure(header_token)
 
+        if request_path && request_path.match(PACKAGES_UPLOAD_REGEX)
+          upload_start_time = Rack::Request.new(env).params['upload_start_time'].to_i
+          if upload_start_time
+            auth                  = env['HTTP_AUTHORIZATION']
+            grace_period          = VCAP::CloudController::Config.config.get(:app_bits_upload_grace_period_in_seconds)
```
philippthun, line 21: I don't think that we should provide the grace period to the token decoder. Using the upload start time seems to be enough.

middleware/security_context_setter.rb
```diff
@@ -1,16 +1,29 @@
 module CloudFoundry
   module Middleware
     class SecurityContextSetter
+      PACKAGES_UPLOAD_REGEX = %r{/v3/packages/[\da-f]{8}-([\da-f]{4}-){3}[\da-f]{12}/upload}i.freeze
```
philippthun, line 4: What about a regex that is also valid for buildpacks (`/v3/buildpacks/.*/upload`) and droplets (`/v3/droplets/.*/upload`)? The form field is there for all upload endpoints and it would be more consistent.

middleware/security_context_setter.rb
```diff
@@ -1,16 +1,29 @@
...
       def call(env)
         header_token = env['HTTP_AUTHORIZATION']
+        request_path = env['REQUEST_PATH']
 
         @security_context_configurer.configure(header_token)
 
+        if request_path && request_path.match(PACKAGES_UPLOAD_REGEX)
+          upload_start_time = Rack::Request.new(env).params['upload_start_time'].to_i
+          if upload_start_time
+            auth                  = env['HTTP_AUTHORIZATION']
```
philippthun, line 20: The `HTTP_AUTHORIZATION` env is already available in variable `header_token`.

middleware/security_context_setter.rb
```diff
@@ -1,16 +1,29 @@
...
       def initialize(app, security_context_configurer)
         @app                         = app
         @security_context_configurer = security_context_configurer
       end
 
       def call(env)
         header_token = env['HTTP_AUTHORIZATION']
+        request_path = env['REQUEST_PATH']
 
         @security_context_configurer.configure(header_token)
```
philippthun, line 15: Is my understanding correct that this line will set an "invalid_token" security context and below it will be overwritten in case the upload start time is provided? If so it would be better to put this into an `else` block. Or do something like this:
```
security_context_configurer = @security_context_configurer

if ...
  security_context_configurer = VCAP::CloudController::Security::SecurityContextConfigurer.new(relaxed_token_decoder)
end

security_context_configurer.configure(header_token)
```

middleware/security_context_setter.rb
will-gant: Yeah I guess that's right. In order to achieve what you're suggesting I think I'd need to use keyword arguments or a hash (and change existing instantiations of the UAA token decoder to make them provide those instead of positional arguments).

middleware/security_context_setter.rb
will-gant: Good point!

middleware/security_context_setter.rb
will-gant: Yep fixing this with keyword arguments.

middleware/security_context_setter.rb
will-gant: Discussed offline, and I've resolved this the way you suggest.

middleware/security_context_setter.rb
will-gant: Yep fair enough. I've added the other endpoints as well.

philippthun (commented):
Looks good from my point of view.

tjvman:
Hey @will-gant, do you have a reliable way to repro this? I'm working on validating the change.

 I tried uploading a large (~1GB) app via the /v3/packages/:guid/upload endpoint right before the token expired, but didn't have much success (also bothersome because of having to wait for the token to get close to expiring)

JenGoldstrich:
Spoke with @stephanme and @tjvman who both gave the thumbs up, given that folks at SAP were able to validate this I am going to merge.

Thanks @will-gant !

Issue #2434, opened 2021-07-22
Implement a clear and user friendly error message and warning informing users they cannot access V2 endpoints while assigned only the space supporter role.


monamohebbi:
#### Issue
Implement a clear and user friendly error message informing users they cannot access V2 endpoints while assigned **only** the space supporter role.

#### Context
We are hoping to avoid untangling the permissions logic in V2 which means the Space Supporter role will not be recognized as a permitted user by any V2 endpoint. This could lead to unexpected or unexplained output if a user with this role were to be using a platform that uses mixed V2 and V3 endpoints. Instead we want to catch any V2 calls early and fail consistently with a helpful error message.

#### Dev Note:
See this [spike branch](https://github.com/cloudfoundry/cloud_controller_ng/commit/a6534f7ed061ba20972889942f3ca559cbc33c75) for inspiration.

#### Expected result
A space supporter with no other roles assigned should fail when curling any endpoint that is prefixed with a v2. The output of this message should inform them why they have failed and tell them to inform their platform admin that their permissions need to be changed. Example:
```
{
  "errors": [
    {
      "code": 10003,
      "detail": "You are not authorized to perform the requested action. See section 'Space Supporter Role in V2' https://docs.cloudfoundry.org/concepts/roles.html",
      "title": "CF-NotAuthorized"
    }
  ]
}
```

A space supporter curling a v2 endpoint with other roles assigned should receive a warning with the following message.
`This endpoint might behaved unpredictably to users assigned the SpaceSupporter role. Check with your admin if you should have this role assigned on a platform using deprecated endpoints. See section 'Space Supporter Role in V2' https://docs.cloudfoundry.org/concepts/roles.html"` 

#### Acceptance
There is a test at the base level that ensures curling endpoints with v2 prefixes as a user with only the space supporter assigned will cause the above error message.

There is a test at the base level that ensures curling endpoints with v2 prefixes as a user with multiple roles including the space supporter assigned will cause the above error message.

#### Documentation
When I browse to any of these endpoints on [valid role types](https://v3-apidocs.cloudfoundry.org/version/3.102.0/index.html#valid-role-types) page of v3 docs, I can see an indication that this role is not recognized by any V2 endpoint, and will fail if a user attempt to curl any v2 endpoint with this role assigned.

[...]

PR #2467, opened 2021-08-16
Prevent users with space supporter role reaching v2 endpoints

monamohebbi:
* applies only to non root and info endpoints
* applies only if the space supporter role is the only role assigned
* applies only if the flag to enable usage of the space supporter is set
to true
* addresses https://github.com/cloudfoundry/cloud_controller_ng/issues/2434

Co-authored-by: Mona Mohebbi <mmohebbi@pivotal.io>
Co-authored-by: Matthew Kocher <mkocher@pivotal.io>
Co-authored-by: Weyman Fung <weymanf@vmware.com>

### DIFF
--- a/docs/v3/source/includes/resources/roles/_valid_roles.md.erb
+++ b/docs/v3/source/includes/resources/roles/_valid_roles.md.erb
@@ -6,4 +6,6 @@
 - `space_auditor`
 - `space_developer`
 - `space_manager`
-- `space_supporter` - This role is under active development and is not supported
+- `space_supporter`
+  * This role is under active development and is not supported
+  * Users with only the space supporter role are not authorized to use the **V2** API
--- a/lib/cloud_controller/rack_app_builder.rb
+++ b/lib/cloud_controller/rack_app_builder.rb
@@ -8,6 +8,7 @@ require 'security_context_setter'
 require 'rate_limiter'
 require 'new_relic_custom_attributes'
 require 'zipkin'
+require 'block_v3_only_roles'
 
 module VCAP::CloudController
   class RackAppBuilder
@@ -41,6 +42,7 @@ module VCAP::CloudController
         use Rack::CommonLogger, logger if logger
 
         map '/' do
+          use CloudFoundry::Middleware::BlockV3OnlyRoles, { logger: Steno.logger('cc.unsupported_roles') }
           run FrontController.new(config)
         end
 
--- a/middleware/block_v3_only_roles.rb
+++ b/middleware/block_v3_only_roles.rb
@@ -1,0 +1,64 @@
+require 'mixins/client_ip'
+
+module CloudFoundry
+  module Middleware
+    class BlockV3OnlyRoles
+      def initialize(app, logger:)
+        @app                   = app
+        @logger                = logger
+      end
+
+      def call(env)
+        if !allowed_path?(env['PATH_INFO']) && v3_only_role_enabled? && v3_only_role?
+          [
+            403,
+            { 'Content-Type' => 'text/html' },
+            'You are not authorized to perform the requested action. See section \'Space Supporter Role in V2\' https://docs.cloudfoundry.org/concepts/roles.html',
+          ]
+        else
+          @app.call(env)
+        end
+      end
+
+      private
+
+      def v3_only_role?
+        current_user = VCAP::CloudController::SecurityContext.current_user
+        current_user.try(:id) && !globally_authenticated? && space_supporter_and_only_space_supporter?(current_user)
+      end
+
+      def allowed_path?(path)
+        ['/v2/info', '/'].include?(path)
+      end
+
+      def globally_authenticated?
+        roles.admin? || roles.admin_read_only? || roles.global_auditor?
+      end
+
+      def v3_only_role_enabled?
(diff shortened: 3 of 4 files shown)


sweinstein22:
Looks like we're getting some unit test failures, y'all mind looking into them? @monamohebbi @mkocher 

```
Auth space supporter user with only space supporter role GET /v2 endpoints does not error when hitting info
     Failure/Error: expect(last_response.status).to eq(200)

       expected: 200
            got: 403


Auth space supporter user with only space supporter role GET /v2 endpoints does not error when hitting root
     Failure/Error: expect(last_response.status).to eq(200)

       expected: 200
            got: 403
```

middleware/block_unsupported_roles.rb
```diff
@@ -0,0 +1,59 @@
...
+            'You are not authorized to perform the requested action. See section \'Space Supporter Role in V2\' https://docs.cloudfoundry.org/concepts/roles.html',
+          ]
+        else
+          @app.call(env)
+        end
+      end
+
+      private
+
+      def allowed_path?(path)
```
sethboyles, line 26: is this method used anywhere?

middleware/block_unsupported_roles.rb
```diff
@@ -0,0 +1,59 @@
...
+                     union(VCAP::CloudController::SpaceSupporter.select(Sequel.as(VCAP::CloudController::RoleTypes::SPACE_SUPPORTER, :type)).where(user_id: current_user.id)).
+                     union(VCAP::CloudController::OrganizationManager.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_MANAGER,
+:type)).where(user_id: current_user.id)).
+                     union(VCAP::CloudController::OrganizationBillingManager.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_BILLING_MANAGER,
+:type)).where(user_id: current_user.id)).
+                     union(VCAP::CloudController::OrganizationAuditor.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_AUDITOR,
+:type)).where(user_id: current_user.id)).
+                     all
+
+        user_roles.count == 1 && user_roles[0][:type] == 'space_supporter'
```
sethboyles, line 51: I think this count check would allow a SpaceSupporter who is a SpaceSupporter for two or more spaces to access these endpoints.  I think it should be more like checking that all roles for the user have `:type` == `space_supporter`.  Let me know if I'm wrong in how the modeling works.

middleware/block_unsupported_roles.rb
```diff
@@ -0,0 +1,59 @@
...
+                     union(VCAP::CloudController::SpaceDeveloper.select(Sequel.as(VCAP::CloudController::RoleTypes::SPACE_DEVELOPER, :type)).where(user_id: current_user.id)).
+                     union(VCAP::CloudController::SpaceAuditor.select(Sequel.as(VCAP::CloudController::RoleTypes::SPACE_AUDITOR, :type)).where(user_id: current_user.id)).
+                     union(VCAP::CloudController::SpaceSupporter.select(Sequel.as(VCAP::CloudController::RoleTypes::SPACE_SUPPORTER, :type)).where(user_id: current_user.id)).
+                     union(VCAP::CloudController::OrganizationManager.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_MANAGER,
+:type)).where(user_id: current_user.id)).
+                     union(VCAP::CloudController::OrganizationBillingManager.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_BILLING_MANAGER,
+:type)).where(user_id: current_user.id)).
+                     union(VCAP::CloudController::OrganizationAuditor.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_AUDITOR,
+:type)).where(user_id: current_user.id)).
+                     all
```
sethboyles, line 49: 🤯  we don't have any easier ways to do this?  Nothing defined in the user model?

middleware/block_unsupported_roles.rb
mkocher: The union operation does a unique on the set (`union all` apparently would not unique the set).

We wrote a test for this to prove to ourselves it was working. We also added a LIMIT 1 clause to all the selects as it makes it obvious and could help the query optimizer out maybe?

middleware/block_unsupported_roles.rb
mkocher: We're not seeing anything in the user model.

We could move this into the user model, or at least a function like `unique_roles` or something. Do you think that would be better?

It seems like Cloud Controller prefers skinny models which frankly is an odd design pattern. Happy to not follow it if you think we should move this into the model.

sethboyles: I would leave it to your discretion.  I don't totally object to this code living here, because it makes it clearer that it is code we can delete if/when we deprecate & delete v2

middleware/block_unsupported_roles.rb
sethboyles: Ah, got it.  Thanks.

mkocher:
Closes #2467

sethboyles (approved):

PR #2476, opened 2021-08-20
Remove ignored dataset.eager() calls

philippthun:
When calling dataset.eager().first, the eager call is ignored by sequel.
Eager loading is intended for loading multiple objects. One can make use
of eager loading also for a single object, but this makes limited sense
(i.e. same amount of SQL queries) unless there is a callback passed to
customize the eagerly loaded dataset. To use eager loading on a single
object, one has to call dataset.eager().all.first.


sethboyles:
Thanks for cleaning these up.

Issue #2489, opened 2021-09-02
No pageable way to gets users by space or org


moleske:
Thanks for submitting an issue to `cloud_controller_ng`. We are always trying to improve! To help us, please fill out the following template.

## Issue

We would like to be able to get users by space or org and the v3 endpoints do not allow that specification in a pageable format.  We could page through the roles endpoints, then process all of the roles to group by users.  But the problem with going through the roles endpoint is that we get inconsistent page sizes when it comes to displaying users and their roles in a table.

Ideally we would have a consistent way to get a page of unique users for a space or org, and then pull all their associated roles

## Context

We are updating an app to use v3 endpoints from v2 and while upgrading `GET /v2/spaces/:guid/user_roles` and `GET /v2/organizations/:guid/user_roles` endpoints we could not find a comparable way in v3 to construct a consistently sized table of users and roles via pagination.  We expect to have lots of users, so we would like to be able to paginate these requests in order to retrieve the information in a performant way.

## Possible Fix

1. Replicate the `user_roles` endpoints
2. Create a `v3/space/:guid/users` and `v3/organizations/:guid/users`
3. Add a `group_by` filter to the roles endpoint

We think 2 is the most feasible and reasonable based on the 1. being a composite endpoint and 3. might be computationally hard

sweinstein22:
Reopening as the org level endpoint has not been completed

JenGoldstrich:
This was released as part of CAPI 1.117.0

PR #2491, opened 2021-09-03
Provide pageable users of space endpoint

moleske:
- /v3/spaces/:guid/users now exists with filters

[#2489]

@ctlong


spec/request/spaces_spec.rb
```diff
@@ -1196,6 +1196,301 @@
...
+          VCAP::CloudController::User.plugin :timestamps, update_on_create: true
+        end
+
+        it 'returns 200 and filters' do
+          org.add_user(resource_1)
+          space1.add_manager(resource_1)
+          get "/v3/spaces/#{space1.guid}/users?updated_ats[lt]=#{resource_3.updated_at.iso8601}", nil, admin_headers
+
+          expect(last_response).to have_status_code(200)
+          # TODO(carson and michael): do we include user.guid and client.guid or not?
```
sweinstein22, line 1368: 👀

app/fetchers/space_users_list_fetcher.rb
```diff
@@ -0,0 +1,40 @@
...
+            label_klass: UserLabelModel,
+            resource_dataset: dataset,
+            requirements: message.requirements,
+            resource_klass: User,
+          )
+        end
+
+        dataset = super(message, dataset, User)
+
+        dataset.inner_join(SpaceDeveloper.table_name, user_id: :id).where(space_id: space.id).select_all(:users).
```
sweinstein22, line 29: Does it make more sense to keep this logic at the model layer? We already have an [established pattern for visible org users](https://github.com/cloudfoundry/cloud_controller_ng/blob/2860386d3f0a8402df451a52858df321f104fd44/app/models/runtime/user.rb#L188)

sweinstein22 (commented):
### Behavior Check

- deployed changes to a bosh lite
- ran `cf curl /v3/users` and `cf curl /v3/spaces/:guid/users` to confirm that only a subset of users was returned. User missing from the space level call was the `network-policy`

```
$ cf curl /v3/spaces/369cc547-b6b8-4ccb-a3c3-987845b04049/users | grep total_results
    "total_results": 3,

$ cf curl /v3/users | grep total_results
    "total_results": 4,
```
- confirmed that permissions for each role are correct, results are identical for all roles except OrgAuditor, which receives a `404`. This matches the error code received by OrgAuditors when curling `/v3/spaces/:guid`.

app/fetchers/space_users_list_fetcher.rb
moleske: I didn't look at user.rb before.  Cool with me to move it if that's the pattern we like

sweinstein22 (approved):
Running through acceptance again as the implementation changed pretty significantly.

### Behavior Check

As a space supporter, confirmed that the list of users returned for a space is a subset of those returned from `/v3/users`:
```
$ cf curl /v3/spaces/369cc547-b6b8-4ccb-a3c3-987845b04049/users | grep total_results
    "total_results": 6,
$ cf curl /v3/users | grep total_results
    "total_results": 8,
```
Also confirmed that output of endpoint looked correct, and then checked to make sure there wasn't a diff in output between roles.
```
$ diff space-supporter space-manager

$ diff space-supporter space-dev

$ diff space-supporter space-auditor

$ diff space-supporter org-manager

```
Confirmed that org auditors get a `404`:
```
$ cf curl /v3/spaces/369cc547-b6b8-4ccb-a3c3-987845b04049/users -v
REQUEST: [2021-09-08T09:58:20-07:00]
GET /v3/spaces/369cc547-b6b8-4ccb-a3c3-987845b04049/users HTTP/1.1
...

RESPONSE: [2021-09-08T09:58:20-07:00]
HTTP/1.1 404 Not Found
...
{
  "errors": [
    {
      "code": 10010,
      "detail": "Space not found",
      "title": "CF-ResourceNotFound"
    }
  ]
}
```

### Code Check

Looks good! Thanks for rearranging things so we would have more consistent patterns, and for introducing a new model to minimize how much data processing we're doing

### Docs Check

New endpoint is documented, renders as expected

PR #2499, opened 2021-09-13
Improve label selection performance

philippthun:
By replacing `JOIN`s with `WHERE` clauses for label selection, the performance can be improved significantly.

A locally running Cloud Controller (with CCDB in a Docker container) was used to measure the impact of this change.

500 spaces have been created (without any labels, i.e. the returned list was always empty). The following command was executed:
```
cf curl "/v3/spaces?label_selector=key0,key1,...,keyn"
```

The captured `X-Runtime` header value:

|# of keys|before change|after change|
|---|---|---|
|5|0.060553|0.007187|
|10|1.110025|0.017798|
|20|49.877381|0.039163|
|50|504 Gateway Time-out|0.349810|

Besides improving the database query performance, it is also wise to limit the number of `label_selector` requirements to prevent the creation of too complex/large queries. 50 selectors should still be enough for almost all use-cases.

* [x] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats) *

\* We cannot run all CATS on our foundations as we do not enable all CF features, but all the executed tests were successful:
```
Ran 136 of 216 Specs
SUCCESS! -- 136 Passed | 0 Failed | 3 Pending | 77 Skipped
```

### DIFF
--- a/app/fetchers/label_selector_query_generator.rb
+++ b/app/fetchers/label_selector_query_generator.rb
@@ -2,65 +2,23 @@ module VCAP::CloudController
   class LabelSelectorQueryGenerator
     class << self
       def add_selector_queries(label_klass:, resource_dataset:, requirements:, resource_klass:)
-        requirements.reduce(nil) do |accumulated_dataset, requirement|
+        resource_guid = Sequel.qualify(resource_klass.table_name, :guid)
+        requirements.each do |requirement|
           case requirement.operator
-          when :in
-            dataset_for_requirement = evaluate_in(label_klass, resource_dataset, requirement, resource_klass)
-          when :notin
-            dataset_for_requirement = evaluate_notin(label_klass, resource_dataset, requirement, resource_klass)
-          when :equal
-            dataset_for_requirement = evaluate_equal(label_klass, resource_dataset, requirement, resource_klass)
-          when :not_equal
-            dataset_for_requirement = evaluate_not_equal(label_klass, resource_dataset, requirement, resource_klass)
+          when :in, :equal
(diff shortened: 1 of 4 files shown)


FloThinksPi:
Nice ! 
I am thinking where we should document this so it shows up in api docs, not in this PR i guess.
Also does it make sense to make the labels limit configureable foundation wide and default it to 50 ?  Could be nice to allow operators to support certain use cases for their foundation  - Opinions on that ?

philippthun:
> I am thinking where we should document this so it shows up in api docs, not in this PR i guess.

I've added a sentence to the `Labels and selectors` section of the API docs.

> Also does it make sense to make the labels limit configureable foundation wide and default it to 50 ? Could be nice to allow operators to support certain use cases for their foundation - Opinions on that ?

I'm pretty sure that 50 requirements in a single selector is enough and thus would hard-code this value.

JenGoldstrich (approved):

JenGoldstrich:
Hey @philippthun 

Thanks for this PR

I validated these changes by reading thru the code, deploying them to a bosh lite, setting 50 labels on an app and then timing running 20 label selectors in a /v3/apps call.  It finished, as your test indicates, in less than a tenth of a second.

Its also worth noting that the max number of labels on a resource is configurable, but [capped at 50](https://bosh.io/jobs/cloud_controller_ng?source=github.com/cloudfoundry/capi-release&version=1.117.0#p%3dcc.max_labels_per_resource) by default so 50 seems like a reasonable limit for the label selector.

Merging this!

PR #2502, opened 2021-09-14
Update membership to make use of SpaceRole and OrgRole models

monamohebbi:

This makes the code more readable and should not affect performance
too much because the new space and org role models will only union the
tables that we would be doing joins or unions on regardless. 

This code is more readable and more sustainable. It eliminates a some repeated sequel unions and joins. This will make this are of the codebase more sustainable moving forward.

### DIFF
--- a/lib/cloud_controller/membership.rb
+++ b/lib/cloud_controller/membership.rb
@@ -1,51 +1,42 @@
 module VCAP::CloudController
   class Membership
-    SPACE_DEVELOPER             = 0
-    SPACE_MANAGER               = 1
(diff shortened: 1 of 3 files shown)


JenGoldstrich:
Thanks for the PR @monamohebbi! 

I validated these changes by pushing them to a bosh lite, pushing an app in an org that my test user didn't have access to, and then trying to `cf curl /v3/app/(app-guid)` that app to ensure that permissions still worked as expected after the refactor, got an app not found error for the user with insufficient permissions as expected.

Merging this!

PR #2505, opened 2021-09-15
Do not count admin requests

philippthun:
As admins are not rate limited, counting their requests in the database is not required and leads to unnecessary `UPDATE` statements on the `request_counts` table. This change skips rate limiting for admins completely, so that they also won't see any `X-RateLimit-*` headers.


JenGoldstrich:
Deployed this to a bosh lite, and validated that admin users would not generate a write to the request_counts table, switched to a non-admin user and was able to validate request_count continues to increment.

Also validated that I was still able to hit the rate limit as a normal user.

Merging, thank you @philippthun !

Issue #2504, opened 2021-09-15
Confusing exception message when using ```cf update-service-broker```


JamesMustafa:
## Issue

When trying to update from one service broker to another using the ```cf update-service-broker``` command I got the following exception message:
```
Updating service broker (broker-name) as (my-username)...
Job (bfe405d8-952d-4bc5-a250-43952a4248d2) failed: An unknown error occurred.
FAILED
```

I tried to find more detailed information about the error and its cause in the logs but unfortunately nothing more useful was introduced there.

## Context

In order to update from one broker to other without making any changes to the offerings I used the same service/plan names in the catalog. I copied the service id as well but unfortunately forgot to replace the plan id from the old broker, which caused this error.

## Steps to Reproduce

1. Deploy a sample broker from [https://github.com/spring-cloud-samples/bookstore-service-broker](https://github.com/spring-cloud-samples/bookstore-service-broker)
2. Register the broker using ```cf create-service-broker``` command and enable the service plan.
3. Change the [plan Id](https://github.com/spring-cloud-samples/bookstore-service-broker/blob/main/src/main/java/org/springframework/cloud/sample/bookstore/servicebroker/config/ServiceCatalogConfiguration.java#L31) to some other guid.
4. Re-deploy the application
5. Use the ```cf update-service-broker``` command with the same parameters as the ```cf create-service-broker``` command in step 2.

## Expected result

More accurate exception title and message. For example: ```Plan name must be unique for the given plan id``` or just ```Plan name must be unique```

## Current result

```
Updating service broker (broker-name) as (my-username)...
Job (bfe405d8-952d-4bc5-a250-43952a4248d2) failed: An unknown error occurred.
FAILED
```

philippthun:
Hi @JamesMustafa! On which CAPI version are you experiencing this error? When you call `api.cf.<your-foundation>/`, what version number is shown for `cloud_controller_v3`? I'm asking as there has been an improvement (PR #2443) with regards to error handling in CAPI 1.115.0 which resolves to `cloud_controller_v3.version = 3.104.0`.

philippthun:
With PR #2510 the error message changes to:
```
Updating service broker (broker-name) as (my-username)...
Job (31bc193d-c769-436e-9e24-ec3c7f54b5f3) failed: Plan names must be unique within a service. Service bookstore already has a plan named standard
FAILED
```

JamesMustafa:
Hi @philippthun, thanks for your support :)

[...]

PR #2510, opened 2021-09-20
Do not treat errors from CloudController modules as UnknownErrors

philippthun:
Every error that was neither a compound error (responding to
'underlying_errors'), nor an api error (responding to 'code'), nor a
service error (type StructuredError) was treated as UnknownError by the
V3ErrorHasher. This means that for all other errors the not so helpful
message "An unknown error occurred." was returned to the client.

But as there are many explicitly defined StandardErrors spread
throughout the cloud controller code base, it makes more sense to
differentiate between these concrete (and known) errors and other e.g.
RuntimeErrors.

With this change all StandardErrors from modules 'CloudController' or
'VCAP::CloudController' are treated as known errors and presented in a
standard way:
 title: CF-<name of error class>
 detail: <error message>
 code: 10001

Additionally this change contains some renaming and refactoring for
consistency.


philippthun:
I've deployed this commit into a dev foundation and verified that the `cf update-service-broker` command (as mentioned in #2504) now fails with a proper error message:
```
Updating service broker (broker-name) as (my-username)...
Job (31bc193d-c769-436e-9e24-ec3c7f54b5f3) failed: Plan names must be unique within a service. Service bookstore already has a plan named standard
FAILED
```

MerricdeLauney:
Hey @philippthun. I'd be curious to hear what made you decide to go in a different direction than this solution? @sweinstein22 and I were thinking that implementing something along these lines would be the the most straightforward fix to #2504 as you previously noted.

philippthun:
This approach still does not solve the problem that a wrong error hasher could be used. That's the main purpose of #2511 and I find it still fundamental that a potential error is formatted according to the request that triggered it (i.e. v2 or v3).

Besides that one could decide to return arbitrary errors instead of the pre-defined ones, but I had the feeling that this was not the original intention.

PR #2521, opened 2021-09-28
Enforce service name uniqueness in shared services in spaces

JenGoldstrich:
Fixes https://github.com/cloudfoundry/cloud_controller_ng/issues/2306

Tested this locally on a bosh lite using the repro steps in the above issue, 

```
A service instance called si has already been shared with target.
FAILED
```

Locally rake bundle exec rake spec and it passed.

Co-authored-by: Brandon Potts <bpotts@vmware.com>
Co-authored-by: Jenna Goldstrich <jgoldstrich@pivotal.io>


MarcPaquette (approved):
LGTM

PR #2525, opened 2021-09-30
Improve efficiency of pagination count query

andy-paine:
Only count the primary key fields for DISTINCT queries

When paginating results, a COUNT is performed to get the total number of
results that would be returned by the query. For queries with a
DISTINCT in their SELECT, this requires the DB to perform a full
sort over the entire SELECT table.* columns (sometimes 15+) which can be
very slow. Since this is only being used for a COUNT, only a unique
identifier is needed to do the count which is already unique so does not
require sorting. This halves the EXPLAIN cost of some of the
service_plans queries since that is currently sorting over 15+ fields
just to return a single COUNT.

### DIFF
--- a/lib/cloud_controller/paging/sequel_paginator.rb
+++ b/lib/cloud_controller/paging/sequel_paginator.rb
@@ -9,15 +9,20 @@ module VCAP::CloudController
       order_direction = pagination_options.order_direction
 
       table_name = sequel_dataset.model.table_name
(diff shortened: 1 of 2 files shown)


andy-paine:
This doesn't actually perform correctly if the `SELECT DISTINCT` query is selecting from multiple tables such as `SELECT DISTINCT a.id, a.b, c.id, c.d`) as it would count identical `a.id` rows as only 1 when in fact they are different as the `c.id` differs between them. Closing in favour of #2527 which achieves better performance improvements whilst keeping the same query semantics in all scenarios

PR #2528, opened 2021-10-01
Only join required tables

philippthun:
`ServicePlanListFetcher` and `ServiceOfferingListFetcher` `JOIN` multiple tables to be used in `WHERE` conditions. The actual set of tables to be joined depends on the provided filters. Instead of joining a superset of tables, this change ensures that only the required tables will be joined.

Unfortunately Postgres does not remove unnecessary `JOIN`s and so they possibly result in slower query execution times.

Before using another table in a `WHERE` clause, one simply has to call `join_<table>(dataset)`. This can even be done multiple times for each table as the underlying `join(dataset, type, table, on)` function ensures that only unique table expressions are added.

### DIFF
--- a/app/fetchers/base_service_list_fetcher.rb
+++ b/app/fetchers/base_service_list_fetcher.rb
@@ -8,15 +8,18 @@ module VCAP::CloudController
       private
 
       def select_readable(dataset, message, omniscient: false, readable_space_guids: [], readable_org_guids: [])
-        dataset = join_tables(dataset, message, omniscient)
-
         if readable_org_guids.any?
+          dataset = join_service_plans(dataset)
+          dataset = join_plan_orgs(dataset)
+          dataset = join_broker_spaces(dataset)
+
           dataset = dataset.where do
             (Sequel[:service_plans][:public] =~ true) |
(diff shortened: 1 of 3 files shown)


philippthun:
We ran `EXPLAIN (ANALYZE, BUFFERS)` for a `SELECT DISTINCT "service_plans".* FROM "service_plans"` query with the following `WHERE` condition:
```
WHERE ((("service_plans"."public" IS TRUE)
  OR ("plan_orgs"."guid" IN ('org_guid'))
  OR ("broker_spaces"."guid" IN ('space_guid')))
AND ("service_instances"."guid" IN ('service_instance_guid')));
```

The main indicators were:
```
Cost: 324,264.94
Buffers: shared hit=67150, temp read=16048 written=16048
Execution Time: 3331.547 ms
```

After removing the (in this case) unnecessary joins `LEFT JOIN "organizations" AS "broker_orgs"` and `LEFT JOIN "spaces" AS "plan_spaces"` the `EXPLAIN (ANALYZE, BUFFERS)` changes to:

```
Cost=32,145.98
Buffers: shared hit=14007
Execution Time: 65.044 ms
```

So omitting unnecessary `JOIN`s can have a huge impact (on Postgres)!

PR #2532, opened 2021-10-05
Simplify roles query for performance

andy-paine:
Replace UNION-ed subqueries with single subquery and WHERE OR clause

The many UNIONs (14 in total) were causing these queries to be extremely complex and slow on our largest landscapes resulting in impact to other queries being made (lots of temp blocks being written). This change simplifies the query so much it no longer needs temp blocks.

## Added context
The backing dataset for the Role class uses a UNION between all the
different org and space roles where the organization_id is -1 for space
roles and vice versa. This means that no role can have both a space_id
and an org_id so doing a SELECT WHERE space_id IN (space_ids) and UNION
that to SELECT WHERE org_id in (org_ids) can be replaced by just a
single select with an OR in the query clause

In our largest environments this reduced the complexity of this query
(using EXPLAIN ANALYZE) by nearly 30x

### DIFF
--- a/app/controllers/v3/roles_controller.rb
+++ b/app/controllers/v3/roles_controller.rb
@@ -159,10 +159,7 @@ class RolesController < ApplicationController
   end
 
   def readable_roles
-    roles_in_visible_spaces = Role.filter(space_id: visible_space_ids)
-    roles_in_visible_orgs = Role.filter(organization_id: visible_org_ids)
-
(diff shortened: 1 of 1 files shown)


MarcPaquette:
Ran CATS on PR. Hit the endpoint a few different ways.  Examined CCNG's logs. Everything looks good!

PR #2555, opened 2021-10-29
Fetch existing service keys count only when quota is set

philippthun:
With this change the existing service keys count is fetched only when the quota for `total_service_keys` is not -1 (i.e. unlimited). This removes an unneeded database query.


philippthun:
Included in #2564

PR #2560, opened 2021-11-03
Don't fetch feature flag twice

philippthun:
When using `FeatureFlag.raise_unless_enabled!` the feature flag was selected twice from the database (i.e. a second time as part of calling `FeatureFlag.enabled?`).

The code has been refactored to ensure that only a single select is performed.


andy-paine (approved):
I don't love the conditional for raising errors but compared to other options this does save a DB roundtrip. I tried a few other options and this is the simplest.

PR #2512, opened 2021-09-20
Implement service_instance rate limit for specific requests

MarcPaquette:
Implement a rate limiter for service-instance PUT/POST/PATCH request. This is to allow platform operators a finer degree of rate limiting for service brokers and their associated services.

* Links to any other associated PRs:
https://github.com/cloudfoundry/capi-release/pull/206

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [x] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats)

PR #2563, opened 2021-11-05
Rate limit endpoints that interact with service brokers

andy-paine:

Limit the maximum number of concurrent requests per-user to a VM that
may interact with a service broker. This can help mitigate the webserver
request queue filling up with requests waiting on slow service brokers.

Rather than sending rate limit headers as the main API rate limiter
does, instead this just returns a nicely formatted error (based on v2 or
v3 formatting) as it is concerned with concurrency rather than max
limits over time with resets etc.

Some requests will hit these endpoints (e.g. v3 user provided service
instances) that don't require access to a broker but these should be so
fast that the concurrency limit is not as much of a problem.

0 or negative configuration for the max value is taken to mean "do not
enable this feature"

#2512 unfortunately isn't as applicable now that rate limiting is done in memory so this is a reimagining of this given the current structure of the code
TODO: CAPI PR to follow

### DIFF
--- a/middleware/service_broker_rate_limiter.rb
+++ b/middleware/service_broker_rate_limiter.rb
@@ -40,9 +40,13 @@ module CloudFoundry
         unless skip_rate_limiting?(env, request)
           return too_many_requests!(env) unless @request_counter.can_make_request?(user_guid, @concurrent_limit)
 
-          status, headers, body = @app.call(env)
-          @request_counter.decrement(user_guid)
-          return [status, headers, body]
+          begin
+            return @app.call(env)
+          rescue => e
+            raise e
+          ensure
+            @request_counter.decrement(user_guid)
+          end
         end
 
         @app.call(env)
--- a/middleware/service_broker_rate_limiter.rb
+++ b/middleware/service_broker_rate_limiter.rb
@@ -1,0 +1,97 @@
+module CloudFoundry
+  module Middleware
+    class ServiceBrokerRequestCounter
+      include Singleton
+
+      def initialize
+        @mutex = Mutex.new
+        @data = {}
+      end
+
+      def can_make_request?(user_guid, limit)
+        @mutex.synchronize do
+          request_count = @data.fetch(user_guid, 0)
+          return false if request_count + 1 > limit
+
+          @data[user_guid] = request_count + 1
+          true
+        end
+      end
+
+      def decrement(user_guid)
+        @mutex.synchronize do
+          @data[user_guid] = @data[user_guid] - 1
+        end
+      end
+    end
+
+    class ServiceBrokerRateLimiter
+      def initialize(app, logger:, concurrent_limit:)
+        @app                               = app
+        @logger                            = logger
+        @concurrent_limit = concurrent_limit
+        @request_counter = ServiceBrokerRequestCounter.instance
+      end
+
+      def call(env)
+        request = ActionDispatch::Request.new(env)
+        user_guid = env['cf.user_guid']
+
+        unless skip_rate_limiting?(env, request)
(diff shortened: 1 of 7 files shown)


sweinstein22 (approved):
Hi @andy-paine , 
@moleske and I have looked through and think this all looks reasonable. Looks like this follows all our usual rate limiting conventions, including ignoring requests from admin users, and is being applied to both v2 and v3 calls.

Once we have a `capi-release` PR we should be good to merge this in. I'm assuming that's the referenced TODO in the message above? I think ideally we default `max_concurrent_service_broker_requests` to `0` so that it's an opt-in feature.

andy-paine:
Thanks! CAPI PR is here https://github.com/cloudfoundry/capi-release/pull/212

middleware/service_broker_rate_limiter.rb
```diff
@@ -0,0 +1,101 @@
...
+      def admin?
+        VCAP::CloudController::SecurityContext.admin? || VCAP::CloudController::SecurityContext.admin_read_only?
+      end
+
+      def is_service_request?(request)
+        [
+          %r{\A/v2/service_instances},
+          %r{\A/v2/service_bindings},
+          %r{\A/v2/service_keys},
+          %r{\A/v3/service_instances},
```
FloThinksPi, line 70: Lets say we have this call POST /v3/service_instances/:guid/relationships/shared_spaces , i would assume it will count the counter yet afaik the endpoint doesnt make a broker request. Do we just say its sufficient to wildcard match the whole  /v3/service_instances endpoint or do we want a more finegrained list so specify each endpoint explicitely ?

See https://v3-apidocs.cloudfoundry.org/version/3.110.0/index.html#share-a-service-instance-to-other-spaces

middleware/service_broker_rate_limiter.rb
```diff
@@ -0,0 +1,101 @@
...
+          %r{\A/v3/service_credential_bindings},
+          %r{\A/v3/service_route_bindings},
+        ].any? { |re| request.fullpath.match re }
+      end
+
+      def rate_limit_method?(request)
+        %w(PATCH PUT POST).include? request.method
```
FloThinksPi, line 77: As addition to above comment, when we would name every endpoint explicitly we might also want to add different methods for some endpoints. E.g. afaik POST /v3/service_instances/:guid/relationships/shared_spaces does just internal CC logic to share the broker with another space while DELETE /v3/service_instances/:guid/relationships/shared_spaces/:space_guid unbinds all apps so makes calls to the broker. 

See https://v3-apidocs.cloudfoundry.org/version/3.110.0/index.html#share-a-service-instance-to-other-spaces 

Dont know if this also applies to other endpoints as well, i just took this one as an example.

middleware/service_broker_rate_limiter.rb
```diff
@@ -0,0 +1,101 @@
...
+        user_guid = env['cf.user_guid']
+
+        unless skip_rate_limiting?(env, request)
+          return too_many_requests!(env) unless @request_counter.can_make_request?(user_guid, @concurrent_limit)
+
+          begin
+            return @app.call(env)
+          rescue => e
+            raise e
+          ensure
```
FloThinksPi, line 47: Leaking the counter could be really bad as its hard to notice over a lot of instances as its in memory and people could get sporadic 429s that are hard to trace. Are we sure that in any error situation the CC code will decrement that counter ? I lack the knowledge if there is a way that the middleware is skipped or the ensure is not called. Like e.g aborted connections or similar low level errors that make answering impossible.
It looks like that the Middleware is called with `@app.call(env)` so in any case the ensure should run ?

PR #2584, opened 2021-12-07
Use an optimised conversion proc for Postgres timestamps

philippthun:
When looking at profiling data one can see that the timestamp conversion is mainly taking place in `Sequel::SequelMethods::convert_input_timestamp` and that the string returned from the database is actually parsed twice. The second parsing is done to check if the given string contains an offset - which is never the case as Postgres field type "timestamp without time zone" is used. So what happens boils down to something like this:
```
t = Time.parse(value)
Date._parse(value).has_key?(:offset)
(t + t.utc_offset).utc
```
Benchmarking this locally (1 million times) shows a (reference) time of 21.6 seconds.

Removing the unnecessary second parsing reduces the time to 12.7s.

As the timestamp format is fix (i.e. no heuristics are needed), the parsing can be done with strptime instead of parse. To distinguish between timestamp strings with and without fractional seconds, it is checked whether they contain a `.`:
```
DateTime.strptime(value, value.include?('.') ? '%Y-%m-%d %H:%M:%S.%N' : '%Y-%m-%d %H:%M:%S').to_time.utc
```
The measured benchmark time for this conversion is 4.0s.

By using `DateTime._strptime` to parse the timestamp and `Time.utc()` to construct a new `Time` object, the time can be reduced to 2.8s (for 1 million conversions, i.e. a factor of ~8):
```
dt = DateTime._strptime(value, value.include?('.') ? '%Y-%m-%d %H:%M:%S.%N' : '%Y-%m-%d %H:%M:%S')
Time.utc(dt[:year], dt[:mon], dt[:mday], dt[:hour], dt[:min], dt[:sec] + (dt[:sec_fraction] || 0))
```

With this change the conversion proc is overwritten by `Sequel::Postgres::DatabaseMethods.add_conversion_proc`.

### DIFF
--- a/lib/cloud_controller/db.rb
+++ b/lib/cloud_controller/db.rb
@@ -41,6 +41,15 @@ module VCAP::CloudController
       db.default_collate = 'utf8_bin' if db.database_type == :mysql
       add_connection_expiration_extension(db, opts)
       add_connection_validator_extension(db, opts)
+
+      if db.database_type == :postgres && Sequel.application_timezone == :utc && Sequel.database_timezone == :utc && Sequel.typecast_timezone == :utc
+        # object identifier '1114' is 'timestamp without time zone' (used for e.g. created_at, updated_at)
+        db.add_conversion_proc(1114) do |value|
+          dt = DateTime._strptime(value, value.include?('.') ? '%Y-%m-%d %H:%M:%S.%N' : '%Y-%m-%d %H:%M:%S')
+          Time.utc(dt[:year], dt[:mon], dt[:mday], dt[:hour], dt[:min], dt[:sec] + (dt[:sec_fraction] || 0))
+        end
+      end
+
       db
     end
 
--- a/spec/unit/lib/cloud_controller/db_spec.rb
+++ b/spec/unit/lib/cloud_controller/db_spec.rb
@@ -1,0 +1,40 @@
+require 'spec_helper'
+
+module VCAP::CloudController
+  RSpec.describe Sequel::Database do
+    describe 'when running with a PostgreSQL database', skip: User.db.database_type != :postgres do
+      let(:conversion_proc) { User.db.conversion_procs[1114] }
+
+      it "does not use the default method for 'timestamp without time zone' conversion" do
+        expect(conversion_proc).not_to be_a(Method)
+      end
+
+      describe 'when reading data from the database' do
+        it 'correctly converts timestamp fields using the custom conversion proc' do
+          t = Time.now
+          User.make
+
+          u = User.first
+          expect(u.created_at).to be_within(10.seconds).of t
+          expect(u.updated_at).to be_within(10.seconds).of t
+        end
+      end
+
+      describe 'when using the custom conversion proc' do
+        it 'converts timestamp strings in the ISO 8601 extended format without fractional seconds' do
+          t = conversion_proc.call('2013-12-11 10:09:08')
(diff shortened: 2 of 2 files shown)


lib/cloud_controller/db.rb
```diff
@@ -41,6 +41,15 @@ def self.connect(opts, logger)
...
       add_connection_validator_extension(db, opts)
+
+      if db.database_type == :postgres && Sequel.application_timezone == :utc && Sequel.database_timezone == :utc && Sequel.typecast_timezone == :utc
+        # object identifier '1114' is 'timestamp without time zone' (used for e.g. created_at, updated_at)
+        db.add_conversion_proc(1114) do |value|
+          dt = DateTime._strptime(value, value.include?('.') ? '%Y-%m-%d %H:%M:%S.%N' : '%Y-%m-%d %H:%M:%S')
+          Time.utc(dt[:year], dt[:mon], dt[:mday], dt[:hour], dt[:min], dt[:sec] + (dt[:sec_fraction] || 0))
+        end
+      end
+
```
sethboyles, line 52: is it worth adding a test for this?

philippthun: Yes, that makes sense. I've added some tests.

Gerg:
Does MySQL have the same problem?

philippthun:
> Does MySQL have the same problem?

@Gerg - short answer: no, long answer: thanks for asking!

So I looked into the topic of timestamp conversion using Sequel and MySQL and couldn't find similar conversions as implemented in Sequel's PostgreSQL adapter. So the conversion is already done "below" in the `mysql2` gem that's using `libmysql` and is already fast. Here a comparison between PostgreSQL and MySQL when fetching a single user record from the database 10000 times:

### PostgreSQL

| user | system | total |
| -- | -- | -- |
|11.120050|18.784816|29.904866|
|11.189934|18.721485|29.911419|

### MySQL

| user | system | total |
| -- | -- | -- |
|0.782401|0.229683|1.012084|
|0.734940|0.210260|0.945200|

When running the same benchmark with the optimisation from this PR, it is slightly faster than MySQL:

### PostgreSQL with optimised conversion proc for timestamps

| user | system | total |
| -- | -- | -- |
|0.685741|0.257086|0.942827|
|0.704644|0.262820|0.967464|

So, no, MySQL does not have this performance problem.

But then I stumbled upon the [PostgreSQL-specific Support in Sequel](https://github.com/jeremyevans/sequel/blob/master/doc/postgresql.rdoc) documentation and learned about `sequel_pg`. By simply adding this gem (and removing my manual optimisation), the benchmark results look as follows:

### PostgreSQL with sequel_pg

| user | system | total |
| -- | -- | -- |
|0.596703|0.256924|0.853627|
|0.591460|0.250780|0.842240|

So I've created another PR #2595 and would suggest to merge that one and close this one.

Gerg:
@philippthun That's great! Looks like there could be some other benefits as well:

> When the postgres adapter is used with the pg driver, Sequel automatically checks for sequel_pg, and loads it if it is available. sequel_pg is a C extension that optimizes the fetching of rows, generally resulting in a ~2x speedup. It is highly recommended to install sequel_pg if you are using the postgres adapter with pg.

> sequel_pg has additional optimizations when using the Dataset map, as_hash, to_hash_groups, select_hash, select_hash_groups, select_map, and select_order_map methods, which avoids creating intermediate hashes and can add further speedups.

philippthun:
Keep this PR as DRAFT until #2595 is merged.

philippthun:
Closed in favor of #2595.

Issue #2582, opened 2021-12-07
Blobstore error messages when pushing CF Docker apps


jochenehret:
## Issue

If a CF Docker application is restaged multiple times (> max_staged_droplets_stored), the droplets will be marked as EXPIRED. Then the delete_expired_droplet_blob coding will attempt to delete the droplet. As the droplet has no "droplet_hash" in the database, the jobs fails with the error "Attempting delete of '' from blobstore 'droplet_blobstore'" which will result in multiple "Failed to perform blobstore operation after three retries." errors.

## Context

We are running CF v17.0.0.

## Steps to Reproduce

1. Push a Docker test app 7 times.

2. Check the "droplets" database table. You should see that the oldest droplet is marked as "EXPIRED":
```
  id  |                 guid                 |         created_at         |         updated_at         |               droplet_hash               |  state  |                        docker_receipt_image                        |               app_guid
------+--------------------------------------+----------------------------+----------------------------+------------------------------------------+---------+--------------------------------------------------------------------+--------------------------------------
 1479 | ef5b189a-eff7-4927-93a0-eb3b18fc866b | 2021-12-06 16:30:44.612357 | 2021-12-06 16:30:44.652223 |                                          | STAGED  | <repository>/cloudfoundry/test-app:latest | 82a95a4d-0c95-4daf-b106-73b6748f5662
 1478 | 03de52d9-d2e3-4049-8240-3ac80e4a22e4 | 2021-12-06 16:29:14.482917 | 2021-12-06 16:29:14.521709 |                                          | STAGED  | <repository>/cloudfoundry/test-app:latest | 82a95a4d-0c95-4daf-b106-73b6748f5662
 1476 | 541b7be2-2c34-47c9-aae4-37e10ee96132 | 2021-12-06 16:28:19.65487  | 2021-12-06 16:28:19.684666 |                                          | STAGED  | <repository>/cloudfoundry/test-app:latest | 82a95a4d-0c95-4daf-b106-73b6748f5662
 1475 | db5d4b1b-de42-45a2-bf27-c1fedc9c4da0 | 2021-12-06 16:25:39.21491  | 2021-12-06 16:25:39.243884 |                                          | STAGED  | <repository>/cloudfoundry/test-app:latest | 82a95a4d-0c95-4daf-b106-73b6748f5662
[...]

PR #2588, opened 2021-12-10
Fix blobstore errors for droplets from CF Docker apps

jochenehret:
PR fixes droplet deletion error messages for Docker apps. For details, see https://github.com/cloudfoundry/cloud_controller_ng/issues/2582

See https://github.com/cloudfoundry/cloud_controller_ng/issues/2582

### DIFF
--- a/lib/cloud_controller/bits_expiration.rb
+++ b/lib/cloud_controller/bits_expiration.rb
@@ -12,7 +12,8 @@ module VCAP::CloudController
     def expire_droplets!(app)
       expirable_candidates = DropletModel.
                              where(state: DropletModel::STAGED_STATE, app_guid: app.guid).
-                             exclude(guid: app.droplet_guid)
+                             exclude(guid: app.droplet_guid).
+                             exclude(droplet_hash: nil)
 
       return if expirable_candidates.count < droplets_storage_count
 
--- a/lib/cloud_controller/bits_expiration.rb
+++ b/lib/cloud_controller/bits_expiration.rb
@@ -12,8 +12,7 @@ module VCAP::CloudController
     def expire_droplets!(app)
       expirable_candidates = DropletModel.
                              where(state: DropletModel::STAGED_STATE, app_guid: app.guid).
-                             exclude(guid: app.droplet_guid).
-                             exclude(droplet_hash: nil)
+                             exclude(guid: app.droplet_guid)
 
       return if expirable_candidates.count < droplets_storage_count
 
@@ -21,7 +20,9 @@ module VCAP::CloudController
 
       droplets_to_expire.each do |droplet|
         droplet.update(state: DropletModel::EXPIRED_STATE)
-        enqueue_droplet_delete_job(droplet.guid)
+        if droplet.droplet_hash
+          enqueue_droplet_delete_job(droplet.guid)
+        end
       end
     end
 
--- a/lib/cloud_controller/bits_expiration.rb
+++ b/lib/cloud_controller/bits_expiration.rb
@@ -20,9 +20,7 @@ module VCAP::CloudController
 
       droplets_to_expire.each do |droplet|
         droplet.update(state: DropletModel::EXPIRED_STATE)
-        if droplet.droplet_hash
-          enqueue_droplet_delete_job(droplet.guid)
-        end
+        enqueue_droplet_delete_job(droplet.guid) if droplet.droplet_hash
       end
     end
 
--- a/spec/unit/jobs/runtime/expired_blob_cleanup_spec.rb
+++ b/spec/unit/jobs/runtime/expired_blob_cleanup_spec.rb
@@ -13,20 +13,13 @@ module VCAP::CloudController
 
       describe 'droplets' do
         context 'expired' do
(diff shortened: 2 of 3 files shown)


spec/unit/lib/cloud_controller/bits_expiration_spec.rb
```diff
@@ -41,19 +41,47 @@ module VCAP::CloudController
...
+          DropletModel.make(
+            app_guid:     app.guid,
+            created_at:   t + i,
+            droplet_hash: nil,
+            docker_receipt_image: 'repo/test-app',
+          )
+        end
+      end
+
+      it 'it does not enqueue a job to delete the blob' do
```
philippthun, line 65: it '~~it~~ does not'

spec/unit/lib/cloud_controller/bits_expiration_spec.rb
```diff
@@ -41,19 +41,47 @@ module VCAP::CloudController
...
+        expect { BitsExpiration.new.expire_droplets!(app) }.not_to change { Delayed::Job.count }.from(0)
```
philippthun, line 66: `not_to change` does not require the `from(0)`

spec/unit/jobs/runtime/expired_blob_cleanup_spec.rb
```diff
@@ -13,21 +13,14 @@ module Jobs::Runtime
 
       describe 'droplets' do
         context 'expired' do
-          let!(:expired_droplet) { DropletModel.make(state: DropletModel::EXPIRED_STATE) }
```
philippthun, line 16: I think this file does not need to be changed (as also `expired_blob_cleanup.rb` did not change). It checks that the `exclude(droplet_hash: nil)` is working.

philippthun (changes requested):

lib/cloud_controller/bits_expiration.rb
```diff
@@ -20,7 +20,9 @@ def expire_droplets!(app)
 
       droplets_to_expire.each do |droplet|
         droplet.update(state: DropletModel::EXPIRED_STATE)
-        enqueue_droplet_delete_job(droplet.guid)
+        if droplet.droplet_hash
```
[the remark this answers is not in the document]
jochenehret: Ok

spec/unit/lib/cloud_controller/bits_expiration_spec.rb
jochenehret: Ok

spec/unit/lib/cloud_controller/bits_expiration_spec.rb
jochenehret: Ok

spec/unit/jobs/runtime/expired_blob_cleanup_spec.rb
jochenehret: Yes, but we want to emphasise that there are two types of droplets: For buildpack apps and for Docker apps. The `expired_blob_cleanup.rb` coding is ok. It correctly filters droplet db entries with `droplet_hash: nil`. This enhancement of the test prevents regressions if someone changes the filtering.

philippthun: But there were already two tests for that:

1. [it 'enqueues a deletion job when droplet_hash is not nil'](https://github.com/cloudfoundry/cloud_controller_ng/blob/a63cfe80d92e8f153db89d10963f10f9173a2f69/spec/unit/jobs/runtime/expired_blob_cleanup_spec.rb#L19-L20)

2. [it 'does nothing when droplet_hash is nil'](https://github.com/cloudfoundry/cloud_controller_ng/blob/a63cfe80d92e8f153db89d10963f10f9173a2f69/spec/unit/jobs/runtime/expired_blob_cleanup_spec.rb#L26-L27)

philippthun (changes requested):

spec/unit/jobs/runtime/expired_blob_cleanup_spec.rb
jochenehret: Yes, but there was no "Docker"-related test object in this spec. At first, droplets were only used for CF buildpack apps. Then later the Docker feature was added. The droplet concept was reused, but with a different attribute configuration (`droplet_hash: nil` and instead a `docker_receipt_image`). We wanted to express this fact more clearly in the spec by adding one Docker droplet test object.

philippthun: Okay, now I understand your intention.

Then I suggest to put the "Docker" droplet into a `let!` and give it a descriptive name (maybe `expired_docker_droplet`). And to distinguish between these two types of droplets, I would also suggest to rename the `expired_droplet` (maybe to `expired_buildpack_droplet`?).

And as you focus more on the semantics (and not the technical details of `droplet_hash` being `nil` or not) the spec could be changed to `it 'enqueues a deletion job for the buildpack droplet only'`.

There is also another context below (`context 'failed'`) with a similar intention. This one should be changed in a similar way to be consistent.

philippthun (approved):

PR #2595, opened 2021-12-14
Add sequel_pg gem

philippthun:
From the README [1]: "sequel_pg overwrites the inner loop of the Sequel postgres adapter row fetching code with a C version. The C version is significantly faster than the pure ruby version that Sequel uses by default."

A local benchmark test (selecting a single user from the database, 10000 times) shows the following improvements:

### without sequel_pg

| user | system | total |
| -- | -- | -- |
|11.120050|18.784816|29.904866|
|11.189934|18.721485|29.911419|

### with sequel_pg

| user | system | total |
| -- | -- | -- |
|0.596703|0.256924|0.853627|
|0.591460|0.250780|0.842240|

According to the "PostgreSQL-specific Support in Sequel" documentation [2] "sequel_pg [..] optimizes the fetching of rows, generally resulting in a ~2x speedup. It is highly recommended to install sequel_pg if you are using the postgres adapter with pg."

[1] https://github.com/jeremyevans/sequel_pg
[2] https://github.com/jeremyevans/sequel/blob/master/doc/postgresql.rdoc#sequel_pg-postgrespg-only-


philippthun:
Converted to DRAFT as the compilation as part of the BOSH release fails.

andy-paine (approved):
I've deployed this to a landscape (using https://github.com/cloudfoundry/capi-release/pull/215) and run CATs on that and all of the ones we run (136) are passing. Since this affects DB querying I think any issues would have been demonstrated by those tests and given the creator of the gem suggests this change also I think this is good to go. Looking forward to seeing the performance improvements with large dataset 🤞

PR #2599, opened 2021-12-17
Try upgrading to rails 6.0

sethboyles:
Co-authored-by: Michael Oleske <moleske@pivotal.io>

- [x] Passed units
- [ ] Passed CATS
- [x] Passed BARAS

This seems like a good intermediate step before upgrading our Rails dependencies to 6.1.  Upgrading to [6.1 directly](https://github.com/cloudfoundry/cloud_controller_ng/pull/2598) results in 283 unit test failures, while this only resulted in 3.  Hopefully this is just because [Rails 6.1 removes a lot of deprecated features](https://guides.rubyonrails.org/6_1_release_notes.html), and we tackle those by turning on deprecation warnings before upgrading to 6.1

### DIFF
--- a/spec/request/space_manifests_spec.rb
+++ b/spec/request/space_manifests_spec.rb
@@ -1029,9 +1029,10 @@ RSpec.describe 'Space Manifests' do
 
         it 'returns an appropriate error' do
           headers = yml_headers(user_header)
-          headers['CONTENT_TYPE'] = 'bogus'
+          headers['CONTENT_TYPE'] = 'application/bogus'
 
           post "/v3/spaces/#{space.guid}/manifest_diff", yml_manifest, headers
+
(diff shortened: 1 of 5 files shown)


spec/request/space_manifests_spec.rb
```diff
@@ -1029,9 +1029,10 @@
 
         it 'returns an appropriate error' do
           headers = yml_headers(user_header)
-          headers['CONTENT_TYPE'] = 'bogus'
+          headers['CONTENT_TYPE'] = 'application/bogus'
```
sethboyles, line 1032: this test started returning [406](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406) when `CONTENT_TYPE` was `bogus`.   Seems reasonable to me.

PR #2603, opened 2021-12-21
Delete binding from db when binding request to service broker fails

philippthun:
When the binding request to the service broker fails (e.g. times out), orphan mitigation was already triggered (i.e. unbind request sent to the service broker) but the database entry was kept. This led to errors when calling `cf bind-service` a second time (`App .. is already bound to service instance ..`).

With this change we remove the binding from the database when the initial binding request to the service broker fails. Thus we also changed an already existing test.


FloThinksPi:
https://cloudfoundry.slack.com/archives/C07C04W4Q/p1640160640057900

philippthun:
Instead of deleting the binding and loosing the possibility to show why it failed (e.g. with `cf service <myservice>`), we are treating it as non-existent and allowing the user to simply "re-bind" it: #2636

Issue #2197, opened 2021-04-19
Help **fog-azure-rm** upgrade (or pin) **azure-storage**


sethboyles:
In response to the Azure regression introduced in https://github.com/cloudfoundry/capi-release/releases/tag/1.97.0

**The Situation**
1. CC uses [fog-azure-rm](https://github.com/fog/fog-azure-rm) for communicating with Azure blobstores
2. fog-azure-rm uses [azure-storage](https://github.com/Azure/azure-storage-ruby) under the hood
3. azure-storage has [a bug](https://github.com/Azure/azure-storage-ruby/issues/122) between `0.15.0.preview` and `1.1.0` that breaks resource matching
4. We use a [cf-specific branch](https://github.com/fog/fog-azure-rm/tree/fog-arm-cf) of fog-azure-rm
5. That branch [depends on azure-storage < 1.0](https://github.com/fog/fog-azure-rm/blob/baaf3f4d684f82561aa8695412fec98f363bdd7f/fog-azure-rm.gemspec#L35), so Bundler keeps trying to upgrade azure-storage to `0.15.0.preview` (which has the bug)
6. Master of fog-azure-rm is pinned to [an even older version of azure-storage](https://github.com/fog/fog-azure-rm/blob/master/fog-azure-rm.gemspec#L35)

**To Do**
Ideally bundler (and dependabot) shouldn't try to upgrade us to a broken version of fog-azure-rm. It would also be cool if in the future we could move to newer versions of fog-azure-rm and azure-storage for new features and security fixes.

Some possible paths forward:
1. Move to the mainline releases of fog-azure-rm, assuming it has the features we need from our special branch
2. Convince the fog-azure-rm maintainers to pin fog-storage on our special branch to a version < `0.15.0.preview`
3. Convince the fog-azure-rm maintainers to bump fog-storage on our special branch to a version >= `1.1.0`
4. Some other, even better, plan

**Deliverable:** One or more of the following:
1. A github issue against fog-azure-rm
2. A github PR against fog-azure-rm
3. A change to the version of the fog-azure-rm that CC uses

**Notes:**
1. If relevant, you can deliver this story when the issue or PR is submitted. No need to wait for the issue/PR to be resolved.
2. There is currently [a PR](https://github.com/fog/fog-azure-rm/pull/418) for bumping azure-storage in fog-azure-rm
3. Related issue: https://github.com/cloudfoundry/cloud_controller_ng/issues/1846

[...]

PR #2623, opened 2022-01-13
[WIP] Ruby 3 upgrade

sethboyles:
Working on Ruby 3 upgrade.

Biggest issue so far was changing keyword args: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/ 

There is the [https://github.com/ruby/ruby2_keywords](https://github.com/ruby/ruby2_keywords) gem, but we figured we would try to adapt to the new, stricter usage first.  If it's too risky, we can try the gem instead

What needs to be done before this is good to go:

- [x] https://github.com/cloudfoundry/cloud_controller_ng/issues/2197 
- [x] https://github.com/fog/fog-google/pull/422 is merged and released
- [x] https://github.com/cloudfoundry/cf-uaa-lib/pull/73 is merged and consumed, preferably in a release rather than pointing at master or a specific branch
- [x] Either have this issue addressed in fog-local, or account for it on our end https://github.com/fog/fog-local/issues/24 
     - [x] (update: PR here: https://github.com/fog/fog-local/pull/25 )
     - [x] merged in and bumped in this branch 
- [ ] Update resources listed here: [Updating-Ruby-Package-in-CAPI-Release](https://github.com/cloudfoundry/cloud_controller_ng/wiki/Updating-Ruby-Package-in-CAPI-Release) 
     - [x] Update capi-release with a PR
     - [x] Updating the capi-migration-compatibility dockerfile
     - [ ] Updating Capi-workspace [non-blocking]
     - [x] Updating capi-runtime-ci dockerfile
     - [x] Updating the capi-ruby-units dockerfile

If the first 3 are done, and ruby 2.7 tests are passing, I think might be able to ahead and merge, and update the package in capi-release after.  But it's probably best to have these ready to go at the same time.

We are unsure if the unit tests have adequate coverage of the different fog gems (e.g fog-aliyun, fog-google, etc).  We have run into a few issue where we were using old versions of the fog gems that were using the now removed [URI.escape](https://docs.knapsackpro.com/2020/uri-escape-is-obsolete-percent-encoding-your-query-string).  We are wondering if might run into similar errors when running CATS and BARAS against these providers.


sethboyles:
I think something about the Sinatra upgrade (2.0 -> 2.1) is causing some tests to fail:

```
 2) Sinatra::VCAP accessing a route that throws a low level exception should return json content type in the header
     Failure/Error: Unable to find /foo to read failed line

     ZeroDivisionError:
       divided by 0
     Shared Example Group: "http header content type" called from ./spec/unit/lib/sinatra/vcap_spec.rb:108
     # /foo:1
     # /bar:2
```

Our test calls this endpoint:
https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/unit/lib/sinatra/vcap_spec.rb#L27-L32

Which divides by zero, and the expectation is that Sinatra returns a 500.  Looking at the Sinatra changelog, I'm not sure why this behavior would have changed...

sethboyles:
Reason for divide by 0 issues: https://github.com/sinatra/sinatra/pull/1214.  

[spec/unit/lib/sinatra/vcap_spec.rb#19 ](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/unit/lib/sinatra/vcap_spec.rb#L19-L21)

doesn't override helpers anymore

sethboyles:
Due to another fog issue, we ended up pinning fog-google to `1.7.1`.  I am not sure that version is compatible with Ruby 3, so we may need to help https://github.com/fog/fog-google/pull/422 move forward so we can move to the newest version of fog-google.

sethboyles:
This has suddenly been made more pressing, as Ruby 3 (3.1?) is required for the new Jammy Jellyfish stemcells: https://bosh.io/docs/jammy-migration/#openssl-3

tjvman:
Closing, superseded by #2703

PR #2636, opened 2022-01-20
Treat failed service-credential-binding and service-credential-key as non-existent / Do not cleanup failed bindings in apply_manifest

philippthun:
### Treat failed binding as non-existent 

When the last operation for a service binding is `create failed`:
- accept request to `POST /v3/service_credential_bindings` instead of rejecting it with "The app is already bound to the service instance"; this allows users to retry the creation without the need to explicitly unbind the service
- create binding in `POST /v3/spaces/:guid/actions/apply_manifest` instead of skipping it; the old behavior resulted in apps using a failed binding

### Treat failed key creation as non-existent 

When the last operation for a service key is `create failed`:
- accept request to `POST /v3/service_credential_bindings` instead of rejecting it with "The service instance already has a key binding with name 'mykey'"; this allows users to retry the creation without the need to explicitly delete the key

When the last operation for a service key is `create in progress` or `create failed`:
- raise an error when `GET /v3/service_credential_bindings/:guid/details` is called instead of showing empty key data

### Do not delete bindings in case of bind errors during apply_manifest

The current implementation has the following drawbacks:
- If the automatically triggered deletion succeeded, the original binding error was not shown anymore (e.g. in `cf service myservice`).
- If also the deletion failed, this status was persisted.
  - A subsequent `cf push` treated this as an existing binding and continued.
  - The app's environment variables contained empty credentials.

The changed implementation has the following advantages:
- `VCAP::Services::ServiceClientProvider` already performs the orphan mitigation.
- There is no need anymore to delete a failed binding as bindings in state `create failed` are treated as non-existent and `cf push` can be re-run without any manual cleanup being necessary.
- Users can still see the state of the 'last operation' (e.g. in `cf service myservice`).
- For a binding in state `create failed` no data is added to the app's environment variables.
- The behavior is consistent with `cf bind-service`.

### DIFF
--- a/app/actions/app_apply_manifest.rb
+++ b/app/actions/app_apply_manifest.rb
@@ -201,7 +201,8 @@ module VCAP::CloudController
     end
 
     def binding_exists?(service_instance, app)
-      ServiceBinding.where(service_instance: service_instance, app: app).present?
+      binding = ServiceBinding.first(service_instance: service_instance, app: app)
+      binding && !binding.create_failed?
     end
 
     def binding_being_deleted!(service_instance, app)
--- a/app/actions/app_apply_manifest.rb
+++ b/app/actions/app_apply_manifest.rb
@@ -1,4 +1,3 @@
-require 'actions/mixins/bindings_delete'
 require 'actions/process_create'
 require 'actions/process_scale'
 require 'actions/process_update'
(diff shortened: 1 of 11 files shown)


MerricdeLauney:
Looks good to me. Thanks for implementing this!

cc: @sweinstein22

PR #2638, opened 2022-01-21
WIP: Reduce db calls in /v3/domains

kathap:
* instead of selecting all org guids for admin user, visible_org_guids is set to :all
* if visible_org_guids is :all (=admin user), no filtering is done, just return all org guids

Signed-off-by: Marc Misoch <marc.misoch@sap.com>
Signed-off-by: Philipp Thun <philipp.thun@sap.com>
Signed-off-by: Marc Misoch <marc.misoch@sap.com>


app/controllers/v3/domains_controller.rb
```diff
@@ -220,4 +220,12 @@ def fetch_router_group(router_group_guid)
   rescue UaaUnavailable, RoutingApi::UaaUnavailable
     service_unavailable!('Communicating with the Routing API failed because UAA is currently unavailable. Please try again later.')
   end
+
+  def visible_org_guids
+    if permission_queryer.can_read_globally?
+      :all
+    else
+      permission_queryer.readable_org_guids
+    end
```
sethboyles, line 229: to have a method return inconsistent types I think can be a bit risky.  Not sure how else to inject this type of information down into the Presenter, though...  perhaps as a separate presenter arg.  Maybe something like

```
def presenter_args
    if permission_queryer.can_read_globally?
      {all_orgs_visible: true} #visible_org_guids ignored in this case
    else
      {all_orgs_visible: false, visible_org_guids: permission_queryer.readable_org_guids} 
    end
end
```

kathap: Thank you, we adapted your suggestion

PR #2651, opened 2022-01-28
Update keyword arguments for Ruby 3.0

sethboyles:
Pulled this out of #2623 to avoid having long living branches and merge conflicts with them. 

See here for why this needs to be changed: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

### DIFF
--- a/lib/cloud_controller/blobstore/fog/error_handling_client.rb
+++ b/lib/cloud_controller/blobstore/fog/error_handling_client.rb
@@ -43,7 +43,7 @@ module CloudController
         error_handling { wrapped_client.cp_r_to_blobstore(*args)    }
       end
 
-      def download_from_blobstore(*args)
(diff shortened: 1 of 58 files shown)


lib/cloud_controller/blobstore/fog/error_handling_client.rb
```diff
@@ -43,7 +43,7 @@ def cp_r_to_blobstore(*args)
         error_handling { wrapped_client.cp_r_to_blobstore(*args)    }
       end
 
-      def download_from_blobstore(*args)
+      def download_from_blobstore(*args, **kwargs)
```
philippthun, line 46: I don't get where these `kwargs` are coming from...

philippthun (approved):

Issue #2637, opened 2022-01-20
Whats the correct way v3 broker endpoints handle operation state ?


philippthun:
# Legend

| Icon | Meaning |
| :-: | :-: |
| :white_check_mark: | fixed/wanted behaviour |
| :x: | not fixed but change of behaviour needed |
| :heavy_minus_sign: | no change of behaviour needed |
| :question: | questionable behavior, needs further analysis |

# V3 API

The introduction of the Failed State for Broker operations does come with some unwanted behaviour in some cases.

#### POST /v3/service_credential_bindings (i.e. cf bind-service) for already existing binding

| Resource State | Observed Behaviour | Wanted Behaviour | PR  | Fixed in main |
| :-: | :-: | :-: | :-: | :-: |
| none (synchronous case) | :question: | | | :heavy_minus_sign: |
| create succeeded | rejects request (**A**) | :white_check_mark: | | :heavy_minus_sign: |
| create in progress | rejects request (**A**) | :white_check_mark: | | :heavy_minus_sign: |
| create failed | rejects request (**A**) | recreates binding | [#2636](https://github.com/cloudfoundry/cloud_controller_ng/issues/2636) | :white_check_mark: |
| delete failed | rejects request (**A**) | raises error (**1**) | [#2657](https://github.com/cloudfoundry/cloud_controller_ng/pull/2657) | :white_check_mark: |
| delete in progress | rejects request (**A**) | raises error (**1**) | [#2657](https://github.com/cloudfoundry/cloud_controller_ng/pull/2657) | :white_check_mark: |

- **A**: `UnprocessableEntity` / `The app is already bound to the service instance`. This is handled as success by cf cli.
- **1**: Even if deletion did not succeed, orphan mitigation should ensure that it eventually will. Thus an error different to `The app is already bound to the service instance` should be raised. This is because cf cli handles this error as a success.

#### DELETE /v3/service_credential_bindings/:guid (i.e. cf unbind-service) for already existing binding

| Resource State | Observed Behaviour | Wanted Behaviour | PR  | Fixed in main |
| :-: | :-: | :-: | :-: | :-: |
| none (synchronous case) | :question: | | | :heavy_minus_sign: |
| create succeeded | accepts request | :white_check_mark: | | :heavy_minus_sign: |
| create in progress | accepts request (**1**) | :white_check_mark: | | :heavy_minus_sign: |
| create failed | accepts request | :white_check_mark: | | :heavy_minus_sign: |
[...]

PR #2657, opened 2022-02-02
Throw an error on sb bind when in binding status is delete_failed or delete_in_progess

FloThinksPi:
This is a fix as part of #2637. It contains changes in behaviour for:
- POST /v3/service_credential_bindings
For both of its workflows, creating a service-credential-binding and a service-credential-key.

When binding is in state delete_in_progess or delete failed we want an
error different from "binding already exists" to be thrown. Fistly this
might not be true since a orphane mitigation might in the meantime
deleted the binding. Secondly the CF CLI interprets this(rightfully) as
success since this message tells it that what CF CLI wanted to to is
already accomplished. So CF CLI will now fail on
create-credential-binding when a binding exists that is in a failed
deletion or pending deletion state.
Same applies for create-credential-key endpoint.

### DIFF
--- a/app/actions/service_credential_binding_app_create.rb
+++ b/app/actions/service_credential_binding_app_create.rb
@@ -21,7 +21,9 @@ module VCAP::CloudController
       PERMITTED_BINDING_ATTRIBUTES = [:credentials, :syslog_drain_url, :volume_mounts].freeze
 
       def precursor(service_instance, app: nil, volume_mount_services_enabled: false, message:)
-        validate!(service_instance, app, volume_mount_services_enabled)
+        validate_service_instance!(app, service_instance, volume_mount_services_enabled)
+        binding = ServiceBinding.first(service_instance: service_instance, app: app)
+        validate_binding!(binding)
 
         binding_details = {
           service_instance: service_instance,
@@ -30,9 +32,6 @@ module VCAP::CloudController
           type: 'app',
           credentials: {}
         }
-
-        binding = ServiceBinding.first(service_instance: service_instance, app: app)
-        already_bound! if binding && !binding.create_failed?
 
         (binding || ServiceBinding.new).tap do |b|
           ServiceBinding.db.transaction do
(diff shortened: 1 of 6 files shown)


app/actions/service_credential_binding_key_create.rb
```diff
@@ -18,17 +18,15 @@ def initialize(user_audit_info, audit_hash)
       PERMITTED_BINDING_ATTRIBUTES = [:credentials].freeze
 
       def precursor(service_instance, message:)
-        validate!(service_instance)
+        key = ServiceKey.first(service_instance: service_instance, name: message.name)
+        validate!(service_instance,key,message)
```
philippthun, line 22: Minor thing: passing `message.name` into the `validate!` function is sufficient.

app/actions/service_credential_binding_app_create.rb
```diff
@@ -50,16 +48,17 @@ def precursor(service_instance, app: nil, volume_mount_services_enabled: false,
...
         app_is_required! unless app.present?
         space_mismatch! unless all_space_guids(service_instance).include? app.space.guid
-
         if service_instance.managed_instance?
           service_not_bindable! unless service_instance.service_plan.bindable?
           service_not_available! unless service_instance.service_plan.active?
           volume_mount_not_enabled! if service_instance.volume_service? && !volume_mount_services_enabled
           operation_in_progress! if service_instance.operation_in_progress?
         end
+        already_bound! if binding && (binding.create_succeeded? || binding.create_in_progress?)
```
philippthun, line 60: Maybe this would be slightly more readable:
```
if binding
  already_bound! if binding.create_succeeded? || binding.create_in_progress?
  ...
end
```

app/models/services/service_binding.rb
```diff
@@ -12,7 +12,7 @@ class ServiceBinding < Sequel::Model
 
     one_to_many :labels, class: 'VCAP::CloudController::ServiceBindingLabelModel', key: :resource_guid, primary_key: :guid
     one_to_many :annotations, class: 'VCAP::CloudController::ServiceBindingAnnotationModel', key: :resource_guid, primary_key: :guid
-    add_association_dependencies labels: :destroy
+    add_association_dependencies labels: :destroy
```
philippthun, line 15: Uh, a whitespace.

FloThinksPi:
Had to split the validate functions as rubocop complains about logic complexity.

app/models/runtime/helpers/service_credential_binding_mixin.rb
```diff
@@ -8,8 +8,8 @@ def operation_in_progress?
       !!last_operation && last_operation.state == 'in progress'
     end
 
-    def create_failed?
-      return true if last_operation&.type == 'create' && last_operation.state == 'failed'
+    def create_succeeded?
```
philippthun, line 11: You can add some tests for the new methods (create_succeeded, delete_failed, delete_in_progress) here: spec/unit/models/services/service_credential_binding_shared.rb

app/actions/service_credential_binding_key_create.rb
```diff
@@ -18,17 +18,16 @@ def initialize(user_audit_info, audit_hash)
       PERMITTED_BINDING_ATTRIBUTES = [:credentials].freeze
 
       def precursor(service_instance, message:)
-        validate!(service_instance)
+        key = ServiceKey.first(service_instance: service_instance, name: message.name)
+        validate_service_instance!(service_instance)
```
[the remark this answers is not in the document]
FloThinksPi: done

philippthun (approved):

PR #2665, opened 2022-02-07
Don't reuse GUIDs when recreating bindings

### LINKED ISSUE #2666: Is reusing the GUID for binding / key recreations correct?
While working on improving how failed bindings are handled (see #2637), we have changed the behavior for bindings / keys in state `create failed`. They are now treated as non-existent, so that subsequent creations (e.g. while pushing an app) will eventually succeed without manual interactions by the user (see #2636).

We decided to update the existing binding / key in the database which results in service broker requests with a previously used GUID, i.e. the same that was used during the failed creation.

In one of our foundations a service broker responded to such a subsequent creation with the status code `409 Conflict`. Although in this particular situation the credential binding has not been created and the service broker should not have kept the GUID in its database, the question was raised whether reusing a GUID is correct or not.

The latest version of the Open Service Broker API spec states (for service instance IDs) that: "Platforms SHOULD NOT reuse IDs, since Service Brokers MAY NOT support this."

A reason why reusing a GUID might not be correct is that orphan mitigation runs asynchronously, i.e. the Cloud Controller schedules a dedicated job that is retried under certain circumstances. That's why another creation with the same GUID could cause conflicts at the service broker side.

philippthun:
When a service credential binding or key is in state `create failed`, subsequent POST requests treat them as non-existent and issue another create request to the service broker. But instead of reusing the previously created GUID, delete the existing binding / key (and its GUID) and create a new one.

Closes #2666.

### DIFF
--- a/app/actions/service_credential_binding_app_create.rb
+++ b/app/actions/service_credential_binding_app_create.rb
@@ -33,13 +33,13 @@ module VCAP::CloudController
           credentials: {}
         }
 
-        (binding || ServiceBinding.new).tap do |b|
+        ServiceBinding.new.tap do |b|
           ServiceBinding.db.transaction do
+            binding.destroy if binding
             b.save_with_attributes_and_new_operation(
               binding_details,
               CREATE_IN_PROGRESS_OPERATION
             )
-
             MetadataUpdate.update(b, message)
           end
         end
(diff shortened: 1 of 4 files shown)


sethboyles (approved):
Makes sense to me--this is probably safer thank reusing the GUID.

FloThinksPi (approved):
Also the new OSBAPI Spec secifically says not to do this. See [Spec](https://github.com/openservicebrokerapi/servicebroker/blob/3b65079b91a27bc8f3af8eecb4e6833dda870176/spec.md#changes-since-v215)

tjvman:
@philippthun @FloThinksPi Please address the merge conflicts and we'll pull this one in.

PR #2667, opened 2022-02-08
Implement exponential backoff for SB operations (CC v3 API)

svkrieger:

Introduce a new parameter `broker_client_async_poll_exponential_backoff_rate` which lets users configure exponential backoff for service related polling jobs. This includes the creation, update and deletion of service instances and the creation and deletion of service credential bindings. 

The CC already offers two parameters with which the polling interval `broker_client_default_async_poll_interval_seconds` and the polling duration `broker_client_max_async_poll_duration_minutes` can be configured. Though, this is not enough to prevent that the CC can be overloaded by a single user e.g. by creating many service instances, while the broker needs a lot of time to handle the request. This will result in many polling jobs, which overload the worker processes.

The newly introduced parameter `broker_client_async_poll_exponential_backoff_rate`, by default, does not change the current behaviour as it is set to 1.0, which means there is no exponential backoff. When set to a higher number, the time until each re-enqueued job will be run will increase and therefore lead to much less polling jobs, which need to be run by the worker processes. 

### DIFF
--- a/spec/unit/jobs/reoccurring_job_spec.rb
+++ b/spec/unit/jobs/reoccurring_job_spec.rb
@@ -75,10 +75,13 @@ module VCAP
         job.polling_interval_seconds = 95
         expect(job.polling_interval_seconds).to eq(95)
 
-        enqueued_time = Time.now
+        enqueued_time = 0
 
-        Jobs::Enqueuer.new(job, queue: Jobs::Queues.generic).enqueue_pollable
-        execute_all_jobs(expected_successes: 1, expected_failures: 0)
+        Timecop.freeze do
+          Jobs::Enqueuer.new(job, queue: Jobs::Queues.generic).enqueue_pollable
+          execute_all_jobs(expected_successes: 1, expected_failures: 0)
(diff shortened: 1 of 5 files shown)


spec/unit/jobs/reoccurring_job_spec.rb
```diff
@@ -98,6 +98,148 @@ def perform
...
+            Timecop.freeze(12.seconds.after(enqueued_time)) do
+              execute_all_jobs(expected_successes: 0, expected_failures: 0)
+            end
+
+            Timecop.freeze(14.seconds.after(enqueued_time)) do
+              execute_all_jobs(expected_successes: 1, expected_failures: 0)
+            end
+          end
+          it 'when changing exponential backoff rate and retry_after from the job' do
+            job = FakeJob.new(retry_after: ['20', '30'])
```
philippthun, line 158: I wonder why the `retry_after` values are strings. Okay, I see that this has been done elsewhere in the code, but then I would rather change it everywhere to numbers.

PR #2680, opened 2022-02-15
Implement exponential backoff for service fetch operations (CC API v2)

svkrieger:

This PR introduces a exponential backoff for service fetch operations in CC API v2, just like in https://github.com/cloudfoundry/cloud_controller_ng/pull/2667

This PR depends on the above one.

### DIFF
--- a/spec/unit/jobs/services/service_instance_state_fetch_spec.rb
+++ b/spec/unit/jobs/services/service_instance_state_fetch_spec.rb
@@ -387,6 +387,64 @@ module VCAP::CloudController
                 end
               end
             end
+
+            context 'when exponential backoff is not set to default' do
+              it 'calculates the polling intervals based on the default interval and the exponential backoff rate' do
+                TestConfig.config[:broker_client_async_poll_exponential_backoff_rate] = 2.0
+                enqueued_time = 0
+
+                Timecop.freeze do
+                  run_job(job)
+                  enqueued_time = Time.now
+                end
+
+                [60, 180, 420, 900, 1860].each do |seconds|
+                  Timecop.freeze((seconds - 1).seconds.after(enqueued_time)) do
+                    execute_all_jobs(expected_successes: 0, expected_failures: 0)
+                  end
+
+                  Timecop.freeze((seconds + 1).seconds.after(enqueued_time)) do
+                    execute_all_jobs(expected_successes: 1, expected_failures: 0)
+                  end
+                end
+              end
+
+              it 'calculates the polling intervals based on the configured interval and the exponential backoff rate' do
+                TestConfig.config[:broker_client_async_poll_exponential_backoff_rate] = 2.0
+                TestConfig.config[:broker_client_default_async_poll_interval_seconds] = 10
+
+                enqueued_time = 0
+
+                Timecop.freeze do
+                  run_job(job)
+                  enqueued_time = Time.now
+                end
+
+                [10, 30, 70, 150, 310].each do |seconds|
+                  Timecop.freeze((seconds - 1).seconds.after(enqueued_time)) do
+                    execute_all_jobs(expected_successes: 0, expected_failures: 0)
+                  end
+
+                  Timecop.freeze((seconds + 1).seconds.after(enqueued_time)) do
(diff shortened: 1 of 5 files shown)


svkrieger:
Regarding the tests we would be glad about feedback, as they are quite different from the v3 tests. The tests can be replicated to the service binding test file after we found some good ones.

spec/unit/jobs/services/service_instance_state_fetch_spec.rb
```diff
@@ -387,6 +387,66 @@ def run_job(job)
                 end
               end
             end
+            context 'when exponential backoff is not set to default' do
+              before do
+                TestConfig.config[:broker_client_async_poll_exponential_backoff_rate] = 2.0
+              end
+
+              describe 'it updates the polling interval' do
```
philippthun, line 395: `context` (and `describe`) should be used to indicate a test scenario (e.g. `context 'when some value is configured'`), `it` describes the expectation (e.g. `it 'returns another value'`). I know that there are files where this has not been done consistently, but this spec adheres to this convention. Thus my proposal is to structure the new tests as follows:
```
context 'when exponential backoff is not set to default' do
  it 'calculates the polling intervals based on the default interval and the exponential backoff rate' do
    # set broker_client_async_poll_exponential_backoff_rate
    # ...
  end
 it 'calculates the polling intervals based on the configured interval and the exponential backoff rate' do
    # set broker_client_async_poll_exponential_backoff_rate + broker_client_default_async_poll_interval_seconds
    # ...
  end
 it 'takes the exponential backoff into account when checking whether the next run would exceed the maximum duration' do
    # ...
  end
end
```

spec/unit/jobs/services/service_instance_state_fetch_spec.rb
```diff
@@ -387,6 +387,66 @@ def run_job(job)
...
+                    run_job(job)
+                    enqueued_time = Time.now
+                  end
+
+                  [10, 30, 70, 150, 310].each do |seconds|
+                    Timecop.freeze((seconds - 1).seconds.after(enqueued_time)) do
+                      execute_all_jobs(expected_successes: 0, expected_failures: 0)
+                    end
+
+                    Timecop.freeze((seconds.ceil + 1).seconds.after(enqueued_time)) do
```
philippthun, line 430: As this test uses whole numbers only, the `.ceil` is not required.

spec/unit/jobs/services/service_instance_state_fetch_spec.rb
```diff
@@ -387,6 +387,66 @@ def run_job(job)
...
+                      execute_all_jobs(expected_successes: 1, expected_failures: 0)
+                    end
+                  end
+                end
+              end
+
+              it 'it takes the exponential backoff into account when checking whether the next run would exceed the maximum duration' do
```
philippthun, line 437: The `it` string can be omitted, i.e. `it 'takes ...'` instead of `it 'it takes ...'`

spec/unit/jobs/services/service_instance_state_fetch_spec.rb
```diff
@@ -387,6 +387,66 @@ def run_job(job)
...
+                TestConfig.config[:broker_client_async_poll_exponential_backoff_rate] = 1.3
```
philippthun, line 438: This test does not check if the calculation of the last polling interval is correct. When changing the new implementation `if Time.now + next_execution_in > end_timestamp` back to the old one (`if Time.now + poll_interval > end_timestamp`), the test still succeeds.

It could be changed to the following:
```
TestConfig.config[:broker_client_async_poll_exponential_backoff_rate] = 1.3
TestConfig.config[:broker_client_max_async_poll_duration_minutes] = 60

job.retry_number = 10
Timecop.freeze(Time.now + 3384.321.ceil.seconds) do
  run_job(job)

  service_instance.reload
  expect(service_instance.last_operation.state).to eq('failed')
  expect(service_instance.last_operation.description).to eq('Service Broker failed to provision within the required time.')
end
```

This is inspired by two other tests below, i.e. "when the job has fetched for more than the max poll duration" and "when enqueuing the job would exceed the max poll duration by the time it runs".

philippthun (approved):

PR #2679, opened 2022-02-14
Dont show service credential keys not in state create succeeded

FloThinksPi:
Basically we want not to show any service key whos state is not in
create succeeded. This is due to all other states cannot be certain that
the key still exists/already exists and orphane mitigation hasnt cleaned
it up even after a initial failed delete.
For more information see https://github.com/cloudfoundry/cloud_controller_ng/issues/2637

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats)

PR #2676, opened 2022-02-11
Only return service bindings in env on creation success

FloThinksPi:
* A short explanation of the proposed change:

We want to only include a service binding in app envs when the binding
was successfull. In any other case it is not guranteed that the binding
really exists on the broker side. E.g without this change a binding in
state failed delete would still show up. However the broker might
already deleted the binding. A full matrix of all cases can be found at

* Links to any other associated PRs

- https://github.com/cloudfoundry/cloud_controller_ng/issues/2637

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats)

PR #2682, opened 2022-02-15
Fix synchronous bind is successfull

FloThinksPi:
In case of a synchronous bind, last_operation is nil but we should consider this a succesfull binding.

### DIFF
--- a/app/actions/service_credential_binding_app_create.rb
+++ b/app/actions/service_credential_binding_app_create.rb
@@ -62,7 +62,7 @@ module VCAP::CloudController
 
       def validate_binding!(binding)
         if binding
-          already_bound! if binding.create_succeeded? || binding.create_in_progress? || binding.last_operation.nil?
+          already_bound! if binding.create_succeeded? || binding.create_in_progress?
           incomplete_deletion! if binding.delete_failed? || binding.delete_in_progress?
         end
       end
--- a/app/actions/service_credential_binding_key_create.rb
+++ b/app/actions/service_credential_binding_key_create.rb
@@ -59,7 +59,7 @@ module VCAP::CloudController
 
       def validate_key!(key, message_name)
         if key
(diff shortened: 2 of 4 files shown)


tjvman:
Please fix the Rubocop errors (looks like it's just an extra blank line); we'll take a look in the meantime.

FloThinksPi:
Hi @tjvman, 
i broke the PR a bit yesterday but now its in reviewable state :).

Its a required change for both:
https://github.com/cloudfoundry/cloud_controller_ng/pull/2676
https://github.com/cloudfoundry/cloud_controller_ng/pull/2679

So i pulled it out into a separate PR

philippthun (commented):
This change makes sense. It also means that the nil checks done [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/a22c2926856f9a44e9f15405f2f9a9b207d8f70b/app/actions/service_credential_binding_app_create.rb#L65) and [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/a22c2926856f9a44e9f15405f2f9a9b207d8f70b/app/actions/service_credential_binding_key_create.rb#L62) are now superfluous; so I suggest that your remove them as part of this PR.

There is also a typo in the commit message. It should be "Treat ...".

FloThinksPi:
> This change makes sense. It also means that the nil checks done [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/a22c2926856f9a44e9f15405f2f9a9b207d8f70b/app/actions/service_credential_binding_app_create.rb#L65) and [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/a22c2926856f9a44e9f15405f2f9a9b207d8f70b/app/actions/service_credential_binding_key_create.rb#L62) are now superfluous; so I suggest that your remove them as part of this PR.
> 
> There is also a typo in the commit message. It should be "Treat ...".

Have adopted that 👍🏼

PR #2687, opened 2022-02-17
Add index for attempts in delayed_jobs table

kathap:
* With adding an index to attempts, the query time can be improved
significantly.
* This will improve SELECT "queue", count(*) AS "count" FROM
  "delayed_jobs" WHERE ("attempts" = 0) GROUP BY "queue";

Signed-off-by: Philipp Thun <philipp.thun@sap.com>

### DIFF
--- a/db/migrations/20220217141700_add_attempts_index_to_delayed_jobs.rb
+++ b/db/migrations/20220217141700_add_attempts_index_to_delayed_jobs.rb
@@ -1,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table :delayed_jobs do
+      add_index :attempts, name: :delayed_jobs_attempts_index
+    end
+  end
+
+  down do
+    alter_table :delayed_jobs do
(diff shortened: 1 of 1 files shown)


philippthun:
We had a look at which queries are executed on the `delayed_jobs` table and how often they are run.

For the most important query (`SELECT * FROM "delayed_jobs" WHERE ... FOR UPDATE`) there is already an index over multiple columns. This query is used to select the next job to process and is executed on a large foundation ~10 million times per day.

There are two additional queries that are executed ~500 thousand times per day that don't use an index. Both queries look like this:
```
SELECT "queue", count(*) AS "count" FROM "delayed_jobs" WHERE (<some condition>) GROUP BY "queue";
```

The condition is either `"attempts" = 0` or `failed_at IS NOT NULL`. On the first one, an index had a significant impact; it reduced the execution time from 7.3ms to 0.4ms. For the second condition, adding an index did not have an effect due to the fact that the majority of the table was returned.

philippthun (approved):

PR #2688, opened 2022-02-17
Raise errors in apply_manifest in certain binding states 

FloThinksPi:
While apply_manifest cannot do asynchronous broker operation(yet?)
it certainly can get the job to care about a binding which
last_operation state can have all states.
This change makes the apply_manifest endpoint correctly handle such
bindings and reply with proper error messages instead of skipping that
binding and erroneously starting an app with an empty binding.
For more details see https://github.com/cloudfoundry/cloud_controller_ng/issues/2637

### DIFF
--- a/app/actions/app_apply_manifest.rb
+++ b/app/actions/app_apply_manifest.rb
@@ -140,8 +140,11 @@ module VCAP::CloudController
       manifest_service_bindings_message.manifest_service_bindings.each do |manifest_service_binding|
         service_instance = app.space.find_visible_service_instance_by_name(manifest_service_binding.name)
         service_instance_not_found!(manifest_service_binding.name) unless service_instance
-        binding_being_deleted!(service_instance, app)
-        next if binding_exists?(service_instance, app)
+        binding = ServiceBinding.first(service_instance: service_instance, app: app)
+        next if binding&.create_succeeded?
+
+        raise_binding_operation_in_progress!(service_instance, binding.last_operation.type) if binding&.operation_in_progress?
+        raise_binding_delete_failed!(service_instance) if binding&.delete_failed?
 
         begin
           binding_message = create_binding_message(service_instance.guid, app.guid, manifest_service_binding)
@@ -193,16 +196,14 @@ module VCAP::CloudController
       raise ServiceBindingError.new(error_message)
     end
 
-    def binding_exists?(service_instance, app)
-      binding = ServiceBinding.first(service_instance: service_instance, app: app)
-      binding && !binding.create_failed?
+    def raise_binding_operation_in_progress!(service_instance, operation)
+      error_message = "A binding is being #{operation}d. Retry this operation later."
+      raise_binding_error!(service_instance, error_message)
     end
 
-    def binding_being_deleted!(service_instance, app)
-      binding = ServiceBinding.first(service_instance: service_instance, app: app)
-      if binding && binding.operation_in_progress? && binding.last_operation.type == 'delete'
-        raise_binding_error!(service_instance, 'An existing binding is being deleted. Try recreating the binding later.')
-      end
+    def raise_binding_delete_failed!(service_instance)
+      error_message = 'A binding failed to be deleted. Resolve the issue with this binding before retrying this operation.'
+      raise_binding_error!(service_instance, error_message)
     end
 
     def app_instance_not_found!(app_guid)
--- a/spec/unit/actions/app_apply_manifest_spec.rb
+++ b/spec/unit/actions/app_apply_manifest_spec.rb
@@ -888,19 +888,48 @@ module VCAP::CloudController
                   }.to raise_error(AppApplyManifest::ServiceBindingError, /For service 'si-name': fake binding error/)
                 end
               end
-            end
 
-            context 'when a delete is in progress for the same binding' do
-              let!(:binding) do
-                binding = ServiceBinding.make(service_instance: service_instance, app: app)
-                binding.save_with_attributes_and_new_operation({}, { type: 'delete', state: 'in progress' })
-                binding
+              context 'when a create is in progress for the same binding' do
+                let!(:binding) do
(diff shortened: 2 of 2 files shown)


app/actions/app_apply_manifest.rb
```diff
@@ -188,21 +193,15 @@ def create_binding_message(service_instance_guid, app_guid, manifest_service_bin
...
-
-    def binding_exists?(service_instance, app)
-      binding = ServiceBinding.first(service_instance: service_instance, app: app)
-      binding && !binding.create_failed?
-    end
-
-    def binding_being_deleted!(service_instance, app)
-      binding = ServiceBinding.first(service_instance: service_instance, app: app)
-      if binding && binding.operation_in_progress? && binding.last_operation.type == 'delete'
-        raise_binding_error!(service_instance, 'An existing binding is being deleted. Try recreating the binding later.')
```
FloThinksPi, line 204: I changed the error message to fit all other error messages better and to not need a huge switch case with custom error messages. However this means the message changes, dont know but maybe this should not be changed as its part of the API and people(cf cli) check this string explicitly ? Opinions ?

FloThinksPi: Also on "create failed" bindings we recreate that binding. Wouldnt it make sense to actually also do this for the "delete failed" case since orphan mitigation anyway tries to delete that binding and one can assume (since the apply_manifest wants it) a new binding is requested ? So we could retrigger a binding on failed operation and throw an error just on "in progress" ones as one has to wait a bit for them to bind completely.

app/actions/app_apply_manifest.rb
```diff
@@ -140,30 +140,35 @@ def create_service_bindings(manifest_service_bindings_message, app)
...
-          binding_message = create_binding_message(service_instance.guid, app.guid, manifest_service_binding)
-          action = V3::ServiceCredentialBindingAppCreate.new(@user_audit_info, binding_message.audit_hash, manifest_triggered: true)
-          binding = action.precursor(
-            service_instance,
-            app: app,
-            volume_mount_services_enabled: volume_services_enabled?,
-            message: binding_message)
+
+        binding = ServiceBinding.first(service_instance: service_instance, app: app)
+        unless binding&.create_succeeded?
```
philippthun, line 145: I think that this:
```
next if some_condition

do_stuff
```
... is more readable compared to that:
```
unless some_condition
  do_stuff
end
```

So this could be changed to `next if binding&.create_succeeded?`.

app/actions/app_apply_manifest.rb
```diff
@@ -140,30 +140,35 @@ def create_service_bindings(manifest_service_bindings_message, app)
...
+
+          raise_invalid_binding_error!(service_instance, binding) unless !binding || binding.create_failed?
```
philippthun, line 147: I find this condition "unless not .. or" hard to understand. By changing this to two dedicated methods and conditions, it becomes more readable:
```
binding_operation_in_progress!(service_instance, binding.last_operation.type) if binding&.operation_in_progress?
binding_deletion_failed!(service_instance) if binding&.delete_failed?
```

app/actions/app_apply_manifest.rb
```diff
@@ -140,30 +140,35 @@ def create_service_bindings(manifest_service_bindings_message, app)
...
+
+              raise ServiceBrokerRespondedAsyncWhenNotAllowed.new('The service broker responded asynchronously, but async bindings are not supported.')
             end
-          rescue ServiceBrokerRespondedAsyncWhenNotAllowed,
-                 V3::ServiceBindingCreate::BindingNotRetrievable
-
-            raise ServiceBrokerRespondedAsyncWhenNotAllowed.new('The service broker responded asynchronously, but async bindings are not supported.')
+          rescue => e
+            error_message = "For service '#{service_instance.name}': #{e.message}"
+            raise ServiceBindingError.new(error_message)
```
philippthun, line 170: I would keep the `raise_binding_error!(service_instance, e.message)` method.

app/actions/app_apply_manifest.rb
```diff
@@ -188,21 +193,15 @@ def create_binding_message(service_instance_guid, app_guid, manifest_service_bin
...
-    def binding_exists?(service_instance, app)
-      binding = ServiceBinding.first(service_instance: service_instance, app: app)
-      binding && !binding.create_failed?
-    end
-
-    def binding_being_deleted!(service_instance, app)
-      binding = ServiceBinding.first(service_instance: service_instance, app: app)
-      if binding && binding.operation_in_progress? && binding.last_operation.type == 'delete'
-        raise_binding_error!(service_instance, 'An existing binding is being deleted. Try recreating the binding later.')
+    def raise_invalid_binding_error!(service_instance, binding)
```
philippthun, line 196: This could then be split up into:
- `def binding_operation_in_progress!(service_instance, operation_type)`
- `def binding_deletion_failed!(service_instance)`

Both would also use the `raise_binding_error!` method.

app/actions/app_apply_manifest.rb
```diff
@@ -193,17 +191,21 @@ def create_binding_message(service_instance_guid, app_guid, manifest_service_bin
...
       raise ServiceBindingError.new(error_message)
     end
 
+    def raise_binding_operation_in_progress!(service_instance, operation)
+      error_message = "A binding is being #{operation}d. Retry this operation later."
+      raise_binding_error!(service_instance, error_message)
+    end
+
+    def raise_binding_deletion_failed!(service_instance, operation)
+      error_message = "A binding failed to be #{operation}d. Resolve the issue with this binding before retrying this operation."
```
philippthun, line 205: One very small thing: the method is called `raise_binding_deletion_failed!`, so it's purpose is to raise an exception when the state is "delete failed". Thus the `operation` parameter is not required; it's always `delete`.

philippthun (approved):

PR #2693, opened 2022-02-22
Allow overwrite of delayed job priorities

johha:
With this change the job priority can be overwritten in the cloud controller config. This allows operators to prioritize e.g. cf push over service broker operations.
It is implemented by an optional 'priorities' config parameter thus it does not change the current behavior/ priorities. The parameter is a hash which contains the jobs display_name and its overwritten priority (int).
An example configuration could look like this:
  ```  
  jobs:
      global:
        timeout_in_seconds: 14400
      priorities:
        space.apply_manifest: -10
        buildpack.delete: 10
  ```
cloudfoundry/capi-release/pull/227

* Related doc PRs
https://github.com/cloudfoundry/docs-cf-admin/pull/214
https://github.com/cloudfoundry/docs-book-cloudfoundry/pull/118

### DIFF
--- a/lib/cloud_controller/job/job_priority_overwriter.rb
+++ b/lib/cloud_controller/job/job_priority_overwriter.rb
@@ -5,14 +5,14 @@ module VCAP::CloudController
     end
 
     def get(job_display_name)
-      if is_overwritten(job_display_name)
+      if overwritten?(job_display_name)
         @overwritten_job_priorities[job_display_name.to_sym]
       end
     end
 
     private
 
-    def is_overwritten(job_display_name)
+    def overwritten?(job_display_name)
       return false if @overwritten_job_priorities.nil? || job_display_name.nil?
 
       @overwritten_job_priorities.key?(job_display_name.to_sym)
--- a/lib/cloud_controller/job/job_priority_overwriter.rb
+++ b/lib/cloud_controller/job/job_priority_overwriter.rb
@@ -1,0 +1,23 @@
+module VCAP::CloudController
+  class JobPriorityOverwriter
+    def initialize(config)
+      @overwritten_job_priorities = config.get(:jobs, :priorities)
+    end
+
+    def get(job_display_name)
+      if is_overwritten(job_display_name)
+        @overwritten_job_priorities[job_display_name.to_sym]
+      end
+    end
+
+    private
+
+    def is_overwritten(job_display_name)
+      return false if @overwritten_job_priorities.nil? || job_display_name.nil?
+
+      @overwritten_job_priorities.key?(job_display_name.to_sym)
+    end
+
+    attr_reader :config
+  end
+end
--- a/spec/unit/jobs/enqueuer_spec.rb
+++ b/spec/unit/jobs/enqueuer_spec.rb
@@ -12,12 +12,12 @@ module VCAP::CloudController::Jobs
           global: {
             timeout_in_seconds: global_timeout,
           },
-          priorities: job_display_name
+          **priorities
         }
       }
     end
     let(:global_timeout) { 5.hours }
-    let(:job_display_name) { nil }
+    let(:priorities) { {} }
 
     before do
       TestConfig.override(**config_override)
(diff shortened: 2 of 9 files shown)


lib/cloud_controller/job/job_priority_overwriter.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def get(job_display_name)
+      if is_overwritten(job_display_name)
+        @overwritten_job_priorities[job_display_name.to_sym]
+      end
+    end
+
+    private
+
+    def is_overwritten(job_display_name)
```
philippthun, line 15: Small Ruby style thing: `is_overwritten` => `overwritten?`

spec/unit/jobs/enqueuer_spec.rb
```diff
@@ -11,11 +11,13 @@ module VCAP::CloudController::Jobs
         jobs: {
           global: {
             timeout_in_seconds: global_timeout,
-          }
+          },
+          priorities: job_display_name
```
philippthun, line 15: As the `priorities` node is optional, this line could be changed to `**priorities`...

spec/unit/jobs/enqueuer_spec.rb
```diff
@@ -11,11 +11,13 @@ module VCAP::CloudController::Jobs
...
         }
       }
     end
     let(:global_timeout) { 5.hours }
+    let(:job_display_name) { nil }
```
philippthun, line 20: And this line to: `let(:priorities) { {} }`

spec/unit/jobs/enqueuer_spec.rb
```diff
@@ -117,6 +128,18 @@ module VCAP::CloudController::Jobs
           end
         end
       end
+
+      context 'priority from config' do
+        let(:job_display_name) { { wrapped_job.display_name.to_sym => 1899 } }
```
philippthun, line 133: And this one to: `let(:priorities) { { priorities: { wrapped_job.display_name.to_sym => 1899 } } }`

philippthun (approved):

sethboyles:
Any idea if jobs with 'higher' priority (e.g. priority 0 or 1) could indefinitely prevent jobs with lower priority (e.g. a job with priority 10)?  Is that something you all have tested or are concerned about?  We read a little bit on the delayed job readme but didn't see a definitive answer there.

johha:
> Any idea if jobs with 'higher' priority (e.g. priority 0 or 1) could indefinitely prevent jobs with lower priority (e.g. a job with priority 10)? Is that something you all have tested or are concerned about? We read a little bit on the delayed job readme but didn't see a definitive answer there.

As the query includes a `run_at` filter this doesn't seem to be an issue and thus most of the jobs will be filtered out. 
```
SELECT * FROM \"delayed_jobs\" WHERE ((((\"run_at\" <= '2022-03-04 07:36:03.671035+0000') AND (\"locked_at\" IS NULL)) OR (\"locked_at\" < '2022-03-04 03:36:03.671035+0000') OR (\"locked_by\" = 'cc_global_worker.cc-worker.10.6')) AND (\"failed_at\" IS NULL) AND (\"queue\" IN ('cc-generic', 'app_usage_events', 'audit_events', 'failed_jobs', 'service_usage_events', 'completed_tasks', 'expired_blob_cleanup', 'expired_resource_cleanup', 'expired_orphaned_blob_cleanup', 'orphaned_blobs_cleanup', 'pollable_job_cleanup', 'pending_droplets', 'pending_builds', 'prune_completed_deployments', 'prune_completed_builds', 'prune_excess_app_revisions', 'request_counts_cleanup'))) ORDER BY \"priority\" ASC, \"run_at\" ASC LIMIT 1 FOR UPDATE
```
Also if a job has multiple attempts its priority will decrease (always doubled: 0 > 1 > 2 > 4 ...) and `run_at` will increase exponentially: https://github.com/cloudfoundry/cloud_controller_ng/blob/4a23a419862ef9b7272f411a4435e33fc7305f30/app/jobs/cc_job.rb#L5

This means that we already have jobs with lower priorities and those are still picked up eventually.

However our plan is to use negative priorities (meaning more important) for certain jobs like cf-push so that we avoid changing the current priorities and defaults too much. In case a job with a negative priority fails (and it has retries) the next priorities would be 0,1,2... like for all other jobs.

PR #2724, opened 2022-03-18
Prefer Numerics in schema config over Floats

sethboyles:
* 'Float' will throw errors if an Integer is given to the config

Authored-by: Seth Boyles <sboyles@pivotal.io>

Passing an integer (e.g. 1) to the config results in a validation error.  See thread here: https://cloudfoundry.slack.com/archives/C07C04W4Q/p1647454358000839

### DIFF
--- a/spec/unit/lib/cloud_controller/config_spec.rb
+++ b/spec/unit/lib/cloud_controller/config_spec.rb
@@ -585,6 +585,60 @@ module VCAP::CloudController
       end
     end
 
+    describe 'broker_client_async_poll_exponential_backoff_rate' do
+      let(:cc_config_file) do
+        config = YAMLConfig.safe_load_file('config/cloud_controller.yml')
+        config['broker_client_async_poll_exponential_backoff_rate'] = backoff_rate
+
+        file = Tempfile.new('cc_config.yml')
+        file.write(YAML.dump(config))
+        file.close
+        file
+      end
+
+      let(:config) do
+        Config.load_from_file(cc_config_file, context: schema_context)
+      end
+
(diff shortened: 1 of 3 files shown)


tcdowney:
One other thing to test is what happens if you give it a [Rational](https://ruby-doc.org/core-2.5.0/Rational.html) or some other even more obscure child of Numeric. Then again -- maybe people just shouldn't configure their capi-release that way. 😅

spec/unit/lib/cloud_controller/config_spec.rb
```diff
@@ -585,6 +585,63 @@ module VCAP::CloudController
...
+        context 'when given a Float' do
+          let(:backoff_rate) { 1.0 }
+
+          it 'succeeds' do
+            expect(config.get(:broker_client_async_poll_exponential_backoff_rate)).to eq 1
+          end
+        end
+      end
+
+      context 'worker schema' do
```
philippthun, line 625: api schema

philippthun (approved):

spec/unit/lib/cloud_controller/config_spec.rb
```diff
@@ -585,6 +585,63 @@ module VCAP::CloudController
       end
     end
 
+    describe 'broker_client_async_poll_exponential_backoff_rate' do
+      let(:backoff_rate) { 1.0 }
+      let(:schema_context) { :worker }
```
philippthun, line 590: These two `let`s are overwritten by the individual tests. Would it make sense to omit them here or set them to `nil`?

PR #2727, opened 2022-03-21
Include specific errors returned by 'unzip' shell out in the response

philippthun:
Concrete errors reported by `unzip` are logged but unavailable to the caller. Instead of returning "Invalid zip archive" only, also provide details in the API error response for some specific errors.

### DIFF
--- a/lib/cloud_controller/app_packager.rb
+++ b/lib/cloud_controller/app_packager.rb
@@ -23,7 +23,7 @@ class AppPackager
 
     unless status.success?
       logger.error("Unzipping had errors\n STDOUT: \"#{output}\"\n STDERR: \"#{error}\"")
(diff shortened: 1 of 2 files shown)


Gerg:
I'm cautious about exposing shell output directly on the API, since it could be used in conjunction with an exploit to expose information about CC's filesystem. If there are specific errors that you care about, mapping them to error messages that we control would be a safer option.

philippthun:
I've added two specific errors now that get returned instead of the shell output. The first one is reported for an empty or non-zip file, the second one is about errors due to mismatching filename encoding. The list can be enhanced as needed.

moleske (approved):
Looks good to us!  Thanks for switching to mapped messages

-michael & @sethboyles

PR #2728, opened 2022-03-21
Send retry-after header when ratelimiting service-broker calls

will-gant:

This adds a Retry-After header to responses sent by the `ServiceBrokerRateLimiter`, which already rate limits users who make too many concurrent requests to endpoints associated with the six API resources below:
* v2/service_instances
* v2/service_bindings
* v2/service_keys
* v3/service_instances
* v3/service_credential_bindings
* v3/service_route_binding

The new header suggests an absolute retry time between 30 and 90 seconds in the future (a time within this range is selected at random at the time of the call).

 This addresses an anomaly whereby this rate limiter does not include such a header while other rate limiters in the Cloud Controller do, and will help clients (some of whom immediately attempt retries) to avoid getting themselves rate limited.

* Related pull requests

Docs: https://github.com/cloudfoundry/docs-running-cf/pull/104

* Some background

While working on this change I generated a heatmap showing response times for endpoints for the above resources from a month of from logs from a large CF foundation that is in heavy use. See screenshot below.

Screenshot 2022-03-22 at 09 42 55

The overwhelming majority of responses occur in less than one second. For three of the API resources responses sometimes take up to 60 seconds, and for one (/v2/service_instances) this exceptionally takes up to 90 seconds (though we're talking about <10 requests out of several million).

### DIFF
--- a/spec/unit/middleware/service_broker_rate_limiter_spec.rb
+++ b/spec/unit/middleware/service_broker_rate_limiter_spec.rb
@@ -9,9 +9,10 @@ module CloudFoundry
       let(:user_env) { { 'cf.user_guid' => 'user_guid', 'PATH_INFO' => path_info } }
       let(:fake_request) { instance_double(ActionDispatch::Request, fullpath: '/v3/service_instances', method: 'POST') }
       let(:concurrent_limit) { 1 }
+      let(:broker_timeout) { 60 }
       let(:middleware) {
         ServiceBrokerRequestCounter.instance.limit = concurrent_limit
-        ServiceBrokerRateLimiter.new(app, logger: logger)
+        ServiceBrokerRateLimiter.new(app, logger: logger, broker_timeout_seconds: broker_timeout)
       }
 
       before(:each) do
@@ -67,7 +68,7 @@ module CloudFoundry
                 'description' => 'Service broker concurrent request limit exceeded',
                 'error_code' => 'CF-ServiceBrokerRateLimitExceeded',
               )
-              expect(response_headers['Retry-After']).to be_within(90.second).of Time.now
+              expect(response_headers['Retry-After']).to be_between(Time.now + broker_timeout * 0.5, Time.now + broker_timeout * 1.5)
             end
           end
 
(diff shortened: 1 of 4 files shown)


will-gant:
I've pushed a new commit implementing the approach outlined in my last comment, and have set this PR back to 'ready for review'.

FloThinksPi (approved):

spec/unit/middleware/service_broker_rate_limiter_spec.rb
```diff
@@ -60,27 +61,42 @@ module Middleware
...
             it 'formats the response error in v2 format' do
-              _, _, body = middleware.call(user_env)
+              _, response_headers, body = middleware.call(user_env)
               json_body = JSON.parse(body.first)
               expect(json_body).to include(
                 'code' => 10016,
                 'description' => 'Service broker concurrent request limit exceeded',
                 'error_code' => 'CF-ServiceBrokerRateLimitExceeded',
               )
+              expect(response_headers['Retry-After']).to be_between(Time.now + broker_timeout * 0.5, Time.now + broker_timeout * 1.5)
```
philippthun, line 71: The [unit test run](https://github.com/cloudfoundry/cloud_controller_ng/actions/runs/2027306272) for the merged commit failed. I would suggest to use `Timecop` to freeze the time before issuing the call.

spec/unit/middleware/service_broker_rate_limiter_spec.rb
```diff
@@ -60,27 +61,42 @@ module Middleware
...
+          context 'when broker_client_timeout_seconds is reduced' do
+            let(:broker_timeout) { 3 }
+            let(:middleware) {
+              ServiceBrokerRequestCounter.instance.limit = concurrent_limit
+              ServiceBrokerRateLimiter.new(app, logger: logger, broker_timeout_seconds: broker_timeout)
+            }
+
+            it 'reduces the suggested delay in the Retry-After header' do
+              _, response_headers, _ = middleware.call(user_env)
+              expect(response_headers['Retry-After']).to be_between(Time.now + broker_timeout * 0.5, Time.now + broker_timeout * 1.5)
```
philippthun, line 99: Should this be rounded (floored / ceiled)? Or should we use a different `broker_timeout`, e.g. `4` for which it is more clear what to expect?

spec/unit/middleware/service_broker_rate_limiter_spec.rb
will-gant: Fixed with #2733

Issue #2713, opened 2022-03-07
State handling in v3 service instance endpoints


svkrieger:
# Legend

| Icon | Meaning |
| :-: | :-: |
| :white_check_mark: | fixed/wanted behaviour |
| :x: | not fixed but change of behaviour needed |
| :heavy_minus_sign: | no change of behaviour needed |
| :question: | questionable behaviour, needs further analysis |

# V3 API

The introduction of the Failed State for Broker operations does come with some unwanted behaviour in some cases.
A first analysis and fix of the service binding endpoints was done here: https://github.com/cloudfoundry/cloud_controller_ng/issues/2637

This issue describes the findings regarding the service instance endpoints.

We made the assumption that a service instance which is in state `update failed` or `delete failed`, is still in a functioning state. The next OSBAPI spec version 2.16 "Allow[s] Service Brokers to indicate if a Service Instance is still usable after a failed update or deprovisioning" (https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#changes-since-v215), which is currently not supported by the Cloud Controller. Furthermore, "If the broker does not indicate in the Error response or Last Operation response whether the Service Instance is usable or not, the Platform SHOULD assume it is still usable" (https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-5).

### POST /v3/service_instances (i.e. cf create-service) for already existing service instance

| Resource State | Observed Behaviour | Wanted Behaviour | PR  | Fixed in main |
| :-: | :-: | :-: | :-: | :-: |
| none (synchronous case) | :question: | | | :heavy_minus_sign: |
| create succeeded | rejects request (**1**) | :white_check_mark: | | :heavy_minus_sign: |
| create in progress | rejects request (**1**) | :white_check_mark: | | :heavy_minus_sign: |
| create failed | rejects request (**1**) | recreates instance | https://github.com/cloudfoundry/cloud_controller_ng/pull/2730 | :white_check_mark: |
| update succeeded | rejects request (**1**) | :white_check_mark: || :heavy_minus_sign: |
| update in progress | rejects request (**1**) | :white_check_mark: || :heavy_minus_sign: |
| update failed | rejects request (**1**) | :white_check_mark: || :heavy_minus_sign: |
| delete failed | rejects request (**1**) | :white_check_mark: || :heavy_minus_sign: |
| delete in progress | rejects request (**1**) | :white_check_mark: || :heavy_minus_sign: |

[...]

PR #2697, opened 2022-02-22
Prevent inconsistencies when cancelling service binding/instance creations

philippthun:
This change adds a new 'create initial' state to service operations to distinguish between requests that have not been responded to and operations that are being processed asynchronously by the broker.

The reasoning for this change is the following scenario:
```mermaid
sequenceDiagram
  participant Client
  participant Cloud Controller
  participant CCDB
  participant Some Routing Component
  participant Service Broker

  Client->>+Cloud Controller: create
  Cloud Controller->>CCDB: insert
  activate CCDB
  Cloud Controller->>+Some Routing Component: provision
  Note over Some Routing Component: some delay

  rect rgba(255, 0, 0, .2)
    Client->>+Cloud Controller: delete
    Cloud Controller->>+Some Routing Component: deprovision
    Some Routing Component->>Service Broker: #
    Service Broker-->>Some Routing Component: 410 Gone
    Some Routing Component-->>-Cloud Controller: #
    Cloud Controller->>-CCDB: delete
    deactivate CCDB
  end

  Some Routing Component->>Service Broker: provision (cont.)
  Service Broker-->>Some Routing Component: 201 Created
  Some Routing Component-->>-Cloud Controller: #
  Cloud Controller->>-CCDB: update
  Note over CCDB: already deleted!
```

When a creation is delayed for some reason (up to 60 seconds until the request times out), it is possible that an impatient client issues a deletion request in parallel that might lead to a situation where an instance is created by the service broker, but the Cloud Controller already removed it from its database. To mitigate this inconsistency, the Cloud Controller would need to trigger an orphan mitigation; but this is solely done within the [service client implementation](/cloudfoundry/cloud_controller_ng/blob/main/lib/services/service_brokers/v2/client.rb), i.e. only as a result of communication errors, not based on state inconsistencies.

The alternative approach implemented in this PR is the introduction of a new 'create initial' state that is set in `ServiceCredentialBindingAppCreate`, `ServiceCredentialBindingKeyCreate`, `ServiceRouteBindingCreate` and `ServiceInstanceCreateManaged`. In most cases the 'initial' state is semantically treated as 'in progress', but when checking for a blocking operation before accepting a delete request, it is treated as such and results in an error being reported to the client:
```mermaid
sequenceDiagram
  participant Client
  participant Cloud Controller
  participant CCDB
  participant Some Routing Component
  participant Service Broker

  Client->>+Cloud Controller: create
  Cloud Controller->>CCDB: insert
  activate CCDB
[...]

PR #2730, opened 2022-03-22
Fix state handling for POST /v3/service_instances

svkrieger:

* Recreate service instance when state is `create failed`
* Better error message for state `delete failed` and `delete in progress`

Links to any other associated PRs
* This PR solves the described issues according to https://github.com/cloudfoundry/cloud_controller_ng/issues/2713


philippthun:
There is a refactoring commit (as part of #2697) that would make the changes in `service_instance.rb` obsolete as `managed_service_instance.rb` would include the `ServiceOperationMixin`. I've extracted this commit into its own PR (#2734) - once merged, this PR could be rebased.

app/actions/v3/service_instance_create_managed.rb
```diff
@@ -38,7 +38,11 @@ def precursor(message:)
           state: ManagedServiceInstance::IN_PROGRESS_STRING
         }
 
-        ManagedServiceInstance.new.tap do |i|
```
philippthun, line 41: This comment is for the beginning of the method...

I had a look where this `precursor` method is invoked. `create_managed` in `ServiceInstancesV3Controller` seems to be the only(?) place. As there the service plan is already fetched from the database, I'm wondering if it makes sense to add `service_plan` as a second (optional) parameter. If provided, fetching it from the database in this method is not necessary anymore, i.e. one query less.

app/actions/v3/service_instance_create_managed.rb
```diff
@@ -185,6 +189,14 @@ def parse_response(fetch_instance, last_operation)
         }
       end
 
+      def validate_service_instance!(instance)
+        if instance
+          instance_already_exists!(instance.name) if instance.create_succeeded? || instance.create_in_progress? || instance.update_succeeded? ||
+            instance.update_in_progress? || instance.update_failed? || instance.last_operation.nil?
```
philippthun, line 195: I think that the check for `last_operation.nil?` is already included in the `ServiceOperationMixin`.

`def create_succeeded?` -> `!last_operation? || ...`

`def last_operation?` -> `!last_operation.nil?`

=> so it's `!!last_operation.nil?` = `last_operation.nil?` :smile:

app/actions/v3/service_instance_create_managed.rb
```diff
@@ -185,6 +189,14 @@ def parse_response(fetch_instance, last_operation)
         }
       end
 
+      def validate_service_instance!(instance)
+        if instance
+          instance_already_exists!(instance.name) if instance.create_succeeded? || instance.create_in_progress? || instance.update_succeeded? ||
```
philippthun, line 194: We basically raise an error when the last operation type is `update` - regardless of the state. It makes sense to enhance the `ServiceOperationMixin` with a method that answers this question. There already is a private method `update?`, so this one could be made public. But I would then also rename it, as I would not understand `instance.update?`. Maybe `instance.last_operation_update?`? Other ideas?

spec/unit/actions/v3/service_instance_create_managed_spec.rb
```diff
@@ -104,16 +104,112 @@ module V3
...
+          let(:instance) { ManagedServiceInstance.make(name: name, space: space) }
+
+          context "when the last operation is in state 'create in progress'" do
+            before do
+              instance.save_with_new_operation({}, { type: 'create', state: 'in progress' })
+            end
+
+            it 'should raise' do
+              expect { action.precursor(message: message) }.to raise_error(
+                ServiceInstanceCreateManaged::InvalidManagedServiceInstance,
```
philippthun, line 117: IntelliJ has a strange sense of aesthetics. I would favor the following formatting:
```
              expect { action.precursor(message: message) }.to raise_error(
                ServiceInstanceCreateManaged::InvalidManagedServiceInstance,
                'The service instance name is taken: si-test-name.'
              )
```

app/actions/v3/service_instance_create_managed.rb
```diff
@@ -38,7 +38,11 @@ def precursor(message:)
           state: ManagedServiceInstance::IN_PROGRESS_STRING
         }
 
-        ManagedServiceInstance.new.tap do |i|
+        space = VCAP::CloudController::Space.first(guid: message.relationships[:space][:data][:guid])
```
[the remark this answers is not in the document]
svkrieger: Good catch, there is no difference.

app/actions/v3/service_instance_create_managed.rb
svkrieger: Added service_plan as a non-optional parameter. I also just found this one caller.

app/actions/v3/service_instance_create_managed.rb
svkrieger: Ehm, what? :D I added a method `last_operation_nil` to the `ServiceOperationMixin`.

app/actions/v3/service_instance_create_managed.rb
svkrieger: Done. Called it `last_operation_is_update` otherwise it might be still confusing.

spec/unit/actions/v3/service_instance_create_managed_spec.rb
svkrieger: Indeed, fixed that.

app/actions/v3/service_instance_create_managed.rb
```diff
@@ -38,7 +38,11 @@ def precursor(message:)
...
+        instance = ManagedServiceInstance.first(name: message.name, space_id: space.id)
```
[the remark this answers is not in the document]
svkrieger: done

app/actions/v3/service_instance_create_managed.rb
philippthun: Sorry, that was confusing...

What I meant was that you don't need this explicit check for `last_operation.nil?`. If the `last_operation` is `nil`, `create_succeeded?` will already return `true`.

The same (i.e. no explicit `nil` check) is done for service bindings, e.g. [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/de07d6efddb933ecdfd710aba64fbd37cfedc1ad/app/actions/service_credential_binding_app_create.rb#L65).

So you can omit this check and also remove the `last_operation_nil` method from `ServiceOperationMixin`.

app/models/runtime/helpers/service_operation_mixin.rb
```diff
@@ -49,6 +49,14 @@ def delete_failed?
       delete? && failed?
     end
 
+    def last_operation_nil?
+      !last_operation?
+    end
+
+    def last_operation_is_update?
```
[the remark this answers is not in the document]
svkrieger: Test added and copy-paste errors fixed

app/actions/v3/service_instance_create_managed.rb
svkrieger: ohh right, changed it accordingly.

spec/unit/models/services/service_operation_shared.rb
```diff
@@ -206,7 +233,7 @@ def update_operation(type, state)
   describe '#update_succeeded?' do
     context 'when there is no operation' do
       it 'returns true' do
```
philippthun, line 235: Very small thing: The description needs to be adapted, i.e. `it 'returns false' do`

svkrieger: done

philippthun (approved):

philippthun (approved):

philippthun:
Will be merged together with the related PRs...

PR #2735, opened 2022-03-24
Install Ruby version inside actions job based on project

FloThinksPi:
With this change one does not need to maintain own OCI images for github
actions. Each run is based on a stock ubuntu image which is setup
acordingly at runtime of a job. This makes it possible to run the unit
tests with exactly what is defined in cloud_controller_ng and not have
to sync a OCI image to the state of the cloud_controller_ng project.

### DIFF
--- a/.github/workflows/composite/setup/action.yml
+++ b/.github/workflows/composite/setup/action.yml
@@ -1,0 +1,43 @@
+name: "setup"
+description: "Configure Ubuntu to run Cloud_Controller_NG and its tests"
+
+inputs:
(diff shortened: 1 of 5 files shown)


FloThinksPi:
@moleske the PRS run in the forks repo as security measures prevent actions to be executed in the main project when the PRS are from a fork I suppose.

Yet I was unable to get the mysql part running before my vacation, had some weid issues where it would try to connect to a Unix socket even when a port was specified. 

But maybe it helps someone to take this a step further In the next week. Will be back on 4.4. and could give it a second try then :)

PR #2755, opened 2022-04-05
Install Ruby version inside actions job based on project

FloThinksPi:

With this change one does not need to maintain own OCI images for github
actions. Each run is based on a stock ubuntu image which is setup
acordingly at runtime of a job. This makes it possible to run the unit
tests with exactly what is defined in cloud_controller_ng and not have
to sync a OCI image to the state of the cloud_controller_ng project.

Simplifies Github action setup.
Reuses the same part of steps to set up an image.
Makes building an own OCI image for CI obsolete.
Installs the ruby version from the code so ruby version bumps are correctly tested.

### DIFF
--- a/.github/workflows/composite/setup/action.yml
+++ b/.github/workflows/composite/setup/action.yml
@@ -1,0 +1,44 @@
+name: "setup"
+description: "Configure Ubuntu to run Cloud_Controller_NG and its tests"
+
+inputs:
+  BOSH_CLI_VERSION:
+    description: "Bosh CLI Version"
+    required: true
+    default: 6.4.17
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v2
+    - run: |
+        sudo apt-get update && \
+        sudo apt-get -y install \
+        build-essential \
+        curl \
+        debconf-utils \
+        git \
+        libcurl4-openssl-dev \
+        libpq-dev \
+        libreadline-dev \
+        libsqlite3-dev \
+        libssl-dev \
+        libxml2-dev \
+        libxslt-dev \
+        libyaml-dev \
+        libv8-dev \
+        software-properties-common \
+        unzip \
+        wget \
+        zip \
+        zlib1g-dev
+        
+        sudo wget -O /usr/local/bin/bosh https://s3.amazonaws.com/bosh-cli-artifacts/bosh-cli-${{ inputs.BOSH_CLI_VERSION }}-linux-amd64 && sudo chmod +x /usr/local/bin/bosh
+      shell: bash
+    - uses: ruby/setup-ruby@v1
+      with:
+        bundler: "2.3.5"
+        bundler-cache: true
+    - name: Install dependencies
+      run: bundle install
+      shell: bash
--- a/.github/workflows/composite/setup/action.yml
+++ b/.github/workflows/composite/setup/action.yml
@@ -6,12 +6,15 @@ inputs:
     description: "Bosh CLI Version"
     required: true
     default: 6.4.17
-
+  WORKING_DIRECTORY:
+    description: "The Current Work Directory from which installs specifically bundle installs should take place"
+    required: true
+    default: "."
 runs:
   using: "composite"
   steps:
-    - uses: actions/checkout@v2
-    - run: |
+    - name: Install OS dependencies
+      run: |
         sudo apt-get update && \
         sudo apt-get -y install \
         build-essential \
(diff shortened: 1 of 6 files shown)


FloThinksPi:
@moleske got it working now :). Btw did you enable actions beeing executed from forks ? It is running now in this repo but i am not sure if e.g. Secrets are passed to the jobs(and thats why they where not executed by default). If that would be the case any person doing a fork and pr could dump the secrets. However since i am not an admin i cannot check the actions config in this repo :).

moleske:
I think we have enabled github actions from "trustworthy" users.  People who have certain administrative rights to a repository can allow certain github users PRs from that user's fork to run github actions.  Though maybe it is per an org since I see the PR is from https://github.com/sap-contributions and not your personal fork

.github/workflows/composite/setup/action.yml
```diff
@@ -0,0 +1,44 @@
...
+        unzip \
+        wget \
+        zip \
+        zlib1g-dev
+        
+        sudo wget -O /usr/local/bin/bosh https://s3.amazonaws.com/bosh-cli-artifacts/bosh-cli-${{ inputs.BOSH_CLI_VERSION }}-linux-amd64 && sudo chmod +x /usr/local/bin/bosh
+      shell: bash
+    - uses: ruby/setup-ruby@v1
+      with:
+        bundler: "2.3.5"
```
moleske, line 40: little surprised to see specific bundler version.  When we were working on [ruby 3 upgrade](https://github.com/cloudfoundry/cloud_controller_ng/pull/2703), we went with matching the bundler version that comes with ruby 3 so we didn't have install/specify a special version.  Probably not a big deal though, just somewhere else we need to keep bundler in sync in addition to the gemfile

moleske (approved):
Functionally looks good to me, just small comment about how concerned we should be about managing a bundler version outside of the gemfile.  Curious to hear from folks who've been working on cloud controller longer than I about any concerns with github actions being very different setup wise from the concourse pipeline

.github/workflows/composite/setup/action.yml
sethboyles: Let me see if I can recall this correctly...

Packaging has two steps [prepackaging](https://github.com/cloudfoundry/capi-release/blob/develop/packages/cloud_controller_ng/pre_packaging) and [packaging](https://github.com/cloudfoundry/capi-release/blob/develop/packages/cloud_controller_ng/packaging)

1. Prepackaging runs on your local machine, and bundles/vendors gems into a local cache
1. Packaging builds CCNG on the BOSH vm, using the cached directory created in prepackaging.  This is done by calling `bosh_bundle_local`, which is [defined here](https://github.com/bosh-packages/ruby-release/blob/efcabdd546478d69a0f42ba40bac2323de0469f0/src/compile-2.7.2.env#L15-L23).  It basically does a version of `bundle install --local`, which is [described as thus](https://bundler.io/v2.3/man/bundle-install.1.html):

```
--local
Do not attempt to connect to rubygems.org. Instead, Bundler will use the gems already present in Rubygems' cache or in vendor/cache. Note that if an appropriate platform-specific gem exists on rubygems.org it will not be found.
```

Packaging, because it runs on the VM (and it needs to be able to run without internet), generally only has access to the version of bundler packaged with the [BOSH Ruby release.](https://github.com/bosh-packages/ruby-release).

So prepackaging with a Bundler version newer than the Bundler version the VM has access to can run into issues.  Normally you get a warning when bundling with an older version, but it's generally OK--but bundling against a local cache made with a newer version of Bundler can be more error prone.  It's been awhile, but I think we had this issue with [the newrelic plugin gem](https://github.com/cloudfoundry/capi-release/commit/d7d74743d739381ab7cb7ca2566ad3785a1738a8).  More details here: https://www.pivotaltracker.com/n/projects/966314/stories/176467695

So I wouldn't recommend bumping the `BUNDLED WITH` beyond what the bosh release has.

sethboyles: As an aside, It's interesting that information about prepackaging has [been made intentionally hard to find](https://github.com/cloudfoundry/bosh/issues/1344#issuecomment-233097369).  In the [bosh.io packaging tutorial](https://bosh.io/docs/create-release/), they say this: 

`Use of the pre_packaging file is not recommended, and is not discussed in this tutorial.`

Instead they package gems manually.  For us, would that mean committing the vendor directory to CCNG?  Not that it would resolve this issue with mismatching Bundler versions

FloThinksPi: Yea so the issue is there is a bug in bundler that lets installs fail on the used ubuntu image and runner combination https://github.com/sap-contributions/cloud_controller_ng/runs/5864281143?check_suite_focus=true 
https://github.com/rubyjs/mini_racer/issues/218
One needs bundler 2.3 for this to work. Yet it is not enough to increase the bundler version just in the docs/v3/gemfile.lock and not in the main gemfile.lock. Would this be fine ?
Added a commit to handle bundler versions properly according to whats in the gemfile.lock

sethboyles: Yeah, I think it's OK for the docs gemfile

PR #2769, opened 2022-04-13
Restructure docs

FloThinksPi:
This includes two changes.

1. Unittests now dont run on PRs that are just markdown.
2. The decitions documentation was moved to /docs as it feels to be the right place.


philippthun (commented):
The `decisions` folder is maintained here: https://github.com/cloudfoundry/cloud_controller_ng/blob/main/.adr-dir

So if you want to move these files, you would also have to adjust the `.adr-dir`. On the actual move I don't have an opinion...

PR #2773, opened 2022-04-22
Enhance filter and tests for degenerated deployments

philippthun:
Instead of filtering for `deploying_web_process_guid` being `nil`, left join the `processes` table and filter for non-existing processes; this also returns deployments that reference a `deploying_web_process` which does not exist anymore.

Ensure that a degenerated deployment is updated (and logged) only once.

Remove unnecessary select from `apps` table.

### DIFF
--- a/lib/cloud_controller/deployment_updater/dispatcher.rb
+++ b/lib/cloud_controller/deployment_updater/dispatcher.rb
@@ -37,7 +37,11 @@ module VCAP::CloudController
         end
 
         def finalize_degenerate_deployments(logger)
-          DeploymentModel.qualify.left_join(:processes, guid: :deployments__deploying_web_process_guid).where(processes__id: nil).each do |d|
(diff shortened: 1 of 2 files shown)


philippthun:
Some fun facts:

On our largest foundation there were more than 10 million `finalized-degenerate-deployment` log entries per day. By changing the query (`exclude(status_reason: DeploymentModel::DEGENERATE_STATUS_REASON)`) it is ensured that every degenerated deployment is processed only once (and not every 5 seconds until cleaned up from the database).

There were also more than 10 million unnecessary `SELECT`s from the database - they will now even be gone for those deployments that need to be updated by not following the `deployments -> apps` relation (`d.app.guid`) but using the already available foreign key (`d.app_guid`).

moleske (approved):
Looks good to me.  Not super thrilled with shared_examples (I personally don't care for them), but that is more of me problem and I see the usefulness for it here.

Issue #2648, opened 2022-01-27
Create metrics endpoint for prom-scraper consumption


sethboyles:
Currently CCNG uses [statsd-injector](https://github.com/cloudfoundry/statsd-injector-release) to gather Statsd style metrics and forward them to loggregator.  

We should build a `/metrics` endpoint to supply those same metrics, so we can use [prom-scraper](https://github.com/cloudfoundry/loggregator-agent-release/blob/main/docs/prom-scraper.md) to forward the metrics instead.  

Doing this will allow us to remove the EventMachine/Thin code we have:

https://github.com/cloudfoundry/cloud_controller_ng/blob/4a23a419862ef9b7272f411a4435e33fc7305f30/lib/cloud_controller/metrics/periodic_updater.rb#L18-L20

and consequently address https://github.com/cloudfoundry/cloud_controller_ng/issues/2254 and move CCNG from Thin to Puma.

sethboyles:
Currently we emit metrics named with periods, e.g. `cc.job_queue_length.total`.  However the [Prometheus data model](https://prometheus.io/docs/concepts/data_model/) prohibits periods and most other non-alphanumeric characters in metric names, so to migrate to prom-scraper would mean having to rename almost all of our metrics.  So  `cc.job_queue_length.total` would become `cc_job_queue_length_total` or similar. 

@stephanme @philippthun do you have any thoughts on how we would handle an upgrade here?

A few options:
1. Do a hard cut where we switch from period-named metrics to underscore metrics in one release, causing a breaking change where monitoring tools would have to be updated
1. Introduce the new non-period metric names in a release to be emitted alongside the period metric names, giving operators a deprecation window to update their monitoring tools
1. Make a configurable option that, when enabled, stops emitting the period metric names, and emits the new metric names

Would appreciate your input here.

philippthun:
Option 1 would work for us; we would need to adapt our monitoring tools beforehand to support both formats and then change them a second time to clean this up (i.e. support only the new format).

But I'm not sure if this works for others as well, thus I would prefer option 2. This also has the advantage that we only have to adapt our monitoring tools once without any disruption.

There might be even an option 4, that combines 2 and 3; i.e. a config parameter that specifies what to emit: either the old, or the new format, or both.

But given the fact that we are not talking about hundreds of metrics, that might be over-engineered; so simply having both formats for one release seems to be the best/easiest solution.

sethboyles:
[...]

PR #2781, opened 2022-04-28
Creates metrics endpoint for use by prom-scraper

klakin-pivotal:
This change resolves Issue #2648 . It adds an endpoint (`/internal/v4/metrics`) which serves Prometheus-style
metrics. The intent is to eventually replace the StatsD-supplied metrics with Prometheus metrics.

It should be noted that the `timing` metric type from StatsD does not have a 1:1 equivalent in Prometheus. We have decided that supplying both a `summary` and a `gauge` metric is our best current course of action. We consulted with the Tanzu Observability team to come to this conclusion. If there's a better thing to do, please do let us know.

Anyway. The two metrics that are so affected are `cc_diego_sync_duration` and `cc_deployments_update_duration`. Each of these metrics also has a `_gauge` variant that reports the most recently collected data point.

I have not run CF Acceptance Tests, but I _have_ tested that the release can deploy multiple instances of the Dora CF App, both with cf-deployment and with VMware Tanzu.


klakin-pivotal:
cc @dalvarado @moleske @MerricdeLauney @tjvman @sethboyles

moleske:
There is apparently some sort of test pollution.  Running with seed 13579, I was able to reproduce locally on my mac with postgres14 the same failure when running the entire test suite.  Have narrowed it down to spec/unit as where the culprit may be `DB=postgres bundle exec rspec spec/unit/ --seed 13579`  I haven't tried with a seed from postgres10 or mysql8, but suspect might be able to reproduce those as well

klakin-pivotal:
Here's some commentary about the "Clear out the Prometheus Client Registry" code:

1) Don't worry about the cost of re-creating the Prometheus Client Registry. The call pretty much just creates a 
new `Mutex` and a new `Hash`.
2) I wanted to explicitly mention where we put the code that re-creates the registry, and ensure that whoever reviews the PR doesn't disagree with the appropriateness of where we put it. The code in question is here:

https://github.com/cloudfoundry/cloud_controller_ng/pull/2781/files#diff-d048418f106e02826d8c42366c82fe4098f10c150beab3e03fd3c11157d6b057R25-R42

We decided to put it here, rather than in -say `spec_helper.rb` for a few reasons:
* So far, the only tests that have had these troubles have been in `prometheus_updater_spec.rb`.
* Adding a function into the `Prometheus::Client` module is sort of a bit of a big deal, and should be made very obvious and hard to miss for anyone who needs to go investigating. In my personal experience, I forget about common files like `spec_helper.rb` all the time.
* Similarly, adding a `before` in something like `spec_helper.rb` that re-creates the Prometheus Client Registry for _every_ test in the system could be convenient for future test writers, it _might_ (maybe) cause problems for tests that would never otherwise care about the Prometheus client library.

Issue #2819, opened 2022-05-26
Feature Request: Include LRP `instance_id` in Process Stats response


tcdowney:
## Issue / Proposed Feature

_This is more of a feature request than it is an issue. I'm going to tweak the rest of the template appropriately._

In addition to instance index, Diego also provides an `instance_id` field that more accurately identifies an individual container. It would be helpful for app operators if the Process Stats endpoint (`GET /v3/apps/:guid/processes/:type/stats`) included this id in its response.

## Context

Developers can see `instance_id` in their logs so they would be able to more easily correlate info from the stats endpoint with what they are seeing there:

Example logs:
```
2022-05-26T16:16:11.09-0600 [RTR/0] OUT dora.dirt-fairy.capi.land - [2022-05-26T22:16:11.073797015Z] "GET / HTTP/2.0" 200 0 13 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:100.0) Gecko/20100101 Firefox/100.0" "24.9.143.210:53261" "10.244.0.139:61002" x_forwarded_for:"24.9.143.210" x_forwarded_proto:"https" vcap_request_id:"e44db4a0-a5b4-4cea-43de-283078631d83" response_time:0.024315 gorouter_time:0.000436 app_id:"34ac676f-5adc-4605-a412-c0a69eb2dbaf" app_index:"0" instance_id:"6907c81e-99c4-4322-4594-ac1d" x_cf_routererror:"-" x_b3_traceid:"107759ce815e7d6518d96017570f490b" x_b3_spanid:"18d96017570f490b" x_b3_parentspanid:"-" b3:"107759ce815e7d6518d96017570f490b-18d96017570f490b"
```

This information is available on the `ActualLRP` resource that the Process Stats endpoint is already referencing under the `instance_guid` field so it already there for Cloud Controller to consume:
```ruby
[5] pry(VCAP::CloudController)> bbs_client.lrp_instances(ProcessModel.first)
[...]

PR #2831, opened 2022-06-09
Add instance_guid field in Process Stats response

tcdowney:
- Display the Diego instance identifier as part of the Process Stats
response. This id is unique for each actual LRP instance and can be used
as a substitute for the integer index id.
- Eirini does not support this in their API at this time, so the OPI
variant of the instances stats reporter returns an empty string for this
field. In the future, if Eirini is enhanced to provide this we can
update that client.
- Issue: https://github.com/cloudfoundry/cloud_controller_ng/issues/2819

## An explanation of the use cases your change solves:
The issue (https://github.com/cloudfoundry/cloud_controller_ng/issues/2819) describes this use case in more detail. The summary though is this can help developers debug their apps (this id is logged by app containers) and provide a pathway for clients to provide other container ids than just the integer index.


tcdowney:
@Gerg some additional context for this change:

On Korifi we're wanting to support workloads other than `StatefulSets` (e.g. `Deployments`, knative `Services`, etc.) so we're working toward reducing the reliance of system components on the numerical indexing of app instances. I'll be putting out a proposal / some sort of design doc to start discussing this in more detail later this week, but I see this change as a first step toward that.

tcdowney:
I'm putting this into draft state. I was looking at the AppRecipeBuilder and noticed that `instance_id` is that metric tag we're adding and it corresponds to instance index. We are calling this the `process_instance_id` on metrics/logs so maybe that name should be used. Worth discussing I think.

https://github.com/cloudfoundry/cloud_controller_ng/blob/7698bc407195cb30de7455ab877d9526b6d4bafe/lib/cloud_controller/diego/app_recipe_builder.rb#L108

Gerg:
It's a little odd that we call index `instance_id` in the metrics tags. Looks like it came from [this story](https://www.pivotaltracker.com/n/projects/966314/stories/162922843).

`index`/`instance_guid` is the most intuitive to me, but I could see other options. `Process.process_instance_id` sounds a bit redundant.

tcdowney:
@Gerg 

Yeah, I started calling it `instance_guid` here in the proposal around this stuff here:
https://docs.google.com/document/d/118yYN4WGj1F2si8gk94xFZsYdewtOKoqO2Dv-1I-PgE/edit

Gerg:
What about just `guid`?

tcdowney:
cc @zrob

tcdowney:
> What about just `guid`?

I went with `instance_guid`. There are other fields prefixed with `instance_` and it felt less ambiguous since this is what the field is called on the ActualLRP.
https://github.com/cloudfoundry/bbs/blob/main/docs/030-lrps.md#instance_guid

tcdowney:
The docs failure is unrelated to my change. It looks like some of the npm packages we're using for the doc link checker are not working with modern node.js? I don't have much experience with all that so not sure, but I feel like we can ignore this for this PR.

Gerg (approved):

sethboyles (approved):

PR #2855, opened 2022-07-06
Improve query to check if broker has service instances

philippthun:
Instead of doing (more than) m times n queries (number of services * number of service plans) a single query joining the required tables is sufficient.

### before
```
1*    SELECT * FROM `services` WHERE (`service_broker_id` = ...)
m*    SELECT * FROM `service_plans` WHERE (`service_id` = ...)
m*n*  SELECT * FROM `service_instances` WHERE (`service_plan_id` = ...)
```

### after
```
SELECT 1 AS `one`
  FROM `service_instances`
  INNER JOIN `service_plans`
    ON (`service_plans`.`id` = `service_instances`.`service_plan_id`)
  INNER JOIN `services`
    ON (`services`.`id` = `service_plans`.`service_id`)
  WHERE (`services`.`service_broker_id` = 1)
  LIMIT 1
```

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats)

PR #2856, opened 2022-07-06
Use Sequel's any_not_empty extension

philippthun:
There are already several PRs (see below [1]) that change `dataset.any?` to `!dataset.empty?` or otherwise mention the difference between these two statements. Whereas the first one fetches all data from the database to then check if the resulting array contains at least one element, the second statement does a `select 1 as one from table limit 1` query that has a far better performance.

Fortunately there is a Sequel extension (see [2]) that 'fixes' this misleading behavior; so instead of searching for more places where we could change the implementation, using `dataset.any?` should now be equivalent to `!dataset.empty?` and everyone can choose freely which one to use.

[1]
#2117
#2533
#2803
#2855

[2] https://sequel.jeremyevans.net/rdoc-plugins/files/lib/sequel/extensions/any_not_empty_rb.html

### DIFF
--- a/app/models/services/service_broker.rb
+++ b/app/models/services/service_broker.rb
@@ -47,11 +47,11 @@ module VCAP::CloudController
     end
 
     def has_service_instances?
-      !VCAP::CloudController::ServiceInstance.
+      VCAP::CloudController::ServiceInstance.
         join(:service_plans, id: :service_plan_id).
         join(:services, id: :service_id).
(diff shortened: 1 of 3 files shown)


moleske (approved):
This seems alright to me.  Also seems better than trying to write some sort of rubocop type rule that says "are you sure you want to use `dataset.any?` and not `dataset.empty?`"

PR #2650, opened 2022-01-28
Adapt presenter_args function for /v3/security_groups and /v3/space_quotas as in /v3/domains

kathap:
* add consistency and use in the former changed /v3/space_quotas and
  /v3/security-groups (#2624) files the same presenter_args function as for the
  /v3/domains changement (#2638) so that instead of selecting all space guids for admin user, all_spaces_visible is set to true and if so no filtering is done, just return all space guids

Thanks for contributing to cloud_controller_ng. To speed up the process of reviewing your pull request please provide us with:

* A short explanation of the proposed change:

* An explanation of the use cases your change solves

* Links to any other associated PRs

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats)

PR #2876, opened 2022-07-19
Don't check org quota visibility if user reads globally

will-gant:

Adopts the approach taken in #2650 and #2638 to eliminate the need for DB queries in the org quotas presenter when the user is a global reader.

This noticeably speeds up [GET /v3/organization_quotas](https://v3-apidocs.cloudfoundry.org/version/3.121.0/index.html#list-organization-quotas) for global readers, which currently drags when constructing the relationships block of each response object. On foundations with thousands of orgs constructing the relationships block currently take a very long time (I measured an average of 117 seconds on one foundation with 65,000 orgs, and that's with `per_page=50` (versus 0.035 seconds for the [equivalent v2 endpoint](https://apidocs.cloudfoundry.org/16.22.0/organization_quota_definitions/creating_a_organization_quota_definition.html))). Currently, a separate DB query is made for each org that org quotas have a relationship with in order to determine whether it is visible to the user - even when we know the user is a global reader.

This change skips these expensive queries for users with an admin, admin read-only or global auditor role, and in a test environment with 20,000 orgs, each assigned its own org quota, `/v3/organization_quotas?per_page=50` now takes 0.533 seconds.

### DIFF
--- a/app/controllers/v3/organization_quotas_controller.rb
+++ b/app/controllers/v3/organization_quotas_controller.rb
@@ -17,9 +17,8 @@ class OrganizationQuotasController < ApplicationController
     unprocessable!(message.errors.full_messages) unless message.valid?
 
     organization_quota = OrganizationQuotasCreate.new.create(message)
-    visible_organizations_guids = permission_queryer.readable_org_guids
 
-    render json: Presenters::V3::OrganizationQuotaPresenter.new(organization_quota, visible_org_guids: visible_organizations_guids), status: :created
+    render json: Presenters::V3::OrganizationQuotaPresenter.new(organization_quota, **presenter_args), status: :created
   rescue OrganizationQuotasCreate::Error => e
     unprocessable!(e.message)
   end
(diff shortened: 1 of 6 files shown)


app/controllers/v3/organization_quotas_controller.rb
```diff
@@ -94,7 +90,7 @@ def apply_to_organizations
     organization_quota = QuotaDefinition.first(guid: hashed_params[:guid])
     resource_not_found!(:organization_quota) unless organization_quota
 
-    OrganizationQuotaApply.new.apply(organization_quota, message)
+    OrganizationQuotaApply.new.apply(organization_quota, message, **presenter_args)
```
philippthun, line 93: Not needed as only admins can reach this...

will-gant: Yep, I was thinking that one might be unnecessary. Reverting this.

philippthun:
Oh, `spec/request/organization_quotas_spec.rb` also checks the error message and needs to be adapted.

philippthun (approved):

PR #2881, opened 2022-07-25
Codespaces

FloThinksPi:

With this change one can use githubs codespaces feature that will create a development environment on a vm. A click in the Github UI will create an instance. Its free for all(single instance) except cooperations. 

image

With a simple click on now ends up with a fully functional development environment. This includes:
- UAA Instance
- Minio S3 Blobstore
- Nginx Reverse Proxy for package uploads
- Mariadb and Postgresdb
- Run Configurations for VSCode and Intellij/Rubymine to start right away
- Tunneling SSH from the container to localhost to be able to connect with Intellij/Rubymines new remote ssh feature

At startup the DBs are set up, migrated and seeded and also all bundle packages get installed. The setup of a codespaces takes rughly 5 minutes. Once a codespace is stopped it wont get recreated at startup so startup times are nearly instant. Also changes to the environment are persisted between restarts. Only if one choses to throw away a codespace and recreate it, thats when the build times come on top of starting times and any temporary change to the environment is gone.

We will use this as our main development environment for CC. In case it gets unmaintained its a simple delete of a folder that is just used by the github codespaces functionality.

Also this enables super simple start on developing in CC for most endpoints (excluding diego and loggregator functionality for now).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -102,4 +102,5 @@ group :development do
   gem 'spork', git: 'https://github.com/sporkrb/spork', ref: '224df49' # '~> 1.0rc'
   gem 'spring'
   gem 'spring-commands-rspec'
+  gem 'debug', '~> 1.8'
 end
--- a/.devcontainer/configs/intellij/.idea/inspectionProfiles/Default_CC_NG.xml
+++ b/.devcontainer/configs/intellij/.idea/inspectionProfiles/Default_CC_NG.xml
@@ -1,0 +1,10 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Default CC_NG" />
+    <inspection_tool class="Rubocop" enabled="true" level="WARNING" enabled_by_default="true">
+      <option name="myUseStandardGemIfPossible" value="true" />
+      <option name="myRunRubocopOnSave" value="true" />
+    </inspection_tool>
+    <inspection_tool class="RubyStringKeysInHashInspection" enabled="true" level="WEAK WARNING" enabled_by_default="true" />
+  </profile>
+</component>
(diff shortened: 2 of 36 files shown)


philippthun:
@FloThinksPi - Rebasing to the current `main` branch fixes mysql:8.0 tests.

Gemfile
```diff
@@ -101,4 +101,6 @@ group :development do
   gem 'spork', git: 'https://github.com/sporkrb/spork', ref: '224df49' # '~> 1.0rc'
   gem 'spring'
   gem 'spring-commands-rspec'
+  gem 'ruby-debug-ide', '~> 0.7.3'
```
moleske, line 104: We removed these in https://github.com/cloudfoundry/cloud_controller_ng/pull/3175 because

> We face an issue where, when bundling for the first time on a new env, the first attempt at installing ruby-debug-ide fails and bundling a second time magically works.

So might be worth checking in on that

moleske (commented):
No strong opinions about adding codespaces stuff, just one question/comment in the gemfile.  Will definitely be curious if people like the codespaces thing!

Gemfile
FloThinksPi: The thing is without this beeing in the gemfile VSCode refuses to work at all forcing everyone that uses that ide to temporarily add it to the gemfile anyway hmmm 

```
bundler: failed to load command: rdebug-ide (/Users/me/.gem/ruby/3.1.2/bin/rdebug-ide)
/Users/me/.gem/ruby/3.1.2/gems/bundler-2.4.10/lib/bundler/rubygems_integration.rb:308:in `block in replace_bin_path': can't find executable rdebug-ide for gem ruby-debug-ide. ruby-debug-ide is not currently included in the bundle, perhaps you meant to add it to your Gemfile? (Gem::Exception)
	from /Users/me/.gem/ruby/3.1.2/gems/bundler-2.4.10/lib/bundler/rubygems_integration.rb:336:in `block in replace_bin_path'
```

I havent had anny issue installing either ruby-debug-ide nor debase. So i would propose we try adding it to the dev dependencies again (i mean they are for real dev dependencies i would say) and see if theres still an issue. Maybe it was just a broken ruby-debug-ide version that wasnt properly installable.

FloThinksPi: I found a way to make vscode use `rdbg` so the default debug gem from ruby 3 👍🏼

kathap (approved):

Issue #1990, opened 2020-12-07
No application summary in v3 API


IvanBorislavovDimitrov:
## Issue

Missing application summary in v3 API

## Context

Hello folks,

We're migrating from v2 to v3 API and our code depends on the application summary resource in several places since everything about the application is required in certain scenarios and it is the fastest and also an elegant way to get the full information about an application with only 1 call (https://apidocs.cloudfoundry.org/15.2.0/apps/get_app_summary.html). If we try to gather this information by using the v3 API and executing separate calls for everything required like env, service instances, routes, domains, buildpacks, packages, etc. it gets much slower and generates quite a lot of calls which may easily result in exceeding the rate limits for some users. Is there a way to achieve the result returned from v2/app/GUID/summary by using the v3 API somehow?
I've checked the "include" parameter but it only supports: 'space', 'org', 'space.organization', which unfortunately is not enough. I've also noticed the "fields" parameters for service but it is only supported for: service instances, shared spaces, service offerings, and service plans. Is it possible for this parameter to be integrated with the apps? Is there a way to bring back the application summary response as it very optimal in certain cases?

Thanks and best regards,
Ivan

## Steps to Reproduce

Try to get the full information about an application with as few calls as possible (preferably 1).

## Expected result

Similar response as in the v2/apps/GUID/summary

## Current result

No way to get similar full information about an application

## Possible Fix

Probably the parameters "fields" may work for apps and may include more resources or "includes" may include more parameters. Or implementing a similar resource like the v2 apps summary?

Gerg:
It is unlikely that we will add an application summary endpoint on v3. That said, we are interested in letting clients build their own summary endpoints using tools like the include parameter. 

What resources would you like to include on applications?

IvanBorislavovDimitrov:
Hi, Gerg

Thanks for the fast reply!
I would like to include the following resources:
-> routes, as well as their respective domain, will be really useful
-> application's process(es) which contains information about: command, application's instances, memory, disk, health-check
-> application's process(es) statistics so we can calculate the running instances
-> application's package
-> application's environment
-> application's services (service name, service plan, last operation)
-> information if ssh is enabled
These resources would be really helpful if they could be included via the "include" parameter.

Thanks and regards,
Ivan

mvadu:
@Gerg any update on this one? I am also looking for getting the app->service dependencies, and other than bound app count (via usage summary) I am not able to find any way to find what service instances are used by app apps.

[...]

Issue #2879, opened 2022-07-22
Enhancement: Include[processes] into GET Apps/{guid} 


KevinJCross:
## Request
We in the [Cf autoscaler team](https://github.com/cloudfoundry/app-autoscaler-release) are migrating to only using the V3 api. Currently there is no direct replacement for the apps/summary endpoint (possibly for valid reasons). The new V3 api allows us to include related resources into the result of the call. 
To reduce our fan out of calls to the cloud controller we would like to include the `apps/{guid}/processes` and `processes/{guid}/stats` as relationships we can include using the V3 `include[relationship]` directives. This would go a long way to replicate the items we require from the summary page namely. 
app state from the `/apps/{guid}`
app instances from `/apps/{guid}/processes`  

the one call is now costing us minimum 2 calls for 2 distictively related items

Gerg:
We have been thinking about including *-to-many relationships for a while, but we never came up with a design we were super happy with. I think the closest we got was [this proposal](https://docs.google.com/document/d/1KxRRGAvE9FXPOA6GFRX5K17lwo5BTHmYzogOUJDrdBw/edit).

philippthun:
@KevinJCross We are in the process of extending this endpoint. See [this issue/comment](https://github.com/cloudfoundry/cloud_controller_ng/issues/1990#issuecomment-1183335580) for more details.

KevinJCross:
@philippthun Great this looks like it is being thought about and looked. I just added a comment to the document too.
Thanks ... ill consider this duplicated then

KevinJCross:
Im happy that this issue is already seen and currently under discussion.

PR #2893, opened 2022-08-01
Add includes of route, domain, process for show app endpoint

svkrieger:

Add includes for route, domain and process to the /v3/apps/:guid endpoint.

The V2 app summary endpoint (/v2/apps/:guid/summary) is no longer available in V3. For a detailed problem description see https://github.com/cloudfoundry/cloud_controller_ng/issues/1990

This PR implements the first step mentioned in the above issue and makes it easier to retrieve domains, routes and processes associated with an app.

### DIFF
--- a/app/controllers/v3/apps_controller.rb
+++ b/app/controllers/v3/apps_controller.rb
@@ -13,6 +13,9 @@ require 'actions/app_assign_droplet'
 require 'decorators/include_space_decorator'
 require 'decorators/include_organization_decorator'
 require 'decorators/include_space_organization_decorator'
+require 'decorators/include_app_route_decorator'
+require 'decorators/include_app_process_decorator'
+require 'decorators/include_app_domain_decorator'
 require 'messages/apps_list_message'
 require 'messages/app_show_message'
 require 'messages/app_update_message'
@@ -73,6 +76,9 @@ eager_loaded_associations: Presenters::V3::AppPresenter.associated_resources)
     decorators = []
     decorators << IncludeSpaceDecorator if IncludeSpaceDecorator.match?(message.include)
     decorators << IncludeOrganizationDecorator if IncludeOrganizationDecorator.match?(message.include)
+    decorators << IncludeAppRouteDecorator if IncludeAppRouteDecorator.match?(message.include)
+    decorators << IncludeAppProcessDecorator if IncludeAppProcessDecorator.match?(message.include)
+    decorators << IncludeAppDomainDecorator if IncludeAppDomainDecorator.match?(message.include)
 
     render status: :ok, json: Presenters::V3::AppPresenter.new(
       app,
--- a/app/controllers/v3/apps_controller.rb
+++ b/app/controllers/v3/apps_controller.rb
@@ -78,7 +78,7 @@ eager_loaded_associations: Presenters::V3::AppPresenter.associated_resources)
     decorators << IncludeOrganizationDecorator if IncludeOrganizationDecorator.match?(message.include)
     decorators << IncludeAppRouteDecorator if IncludeAppRouteDecorator.match?(message.include)
     decorators << IncludeAppProcessDecorator if IncludeAppProcessDecorator.match?(message.include)
-    decorators << IncludeAppDomainDecorator if IncludeAppDomainDecorator.match?(message.include)
+    decorators << IncludeAppDomainDecorator.new(presenter_args) if IncludeAppDomainDecorator.match?(message.include)
 
     render status: :ok, json: Presenters::V3::AppPresenter.new(
       app,
@@ -403,5 +403,13 @@ eager_loaded_associations: Presenters::V3::AppPresenter.associated_resources)
 
   def instances_reporters
     CloudController::DependencyLocator.instance.instances_reporters
+  end
+
+  def presenter_args
+    if permission_queryer.can_read_globally?
+      { all_orgs_visible: true }
+    else
+      { visible_org_guids: permission_queryer.readable_org_guids }
+    end
   end
 end
--- a/app/decorators/include_app_domain_decorator.rb
+++ b/app/decorators/include_app_domain_decorator.rb
@@ -7,11 +7,15 @@ module VCAP::CloudController
 
       def decorate(hash, apps)
         hash[:included] ||= {}
-        routes = apps.map(&:routes).flatten
-        domain_ids = routes.map(&:domain_id).uniq
 
-        domains = Domain.where(id: domain_ids).order(:created_at).
-          eager(Presenters::V3::DomainPresenter.associated_resources).all
+        app_guids = apps.map(&:guid)
+        domains = Domain.select_all(:domains).distinct.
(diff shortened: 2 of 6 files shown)


svkrieger:
One topic to discuss is how to handle large result sets of included resources. E.g  if the app has 1000 routes. 

An idea would be to set a limit for included resources e.g. if the app has up to 50 associated routes, the request will be processed and returns the routes. This could be either hardcoded or configurable.

If the app has more resources than allowed, the request would be rejected with a 403 - Forbidden with an according message that the result set is too large and that e.g. the /v3/apps/:guid/routes endpoint shall be used to retrieve the routes of the app, which supports pagination.

I understand the purpose of includes in general and also this PR to get an overview of an app's associated resources with a single API call. If we would add pagination to included resources, the client would have to make several calls, which is not different from using the already existing endpoints. So IMHO limiting the retrievable result sets in the includes seems appropriate. Also this should be fine for most apps and use cases.

I'm happy for comments and other suggestions.

sethboyles:
Curious to know if the use case for an app with 1000 routes is a real one you all have encountered.  If so, can you tell us a little about how apps use 1000 routes?

I agree that pagination on include statements would not be ideal.  Perhaps we could limit the number of returned routes and then provide a link to routes to get the rest of the records, already prepopulated with the pagination parameters.  I'm not sure a 403 is meant for such cases, and in general returning an error on a valid request like that seems a bit unorthodox.

@Gerg do you have any thoughts here?

@svkrieger I know you all were also thinking about an 'includes' for process stats--have you given any more thought as to how that might work?

Gerg:
Re-posting this from https://github.com/cloudfoundry/cloud_controller_ng/issues/2879

----

We have been thinking about including *-to-many relationships for a while, but we never came up with a design we were super happy with. I think the closest we got was [this proposal](https://docs.google.com/document/d/1KxRRGAvE9FXPOA6GFRX5K17lwo5BTHmYzogOUJDrdBw/edit).

Gerg:
Also, the thinking has generally been that include paths should correspond to relationship names on resources.

So, you wouldn't do something like `GET /v3/apps?include=organization`, you would do `GET /v3/apps?include=space.organization`, since apps don't have an `organization` relationship. This is important since resources can have multiple relationship paths to the same type of resource. For example, service instances (and soon routes) have an "owning" space, but can also be shared with other spaces.

For the routes/domains case here, destinations will probably complicate things. I wish I had a good idea on how to solve that, but I don't 😄.

Docs:
- https://github.com/cloudfoundry/cloud_controller_ng/blob/main/docs/v3_style_guide.md#including-related-resources
- https://v3-apidocs.cloudfoundry.org/version/release-candidate/#include

svkrieger:
@sethboyles
- On our biggest landscapes we can see apps with 300+ routes. This is due to multitenancy.
- As the [hard limit for routes per app](https://docs.cloudfoundry.org/devguide/deploy-apps/routes-domains.html#routes:~:text=The%20number%20of%20routes%20that%20can%20be%20mapped%20to%20each%20app%20is%20approximately%201000%20(128%C2%A0KB).) seems to be 1000 and the max_page_size is [hardcoded to 5000](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/cloud_controller/paging/pagination_options.rb#L10), we could consider to not implement any limit or pagination mechanism here.
- If we decide to implement a limit, your suggestion to show the first e.g. 50 and provide a link to the paginated result sounds reasonable.

@Gerg
- I think we don't want to go down the road of implementing *-to-many relationships. The requirement from users we have right now is to provide an endpoint which enables them to retrieve a single app and it's associated resources. Currently we also do not plan to implement this in the list endpoint, as this requires more thoughts on performance.
- Thanks for pointing me to the style guide regarding includes! Though that guideline seems to be breached already, as you can also retrieve the org of an app via `include=org`. But it makes sense to stick to `route.domain` instead of `domain`. The question would be, do we want to remove the `include=org`? 
Also at the moment when including `space.organization` both the space and the organization will be included. Is this on purpose? Or should it rather be like `space` includes the space and `space.organization` includes the org only and if you want both you'd need to specify `include=space,space.organization`?
- Destinations is a good point. I didn't check whether they are filtered for visibility yet. I guess a user should only be able to see the destinations, which the user has permissions for.
- Would `include=route,route.domain` work for you? I know that in the background the connection between app and route is made through destinations (DB table route_mappings), but not sure whether we need to add those here.

Gerg:
> I think we don't want to go down the road of implementing *-to-many relationships. The requirement from users we have right now is to provide an endpoint which enables them to retrieve a single app and it's associated resources. Currently we also do not plan to implement this in the list endpoint, as this requires more thoughts on performance.

Processes and Routes (and, by extension, Domains) are one-to-many relationships for Apps. The patterns we define for this endpoint should apply to other endpoints as well.

> ...Though that guideline seems to be breached already, as you can also retrieve the org of an app via `include=org`.... The question would be, do we want to remove the `include=org`? 

`GET /v3/apps/:guid?include=org` was a previous implementation that we continue to support for backwards-compatibility reasons. It is not documented in the API docs, and it is not recommended for use.

> Also at the moment when including `space.organization` both the space and the organization will be included. Is this on purpose? Or should it rather be like `space` includes the space and `space.organization` includes the org only and if you want both you'd need to specify `include=space,space.organization`?

We intentionally include space so that the included resources are "fully linked" to the root resource. This is primarily useful for list endpoints. Otherwise, you wouldn't be able to tell which apps are in which organizations. See [the relevant section of this proposal](https://docs.google.com/document/d/1vNPdobA21D_eLt73z5erDDDbd3ScSC3eCMX62Xi_lfo/edit#heading=h.dfz6o8v0298f).

> Would `include=route,route.domain` work for you? I know that in the background the connection between app and route is made through destinations (DB table route_mappings), but not sure whether we need to add those here.

Thinking about the list case (`GET /v3/apps?include=route`), how would clients associate routes back to the root apps? I suppose the app guids in the route's destinations would be serviceable, though maybe not ideal. See [the relevant section of this proposal](https://docs.google.com/document/d/1vNPdobA21D_eLt73z5erDDDbd3ScSC3eCMX62Xi_lfo/edit#heading=h.a6qxeg32c353).

svkrieger:
We tried to come up with a concept which would work for show and list endpoints as well. We liked the idea to use the already present pagination for included resources like when including spaces for orgs, the pagination link would look like `/v3/spaces?organization_guids=<list of org guids>&per_page=50&page=2`. Though, this can lead to long URLs when querying for a big amount of orgs. These lead to `URI too long` errors, which we could bypass to a certain amount with enlarging the buffer size in the nginx config.

Nevertheless we would drop the idea of implementing *-to-many relationships for now. We spoke to one of the biggest consumers of the API on one of our landscapes and came to a conclusion that additional calls to the API for fetching app-related data is ok. What they would like to see included in the `/v3/apps:guid` endpoint is the ssh_enabled feature flag as well as environment variables. These are viable options for us. When we got a draft PR ready it will be linked to https://github.com/cloudfoundry/cloud_controller_ng/issues/1990

PR #2900, opened 2022-08-03
Feat: log rate limits

ctlong:
Introduce log rate limits to org and space quotas, and tasks and processes.

https://github.com/cloudfoundry/capi-release#245

cc/ @sethboyles


mkocher:
@sethboyles we think this is ready to go! Can you please take a look?

sethboyles:
Awesome!  I'll go ahead and mark it as 'ready for review'

lib/diego/bbs/models/desired_lrp_pb.rb
```diff
@@ -15,6 +15,7 @@
 require 'image_layer_pb'
 require 'metric_tags_pb'
 require 'sidecar_pb'
+require 'log_rate_limit_pb'
```
moleske, line 18: This and the other `*_pb.rb` files are generated by this [job](https://ci.cake.capi.land/teams/main/pipelines/capi/jobs/bump-bbs-protos/builds/23) in our pipeline.  They will likely get overridden until a diego release is cut containing your [bbs changes](https://github.com/cloudfoundry/bbs/commit/3a12fde9c836a527f5f32d9edb402c9c36f28a17).  Is that a problem?

moleske (commented):
I haven't looked at this in detail but have some questions just from a 5 minute skim

lib/diego/bbs/models/desired_lrp_pb.rb
moleske: looking close at our pipeline, it only triggers on diego releases, so it is probably ok.  Unless diego cuts a release without your bbs changes, then it will be a sad time

db/migrations/20220606172913_add_log_rate_limit_to_quota_definitions.rb
```diff
@@ -0,0 +1,5 @@
+Sequel.migration do
```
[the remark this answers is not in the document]
acrmp: 👍 I pushed a [change](https://github.com/cloudfoundry/cloud_controller_ng/pull/2900/commits/e8afa9c73a51f7276fd0b8de1eaa96e6100304c5) to combine the migrations.

jdgonzaleza:
Hello! Please let the cli team when this PR since it's related with https://github.com/cloudfoundry/cli/pull/2303

ctlong:
Those last commits are not us continuing to develop on this PR - we just needed to correct some history that was lost.

Benjamintf1:
we added docs

ctlong:
@sethboyles I'd lean towards no, but defer to your process for introducing new features. The Diego work has been merged, and released in [v2.66.0](https://github.com/cloudfoundry/diego-release/releases/tag/v2.66.0).

sethboyles:
That's good enough for me. Thanks!

sethboyles:
Hey @ctlong ,

We are seeing a couple of errors with manifests.  

When supplying `-1` or `0` for `log-rate-limit-per-second`, we are getting 500s from the `manifest_diff` and `apply_manifest` endpoints.  It seems like `strip` is being called on the value, which errors because `0` and `-1` are integers, not strings.

The other issue we are seeing is if we set `disk_quota` to `0`.  Normally this errors with 

```
For application 'dora2': Process "web": Disk quota must use a supported unit: B, K, KB, M, MB, G, GB, T, or TB
FAILED
```

However, with this branch, we are seeing this error rendered in the `manifest_diff`:

```
Pushing app dora to org org / space space as admin...
Applying manifest file dora-manifest2.yml...

Updating with these attributes...
  ---
  applications:
  - name: dora
    processes:
-   - disk_quota: 1024M
+   - disk_quota: 'disk_quota must use a supported unit: B, K, KB, M, MB, G, GB, T, or TB'
      health-check-type: port
      instances: 1
      memory: 256M
      type: web
```

And then the error is not picked up by `apply_manifest` and then a deployment continues, despite having an invalid manifest.  Weirdly this only happens with `0` and not `-1` or other integers.  We double checked that this error didn't occur on envs with our latest release.

We looked at the code a little and are not sure how this odd error happened

acrmp:
@sethboyles Thanks! We pushed some more changes to address these issues.

sethboyles:
@acrmp thanks!  Confirmed that both issues are fixed.

Sending an int value like `9999` for `log-rate-limit-per-second` causes a 500.  This is the error:

```
undefined method `strip' for 9999:Integer
byte_converter.convert_to_b(human_readable_byte_value.strip)
```

I think coercing `human_readable_byte_value` to a string at the beginning of that method would protect against that.

(oh the joys of yaml...)

acrmp:
@sethboyles Thanks. I pushed a fix.

sethboyles:
Noticed a 500 on processes/scale when sending `null`: 

```
Mysql2::Error: Column 'log_rate_limit' cannot be null
```

sethboyles:
checked the other endpoints:
builds
tasks
space_quotas
org_quotas
--they all seemed fine 👍

duanemay:
@sethboyles While looking at the issue with sending `null` for `log_rate_limit_in_bytes_per_second` we also noticed an error with sending `null` for `memory_in_mb`

```cf curl /v3/apps/$(cf app dora --guid)/processes/web/actions/scale -X POST -d '{ "memory_in_mb": null}'```

sethboyles:
We noticed it with `instances`, too.  I think that is related to the memory since it tries to do some math with those two values.  I didn't think `null` is supposed to be valid for `instances`, not sure about `memory_in_mb`.  Is `null` valid in the `scale` endpoint for `log_rate_limit_in_bytes_per_second`?

sethboyles:
that is to say I'm not super concerned if it's not supposed to be valid.  If it's not, we can toss a generic story in the backlog to deal with all three of these fields if you'd rather get this PR merged sooner than later

duanemay:
@sethboyles We don't think null is valid in this case. But we have put in a fix so that the error is not produced in this case.

@acrmp & Duane

PR #2905, opened 2022-08-05
Add delete on cascade to fk_svc_inst_op_svc_instance_id

kathap:
We can see double entries for the same service_instance_id
in the service_instance_operations table. This leads to:
PG::ForeignKeyViolation: ERROR:  update or delete on table
"service_instances" violates foreign key constraint
"fk_svc_inst_op_svc_instance_id" on table "service_instance_operations"

For service bindings, service keys and route bindings the foreign
key constraint from operation to resource contains ON DELETE CASCADE;
for service instances this is missing.

Adding ON DELETE CASCADE to FK fk_svc_inst_op_svc_instance_id in
service_instance_operations

The root cause is fixed in this PR: #2916
#2929

### DIFF
--- a/db/migrations/20220805145100_add_delete_on_cascade_to_fk_svc_inst_op_svc_instance_id.rb
+++ b/db/migrations/20220805145100_add_delete_on_cascade_to_fk_svc_inst_op_svc_instance_id.rb
@@ -1,6 +1,6 @@
 Sequel.migration do
-  # Add delete cascade to fk_svc_inst_op_svc_instance_id, without there won't be cleaned up all ocurrencies in
-  # the service_instance_operations table
+  # Add DELETE CASCADE to foreign key fk_svc_inst_op_svc_instance_id to ensure that 'service instance operations'
+  # are deleted together with the referenced 'service instance'.
 
   up do
     alter_table :service_instance_operations do
--- a/db/migrations/20220805145100_add_delete_on_cascade_to_fk_svc_inst_op_svc_instance_id.rb
+++ b/db/migrations/20220805145100_add_delete_on_cascade_to_fk_svc_inst_op_svc_instance_id.rb
@@ -1,0 +1,18 @@
+Sequel.migration do
+  # Add delete cascade to fk_svc_inst_op_svc_instance_id, without there won't be cleaned up all ocurrencies in
+  # the service_instance_operations table
+
+  up do
+    alter_table :service_instance_operations do
+      drop_constraint :fk_svc_inst_op_svc_instance_id, type: :foreign_key
+      add_foreign_key [:service_instance_id], :service_instances, key: :id, name: :fk_svc_inst_op_svc_instance_id, on_delete: :cascade
+    end
+  end
+
+  down do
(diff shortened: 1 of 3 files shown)


db/migrations/20220805145100_add_delete_on_cascade_to_fk_svc_inst_op_svc_instance_id.rb
```diff
@@ -0,0 +1,18 @@
+Sequel.migration do
+  # Add delete cascade to fk_svc_inst_op_svc_instance_id, without there won't be cleaned up all ocurrencies in
```
philippthun, line 2: I find this sentence hard to read. What about `Add DELETE CASCADE to foreign key fk_svc_inst_op_svc_instance_id to ensure that 'service instance operations' are deleted together with the referenced 'service instance'.`?

spec/unit/models/services/service_instance_spec.rb
```diff
@@ -183,6 +183,31 @@ module VCAP::CloudController
         expect(ServiceInstanceAnnotationModel.where(id: annotation.id)).to be_empty
       end
 
+      it 'cascade deletes all ServiceInstanceOperations for this instance' do
```
philippthun, line 186: In theory there should only be a single operation - and this test also instantiates a single operation, thus I would change the text to `it 'cascade deletes the related ServiceInstanceOperation for this instance'`.

spec/unit/models/services/service_instance_spec.rb
```diff
@@ -183,6 +183,31 @@ module VCAP::CloudController
...
+        last_operation = ServiceInstanceOperation.make
+        service_instance.service_instance_operation = last_operation
+
+        service_instance.destroy
+
+        expect(ServiceInstance.find(guid: service_instance.guid)).to be_nil
+        expect(ServiceInstanceOperation.find(guid: last_operation.guid)).to be_nil
+      end
+
+      it 'cascades deletion of related dependencies' do
```
philippthun, line 196: I would remove this test as it is unrelated to the changes in this PR.

philippthun (commented):
The newly added test passes even without the migration. The reason for this is the following line in `app/models/services/service_instance.rb`:
```
add_association_dependencies service_instance_operation: :destroy
```

I would propose to remove this 'association dependency' (then the test should fail without the new migration). This would mean that service instance operations would be consistent with service binding operations, route binding operations and service key operations - they also do not have such an 'association dependency' and use DELETE CASCADE instead.

db/migrations/20220805145100_add_delete_on_cascade_to_fk_svc_inst_op_svc_instance_id.rb
kathap: ok, sounds good.

spec/unit/models/services/service_instance_spec.rb
kathap: ok

philippthun (approved):

PR #2915, opened 2022-08-09
Allow bound routes to be moved if the route sharing flag is enabled

MerricdeLauney:
This validation comes from v2 where we allowed users to move a routes
space as long as the route wasnt bound. This will change v2 behavior IF
the user enabled route_sharing. However Route sharing is a v3 feature
disabled by default so I belive this to be acceptable.

Authored-by: Merric de Launey <mdelauney@pivotal.io>

### DIFF
--- a/app/models/runtime/route.rb
+++ b/app/models/runtime/route.rb
@@ -167,7 +167,9 @@ module VCAP::CloudController
     end
 
     def validate_changed_space(new_space)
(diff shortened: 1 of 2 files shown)


moleske (approved):
This looks good to me.  Seems like a reasonable compromise with the V2 api.  We may want to revisit this if we change the default value of the feature flag

PR #2922, opened 2022-08-15
Docs: Add missing `guids` query parameter of list service instances

svkrieger:
Add missing `guids` query parameter of service instance list endpoint to docs


johha (approved):
LGTM

PR #2929, opened 2022-08-18
Add an unique index to service_instance_operations

kathap:
We can see double entries for the same service_instance_id
in the service_instance_operations table. This leads to:
PG::ForeignKeyViolation: ERROR: update or delete on table
"service_instances" violates foreign key constraint
"fk_svc_inst_op_svc_instance_id" on table "service_instance_operations".

Compared to service bindings, service keys and route bindings there is a UNIQUE index
defined for the "resource_id" field; for service instances the UNIQUE keyword is missing. An unique index would prevent to get duplicates, therefore:

* Add unique index on column service_instance_id to prevent double entries for the same
  service_instance
* Remove duplicate entries for one service_instance_id to prepare for adding a uniqueness constraint
* Drop old index before creating the unique one

#2916
#2905

### DIFF
--- a/db/migrations/20220818142407_add_unique_index_to_service_instance_operations_service_instance_id.rb
+++ b/db/migrations/20220818142407_add_unique_index_to_service_instance_operations_service_instance_id.rb
@@ -1,22 +1,77 @@
 Sequel.migration do
   up do
     # Remove duplicate service_instance_operations.service_instance_id to prepare for adding a uniqueness constraint
-    max_updated_at_group = self[:service_instance_operations].
-                           select(Sequel.function(:max, :updated_at)).
-                           group_by(:service_instance_id).
-                           having { count.function.* >= 1 }
-    service_instance_id_group = self[:service_instance_operations].
-                                select(:service_instance_id).
-                                group_by(:service_instance_id).
-                                having { count.function.* >= 1 }
-    self[:service_instance_operations].exclude(id: self[:service_instance_operations].select(:id).
-      where { Sequel.&(Sequel.|({ service_instance_id: service_instance_id_group }, { service_instance_id: nil }),
-                       { updated_at: max_updated_at_group })
-      }).each do |row|
-      self[:service_instance_operations].where(id: row[:id]).delete
+
+    dup_groups = self[:service_instance_operations].exclude(service_instance_id: nil).
+                 select(:service_instance_id).
+                 group_by(:service_instance_id).
+                 having { count.function.* > 1 }
+
+    dup_groups.each do |group|
+      sorted_ids = self[:service_instance_operations].
+                   select(:id).
+                   where(service_instance_id: group[:service_instance_id]).
+                   map(&:values).
+                   flatten.
+                   sort
+      ids_to_remove = sorted_ids
+
+      same_si_same_updated_at_take_max_id = self[:service_instance_operations].
+                                            select(Sequel.function(:max, :id)).
+                                            where(service_instance_id: group[:service_instance_id]).
+                                            group_by(:updated_at, :service_instance_id).
+                                            having { count.function.* > 1 }.
+                                            map(&:values).
+                                            flatten.
+                                            sort
+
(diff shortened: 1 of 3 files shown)


db/migrations/20220818142407_add_unique_index_to_service_instance_operations_service_instance_id.rb
```diff
@@ -0,0 +1,35 @@
+Sequel.migration do
+  up do
+    # Remove duplicate service_instance_operations.service_instance_id to prepare for adding a uniqueness constraint
+    max_updated_at_group = self[:service_instance_operations].
```
philippthun, line 4: I find this hard to understand and thus difficult to verify. So I've pushed a commit with a test for correctly removing duplicates. When executing this test, it fails.

Maybe a different approach would be easier. We don't expect many duplicates, so first finding those service_instance_ids for which there are duplicates and then (for each of them) finding the duplicates that should be removed (keep latest or the one with the highest id) might be easier to implement and understand.

db/migrations/20220818142407_add_unique_index_to_service_instance_operations_service_instance_id.rb
```diff
@@ -0,0 +1,90 @@
...
+    dup_groups = self[:service_instance_operations].exclude(service_instance_id: nil).
+                 select(:service_instance_id).
+                 group_by(:service_instance_id).
+                 having { count.function.* > 1 }
+
+    dup_groups.each do |group|
+      sorted_ids = self[:service_instance_operations].
+                   select(:id).
+                   where(service_instance_id: group[:service_instance_id]).
+                   map(&:values).
```
philippthun, line 14: `.select(:id).map(&:values)` -> `.select_map(:id)`

spec/migrations/20220818142407_add_unique_index_to_service_instance_operations_service_instance_id_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    VCAP::CloudController::ServiceInstanceOperation.instance_exec do
+      @allow_manual_timestamp_update = true
+    end
+
+    # Revert the given migration, i.e. remove the uniqueness constraint.
+    Sequel::Migrator.run(db, tmp_migrations_dir, target: 0, allow_missing_migration_files: true)
+  end
+
+  it 'removes duplicate service instance operations' do
+    # Two operations that do not reference a service instance (i.e. service_instance_id is nil);
```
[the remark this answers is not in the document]
kathap: operations with svc_inst_id = nil now don't get deleted at all

spec/unit/models/services/service_instance_operation_spec.rb
```diff
@@ -51,5 +51,37 @@ module VCAP::CloudController
...
+    describe 'when two are created with the same id' do
+      describe 'when a ServiceInstanceOperation exists' do
+        let(:service_instance_attrs) do
+          {
+            name: 'bommel_instance',
+            space: VCAP::CloudController::Space.make
+          }
+        end
+
+        let(:service_instance) { ServiceInstance.create(service_instance_attrs) }
```
philippthun, line 64: You can use `ServiceInstance.make` and remove the `service_instance_attrs`.

spec/unit/models/services/service_instance_operation_spec.rb
```diff
@@ -51,5 +51,37 @@ module VCAP::CloudController
...
+            state: 'in progress',
+            description: '50% all the time',
+            type: 'create',
+            proposed_changes: {
+              name: 'pizza',
+              service_plan_guid: '1800-pizza',
+            },
+          }
+        end
+        before { ServiceInstanceOperation.create(operation_attributes2) }
```
philippthun, line 77: You can use `ServiceInstanceOperation.make(service_instance_id: service_instance.id)` and remove `operation_attributes2`.

philippthun (approved):

Issue #2937, opened 2022-08-23
Service Broker locked in state `SYNCHRONIZING` after erroneous request


svkrieger:
Thanks for submitting an issue to `cloud_controller_ng`. We are always trying to improve! To help us, please fill out the following template.

## Issue

Sending a broker update including empty strings for `username` and `password` will cause the broker to be locked in state `SYNCHRONIZING`.

## Context

When updating a service broker with empty strings for `username` or `password` like so:
```
$ cf curl -X PATCH "/v3/service_brokers/08f7ad8a-aee8-4211-b605-3b4c88a874b1" -d '{"authentication": {"type":"basic", "credentials":{"username":"", "password":""}}}'
```

the async job will fail with a 500 - unknown error:
```
$ cf curl "/v3/jobs/8c88693f-0958-4ff1-ae5f-979b00ee8198"
{
  "guid": "8c88693f-0958-4ff1-ae5f-979b00ee8198",
  "created_at": "2022-08-22T08:55:07Z",
  "updated_at": "2022-08-22T08:55:07Z",
  "operation": "service_broker.update",
  "state": "FAILED",
  "errors": [
    {
      "title": "UnknownError",
      "detail": "An unknown error occurred.",
      "code": 10001
    }
  ],
  "warnings": [],
  "links": {
    "self": {
      "href": "https://api.cf.dev22.aws.bndl.sapcloud.io/v3/jobs/8c88693f-0958-4ff1-ae5f-979b00ee8198"
    },
    "service_brokers": {
      "href": "https://api.cf.dev22.aws.bndl.sapcloud.io/v3/service_brokers/08f7ad8a-aee8-4211-b605-3b4c88a874b1"
    }
  }
}
```
The service broker stays in state `SYNCHRONIZING`:
```
-[ RECORD 2 ]---------+-------------------------------------------------------
id                    | 2
guid                  | 08f7ad8a-aee8-4211-b605-3b4c88a874b1
created_at            | 2022-08-22 08:42:36.700996
updated_at            | 2022-08-22 08:55:07.550528
name                  | ta-sb
broker_url            | https://at-sb.cfapps.dev22.aws.bndl.sapcloud.io
auth_password         | PLbOfqnbHa/Cb/zmAegTHg==
salt                  | f88f636820c17ac2
auth_username         | admin
space_id              |
encryption_key_label  | encryption_key_0
encryption_iterations | 2048
state                 | SYNCHRONIZING
```
[...]

PR #2939, opened 2022-08-23
Add validation for credentials in SB update to be not blank

svkrieger:

This PR consists of two fixes.
One is adding a validation check to make sure that credentials, which have been sent with a SB update request, are not empty strings. From now on `username` and `password` must be non-empty strings, otherwise the validation will fail. This coding is executed by the API processes.

The other ensures that after a model layer validation error, the SB can be reset to its previous state. This coding is executed by the CC worker processes and is run during the delayed job execution of the broker.update job.

The issue is explained in #2937.

Registering a broker in CF with empty credentials never worked. In V2 and V3 an error will be returned. So it is okay to make non-empty credentials obligatory. The OSBAPI Spec says "While the communication between a Platform and Service Broker MAY be unsecure, it is RECOMMENDED that all communications between a Platform and a Service Broker are secured via TLS and authenticated."([source](https://github.com/openservicebrokerapi/servicebroker/blob/v2.15/spec.md#platform-to-service-broker-authentication:~:text=While%20the%20communication%20between%20a%20Platform%20and%20Service%20Broker%20MAY%20be%20unsecure%2C%20it%20is%20RECOMMENDED%20that%20all%20communications%20between%20a%20Platform%20and%20a%20Service%20Broker%20are%20secured%20via%20TLS%20and%20authenticated.))

### DIFF
--- a/app/jobs/v3/services/update_broker_job.rb
+++ b/app/jobs/v3/services/update_broker_job.rb
@@ -53,6 +53,8 @@ module VCAP::CloudController
         end
 
         def perform
+          raise CloudController::Errors::V3::ApiError.new_from_details('ServiceBrokerGone') if broker.nil?
+
           ServiceBroker.db.transaction do
             broker.update(update_params)
 
@@ -64,11 +66,7 @@ module VCAP::CloudController
 
           @warnings
         rescue => e
-          begin
-            broker.update(state: previous_broker_state)
-          rescue
-            raise CloudController::Errors::V3::ApiError.new_from_details('ServiceBrokerGone') if broker.nil?
-          end
+          ServiceBroker.where(id: update_request.service_broker_id).update(state: previous_broker_state)
 
           if e.is_a?(Sequel::ValidationFailed)
             raise V3::ServiceBrokerUpdate::InvalidServiceBroker.new(e.message)
--- a/spec/unit/jobs/v3/services/update_broker_job_spec.rb
+++ b/spec/unit/jobs/v3/services/update_broker_job_spec.rb
@@ -182,8 +182,8 @@ module VCAP
           describe 'when there is a model level validation error' do
             before do
               allow_any_instance_of(ServiceBroker).to receive(:validate) do |record|
-                # Make setting the state to 'AVAILABLE' fail, but nothing else
-                if record.state == 'AVAILABLE'
+                # Make setting the name to 'new-name' fail, but nothing else
+                if record.name == 'new-name'
                   record.errors.add(:something, 'is not right')
                 end
               end
@@ -198,6 +198,31 @@ module VCAP
 
               broker.reload
               expect(broker.state).to eq(ServiceBrokerStateEnum::DELETE_FAILED)
+            end
+          end
+
+          describe 'when there is a model level validation error i.e. blank credentials' do
+            let(:update_broker_request) do
+              ServiceBrokerUpdateRequest.create(
+                authentication: '{"credentials":{"username":"","password":""}}',
+                service_broker_id: broker.id
+              )
+            end
+
+            it 'fails to update and raises InvalidServiceBroker' do
+              expect {
+                job.perform
+              }.to raise_error(V3::ServiceBrokerUpdate::InvalidServiceBroker) { |error|
+                expect(error.message).to include('auth_username presence')
+                expect(error.message).to include('auth_password presence')
+              }
+
+              broker.reload
+              expect(broker.name).to eq('test-broker')
+              expect(broker.broker_url).to eq('http://example.org/broker-url')
+              expect(broker.auth_username).to eq('username')
+              expect(broker.auth_password).to eq('password')
+              expect(broker.state).to eq(ServiceBrokerStateEnum::AVAILABLE)
             end
           end
 
--- a/spec/unit/messages/service_broker_create_message_spec.rb
+++ b/spec/unit/messages/service_broker_create_message_spec.rb
@@ -119,18 +119,37 @@ module VCAP::CloudController
 
         context 'when authentication is not valid' do
           let(:request_body) do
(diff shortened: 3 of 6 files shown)


spec/unit/messages/service_broker_create_message_spec.rb
```diff
@@ -119,18 +119,37 @@ module VCAP::CloudController
...
           let(:request_body) do
+            valid_body[:authentication] = valid_body[:authentication].except(:type)
             valid_body
           end
 
-          before do
-            allow_any_instance_of(VCAP::CloudController::Validators::AuthenticationValidator).to receive(:validate) do |_, record|
-              record.errors.add(:authentication, 'this authentication is not valid!')
-            end
+          it 'is not valid' do
```
svkrieger, line 126: This is just a fix of a test which was wrong.
Context is 'when authentication is not valid' but the test ran with a valid_body. It was just green because the passed body didn't matter, because the validation was mocked. I changed the body to include an invalid `authentication` object by removing the mandatory `type` field. Also I removed the mock and checked for the correct error message, which is thrown by the validation in this case.

spec/unit/messages/service_broker_update_message_spec.rb
```diff
@@ -149,18 +149,37 @@ module VCAP::CloudController
...
           end
 
-          before do
-            allow_any_instance_of(VCAP::CloudController::Validators::AuthenticationValidator).to receive(:validate) do |_, record|
-              record.errors.add(:authentication, 'this authentication is not valid!')
-            end
+          it 'is not valid' do
+            expect(message).not_to be_valid
+            expect(message.errors_on(:authentication)).to include('authentication.type must be one of ["basic"]')
+          end
```
svkrieger, line 159: Same fix as in the other test file.

spec/unit/jobs/v3/services/update_broker_job_spec.rb
```diff
@@ -182,8 +182,8 @@ module V3
           describe 'when there is a model level validation error' do
             before do
               allow_any_instance_of(ServiceBroker).to receive(:validate) do |record|
-                # Make setting the state to 'AVAILABLE' fail, but nothing else
-                if record.state == 'AVAILABLE'
+                # Make setting the name to 'new-name' fail, but nothing else
+                if record.name == 'new-name'
```
svkrieger, line 186: This test unfortunately tested the only case in which the state of the SB could be reset to its previous state after a failed model validation. This is because in the rescue call, the broker's state will be explicitly set to the previous state and will overwrite the state `AVAILABLE` which caused the model violation to fail.
If the cause for the model violation is something else, e.g. a name, then the test would fail, because the broker object would still have the name `new-name` which caused the model violation. This leads to a failed broker.update which is meant to reset the brokers state to its previous state.

app/jobs/v3/services/update_broker_job.rb
```diff
@@ -65,7 +68,7 @@ def perform
           @warnings
         rescue => e
           begin
-            broker.update(state: previous_broker_state)
+            initial_broker.update(state: previous_broker_state)
```
svkrieger, line 71: This PR fixes the issue, that the previous state could not be reset. Though I think it does not always make sense to let the broker return to the previous state.
For example when `@catalog_updater.refresh` fails to contact the broker to update the catalog and the previous_state was `AVAILABLE` the broker would return to state `AVAILABLE` even it is not reachable at the moment. I think in this case setting the brokers state to `SYNCHRONIZATION_FAILED` would be more appropriate. This still allows for broker.update request, but will let all service related operations fail fast.

spec/unit/jobs/v3/services/update_broker_job_spec.rb
```diff
@@ -201,6 +201,30 @@ module V3
             end
           end
 
+          describe 'when there is a model level validation error i.e. blank credentials' do
+            it 'fails to update and raises InvalidServiceBroker' do
+              update_broker_request = ServiceBrokerUpdateRequest.create(
+                authentication: '{"credentials":{"username":"","password":""}}',
+                service_broker_id: broker.id
+              )
+              job = UpdateBrokerJob.new(update_broker_request.guid, broker.guid, previous_state, user_audit_info: user_audit_info)
```
philippthun, line 210: You should use `let(:update_broker_request)` in this context (describe); then the line `job = UpdateBrokerJob.new...` can be omitted.

I see that it is done wrong throughout the whole file, but we should not stick to this bad pattern.

app/jobs/v3/services/update_broker_job.rb
philippthun, line 71: First I thought that `broker.reload.update...` could be an alternative to using `dup`; but this would result in an additional `SELECT`.

I think the best way to set the previous state would be:
```
ServiceBroker.where(id: update_request.service_broker_id).update(state: previous_broker_state)
```

spec/unit/jobs/v3/services/update_broker_job_spec.rb
svkrieger: True, that looks better

app/jobs/v3/services/update_broker_job.rb
svkrieger: Jep, I also thought about broker.reload first, but wanted to prevent the additional select. Indeed the combined statement is better than using dup.

app/jobs/v3/services/update_broker_job.rb
```diff
@@ -65,7 +68,7 @@ def perform
...
           rescue
             raise CloudController::Errors::V3::ApiError.new_from_details('ServiceBrokerGone') if broker.nil?
```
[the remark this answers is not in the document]
svkrieger: Right, when we rewrite the second broker.update, we need to restructure it anyways, otherwise the unit test fails, which tests the case that the broker got deleted meanwhile.

philippthun (approved):

PR #2946, opened 2022-08-30
Service instance permissions

will-gant:

Adds a `/v3/service_instance/:guid/permissions` endpoint that enables service_instance dashboard functionality with the V3 API

As a service developer there is a functionality called service dashboards (https://docs.cloudfoundry.org/services/dashboard-sso.html#checking-user-permissions). Users can manage their service instances via these dashboards, which call the CF API to check if a user is allowed to alter or read a given service instance. This can currently only be done via `/v2/service_instance/:guid/permissions`.

A corresponding V3 endpoint does not exist, which means this functionality won't work when V2 is disabled.

This PR adds an equivalent endpoint to V3.

See also:
https://cloudfoundry.slack.com/archives/C07C04W4Q/p1657878898740599
https://github.com/cloudfoundry/cf-acceptance-tests/issues/565

cloudfoundry/cf-acceptance-tests#593

### DIFF
--- a/app/controllers/v3/application_controller.rb
+++ b/app/controllers/v3/application_controller.rb
@@ -163,16 +163,20 @@ class ApplicationController < ActionController::Base
     !READ_SCOPE_HTTP_METHODS.include?(request.method)
   end
 
-  def check_read_permissions!
-    read_scope = SecurityContext.scopes.include?('cloud_controller.read')
-    admin_read_only_scope = SecurityContext.scopes.include?('cloud_controller.admin_read_only')
-    global_auditor_scope = SecurityContext.scopes.include?('cloud_controller.global_auditor')
+  def read_scope
+    roles.cloud_controller_reader?
+  end
 
-    raise CloudController::Errors::ApiError.new_from_details('NotAuthorized') if !roles.admin? && !read_scope && !admin_read_only_scope && !global_auditor_scope
+  def write_scope
+    roles.cloud_controller_writer?
+  end
+
+  def check_read_permissions!
+    raise CloudController::Errors::ApiError.new_from_details('NotAuthorized') if !roles.admin? && !roles.admin_read_only? && !roles.global_auditor? && !read_scope
   end
 
   def check_write_permissions!
-    write_scope = SecurityContext.scopes.include?('cloud_controller.write')
+    roles.cloud_controller_writer?
     raise CloudController::Errors::ApiError.new_from_details('NotAuthorized') if !roles.admin? && !write_scope
   end
 
--- a/app/controllers/v3/application_controller.rb
+++ b/app/controllers/v3/application_controller.rb
@@ -176,7 +176,6 @@ class ApplicationController < ActionController::Base
   end
 
   def check_write_permissions!
-    roles.cloud_controller_writer?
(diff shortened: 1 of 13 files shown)


spec/request/service_instances_spec.rb
```diff
@@ -4044,6 +4046,95 @@ def create_bindings(instance, space:, count:)
...
+
+      before do
+        org.add_user(user)
+        space.add_developer(user)
+      end
+
+      it 'returns a 200' do
+        get "/v3/service_instances/#{guid}/permissions", nil, user_headers
+        expect(last_response.status).to eq(200)
+        expect(JSON.parse(last_response.body)).to eq({ 'manage' => true, 'read' => true })
```
FloThinksPi, line 4096: Can we make use of `read_and_write` variable here ?

FloThinksPi (approved):
LGTM, maybe squash it. My commit is more than useless here 😄.
I guess doc changes are part of another PR then :)

spec/request/service_instances_spec.rb
will-gant: Yep, good spot

will-gant:
> LGTM, maybe squash it. My commit is more than useless here 😄. I guess doc changes are part of another PR then :)

Nope, just meant to do it and forgot! Thanks for the reminder. I'll make a separate PR to tweak [the docs on docs.cloudfoundry.org](https://docs.cloudfoundry.org/services/dashboard-sso.html) to replace (some of the) references to the v2 endpoint.

app/controllers/v3/service_instances_controller.rb
```diff
@@ -217,6 +217,16 @@ def parameters
     end
   end
 
+  def show_permissions
+    service_instance = ServiceInstance.first(guid: hashed_params[:guid])
+    service_instance_not_found! unless service_instance
+
+    render status: :ok, json: {
+      manage: is_space_active?(service_instance.space) ? can_write_to_active_space?(service_instance.space) : permission_queryer.can_write_globally?,
```
philippthun, line 225: There is a helper method `admin?` defined below that can be used instead of `permission_queryer.can_write_globally?`.

app/controllers/v3/application_controller.rb
```diff
@@ -163,16 +163,20 @@ def enforce_write_scope?
...
+    roles.cloud_controller_writer?
+  end
+
+  def check_read_permissions!
+    raise CloudController::Errors::ApiError.new_from_details('NotAuthorized') if !roles.admin? && !roles.admin_read_only? && !roles.global_auditor? && !read_scope
   end
 
   def check_write_permissions!
-    write_scope = SecurityContext.scopes.include?('cloud_controller.write')
+    roles.cloud_controller_writer?
```
philippthun, line 179: This line is not needed.

will-gant: Ah yeah good point. I would've had every Cloud Controller in the world double-checking permissions on every POST, PUT, etc!

app/controllers/v3/service_instances_controller.rb
will-gant: Yeah fair enough

philippthun (approved):

PR #2948, opened 2022-09-01
Prevent quota updates or assignment with finite log rates

acrmp:
This PR builds on #2900 to prevent updates to or assignment of org and space quotas with a finite log rate limit where they would apply to orgs or spaces that contain processes with unlimited log rate limits.

- If an existing deployment is upgraded to gain log rate limiting support then existing application processes will default to -1 (unlimited).
- If an org or space quota is subsequently updated to include a finite log rate limit then processes in the affected orgs or spaces will be unable to restart without having their log rate limit changed.
- Block org and space quota finite log rate limit updates until processes under the affected orgs or spaces have been updated with a finite log rate limit. Specify the names of the first two orgs or spaces and a total count in the error.
- Block assignment of org or space quotas with finite log rate limits to orgs or spaces that contain processes with unlimited log rate limits.

### DIFF
--- a/spec/request/organization_quotas_spec.rb
+++ b/spec/request/organization_quotas_spec.rb
@@ -462,6 +462,20 @@ module VCAP::CloudController
           expect(last_response).to include_error_message("Organization Quota '#{organization_quota.name}' already exists.")
         end
       end
+
+      context 'when trying to set a log rate limit and there are apps with unlimited log rates' do
+        let!(:app_model) { VCAP::CloudController::AppModel.make(name: 'name1', space: space) }
+        let!(:process_model) { VCAP::CloudController::ProcessModel.make(app: app_model, log_rate_limit: -1) }
+
+        it 'returns 422' do
+          patch "/v3/organization_quotas/#{organization_quota.guid}", params.to_json, admin_header
+
+          expect(last_response).to have_status_code(422)
+          expect(last_response).to include_error_message(
+            'Current usage exceeds new quota values. ' \
+            "Org '#{org.name}' assigned this quota contains apps running with an unlimited log rate limit.")
+        end
+      end
     end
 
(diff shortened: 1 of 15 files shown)


spec/request/organization_quotas_spec.rb
```diff
@@ -462,6 +462,20 @@ module VCAP::CloudController
...
+        let!(:app_model) { VCAP::CloudController::AppModel.make(name: 'name1', space: space) }
+        let!(:process_model) { VCAP::CloudController::ProcessModel.make(app: app_model, log_rate_limit: -1) }
+
+        it 'returns 422' do
+          patch "/v3/organization_quotas/#{organization_quota.guid}", params.to_json, admin_header
+
+          expect(last_response).to have_status_code(422)
+          expect(last_response).to include_error_message(
+            'Current usage exceeds new quota values. ' \
+            "Org '#{org.name}' assigned this quota contains apps running with an unlimited log rate limit.")
```
MerricdeLauney, line 476: This error message is upon first read. Maybe adding commas would help?

acrmp: I'm not sure. We did struggle a little with the wording here (and varying the messages based on the number of affected orgs).

Can you provide an alternative error message that you think is less confusing?

MerricdeLauney: I agree theres no good way to word it. Personally I find one of the following _slightly_ easier to understand.

"Org '#{org.name},' assigned this quota, contains apps running with an unlimited log rate limit."
or
"This quota is applied to org '#{org.name}' which contains apps running with an unlimited log rate limit."

acrmp: @MerricdeLauney Thanks. I pushed a commit to reword following the second option you suggested.

PR #2965, opened 2022-09-16
Puma thin refactor

sethboyles:
This PR enables switching to using Puma webserver (experimental) via a configuration option.

Metrics, particularly request metrics and vitals metrics (e.g. cpu, memory) don't necessarily behave in helpful ways, and we will probably need to rethink what metrics we supply in order to help inform scaling decisions.  But we hope this PR will enabled some initial explorations into performance testing and scaling guidelines.

https://github.com/cloudfoundry/capi-release/pull/269


lib/cloud_controller/runners/puma_runner.rb
```diff
@@ -0,0 +1,51 @@
...
+        # this is actually called everytime a worker is started
+        # https://github.com/puma/puma/blob/5f3f489ee867317c47724d0fc5b1d906f1b23de6/lib/puma/dsl.rb#L607
+        # we probably want to come up with a different way to do this.  Perhaps a singleton?
+        conf.after_worker_fork {
+          Thread.new do
+            EM.run do
+              @periodic_updater.setup_updates
+            end
+          end
+        }
```
Gerg, line 21: Why do we still need to do this? I thought switching from statsd to Prometheus meant we don't need to periodically emit metrics?

sethboyles: Neither Prometheus nor statsd style metrics are going to work well with Puma clusters yet.  A lot of the current metrics aren't set up to consider multiple processes.  

Prometheus metrics are going to require us to figure out a caching system that the processes will share if we want to emit metrics via a Prometheus endpoint, since each process tracks metrics and you can't guarantee which process will respond to the prometheus query.

The statsd_injector acts as a shared memory store, so we thought that it might be a better option for now, but after some testing, we found there is no simple way of migrating our existing metrics to a format that works for multiple processes.

Our current thinking is to either use the imperfect statsd metrics for now, or just remove the metrics entirely until we are able to do a thorough rethinking of what metrics will be valuable for a Puma cluster deployment of CCNG--some of the metrics we have right now were designed to help operators scale CCNG on Thin, for example, and aren't necessarily useful for a Puma deployment.

PR #2967, opened 2022-09-19
Remove redundant locks and transactions

philippthun:
- When locking an instance, there is no need to explicitly lock its corresponding operation (dependent 1:1 association).
- As `save_with_new_operation` (`ServiceInstance`) now locks the instance and updates itself and the associated operation in a transaction (see [1]), there is no need to wrap it into another transaction in the calling code.

[1] https://github.com/cloudfoundry/cloud_controller_ng/pull/2916


svkrieger (approved):
Good idea to remove the redundant parts here!

PR #2971, opened 2022-09-20
Optimisations for /v3/roles API calls

samcolson4:
This PR optimizes the speed of requests to `/v3/roles` by:
- Removing unnecessary checks when creating space roles, reducing the number of expensive DB queries.
- Avoiding a number of expensive DB queries that were run on all `create` calls by where a user guid was provided, which were aimed at determining whether the assignee user was in the same org as the user trying to assign them a role.
- Retrieving the `org_guid` used in destroying a role more efficiently.

The second commit ([ed19bdbe](https://github.com/sap-contributions/cloud_controller_ng/commit/ed19bdbe7fbfbc27936ba7a0f192bd58cc238c47)) ends an inconsistency whereby org managers received a 201 if they assigned an org role to an unaffiliated user by providing that user's username, but a 422 CF-UnprocessableEntity if they tried to do the same by providing the user's guid. See an example below of the previous behaviour (now corrected).

Org manager setting an org role on a user outside the organizaition when providing the user's guid:

```
$ cf curl -X POST /v3/roles -d '{
>       "type": "organization_auditor",
>       "relationships": {
>         "user": {
>           "data": {
>             "guid": "6823285a-51d3-47de-989b-51d6b9bd834f"
>           }
>         },
>         "organization": {
>           "data": {
>             "guid": "550d3635-3ae8-481e-ad2a-c150ec45d910"
>           }
>         }
>       }
>     }'
{
  "errors": [
    {
      "detail": "Invalid user. Ensure that the user exists and you have access to it.",
      "title": "CF-UnprocessableEntity",
      "code": 10008
    }
  ]
}
```

Org manager setting the same org role on the same user, still outside their org, by providing their username:
```
$ cf curl -X POST /v3/roles -d '{
>       "type": "organization_auditor",
>       "relationships": {
>         "user": {
>           "data": {
>             "username": "lame"
>           }
>         },
>         "organization": {
>           "data": {
>             "guid": "550d3635-3ae8-481e-ad2a-c150ec45d910"
>           }
>         }
>       }
>     }'
{
  "guid": "277ff39c-d8ac-4c1e-a118-32cee054e882",
  "created_at": "2022-09-15T13:54:32Z",
  "updated_at": "2022-09-15T13:54:32Z",
  "type": "organization_auditor",
  "relationships": {
    "user": {
      "data": {
        "guid": "6823285a-51d3-47de-989b-51d6b9bd834f"
      }
    },
    "organization": {
      "data": {
        "guid": "550d3635-3ae8-481e-ad2a-c150ec45d910"
      }
    },
    "space": {
      "data": null
    }
  },
  "links": {
    "self": {
      "href": "https://api.cf.azure-dev24.azure.cfp.sapcloud.io/v3/roles/277ff39c-d8ac-4c1e-a118-32cee054e882"
    },
    "user": {
      "href": "https://api.cf.azure-dev24.azure.cfp.sapcloud.io/v3/users/6823285a-51d3-47de-989b-51d6b9bd834f"
    },
    "organization": {
      "href": "https://api.cf.azure-dev24.azure.cfp.sapcloud.io/v3/organizations/550d3635-3ae8-481e-ad2a-c150ec45d910"
    }
  }
}
```

### DIFF
--- a/app/controllers/v3/roles_controller.rb
+++ b/app/controllers/v3/roles_controller.rb
@@ -116,7 +116,6 @@ class RolesController < ApplicationController
     user_guid = message.user_guid || lookup_user_guid_in_uaa(message.username, message.user_origin)
 
     user = User.first(guid: user_guid) || create_cc_user(user_guid)
-    unprocessable_user! unless user
 
     RoleCreate.new(message, user_audit_info).create_organization_role(
       type: message.type,
@@ -167,10 +166,6 @@ class RolesController < ApplicationController
 
   def unprocessable_organization!
     unprocessable!('Invalid organization. Ensure that the organization exists and you have access to it.')
-  end
-
-  def unprocessable_user!
-    unprocessable!('Invalid user. Ensure that the user exists and you have access to it.')
   end
 
   def unprocessable_space_user!
--- a/app/controllers/v3/roles_controller.rb
+++ b/app/controllers/v3/roles_controller.rb
@@ -114,29 +114,16 @@ class RolesController < ApplicationController
     unauthorized! unless permission_queryer.can_write_to_active_org?(message.organization_guid)
     suspended! unless permission_queryer.is_org_active?(message.organization_guid)
 
-    if message.user_guid
-      user_guid = message.user_guid
-      unprocessable_user! if user_in_db?(user_guid) && !user_is_readable?(user_guid)
-    else
-      user_guid = lookup_user_guid_in_uaa(message.username, message.user_origin)
-      unprocessable_user! unless user_guid
-    end
+    user_guid = message.user_guid || lookup_user_guid_in_uaa(message.username, message.user_origin)
 
     user = User.first(guid: user_guid) || create_cc_user(user_guid)
+    unprocessable_user! unless user
 
     RoleCreate.new(message, user_audit_info).create_organization_role(
       type: message.type,
       user: user,
       organization: org
     )
-  end
-
-  def user_in_db?(user_guid)
-    !User.where(guid: user_guid).empty?
-  end
-
-  def user_is_readable?(user_guid)
-    !readable_users.where(guid: user_guid).empty?
   end
 
   def fetch_readable_user(user_guid)
--- a/app/controllers/v3/roles_controller.rb
+++ b/app/controllers/v3/roles_controller.rb
@@ -94,9 +94,6 @@ class RolesController < ApplicationController
     unprocessable_space! unless space
     org = space.organization
 
-    unprocessable_space! if permission_queryer.can_read_from_org?(org.guid) &&
-      !permission_queryer.can_read_from_space?(message.space_guid, org.guid)
-
     unauthorized! unless permission_queryer.can_update_active_space?(message.space_guid, org.guid)
     suspended! unless permission_queryer.is_space_active?(message.space_guid)
 
--- a/app/controllers/v3/roles_controller.rb
+++ b/app/controllers/v3/roles_controller.rb
@@ -65,7 +65,8 @@ class RolesController < ApplicationController
     resource_not_found!(:role) unless role
 
     if role.for_space?
-      org_guid = Space.find(guid: role.space_guid).organization.guid
+      org_guid = Organization.join(:spaces, organization_id: :id).select(:organizations__guid).where(spaces__guid: role.space_guid)
+
       unauthorized! unless permission_queryer.can_update_active_space?(role.space_guid, org_guid)
       suspended! unless permission_queryer.is_space_active?(role.space_guid)
     else
--- a/app/controllers/v3/roles_controller.rb
+++ b/app/controllers/v3/roles_controller.rb
@@ -65,9 +65,7 @@ class RolesController < ApplicationController
     resource_not_found!(:role) unless role
 
     if role.for_space?
-      org_guid = Organization.join(:spaces, organization_id: :id).select(:organizations__guid).where(spaces__guid: role.space_guid)
-
-      unauthorized! unless permission_queryer.can_update_active_space?(role.space_guid, org_guid)
+      unauthorized! unless permission_queryer.can_update_active_space?(role.space_guid, role.organization_guid)
       suspended! unless permission_queryer.is_space_active?(role.space_guid)
     else
       unauthorized! unless permission_queryer.can_write_to_active_org?(role.organization_guid)
--- a/app/decorators/include_role_organization_decorator.rb
+++ b/app/decorators/include_role_organization_decorator.rb
@@ -9,7 +9,7 @@ module VCAP::CloudController
 
       def decorate(hash, roles)
         hash[:included] ||= {}
-        organization_guids = roles.map(&:organization_guid).uniq
+        organization_guids = roles.map { |role| role.organization_guid if !role.for_space? }.uniq
         organizations = Organization.where(guid: organization_guids).
                         order(:created_at).eager(Presenters::V3::OrganizationPresenter.associated_resources).all
 
--- a/app/decorators/include_role_organization_decorator.rb
+++ b/app/decorators/include_role_organization_decorator.rb
@@ -9,7 +9,7 @@ module VCAP::CloudController
 
       def decorate(hash, roles)
         hash[:included] ||= {}
-        organization_guids = roles.map { |role| role.organization_guid if !role.for_space? }.uniq
+        organization_guids = roles.map { |role| role.organization_guid unless role.for_space? }.uniq
         organizations = Organization.where(guid: organization_guids).
                         order(:created_at).eager(Presenters::V3::OrganizationPresenter.associated_resources).all
 
--- a/app/models/runtime/role.rb
+++ b/app/models/runtime/role.rb
@@ -103,11 +103,13 @@ module VCAP::CloudController
     end
 
     def organization_guid
-      organization&.guid
+      return organization.guid unless organization_id == SPACE_OR_ORGANIZATION_NOT_SPECIFIED
+
+      space.organization_guid
     end
 
     def space_guid
-      space&.guid
+      space.guid unless space_id == SPACE_OR_ORGANIZATION_NOT_SPECIFIED
     end
 
     def for_space?
--- a/spec/request/roles_spec.rb
+++ b/spec/request/roles_spec.rb
@@ -79,8 +79,6 @@ RSpec.describe 'Roles Request' do
           code: 201,
           response_object: expected_response
         }
-        h['org_auditor'] = { code: 422 }
-        h['org_billing_manager'] = { code: 422 }
         h
       end
 
@@ -332,8 +330,6 @@ RSpec.describe 'Roles Request' do
             code: 201,
             response_object: expected_response
           }
-          h['org_auditor'] = { code: 422 }
-          h['org_billing_manager'] = { code: 422 }
           h
         end
 
@@ -525,8 +521,6 @@ RSpec.describe 'Roles Request' do
           code: 201,
           response_object: expected_response
         }
-        h['org_auditor'] = { code: 422 }
-        h['org_billing_manager'] = { code: 422 }
         h
       end
 
@@ -567,8 +561,6 @@ RSpec.describe 'Roles Request' do
             response_object: expected_response
           }
 
-          h['org_auditor'] = { code: 422 }
-          h['org_billing_manager'] = { code: 422 }
           h
         end
         it_behaves_like 'permissions for single object endpoint', ALL_PERMISSIONS
--- a/spec/request/roles_spec.rb
+++ b/spec/request/roles_spec.rb
@@ -271,6 +271,62 @@ RSpec.describe 'Roles Request' do
           )
         end
       end
+      context 'creating role by user GUID for unaffiliated user' do
+        let(:params) do
+          {
+            type: 'organization_auditor',
+            relationships: {
+              user: {
+                data: { guid: user_unaffiliated.guid }
+              },
+              organization: {
+                data: { guid: org.guid }
+              }
+            }
+          }
+        end
+
+        let(:expected_response) do
+          {
+            guid: UUID_REGEX,
+            created_at: iso8601,
+            updated_at: iso8601,
+            type: 'organization_auditor',
+            relationships: {
+              user: {
+                data: { guid: user_unaffiliated.guid }
+              },
+              space: {
+                data: nil
+              },
+              organization: {
+                data: { guid: org.guid }
+              }
+            },
+            links: {
+              self: { href: %r(#{Regexp.escape(link_prefix)}\/v3\/roles\/#{UUID_REGEX}) },
+              user: { href: %r(#{Regexp.escape(link_prefix)}\/v3\/users\/#{user_unaffiliated.guid}) },
+              organization: { href: %r(#{Regexp.escape(link_prefix)}\/v3\/organizations\/#{org.guid}) },
+            }
+          }
+        end
+
+        let(:expected_codes_and_responses) do
+          h = Hash.new(code: 403)
+          h['admin'] = h['org_manager'] = {
+            code: 201,
+            response_object: expected_response
+          }
+          h
+        end
+
+        before do
+          allow(uaa_client).to receive(:users_for_ids).with([user_unaffiliated.guid]).and_return({ user_unaffiliated.guid => { 'username' => user_unaffiliated.username } })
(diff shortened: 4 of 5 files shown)


app/controllers/v3/roles_controller.rb
```diff
@@ -65,7 +65,8 @@ def destroy
     resource_not_found!(:role) unless role
 
     if role.for_space?
-      org_guid = Space.find(guid: role.space_guid).organization.guid
+      org_guid = Organization.join(:spaces, organization_id: :id).select(:organizations__guid).where(spaces__guid: role.space_guid)
```
sethboyles, line 68: is this really faster?  I feel like it may be easier to call
```
org_guid = Space.find(guid: role.space_guid).organization_guid
```

as that will not query the DB to pull up the entire organization

sethboyles: hmm, can we not just call `role.organization_guid`?

sethboyles:
based on [this comment](https://www.pivotaltracker.com/story/show/169353303/comments/208378113) from the initial implementation of the 'assign by username' story

```
 If an org manager tries to add an org role to an unaffiliated user by guid I think either of the following is fine:

1. They get a 422 and "Invalid user..." error
1. The requests succeeds and the role is created
I would go with whichever is easier for y'all. It should be basically impossible for an Org Manager to discover an unaffiliated user guid by themselves, so this case should be very rare.
```

I think it's fine to open up the user guid as choice for adding users outside the org.

sethboyles (changes requested):
address [this comment](https://github.com/cloudfoundry/cloud_controller_ng/pull/2971/files#r984017672) and I think this is good to go

app/controllers/v3/roles_controller.rb
philippthun: I ran a test with three different versions:

1. `Space.find(guid: role.space_guid).organization.guid`

```
(app/models/runtime/role.rb:110:in `space_guid') SELECT * FROM `spaces` WHERE `id` = 1
(app/controllers/v3/roles_controller.rb:68:in `destroy') SELECT * FROM `spaces` WHERE (`guid` = 'big-space') LIMIT 1
(app/controllers/v3/roles_controller.rb:68:in `destroy') SELECT * FROM `organizations` WHERE `id` = 2
(lib/cloud_controller/membership.rb:22:in `has_any_roles?') SELECT 1 AS `one` FROM (...) AS `t1` WHERE ((`type` IN ('space_manager')) AND (`user_id` = 3) AND (`space_id` IN (SELECT `id` FROM `spaces` WHERE (`guid` = 'big-space')))) LIMIT 1
(lib/cloud_controller/membership.rb:27:in `has_any_roles?') SELECT 1 AS `one` FROM (...) AS `t1` WHERE ((`type` IN ('organization_manager')) AND (`user_id` = 3) AND (`organization_id` IN (SELECT `id` FROM `organizations` WHERE (`guid` = 'big-org')))) LIMIT 1
```

There are two queries to get the org guid:
```
SELECT * FROM `spaces` WHERE (`guid` = 'big-space') LIMIT 1
SELECT * FROM `organizations` WHERE `id` = 2
```

2. `Space.find(guid: role.space_guid).organization_guid`

```
(app/models/runtime/role.rb:110:in `space_guid') SELECT * FROM `spaces` WHERE `id` = 1
(app/controllers/v3/roles_controller.rb:68:in `destroy') SELECT * FROM `spaces` WHERE (`guid` = 'big-space') LIMIT 1
(lib/sequel_plugins/vcap_relations.rb:124:in `block in define_guid_reader') SELECT * FROM `organizations` WHERE `id` = 2
(lib/cloud_controller/membership.rb:22:in `has_any_roles?') SELECT 1 AS `one` FROM (...) AS `t1` WHERE ((`type` IN ('space_manager')) AND (`user_id` = 3) AND (`space_id` IN (SELECT `id` FROM `spaces` WHERE (`guid` = 'big-space')))) LIMIT 1
(lib/cloud_controller/membership.rb:27:in `has_any_roles?') SELECT 1 AS `one` FROM (...) AS `t1` WHERE ((`type` IN ('organization_manager')) AND (`user_id` = 3) AND (`organization_id` IN (SELECT `id` FROM `organizations` WHERE (`guid` = 'big-org')))) LIMIT 1
```

The exact same two queries are executed, but this time the second query comes from the `VcapRelations` sequel plugin as there is no `organization_guid` column in the `spaces` table. I wasn't aware of this magic...

3. `Organization.join(:spaces, organization_id: :id).select(:organizations__guid).where(spaces__guid: role.space_guid)`

```
(app/models/runtime/role.rb:110:in `space_guid') SELECT * FROM `spaces` WHERE `id` = 1
(lib/cloud_controller/membership.rb:22:in `has_any_roles?') SELECT 1 AS `one` FROM (...) AS `t1` WHERE ((`type` IN ('space_manager')) AND (`user_id` = 3) AND (`space_id` IN (SELECT `id` FROM `spaces` WHERE (`guid` = 'big-space')))) LIMIT 1
(lib/cloud_controller/membership.rb:27:in `has_any_roles?') SELECT 1 AS `one` FROM (...) AS `t1` WHERE ((`type` IN ('organization_manager')) AND (`user_id` = 3) AND (`organization_id` IN (SELECT `id` FROM `organizations` WHERE (`guid` IN (SELECT `organizations`.`guid` FROM `organizations` INNER JOIN `spaces` ON (`spaces`.`organization_id` = `organizations`.`id`) WHERE (`spaces`.`guid` = 'big-space')))))) LIMIT 1
```

Now, there is a single query to get the org guid (`SELECT 'organizations'.'guid' FROM 'organizations' INNER JOIN 'spaces' ON ('spaces'.'organization_id' = 'organizations'.'id') WHERE ('spaces'.'guid' = 'big-space')`) and this query is even inlined into the roles check and thus sometimes not even executed (if the user is authorized by space role). So for this specific query, I think that Sam's proposal is the most efficient one.

@samcolson4: I would prefer if the variable name is changed from `org_guid` to `org_guid_query` to be consistent with other places.

P.S. We should think about enhancing `VCAP::CloudController::Permissions` and `VCAP::CloudController::Membership` to (alternatively) deal with space and org ids instead of guids. Now we often have an id, get the guid for this id and finally get the id for this guid...

philippthun: What actually would work is a combination of @sethboyles' ideas:
```
unauthorized! unless permission_queryer.can_update_active_space?(role.space_guid, role.space.organization_guid)
```

`role.space_guid` loads the space, and then `role.space.organization_guid` loads the org.

This would mean 4 queries (i.e. one less than no. 1 and 2 above / one more than no. 3). So it would still be an improvement and a very readable solution...

sethboyles: ah, I wasn't aware of the magic with `organization_guid`, either. 

@philippthun I think your combination solution with the additional commit looks good to me.  

Just curious--did you all notice a performance issue with this endpoint?  I would have thought that being a DELETE endpoint it would not be called all too often and cause too many issues.

will-gant: I'm happy to adopt that change for the sake readability. I've cherry-picked Philipp's [commit](https://github.com/sap-contributions/cloud_controller_ng/commit/aadce573387cd6c8d5aec3fba67d8d12ae59af25) into the branch, and have switched to `role.organization_guid` for space roles.

As a result of that I've tweaked [app/decorators/include_role_organization_decorator.rb](https://github.com/cloudfoundry/cloud_controller_ng/pull/2971/files#diff-285ddf85dfb392e2996c3040308af7407fac13b0077f84c6efef4b9fc8d06c8e) so as to avoid retrieving orgs for space roles when hitting `/v3/roles?include=organization` (previously this retrieved the org for any role for whom `organization_guid` wasn't nil).

will-gant: @sethboyles I threw a visualization together for one of our larger foundations giving the _sum_ of response times for different endpoints and request methods, excluding what I reckoned were most those whose slowness could plausibly be attributed to some external source (i.e. a service broker or Diego). On this measure the CREATE/DELETE endpoints for roles were both in the top ten, and since they concerned the same controller I decided to tackle them together.

app/controllers/v3/roles_controller.rb
```diff
@@ -117,15 +113,10 @@ def create_org_role(message)
...
-      user_guid = message.user_guid
-      unprocessable_user! if user_in_db?(user_guid) && !user_is_readable?(user_guid)
-    else
-      user_guid = lookup_user_guid_in_uaa(message.username, message.user_origin)
-      unprocessable_user! unless user_guid
-    end
+    user_guid = message.user_guid || lookup_user_guid_in_uaa(message.username, message.user_origin)
 
     user = User.first(guid: user_guid) || create_cc_user(user_guid)
+    unprocessable_user! unless user
```
philippthun, line 119: Is this even possible, i.e. can `user` be `nil`? Or can this line as well as the `unprocessable_user!` method be removed?

sethboyles: good catch. I think to keep the intent with the original code it should read 

```
   unprocessable_user! unless user_guid
   user = User.first(guid: user_guid) || create_cc_user(user_guid)
```

But rereading the code I'm less sure of this change.  Seems like the previous code only created new users if the user already existed in UAA, with this new code it will create new users even if the user doesn't exist at all?

sethboyles: I think the code needs to go back to two branches:

```
    if message.user_guid
      user_guid = message.user_guid
      unprocessable_user! if user_in_db?(user_guid)
    else
      user_guid = lookup_user_guid_in_uaa(message.username, message.user_origin)
      unprocessable_user! unless user_guid
    end
  ```

That way when creating a role and supplying a non-existent user_guid it still fails, instead of trying to create a user that doesn't exist in UAA

Comment (will-gant) [reply app/controllers/v3/roles_controller.rb:119]: Nah you're right. `User.first(guid: user_guid)` can be nil, but in that case `create_cc_user(user_guid)` won't be. 

I guess that was an attempt to handle the case where `create_cc_user` fails, but probably not necessary since that raises its own errors. I'll cut that now.

Comment (sethboyles) [reply app/controllers/v3/roles_controller.rb]: @will-gant I think throughout the refactor the original purpose of the `unprocessable_user!` call was lost.  I think it's still necessary for cases where the `user_guid` is sent via the message and does not exist in the db

Comment (philippthun) [reply app/controllers/v3/roles_controller.rb]: > @will-gant I think throughout the refactor the original purpose of the `unprocessable_user!` call was lost. I think it's still necessary for cases where the `user_guid` is sent via the message and does not exist in the db

This case (i.e. provided `user_guid` does not exist in UAA) wasn't handled before.
```
unprocessable_user! if user_in_db?(user_guid) && !user_is_readable?(user_guid)
```
The original check was for the "unaffiliated" user only.

So from my point of view the refactoring is now complete. But the question that remains is, if we might want to add a check (e.g. to `create_cc_user`) if the given `user_guid` is valid (this could leverage [User.uaa_users_info](https://github.com/cloudfoundry/cloud_controller_ng/blob/04aa8e89f3934823245981d0df30079e66bf1a34/app/models/runtime/user.rb#L211-L214)). Currently I'm able to add an arbitrary guid (e.g. `bommel`) and this will be displayed as follows:
```
Getting users in org my-org as me@example.com...

ORG MANAGER
  me@example.com (my-origin)

BILLING MANAGER
  No BILLING MANAGER found

ORG AUDITOR
  bommel (client)
```

Comment (will-gant) [reply app/controllers/v3/roles_controller.rb]: Ah apologies for the out-of-order reply. As you possibly guessed, I was responding to what Phillip wrote above.

In fact the current behaviour is that we validate that the user exists in UAA only if a _username_ is provided, not a guid:

```ruby
user_guid = lookup_user_guid_in_uaa(message.username, message.user_origin)
unprocessable_user! unless user_guid
```

Where a user _guid_ is provided, we emit `unprocessable_user!` only if the user already exists in the Users table but does not already have an org role in the requesting user's org. There's no check against UAA:

```ruby
unprocessable_user! if user_in_db?(user_guid) && !user_is_readable?(user_guid)
```

That permits the current bizarre behaviour:

```console
$ cf curl -X POST /v3/roles -d '{
>   "type": "organization_user",
>   "relationships": {
>     "user": {
>       "data": {
>         "guid": "not-a-real-guid"
>       }
>     },
>     "organization": {
>       "data": {
>         "guid": "459381c3-355a-4cef-903a-598ebed19757"
>       }
>     }
>   }
> }'
{"guid":"4c67678f-428e-4ac8-8840-80328b244ed1","created_at":"2022-10-05T08:18:41Z","updated_at":"2022-10-05T08:18:41Z","type":"organization_user","relationships":{"user":{"data":{"guid":"not-a-real-guid"}},"organization":{"data":{"guid":"459381c3-355a-4cef-903a-598ebed19757"}},"space":{"data":null}},"links":{"self":{"href":"https://api.cf.dev22.aws.bndl.sapcloud.io/v3/roles/4c67678f-428e-4ac8-8840-80328b244ed1"},"user":{"href":"https://api.cf.dev22.aws.bndl.sapcloud.io/v3/users/not-a-real-guid"},"organization":{"href":"https://api.cf.dev22.aws.bndl.sapcloud.io/v3/organizations/459381c3-355a-4cef-903a-598ebed19757"}}}

$ cf curl /v3/users/not-a-real-guid
{"guid":"not-a-real-guid","created_at":"2022-10-05T08:18:41Z","updated_at":"2022-10-05T08:18:41Z","username":null,"presentation_name":"not-a-real-guid","origin":null,"metadata":{"labels":{},"annotations":{}},"links":{"self":{"href":"https://api.cf.dev22.aws.bndl.sapcloud.io/v3/users/not-a-real-guid"}}}
```

The acceptance criteria in the [original story](https://www.pivotaltracker.com/n/projects/2196383/stories/170061485) and [accompanying flowchart](https://www.pivotaltracker.com/file_attachments/101221754/download?inline=true) make clear this was not intentional, and that a 422 should be returned in this case, so it looks like this was just an oversight.

philippthun: @will-gant So after looking at the original story and the flowchart you referenced, I get the feeling that this is intentional. The acceptance block at the end contains a "Not in UAA - GUID" section showing this "bizarre" behaviour and for @Gerg this was correct. And the fact that the cf cli displays such a "user" as `bommel (client)` (see my comment above) implies that this is some special case we should not change.

will-gant: Small extra thing is I think we don't need to throw `unprocessable!` if `lookup_user_guid_in_uaa` doesn't find a user by their username, as the method [already does that](https://github.com/cloudfoundry/cloud_controller_ng/blob/5054365d7ebe44f31d8ab17f2ef0328a4a08e53e/app/controllers/v3/roles_controller.rb#L195-L197).

will-gant: @philippthun So yeah, from the flowchart it looks like they settled on just checking the CCDB for users when a role was being created by user guid, instead of looking them up in the UAA.

Screenshot 2022-10-05 at 12 17 12

It's not totally clear to me, though from [this slack chat](https://cloudfoundry.slack.com/archives/CBNDLQBB5/p1576025705243800) I _think_ they wanted to check the UAA but reckoned the UAA client the CC uses for this didn't have the right authorities.

sethboyles: My mistake, I misread that condition and thought that behavior was introduced by the PR, and am surprised to see that it was already present and intentional.  

If that's the case then whether or not we change the behavior is out of scope of the PR , IMO, unless you feel inclined to investigate it further.  Thanks for digging up that history.

will-gant: No worries, and thanks for being so thorough! 

Just a note for posterity, in case someone actually does decide to address this issue in future and reads this PR. I reached out to @Gerg, who worked on this logic in 2019. Although he can't remember why the decision was taken to not check the UAA in the case that a user guid is provided, he speculates that it's due to the fact CF users can represent either a user or client in UAA.

CF users always have a guid, regardless of whether they map to a UAA user or a UAA client. But only CF users that represent a UAA user can have a username (which isn't stored by the CC).

Therefore if a **username** is provided for a CF user, only one call to the UAA, to check for the existence of a UAA user, is necessary.

By contrast the **guid** of a CF user could be either the id of a UAA user or the client_id of a UAA client. You'd need to be prepared to check both of these if you wanted to validate it. From the slack conversation there seemed to be a wariness of the hassle involved in getting the [`cloud_controller_username_lookup`](https://github.com/cloudfoundry/cf-deployment/blob/6ec2aca405f23a2eb32ec4108d3385edcfdb9b22/cf-deployment.yml#L719-L723) the `client.read` authority it'd need for this, and there also seemed to be a concern to minimise UAA queries (though I'd argue most org role assignments will be for UAA users, those extra queries will be infrequent).

PR #2983, opened 2022-09-26
Remove protobuf as dependency

MerricdeLauney:
We stopped using it as of https://github.com/cloudfoundry/cloud_controller_ng/pull/2900

Co-authored-by: Joseph Palermo <jpalermo@pivotal.io>


jpalermo (approved):
👍

johha:
I don't have much knowledge about protobuf but there seems to be references in some models like bbs: https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/diego/bbs/models/lrp_convergence_request.pb.rb#L6 
Are these files still used?

jpalermo:
Thanks for the catch @johha, we also don't have a lot of knowledge about the protobuf stuff. We were surprised that none of the tests failed even though those `requires` are clearly there.

We traced through the protobuf generation code and it looks like the three files that still had the `require` actually aren't even generated anymore, they were just getting left behind because the generation doesn't `rm -rf` before it regenerates.

We updated the PR to also remove those three files which we think is correct.

moleske:
For where proto stuff is used in capi, there's this [job](https://github.com/cloudfoundry/capi-ci/blob/main/ci/pipeline.yml#L620) in the pipeline that keeps bbs protos in sync.  This is helpful for the rare times that bbs changes some of the definitions of things like ActualLRP and DesiredLRP.  There used to be traffic controller generated code and a script file that was not configured in the pipeline, but that was removed as described above

jpalermo:
The job you linked to is actually the one that used to generate the files we're deleting here(didn't know there was another one).

That job now spits out this warning:
```
Omitting proto2 dependency 'github.com/gogo/protobuf/gogoproto/gogo.proto' from proto3 output file 'VARIOUS FILES HERE' because we don't support proto2 and no proto2 types from that file are being used.
```
which implies something changed such that those libraries can no longer be used, but also, it sounds like they aren't being used anyway so it's fine...

moleske:
Yeah that sounds good then.  There's some other protbuf issues such as the fact that CI uses [an old protobuf version](https://github.com/cloudfoundry/capi-dockerfiles/blob/main/capi-ruby-units/Dockerfile#L54) to generate the files.  I think someday that will bite us, but not sure when/where/how

moleske (approved):

PR #2993, opened 2022-09-29
V2 rate limit exclusions

samcolson4:
This PR allows users to be exempted from the V2 api rate limiter. This can be used by certain components that are not migrating to the V3 api.

### DIFF
--- a/spec/unit/middleware/rate_limiter_v2_api_spec.rb
+++ b/spec/unit/middleware/rate_limiter_v2_api_spec.rb
@@ -113,11 +113,12 @@ module CloudFoundry
         end
       end
 
-      describe 'headers when the user is exempt from v2 rate limiting' do
+      describe "when the user has the 'cloud_controller.v2_api_rate_limit_exempt' scope" do
         before do
           allow(VCAP::CloudController::SecurityContext).to receive(:v2_rate_limit_exempted?).and_return(true)
         end
-        it "doesn't decrease the count" do
+
+        it 'exempts them from rate limiting' do
           _, response_headers, _ = middleware.call(user_1_env)
 
           expect(request_counter).not_to have_received(:get)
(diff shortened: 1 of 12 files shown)


spec/unit/middleware/rate_limiter_v2_api_spec.rb
```diff
@@ -342,6 +361,24 @@ module Middleware
           end
         end
       end
+
+      context 'when the user is exempt from rate limiting' do
```
philippthun, line 365: I would put this context into `context 'when limit has exceeded' do` as it tests the behavior in case the limit has been reached. `path_info`, `middleware_env` and `before` can then be reused from the parent context.

spec/unit/middleware/rate_limiter_v2_api_spec.rb
```diff
@@ -68,6 +68,25 @@ module Middleware
             _, response_headers, _ = middleware.call(user_2_env)
             expect(response_headers['X-RateLimit-Remaining-V2-API']).to eq('80')
           end
+
+          context 'the user is exempt from v2 rate limiting' do
```
philippthun, line 72: I would move this context out of `describe 'X-RateLimit-Remaining-V2-API' do` and `describe 'headers as regular user' do` as it seems misplaced to me. Instead it would make a good top level context.

spec/unit/middleware/rate_limiter_v2_api_spec.rb
```diff
@@ -68,6 +68,25 @@ module Middleware
...
+            before do
+              allow(VCAP::CloudController::SecurityContext).to receive(:v2_rate_limit_exempted?).and_return(true)
+            end
+
+            let(:fake_request) { instance_double(ActionDispatch::Request, fullpath: '/v2/apps') }
+
+            it "doesn't decrease the count" do
+              allow(ActionDispatch::Request).to receive(:new).and_return(fake_request)
```
philippthun, line 80: This mock (and `fake_request`) seem unnecessary. I would propose to simply use `user_1_env`.

PR #3002, opened 2022-10-03
Diego::Client targets active bbs instance with id from locket

will-gant:
* **A short explanation of the proposed change:**

Get the bosh instance id for the currently-active `bbs` process from `locket`, so that [Diego::Client](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/diego/client.rb) can use an instance-specific bosh-dns URL instead of one that resolves to all IPs in the instance group (i.e. `fd0458bf-71f9-4df2-9ab1-5fb523a92e56.bbs.service.cf.internal`, not `bbs.service.cf.internal`).

Mitigates the risk of `Runner is unavailable` errors during a `cf push` when the client is unable to reach a `bbs` instance.

`bbs` operates as a cluster. At any one time one instance is 'active', and able to serve responses to API calls, while others are inactive. An active instance holds the `bbs` lock from `locket`.

At present, whenever the Diego::Client needs to communicate with `bbs` it hits the URL passed in from [config.diego.bbs.url](https://github.com/cloudfoundry/cloud_controller_ng/blob/cce461e7c3fb97161bf7501d6ab2d51953d95f89/config/cloud_controller.yml#L294-L295). The [default value](https://github.com/cloudfoundry/capi-release/blob/282227c144b19b0c88839555aadf10b5182d0f71/jobs/cloud_controller_ng/spec#L1083-L1085) passed in for this from the capi-release is `https://bbs.service.cf.internal:8889`. This domain is [defined by cf-deployment](https://github.com/cloudfoundry/cf-deployment/blob/6ec2aca405f23a2eb32ec4108d3385edcfdb9b22/cf-deployment.yml#L209-L215) as a bosh-dns alias for the `diego-api` instance group with query [q-s4](https://bosh.io/docs/dns/#constructing-queries) (i.e. resolving to all instances, whether healthy or unhealthy, in a random order).

When all `bbs` instances are reachable, you either hit the active instance first, or you hit an inactive one (`connection refused`) and the http client (implicitly) tries the next IP in the list. But if the first IP in the list is unreachable, the request hangs until it times out ([by default](https://github.com/cloudfoundry/capi-release/blob/282227c144b19b0c88839555aadf10b5182d0f71/jobs/cloud_controller_ng/spec#L1089-L1091) after 10s) and then fails. The Diego::Client [makes three attempts](https://github.com/cloudfoundry/cloud_controller_ng/blob/cce461e7c3fb97161bf7501d6ab2d51953d95f89/lib/diego/client.rb#L163-L169) to reach the active `bbs` instance before giving up and raising an error, with our [`httpclient`](https://www.rubydoc.info/gems/httpclient) re-resolving DNS each time. 

Say you've got two `diego-api` instances, and a network issue makes one of them unreachable. In such a case, every API call currently has a 1/8 chance of that the first IP in the list is the unreachable one three times in a row, which will cause `cf push` and some other commands to fail.

Note:
* We considered switching to a `q-s3` bosh-dns query (healthy instances only) but decided against this. If some process other than `bbs` was failing on an active bbs instance clients would be stuck unable to reach the Diego API (DNS would resolve only to inactive instances, while nothing would trigger a new leadership election).
* `locket` is colocated with `bbs` on the `diego-api` instance. Retries are handled automatically by gRPC (and with a shorter 5-second timeout).

cloudfoundry/capi-release#275 (should be merged before this PR)

* [x] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)*
    \* Run against the following cf-acceptance-test suites: apps, container_networking, detect, docker, internet_dependent, routing, service_instance_sharing, services, ssh, sso, tasks, v3

### DIFF
--- a/lib/cloud_controller/config_schemas/base/worker_schema.rb
+++ b/lib/cloud_controller/config_schemas/base/worker_schema.rb
@@ -77,6 +77,15 @@ module VCAP::CloudController
               fog_aws_storage_options: Hash
             },
 
+            locket: {
+              host: String,
+              port: Integer,
+              ca_file: String,
+              cert_file: String,
+              key_file: String,
+              diego_client_timeout: Integer
+            },
+
             packages: {
               max_package_size: Integer,
               app_package_directory_key: String,
--- a/lib/diego/client.rb
+++ b/lib/diego/client.rb
@@ -1,27 +1,30 @@
 require 'diego/bbs/bbs'
 require 'diego/errors'
 require 'diego/routes'
+require 'uri'
+require 'locket/client'
 
 module Diego
   class Client
     PROTOBUF_HEADER = { 'Content-Type'.freeze => 'application/x-protobuf'.freeze }.freeze
 
     def initialize(url:, ca_cert_file:, client_cert_file:, client_key_file:,
-      connect_timeout:, send_timeout:, receive_timeout:)
+      connect_timeout:, send_timeout:, receive_timeout:, locket_config:)
       ENV['PB_IGNORE_DEPRECATIONS'] ||= 'true'
       @client = build_client(
-        url,
         ca_cert_file,
         client_cert_file,
         client_key_file,
         connect_timeout,
         send_timeout,
         receive_timeout)
+      @locket_config = locket_config
+      @base_url = url
     end
 
     def ping
       response = with_request_error_handling do
-        client.post(Routes::PING)
+        client.post(URI.join(bbs_url, Routes::PING))
       end
 
       validate_status!(response: response, statuses: [200])
@@ -32,7 +35,7 @@ module Diego
       request = protobuf_encode!({ domain: domain, ttl: ttl.to_i }, Bbs::Models::UpsertDomainRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::UPSERT_DOMAIN, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::UPSERT_DOMAIN), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -43,7 +46,7 @@ module Diego
       request = protobuf_encode!({ task_definition: task_definition, domain: domain, task_guid: task_guid }, Bbs::Models::DesireTaskRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::DESIRE_TASK, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::DESIRE_TASK), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -54,7 +57,7 @@ module Diego
       request = protobuf_encode!({ task_guid: task_guid }, Bbs::Models::TaskByGuidRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::TASK_BY_GUID, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::TASK_BY_GUID), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -65,7 +68,7 @@ module Diego
       request = protobuf_encode!({ domain: domain, cell_id: cell_id }, Bbs::Models::TasksRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::LIST_TASKS, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::LIST_TASKS), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -76,7 +79,7 @@ module Diego
       request = protobuf_encode!({ task_guid: task_guid }, Bbs::Models::TaskGuidRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::CANCEL_TASK, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::CANCEL_TASK), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -87,7 +90,7 @@ module Diego
       request = protobuf_encode!({ desired_lrp: lrp }, Bbs::Models::DesireLRPRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::DESIRE_LRP, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::DESIRE_LRP), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -98,7 +101,7 @@ module Diego
       request = protobuf_encode!({ process_guid: process_guid }, Bbs::Models::DesiredLRPByProcessGuidRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::DESIRED_LRP_BY_PROCESS_GUID, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::DESIRED_LRP_BY_PROCESS_GUID), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -109,7 +112,7 @@ module Diego
       request = protobuf_encode!({ process_guid: process_guid, update: lrp_update }, Bbs::Models::UpdateDesiredLRPRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::UPDATE_DESIRED_LRP, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::UPDATE_DESIRED_LRP), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -120,7 +123,7 @@ module Diego
       request = protobuf_encode!({ process_guid: process_guid }, Bbs::Models::RemoveDesiredLRPRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::REMOVE_DESIRED_LRP, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::REMOVE_DESIRED_LRP), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -131,7 +134,7 @@ module Diego
       request = protobuf_encode!({ actual_lrp_key: actual_lrp_key }, Bbs::Models::RetireActualLRPRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::RETIRE_ACTUAL_LRP, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::RETIRE_ACTUAL_LRP), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -142,7 +145,7 @@ module Diego
       request = protobuf_encode!({ domain: domain }, Bbs::Models::DesiredLRPsRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::DESIRED_LRP_SCHEDULING_INFOS, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::DESIRED_LRP_SCHEDULING_INFOS), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -153,7 +156,7 @@ module Diego
       request = protobuf_encode!({ process_guid: process_guid }, Bbs::Models::ActualLRPsRequest)
 
       response = with_request_error_handling do
-        client.post(Routes::ACTUAL_LRPS, request, PROTOBUF_HEADER)
+        client.post(URI.join(bbs_url, Routes::ACTUAL_LRPS), request, PROTOBUF_HEADER)
       end
 
       validate_status!(response: response, statuses: [200])
@@ -164,13 +167,29 @@ module Diego
       tries ||= 3
       yield
     rescue => e
+      @cached_active_bbs_id = nil
       retry unless (tries -= 1).zero?
       raise RequestError.new(e.message)
     end
 
+    def bbs_url
+      uri = URI(base_url)
+      uri.host = "#{active_bbs_id}.#{uri.host}"
+      uri.to_s
+    end
+
+    def active_bbs_id
+      tries ||= 3
+      @cached_active_bbs_id ||= latest_active_bbs_id
+    rescue GRPC::BadStatus => e
+      retry unless (tries -= 1).zero?
+      raise e
+    end
+
     private
 
-    attr_reader :client
+    attr_reader :client, :base_url, :locket_config
+    attr_accessor :cached_active_bbs_id
 
     def protobuf_encode!(hash, protobuf_message_class)
       # See below link to understand proto3 message encoding
@@ -190,15 +209,30 @@ module Diego
       raise DecodeError.new(e.message)
     end
 
-    def build_client(url, ca_cert_file, client_cert_file, client_key_file,
+    def build_client(ca_cert_file, client_cert_file, client_key_file,
       connect_timeout, send_timeout, receive_timeout)
-      client                 = HTTPClient.new(base_url: url)
+      client                 = HTTPClient.new
       client.connect_timeout = connect_timeout
       client.send_timeout    = send_timeout
       client.receive_timeout = receive_timeout
       client.ssl_config.set_client_cert_file(client_cert_file, client_key_file)
       client.ssl_config.set_trust_ca(ca_cert_file)
       client
+    end
+
+    def latest_active_bbs_id
+      locket_client.fetch(key: 'bbs').resource.owner
+    end
+
+    def locket_client
+      Locket::Client.new(
+        host: locket_config[:host],
+        port: locket_config[:port],
+        client_ca_path: locket_config[:ca_file],
+        client_key_path: locket_config[:key_file],
+        client_cert_path: locket_config[:cert_file],
+        timeout: locket_config[:diego_client_timeout]
+      )
     end
   end
 end
--- a/lib/diego/client.rb
+++ b/lib/diego/client.rb
@@ -189,7 +189,6 @@ module Diego
     private
 
     attr_reader :client, :base_url, :locket_config
-    attr_accessor :cached_active_bbs_id
 
     def protobuf_encode!(hash, protobuf_message_class)
       # See below link to understand proto3 message encoding
--- a/lib/locket/client.rb
+++ b/lib/locket/client.rb
@@ -6,16 +6,21 @@ module Locket
     class Error < StandardError
     end
 
-    def initialize(host:, port:, client_ca_path:, client_cert_path:, client_key_path:)
+    def initialize(host:, port:, client_ca_path:, client_cert_path:, client_key_path:, timeout: nil)
       client_ca = File.open(client_ca_path).read
       client_key = File.open(client_key_path).read
       client_cert = File.open(client_cert_path).read
 
       @service = Models::Locket::Stub.new(
         "#{host}:#{port}",
-        GRPC::Core::ChannelCredentials.new(client_ca, client_key, client_cert)
+        GRPC::Core::ChannelCredentials.new(client_ca, client_key, client_cert),
+        timeout: timeout
       )
       @lock_acquired = false
+    end
+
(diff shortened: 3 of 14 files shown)


lib/diego/client.rb
```diff
@@ -9,19 +11,28 @@ class Client
...
+      @config = VCAP::CloudController::Config.config
       @client = build_client(
-        url,
         ca_cert_file,
         client_cert_file,
         client_key_file,
         connect_timeout,
         send_timeout,
         receive_timeout)
+      @locket_client = Locket::Client.new(
```
philippthun, line 22: During my "experiments" (*) I got the impression that the DNS resolution is done only once for a locket client. The BOSH DNS query type is different compared to `bbs.service.cf.internal`, i.e. it "should" return healthy instances only, but I think that it might be better to move the locket client creation into a dedicated method and initialize a fresh client for every `fetch` call.

(*) What I tested was adding a general network delay to one locket VM and I could see that the client was either using the fast or the slow locket service throughout its lifetime.

lib/diego/client.rb
```diff
@@ -168,9 +179,23 @@ def with_request_error_handling(&blk)
...
+    def active_bbs_url
+      uri = URI(base_url)
+      uri.host = "#{active_bbs_id}.#{uri.host}"
+      uri.to_s
+    end
+
+    def active_bbs_id
+      tries ||= 3
+      locket_client.fetch('bbs').resource.owner
+    rescue GRPC::BadStatus => e
```
philippthun, line 191: Are these definitely the only errors that could happen? Or should we make this `rescue` more general?

lib/diego/client.rb
```diff
@@ -168,9 +179,23 @@ def with_request_error_handling(&blk)
       raise RequestError.new(e.message)
     end
 
+    def active_bbs_url
+      uri = URI(base_url)
+      uri.host = "#{active_bbs_id}.#{uri.host}"
+      uri.to_s
+    end
+
+    def active_bbs_id
```
philippthun, line 188: Querying the active bbs instance for every call seems rather costly. I would propose to cache the `active_bbs_url` (in some instance field) and clear it in the `rescue` block in `with_request_error_handling`.

philippthun (commented):
I like the idea of using locket to determine the active bbs instance!

I've added some initial feedback to your changes. (So far I did not look at the specs.)

lib/locket/client.rb
```diff
@@ -52,7 +54,7 @@ def lock_acquired?
 
     attr_reader :service, :lock_acquired, :key, :owner
```
philippthun, line 55: `key` and `owner` are no longer instance fields...

lib/locket/client.rb
```diff
@@ -2,32 +2,34 @@
...
-        GRPC::Core::ChannelCredentials.new(client_ca, client_key, client_cert)
+        GRPC::Core::ChannelCredentials.new(client_ca, client_key, client_cert),
+        timeout: timeout
       )
       @lock_acquired = false
+    end
 
-      @key = key
-      @owner = owner
+    def fetch(key)
```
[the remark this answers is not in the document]
will-gant: Yep fair enough

lib/locket/client.rb
```diff
@@ -65,6 +67,14 @@ def build_lock_request
       )
     end
 
+    def build_fetch_request(key)
```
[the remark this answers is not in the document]
will-gant: Done

lib/diego/client.rb
```diff
@@ -9,19 +11,28 @@ class Client
...
         connect_timeout,
         send_timeout,
         receive_timeout)
+      @locket_client = Locket::Client.new(
+        host: config.get(:locket, :host),
+        port: config.get(:locket, :port),
+        client_ca_path: config.get(:locket, :ca_file),
+        client_key_path: config.get(:locket, :key_file),
+        client_cert_path: config.get(:locket, :cert_file),
+        timeout: 5,
```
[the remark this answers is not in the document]
will-gant: Yeahh I thought about that. I think the sticking point for me was that an operator would logically expect that that'd control the timeout for the Locket::Client everywhere that it's used - both in the new usage in Diego::Client and [the existing usage](https://github.com/sap-contributions/cloud_controller_ng/blob/69587e9e5a5a80172675819775b11c80760a7294/lib/cloud_controller/deployment_updater/scheduler.rb#L27-L33) in DeploymentUpdater::Scheduler.

As you can see I've deliberately not passed a timeout for the latter, which means [we pass `nil` by default](https://github.com/sap-contributions/cloud_controller_ng/blob/69587e9e5a5a80172675819775b11c80760a7294/lib/locket/client.rb#L9) when instantiating the gRPC 'stub'. Ultimately [that means an infinite timeout](https://rubydoc.info/gems/grpc/GRPC/ClientStub). That was the situation previously, and that's still the case now.

I'm unfamiliar with DeploymentUpdater::Scheduler's usage of Locket, and wasn't comfortable introducing a new timeout for that as part of these changes.

lib/diego/client.rb
will-gant: Ah that's already the widest error possible - all other the errors in the `grpc` gem inherit from `GRPC::BadStatus`. It inherits from the standard library's `StandardError`.

I guess I could `rescue StandardError` instead, but I'd be concerned we might try to recover from some unrecoverable errors... What do you think?

lib/locket/client.rb
will-gant: True dat. I'll update this.

lib/diego/client.rb
will-gant: Yeah that seems reasonable. I'll take a look at that now.

lib/diego/client.rb
will-gant: If I understand correctly, you want to ensure that Locket::Client re-resolves DNS on each of the three retries, and your concern is that, without instantiating a new 'stub' each time, those retries would just keep trying the same unavailable server?

lib/diego/client.rb
philippthun: Okay, if that's the "root" GRPC error, that should be fine.

lib/diego/client.rb
will-gant: Done this. We store the active instance id in a variable and keep reusing that until it fails, then get a new one from locket.

lib/diego/client.rb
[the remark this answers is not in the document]
will-gant: Yeah that's a good solution - I'll do that then.

lib/diego/client.rb
```diff
@@ -175,6 +175,7 @@ def with_request_error_handling(&blk)
       tries ||= 3
       yield
     rescue => e
+      @cached_active_bbs_id = current_active_bbs_id
```
philippthun, line 178: Here I would set `@cached_active_bbs_id` to `nil`. Then the next call to `active_bbs_id` gets the current one.

lib/locket/client.rb
```diff
@@ -7,15 +7,12 @@ class Error < StandardError
     end
 
     def initialize(host:, port:, client_ca_path:, client_cert_path:, client_key_path:, timeout: nil)
-      client_ca = File.open(client_ca_path).read
-      client_key = File.open(client_key_path).read
-      client_cert = File.open(client_cert_path).read
-
-      @service = Models::Locket::Stub.new(
```
philippthun, line 14: Changing this here has an impact on the other usage of this client as well. It might be better to only do this (i.e. create new client for each call) in the diego client.

From my point of view in the diego client this is justified as we now also cache the result (active bbs instance). So it's okay if retrieving gets a little bit more expensive (due to the recreation of the locket client) but therefore more resilient to temporary network issues.

will-gant: Yep, another good point. I've switched this around so we remake it only when Diego::Client does a `fetch`.

lib/diego/client.rb
will-gant: Yep, agreed, that's better

lib/diego/client.rb
```diff
@@ -9,19 +11,28 @@ class Client
     def initialize(url:, ca_cert_file:, client_cert_file:, client_key_file:,
       connect_timeout:, send_timeout:, receive_timeout:)
       ENV['PB_IGNORE_DEPRECATIONS'] ||= 'true'
+      @config = VCAP::CloudController::Config.config
```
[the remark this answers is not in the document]
will-gant: Yeah alright. I'm going with option 3. Seems nicer to maintain the (sort of) pattern of getting values from config in `DependencyLocator`, and if I do option 1 then I'll have to pass in about a dozen args when initializing Diego::Client.

lib/cloud_controller/config_schemas/base/worker_schema.rb
```diff
@@ -77,6 +77,14 @@ class WorkerSchema < VCAP::Config
               fog_aws_storage_options: Hash
             },
 
+            locket: {
```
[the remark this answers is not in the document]
will-gant: Hmm so I can see that it was [made optional](https://github.com/cloudfoundry/cloud_controller_ng/commit/47c75de296973bc72e3984b3e633717978678630) for the `DeploymentUpdater::Scheduler` three years ago as part of [this](https://www.pivotaltracker.com/n/projects/966314/stories/170160715) user story. From the description and [one of the other commits](https://github.com/cloudfoundry/capi-k8s-release/commit/49a1cafe0cf39119b2f1d7bedf9cd6b7b41c63e3) it looks like they did this so the scheduler could be deployed on Kubernetes as part of the [capi-k8s-release](https://github.com/cloudfoundry/capi-k8s-release)... Which is now deprecated. I guess therefore we don't need to care about breaking it.

lib/diego/client.rb
will-gant: Ok this is sorted

mkocher:
Locket seems like an implementation detail of Diego, I'd hesitate to use it for service discovery from another component without thinking deeply about it.

Could this be solved by going through the list of IPs returned instead of picking one randomly? Presumably the successful IP could be cached in memory and used until Diego goes through another leader election?

will-gant:
> _Locket seems like an implementation detail of Diego, I'd hesitate to use it for service discovery from another component without thinking deeply about it._

Definitely a valid concern. At the same time it's worth weighing against that the fact that aside from Diego itself there are four other components that we package in cf-deployment with clients for speaking to Locket. Those are the [routing-api](https://github.com/cloudfoundry/cf-deployment/blob/b2c9749c516fee106974bd336e0b11c196e46dec/cf-deployment.yml#L1069-L1100), [tps](https://github.com/cloudfoundry/cf-deployment/blob/b2c9749c516fee106974bd336e0b11c196e46dec/cf-deployment.yml#L1299-L1312), [policy-server-asg-syncer](https://github.com/cloudfoundry/cf-deployment/blob/b2c9749c516fee106974bd336e0b11c196e46dec/cf-deployment.yml#L1125-L1137) and [cc_deployment_updater](https://github.com/cloudfoundry/cf-deployment/blob/b2c9749c516fee106974bd336e0b11c196e46dec/cf-deployment.yml#L1273-L1283). The latter is part of cloud_controller_ng, and has had a [dependency](https://github.com/cloudfoundry/cloud_controller_ng/commit/e3d9d339756efcd8c7fff4010d8b6e0bc28a3bf2) on locket since CAPI 1.63.0, in 2018. 

The risk is lowered a bit more thanks to the fact that we're doing this with a quite modest tweak to a pre-existing wrapper around a gRPC client whose ruby code has been generated from a [protofile](https://github.com/cloudfoundry/locket/blob/master/models/locket.proto). That provides some insurance against things changing.

Given the above, and the relatively superficial nature of the dependency I've added, I feel on balance it's unlikely that future changes in the way Diego uses Locket are going to be drastic enough to risk serious disruption for us (or that the CC's usage would deter such refactoring). 


> _Could this be solved by going through the list of IPs returned instead of picking one randomly? Presumably the successful IP could be cached in memory and used until Diego goes through another leader election?_

Currently the CC uses the [`httpclient`](https://www.rubydoc.info/gems/httpclient), which handles DNS resolution. The alternative would, I guess, be for Diego::Client to use another library to do an 'explicit' lookup with bosh-dns, loop over the IP list returned, and then try a `httpclient.post` on each IP, rescuing on the failures. Then cache and reuse the IP for the active server. This would work, though I think having the client doing DNS lookups like this and using trial-and-error to discover the server aren't ideal.

It'd be good to hear from others, as the fact that I've just finished working on a different solution biases my opinion somewhat. I think, for the reasons I gave a few paragraphs above, I don't see this dependency on Locket as posing much risk.

@mkocher

will-gant:
@sethboyles If you get time, it'd be great to get your thoughts on this!

will-gant:
Also threw together a couple of diagrams. First one shows what can happen today if there's an outage and the first IP in the randomly-ordered list that `bbs.service.cf.internal` resolves to is, by chance, an unreachable `diego-api` instance three times in a row:

diego-bbs-comms-1

The second one shows how this would work with the change under my PR:

 
diego-bbs-comms-2

spec/diego/client_spec.rb
```diff
@@ -609,6 +647,63 @@ module Diego
...
+      context 'when the last call to bbs failed' do
+        before do
+          call_count = 0
+          allow(locket_client).to receive(:fetch).with(key: key) do
+            call_count += 1
+            call_count == 1 ? fetch_response : fetch_response2
+          end
+          stub_request(:post, 'https://bbs-1.bbs.example.com:4443/v1/ping').to_timeout
+          stub_request(:post, 'https://bbs-2.bbs.example.com:4443/v1/ping').to_return(status: 200,
+body: Bbs::Models::PingResponse.encode(Bbs::Models::PingResponse.new(available: true)).to_s)
```
philippthun, line 686: This line should be indented.

spec/diego/client_spec.rb
```diff
@@ -1,16 +1,54 @@
...
   RSpec.describe Client do
     let(:bbs_url) { 'https://bbs.example.com:4443' }
     let(:ca_cert_file) { File.join(Paths::FIXTURES, 'certs/bbs_ca.crt') }
     let(:client_cert_file) { File.join(Paths::FIXTURES, 'certs/bbs_client.crt') }
     let(:client_key_file) { File.join(Paths::FIXTURES, 'certs/bbs_client.key') }
+    let(:locket_client) { instance_double(Locket::Client) }
+    let(:test_config) { TestConfig.config_instance }
+    let(:locket_config) { test_config.get(:locket) }
+    let(:key) { 'bbs' }
+    let(:fetch_request) do
```
philippthun, line 15: This `fetch_request` is unused.

lib/diego/client.rb
```diff
@@ -164,13 +167,29 @@ def with_request_error_handling(&blk)
...
+    rescue GRPC::BadStatus => e
+      retry unless (tries -= 1).zero?
+      raise e
+    end
+
     private
 
-    attr_reader :client
+    attr_reader :client, :base_url, :locket_config
+    attr_accessor :cached_active_bbs_id
```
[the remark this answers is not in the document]
will-gant: Ah yes, cheers. I was trying out a couple of different approaches and forgot to take this out after settling on `@...`

spec/diego/client_spec.rb
will-gant: Good spot

spec/unit/lib/locket/lock_worker_spec.rb
```diff
@@ -13,32 +15,32 @@
...
-        allow(lock_runner).to receive(:lock_acquired?).and_return(false)
+        allow(client).to receive(:lock_acquired?).and_return(false)
 
-        lock_worker.acquire_lock_and_repeatedly_call {}
+        lock_worker.acquire_lock_and_repeatedly_call(owner: owner, key: key, &{})
         expect(lock_worker).to have_received(:sleep).with(1)
       end
     end
 
     describe 'when it does not have the lock' do
```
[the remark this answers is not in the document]
will-gant: Yep true

sethboyles:
> At the same time it's worth weighing against that the fact that, aside from Diego itself, there are four other components packaged in cf-deployment with clients that talk to Locket

Both TPS and the deployment-updater are a part of CAPI, but as far as I can tell both of them use Locket to register their own locks, not query information about Diego.  

I don't have much of an opinion yet on whether or not using Locket to query information about Diego is a pattern we should avoid or not, but I know little about Locket.  Perhaps we should reach out to the Diego team for their thoughts?

will-gant:
Yep I'll ask. @ameowlia, @geofffranks, @jrussett - I wonder if any of yourselves or your other Diego colleagues would have an opinion on the proposed use of Locket here?

geofffranks:
This seems like a great poposal to me. The data is authoritative about who the active BBS VM is. There's precedent for this in the `cfdot locks` command (pulls data from locket to display the active BBS + Aucitoneer VMs).

jpalermo:
Wait, hold on, it feels very wrong to have a service running on the platform that a client would have to inspect which instances of the service are healthy and then target them directly.

`bbs.service.cf.internal` should not return instances that do not actually work, which is what it sounds like current exists.

My first thought was that bosh-dns [healthiness](https://bosh.io/docs/dns/#healthiness) could be used to ensure the "unhealthy" VMs aren't actually returned by `bbs.service.cf.internal`, and that does work great, right up until you co-locate the `bbs` job on a VM with other jobs, now the whole VM is unhealthy for bosh dns query purposes, which is probably bad. 

Using process level healthiness does "solve" the problem with s3 queries mentioned above, because you could just switch to s0 and in the case where another job on the active bbs VM is failing, they would all be in an "unhealthy" state, and you'd basically fall back to the behavior we have now. But again, people do co-locate jobs differently than is done in cf-deployment, so until we add job level healthiness to bosh-dns-aliases, this seems like probably a bad idea.

So the other option would be to put this healthiness logic in the bbs itself, and then when a non-active VM got a request, it could proxy it to a healthy one or maybe it could just return a `302` and the clients would follow that to the proper VM? That would mean clients don't need to do any special logic and can just query `bbs.service.cf.internal`.

philippthun:
The origin of the problem is that the Ruby HTTP client that is used for talking to BBS is bad in handling **unreachable** (i.e. isolated from a network point of view) endpoints.

This is not a general problem; when you use cURL to retrieve data from `bbs.service.cf.internal` and there is an unreachable VM, the request to this VM times out very fast and the next IP address is used. So in general this is something that a client could handle.

I don't think that Diego/BBS is doing something wrong; rejecting connections on an inactive VM seems like a reasonable approach and works well - the issue we are seeing is only related to unreachable VMs.

In general I see three different possibilities to solve this:
1. Fix/enhance the currently used Ruby HTTP client.
2. Identify and use a different Ruby HTTP client that behaves more like cURL.
3. Use an alternative approach to identify the currently active BBS instance to circumvent issues with unreachable VMs.

---

1. We tested a monkey patch (see [1]) of the `TCPSocket` class which introduced a `connect_timeout` parameter with Ruby 3.0. Setting this to a lower value than the `HTTPClient.connect_timeout` (default: 10 seconds), results in a similar behavior as for a refused connection, i.e. the next IP address is tried implicitly. We didn't dare to apply this monkey patch as it would affect all TCP sockets created in Ruby and we are also not sure if a contribution to the `httpclient` gem would be feasible.

2. We did not go down this road so far.

3. That's the purpose of this PR. It can be seen as a workaround for the deficient Ruby HTTP client, but it prevents the issues we are seeing with unreachable VMs.

[1]
```
module DefaultConnectTimeout
  DEFAULT_CONNECT_TIMEOUT = 2 # seconds

  def initialize(remote_host, remote_port, local_host = nil, local_port = nil, connect_timeout: nil)
    super(remote_host, remote_port, local_host, local_port, connect_timeout: connect_timeout || DEFAULT_CONNECT_TIMEOUT)
  end
end

class TCPSocket
  prepend DefaultConnectTimeout
end
```

jpalermo:
Thanks for those details @philippthun that does make it seem less like a Diego problem than I was saying and I'm no longer thinking the problem should be solved elsewhere.

That's unfortunate that the timeout isn't configurable in the http client, that would be an ideal solution.

sethboyles:
I think it's worth evaluating other HTTP clients to see if there is support for this before adding a workaround on our side.

sethboyles:
httpclient hasn't had a release in 6 years: https://rubygems.org/gems/httpclient

will-gant:
Ok given the mixed reaction to this I'm going to close it and we'll take a look at alternative HTTP clients.

will-gant:
Picking this story up again. Just noting that as far as I can tell there doesn't seem to be any (maintained) ruby HTTP client that, for a domain that resolves to multiple IPs, will try the next IP in the list if the first one it tries times out.

So I think we'll just need to resolve the DNS ourselves as @mkocher suggested.

PR #3021, opened 2022-10-14
Check space/org permissions by id, not guid

will-gant:

Optimizes the way controllers check user permissions by cutting out _lots and lots_ of duplicate DB queries for spaces and orgs. An idea shamelessly stolen from @philippthun (bottom of [this comment](https://github.com/cloudfoundry/cloud_controller_ng/pull/2971#discussion_r984594512)).

**An explanation of the use cases your change solves:**

Until now we've been doing something like this for a large chunk of API calls:

```ruby
# app/controllers/v3/foo_controller.rb
def create
  foo = Foo.find(guid: hashed_params[:guid])
  org = foo.organization
  foo_not_found! unless @foo && permission_queryer.can_read_whatever?(org.guid)
  ...
end
```
```ruby
# lib/cloud_controller/permissions.rb
def can_read_whatever?(org_guid)
  can_read_globally? || membership.has_any_roles?(ROLES_FOR_WHATEVER, nil, org_guid)
end
```
```ruby
# lib/cloud_controller/membership.rb
def has_any_roles?(roles, space_guid=nil, org_guid=nil)
  org_id = Organization.where(guid: org_guid).select(:id)
  ...
end
```

Basically a controller action does a DB query to look up an org or a space purely so it can pass the guid of whichever it is through (ultimately) to `VCAP::CloudController::Membership`. The latter then uses the guid to do another DB query to look up the same row to get that thing's id. Silly.

The changes in summary:
* Where we want to check a user's permissions in relation to a specific org or space, the controller now passes the org/space's id instead of its guid 
* Since we no longer need org guids for most permissions checks, if we've already looked up a space we can skip looking up its org in the controller entirely - rows in the `spaces` table have an `organization_id` column.
* Fetchers. Some of these (e.g. [AppFetcher](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/fetchers/app_fetcher.rb)) have `fetch` methods whose return values previously included an org. For some fetchers the org was only ever retrieved so its guid could passed for a permissions check. Such fetchers now no longer return orgs.

**Links to any other associated PRs:**

n/a


app/controllers/v3/deployments_controller.rb
```diff
@@ -59,10 +62,12 @@ def create
 
   def update
     deployment = DeploymentModel.find(guid: hashed_params[:guid])
-    resource_not_found!(:deployment) unless deployment &&
-      permission_queryer.can_read_from_space?(deployment.app.space.guid, deployment.app.space.organization.guid)
-    unauthorized! unless permission_queryer.can_write_to_active_space?(deployment.app.space.guid)
-    suspended! unless permission_queryer.is_space_active?(deployment.app.space.guid)
+    resource_not_found!(:deployment) unless deployment
+
+    space = deployment.app.space
```
philippthun, line 67: > In some places I noticed us doing multiple DB queries in order to get different attributes of the same thing ...

`deployment.app.space.id` and `deployment.app.space.organization_id` would not lead to two DB queries; sequel fetches data from the database into the object (e.g. `deployment.app.space`) when it is accessed for the first time.

But using a variable `space = deployment.app.space` here makes the code definitely more readable. In other places it might not be necessary.

will-gant: Ah right, my mistake. Thanks for that!

will-gant:
Cool, just ran this successfully through the [cf-acceptance-tests](https://github.com/cloudfoundry/cf-acceptance-tests) (a partial run), swapping this PR in for `src/cloud_controller_ng` in CAPI 1.140.0. Ran the following from CATs:

apps
container_networking
detect
docker
internet_dependent
routing
service_instance_sharing
services
ssh
sso
tasks
v3

lib/cloud_controller/permissions.rb
```diff
@@ -276,7 +267,7 @@ def readable_space_scoped_spaces_query
     end
   end
 
-  def can_read_route?(space_guid, org_guid)
+  def can_read_route?(space_guid)
```
philippthun, line 270: This method could also be changed to accept `space_id` instead.

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -429,115 +429,120 @@ module VCAP::CloudController
     end
 
     describe '#can_read_from_space?' do
+      let(:membership) { instance_double(Membership) }
```
philippthun, line 432: Not needed (also the `allow(Membership)` below).

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -429,115 +429,120 @@ module VCAP::CloudController
...
         it 'returns true for org manager' do
           org.add_manager(user)
-          expect(permissions.can_read_from_space?(space_guid, org_guid)).to be true
+          expect(permissions.can_read_from_space?(space.id, org.id)).to be true
         end
       end
     end
 
     describe '#can_read_secrets_in_space?' do
+      let(:membership) { instance_double(Membership) }
```
philippthun, line 500: Not needed (also the `allow(Membership)` below).

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -429,115 +429,120 @@ module VCAP::CloudController
...
         context 'and user is not an admin' do
           it 'return false' do
             set_current_user(user)
-            expect(permissions.can_read_secrets_in_space?(space_guid, org_guid)).to be false
+            expect(permissions.can_read_secrets_in_space?(space.id, org.id)).to be false
           end
         end
       end
 
       context 'user has valid membership' do
```
philippthun, line 538: The previous checks for different roles also worked fine. Any reason why not to keep them?

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -661,121 +666,103 @@ module VCAP::CloudController
     end
 
     describe '#can_write_to_active_space?' do
+      let(:membership) { instance_double(Membership) }
```
philippthun, line 669: Not needed (also the `allow(Membership)` below).

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -661,121 +666,103 @@ module VCAP::CloudController
...
           it 'return false' do
             set_current_user(user)
-            expect(permissions.can_write_to_active_space?(space_guid)).to be false
+            expect(permissions.can_write_to_active_space?(space.id)).to be false
           end
         end
       end
 
       context 'user has valid membership' do
-        it 'returns true for space developer' do
```
philippthun, line 695: The previous checks for different roles also worked fine. Any reason why not to keep them?

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -661,121 +666,103 @@ module VCAP::CloudController
...
-          org.add_manager(user)
-          expect(permissions.can_write_to_active_space?(space_guid)).to be false
+          space.add_developer(user)
+          expect(permissions.can_write_to_active_space?(space.id)).to be true
         end
       end
     end
 
     describe '#can_manage_apps_in_active_space?' do
+      let(:membership) { instance_double(Membership) }
```
philippthun, line 720: Not needed (also the `allow(Membership)` below).

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -661,121 +666,103 @@ module VCAP::CloudController
...
           it 'return false' do
             set_current_user(user)
-            expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be false
+            expect(permissions.can_manage_apps_in_active_space?(space.id)).to be false
           end
         end
       end
 
       context 'user has valid membership' do
-        it 'returns true for space developer' do
```
philippthun, line 752: The previous checks for different roles also worked fine. Any reason why not to keep them?

app/controllers/v3/deployments_controller.rb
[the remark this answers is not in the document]
will-gant: Yeah fair enough

app/controllers/v3/domains_controller.rb
```diff
@@ -220,6 +222,18 @@ def unprocessable_org!(org_guid)
     unprocessable!("Organization with guid '#{org_guid}' does not exist or you do not have access to it.")
   end
 
+  def not_shared_to_org!(org_name)
+    unprocessable!("Unable to unshare domain from organization with name '#{org_name}'. Ensure the domain is shared to this organization.")
+  end
+
+  def can_write_to_org?(org_id)
```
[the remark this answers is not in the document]
will-gant: Yeah ok

app/controllers/v3/domains_controller.rb
```diff
@@ -140,28 +140,18 @@ def delete_shared_org
...
 
-    permission_check = lambda do |owning_org_guid, shared_org_guid|
-      can_write_to_owning_org = permission_queryer.can_write_to_active_org?(owning_org_guid)
-      return if can_write_to_owning_org && permission_queryer.is_org_active?(owning_org_guid)
-
-      can_write_to_shared_org = permission_queryer.can_write_to_active_org?(shared_org_guid)
-      return if can_write_to_shared_org && permission_queryer.is_org_active?(shared_org_guid)
+    shared_org = Organization.find(guid: message.org_guid)
+    unprocessable_org!(message.org_guid) unless shared_org
+    not_shared_to_org!(shared_org.name) unless domain.private?
```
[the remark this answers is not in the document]
will-gant: Good spot - that's something I added while trying out another way of reorganising this, and left in by mistake.

app/controllers/v3/domains_controller.rb
```diff
@@ -140,28 +140,18 @@ def delete_shared_org
...
-      else
-        suspended!
-      end
-    end
-    permission_check.call(domain.owning_organization_guid, message.org_guid)
+    check_unshare_domain_permissions!(domain.owning_organization_id, shared_org.id)
 
-    shared_org = Organization.find(guid: message.org_guid)
-    unprocessable_org!(message.org_guid) unless shared_org && permission_queryer.can_read_from_org?(shared_org.guid)
+    unprocessable_org!(message.org_guid) unless permission_queryer.can_read_from_org?(shared_org.id)
```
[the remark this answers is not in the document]
will-gant: This one's a bit awkward. I've done it this way to maintain the previous logic and stray as little as possible from switching guids for ids.

[The unit test that'll fail](https://github.com/cloudfoundry/cloud_controller_ng/blob/12d50221c17d485c6e3d68f18114f14bee5cd0e6/spec/request/domains_spec.rb#L1817-L1834) if we emit that error before running `check_unshare_domain_permissions!` is one that checks response codes when an org is suspended. That test expects that an authorized user (i.e. an org manager) receives a 403 `OrgSuspended` error, and that everyone else should get a 403 `NotAuthorized`. But if we move `can_read_from_org?` up, all users instead get a 422 `UnprocessableEntity`.

I felt while reorganising this that there might be a case for refactoring that, but I didn't want to get sucked in as part of this PR...

app/controllers/v3/roles_controller.rb
```diff
@@ -65,11 +65,12 @@ def destroy
     resource_not_found!(:role) unless role
 
     if role.for_space?
-      unauthorized! unless permission_queryer.can_update_active_space?(role.space_guid, role.organization_guid)
-      suspended! unless permission_queryer.is_space_active?(role.space_guid)
+      space = role.space
+      unauthorized! unless permission_queryer.can_update_active_space?(space.id, space.organization_id)
```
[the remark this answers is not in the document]
will-gant: Nice! Wasn't aware of that one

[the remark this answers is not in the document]
will-gant: (despite the obvious example two lines below ;))

app/controllers/v3/security_groups_controller.rb
```diff
@@ -172,12 +172,12 @@ def check_unwritable_spaces(space_guids)
     unauthorized_space = false
     suspended_space = false
     space_guids.each do |space_guid|
-      org = Space.find(guid: space_guid)&.organization
-      if org
-        if !permission_queryer.can_update_active_space?(space_guid, org.guid)
+      space = Space.find(guid: space_guid)
+      if space&.organization
```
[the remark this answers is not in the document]
will-gant: Agreed

lib/cloud_controller/permissions.rb
will-gant: Agreed, though I've already made that change in another branch I'm working on at the moment, so to avoid making it in two places I'll maybe skip it here.

spec/unit/lib/cloud_controller/permissions_spec.rb
will-gant: So yeah, as you can see my intention was to mock `Membership` - I felt it that was better than testing the behaviour of both classes in `permissions_spec.rb` - and then stub its various methods to always return the same thing. That meant that it was only necessary to test _one_ of the non-global reader roles, since they'd all just give the same answer.

But if you'd rather I don't mock `Membership` then it makes sense to leave these in.

spec/unit/lib/cloud_controller/permissions_spec.rb
will-gant: See my other comment [here](https://github.com/cloudfoundry/cloud_controller_ng/pull/3021/files#r1004261111)

will-gant: (but I'm switching this back in the way you suggest)

app/controllers/v3/roles_controller.rb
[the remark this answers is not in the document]
will-gant: Just double-checked this, and only org roles have an `organization_id`. We're deleting a space role here, so I think we _have_ to look up the space to get the org id.

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -752,30 +735,30 @@ module VCAP::CloudController
...
           org.add_user(user)
           space.add_auditor(user)
-          expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be false
+          expect(permissions.can_write_to_active_space?(space.id)).to be false
         end
 
         it 'returns false for org manager' do
           org.add_manager(user)
-          expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be false
+          expect(permissions.can_write_to_active_space?(space.id)).to be false
```
philippthun, line 755: Was `expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be false` before.

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -752,30 +735,30 @@ module VCAP::CloudController
...
           space.add_manager(user)
-          expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be false
+          expect(permissions.can_write_to_active_space?(space.id)).to be false
         end
 
         it 'returns false for space auditor' do
           org.add_user(user)
           space.add_auditor(user)
-          expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be false
+          expect(permissions.can_write_to_active_space?(space.id)).to be false
```
philippthun, line 750: Was `expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be false` before.

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -752,30 +735,30 @@ module VCAP::CloudController
...
-          space.add_supporter(user)
-          expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be true
+          expect(permissions.can_write_to_active_space?(space.id)).to be true
         end
 
         it 'returns false for space manager' do
           org.add_user(user)
           space.add_manager(user)
-          expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be false
+          expect(permissions.can_write_to_active_space?(space.id)).to be false
```
philippthun, line 744: Was `expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be false` before.

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -752,30 +735,30 @@ module VCAP::CloudController
...
           org.add_user(user)
           space.add_developer(user)
-          expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be true
-        end
-
-        it 'returns true for space supporter' do
-          org.add_user(user)
-          space.add_supporter(user)
-          expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be true
+          expect(permissions.can_write_to_active_space?(space.id)).to be true
```
philippthun, line 738: Was `expect(permissions.can_manage_apps_in_active_space?(space_guid)).to be true` before.

spec/unit/lib/cloud_controller/permissions_spec.rb
```diff
@@ -665,54 +665,37 @@ module VCAP::CloudController
...
             set_current_user(user)
-            expect(permissions.can_write_to_active_space?(space_guid)).to be false
+            expect(permissions.can_write_to_active_space?(space.id)).to be false
           end
         end
       end
 
       context 'user has valid membership' do
-        it 'returns true for space developer' do
+        it 'returns true' do
```
philippthun, line 695: Here you still have this single test instead of separate ones for space developer, space manager, space auditor and org manager.

spec/unit/lib/cloud_controller/permissions_spec.rb
will-gant: Doh, yes, good spot

philippthun (approved):

PR #3034, opened 2022-10-25
Enable CC to process large manifests

svkrieger:

Enable the CC to process large requests (caused by large manifests) up to 1MB.

Until now, when a user does a cf push with an app manifest larger than ~130KB, the CC will error with a 500 - unknown error. This is due to Thin ([see here](https://github.com/macournoyer/thin/blob/02f0a3a72620313c69a1d797671cc522399d9b4c/lib/thin/request.rb#L12))  which outsources large request bodies/headers into Tempfiles, instead of in-memory StreamIOs (Puma does the same [see here](https://github.com/puma/puma/blob/32d999708d7019b85b0518c8e0693773b0bd4ff4/lib/puma/const.rb#L147)). Currently the CC coding calls a `.string` method on the request body, which is not present for Tempfile objects. This commit changes to the `.read` method, which is available on both classes.

### DIFF
--- a/app/controllers/v3/application_controller.rb
+++ b/app/controllers/v3/application_controller.rb
@@ -89,8 +89,11 @@ class ApplicationController < ActionController::Base
   def parsed_yaml
     return @parsed_yaml if @parsed_yaml
 
+    bad_request!('Manifest size is too large. The maximum supported size is 1MB.') if request.body.size > 1.megabyte
+
     allow_yaml_aliases = false
-    yaml = Psych.safe_load(request.body.string, permitted_classes: [], permitted_symbols: [], aliases: allow_yaml_aliases, strict_integer: true)
+    yaml = Psych.safe_load(request.body.read, permitted_classes: [], permitted_symbols: [], aliases: allow_yaml_aliases, strict_integer: true)
+
     message_parse_error!('invalid request body') if !yaml.is_a? Hash
     @parsed_yaml = yaml
   rescue Psych::BadAlias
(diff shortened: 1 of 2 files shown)


app/controllers/v3/application_controller.rb
```diff
@@ -90,7 +90,19 @@ def parsed_yaml
...
+    # Thin and also Puma outsource large request bodies/headers to a tempfile.
+    # So we need to check here whether it is a StringIO or a Tempfile and load it accordingly.
+    if request.body.instance_of? Tempfile
+      if request.body.size > 1.megabyte
+        bad_request!('Manifest size is too large. The maximum supported size is 1MB.')
+      else
+        yaml = Psych.load_file(request.body, permitted_classes: [], permitted_symbols: [], aliases: allow_yaml_aliases, strict_integer: true)
+      end
+    else
+      yaml = Psych.safe_load(request.body.string, permitted_classes: [], permitted_symbols: [], aliases: allow_yaml_aliases, strict_integer: true)
```
philippthun, line 103: You can use `request.body.read` in both cases, so the `if-else` is not necessary. The check for the size still makes sense and should be done before.

philippthun (commented):
I've pushed a commit with a test and added a comment on the change that can be simplified.

philippthun (commented):
I think the change looks good now. There is only one tiny thing; the commit message still says  the following: "This commit adds a check for the object type and loads the manifest accordingly."

philippthun (approved):

PR #3037, opened 2022-10-27
route fetcher checks shared routes

MerricdeLauney:

### DIFF
--- a/app/fetchers/route_fetcher.rb
+++ b/app/fetchers/route_fetcher.rb
@@ -37,10 +37,6 @@ module VCAP::CloudController
           dataset = dataset.where(space_id: space_ids)
         end
 
-        if message.requested?(:space_guids)
-          dataset = dataset.where(space_id: Space.where(guid: message.space_guids).select(:id))
-        end
-
         if message.requested?(:domain_guids)
           dataset = dataset.where(domain_id: Domain.where(guid: message.domain_guids).select(:id))
         end
@@ -66,6 +62,13 @@ module VCAP::CloudController
             requirements: message.requirements,
             resource_klass: Route,
           )
+        end
+
+        if message.requested?(:space_guids)
+          dataset = dataset.where do
+            (Sequel[:spaces][:guid] =~ message.space_guids) |
+              (Sequel[:route_shares][:target_space_guid] =~ message.space_guids)
+          end
         end
 
         super(message, dataset, Route)
--- a/app/fetchers/route_fetcher.rb
+++ b/app/fetchers/route_fetcher.rb
@@ -3,8 +3,17 @@ require 'fetchers/base_list_fetcher'
 module VCAP::CloudController
   class RouteFetcher < BaseListFetcher
     class << self
-      def fetch(message, readable_route_dataset, eager_loaded_associations: [])
-        dataset = readable_route_dataset.eager(eager_loaded_associations).qualify
+      def fetch(message, readable_space_guids_dataset: nil, eager_loaded_associations: [], omniscient: false)
+        dataset = Route.dataset.eager(eager_loaded_associations).
+                  join(:spaces, id: Sequel[:routes][:space_id]).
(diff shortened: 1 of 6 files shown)


lib/cloud_controller/permissions.rb
```diff
@@ -237,6 +241,14 @@ def can_read_route?(space_id)
       @user.audited_organizations.map(&:id).include?(space.organization_id)
   end
 
+  def space_guids_with_readable_routes_query
+    if can_read_globally?
+      VCAP::CloudController::SPACE.select(:guid)
```
philippthun, line 246: This line should never be reached. Otherwise it would be very expensive to first select all `space_guids` and then use this giant list in some filter. Thus I would recommend to raise an exception like we did e.g. [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/8173f4677b350b79dc5c985a3005119b9644919a/lib/cloud_controller/permissions.rb#L157).

app/fetchers/route_fetcher.rb
```diff
@@ -3,8 +3,17 @@
 module VCAP::CloudController
   class RouteFetcher < BaseListFetcher
     class << self
-      def fetch(message, readable_route_dataset, eager_loaded_associations: [])
-        dataset = readable_route_dataset.eager(eager_loaded_associations).qualify
+      def fetch(message, readable_space_guids_dataset: nil, eager_loaded_associations: [], omniscient: false)
+        dataset = Route.dataset.eager(eager_loaded_associations).
+                  join(:spaces, id: Sequel[:routes][:space_id]).
```
philippthun, line 8: Personally I'm in favor of a different syntax: `id: :routes__space_id` instead of `id: Sequel[:routes][:space_id]`. I find this easier to read.

MerricdeLauney: After discussing it with @sethboyles and @moleske, we felt that Sequel[:routes][:space_id] was easier to read for us so i'm leaving this for now. If we want to have a dedicated conversation about it i'm open to it.

lib/cloud_controller/permissions.rb
MerricdeLauney: I am surprised to hear that calling `space_guids_with_readable_routes_query` as an admin would lead to expensive queries. I naively assumed that it was safe because it returns a dataset rather than an in memory object. 

I switched to throwing an error because I know y'all have been doing a lot of work to optimize performance but would love to hear your thoughts.

philippthun: The call to `space_guids_with_readable_routes_query` as admin itself would not be expensive, as no DB query would be executed, but the resulting dataset could then be used by the caller in some other DB query that will be executed. And returning all space guids and using them in another WHERE condition just does not make sense, i.e. `SELECT * FROM apps WHERE space_guid IN (SELECT guid FROM spaces)` is semantically the same as `SELECT * FROM apps`.

Thus it's always better to support the two different scenarios (i.e. admin vs. non-admin) directly in the DB query that's going to be executed (as you did [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/d58cb6afb8fa90f14bfb3205e33c7a305c50397a/app/fetchers/route_fetcher.rb#L11-L16).

The exception should only be there to ensure that no one accidentally forgets this when adding some new functionality.

app/fetchers/route_fetcher.rb
philippthun: In the end consistency is more important to me than a specific syntax. This line now uses two different ways to express a "thing", i.e. table `:spaces` and column `Sequel[:routes][:space_id]`.

When I look through the code, I can see several different identifiers (I guess this list is not even complete):
```
:spaces
Sequel[:spaces]
Space.table_name
:apps__space_guid
Sequel[:apps][:space_guid]
"#{AppModel.table_name}__space_guid".to_sym
```

I would prefer if we could agree on a single syntax that's used throughout the codebase. And to me the symbol syntax (i.e. `:spaces` and `:apps__space_guid` is the shortest, simplest and most concise syntax.

PR #3041, opened 2022-10-31
Replace .include? / .map with more efficient SQL statements

philippthun:


app/access/space_quota_definition_access.rb
```diff
@@ -70,7 +70,7 @@ def delete?(space_quota_definition, params=nil)
     def read?(space_quota_definition, *_)
       context.admin_override || (
         !context.user.nil? && (
-          (context.user.managed_organizations.include? space_quota_definition.organization) ||
+          context.user.managed_organizations_dataset.where(id: space_quota_definition.organization_id).any? ||
           !(context.user.managed_spaces & space_quota_definition.spaces).empty? ||
```
will-gant, line 74: This is definitely an improvement, but you can go further and cut out organizations table by doing something like

`OrganizationManager.where(user_id: context.user.id, organization_id: space_quota_definition.organization.id).any?`

lib/cloud_controller/permissions.rb
```diff
@@ -233,8 +233,8 @@ def can_read_route?(space_id)
     space = VCAP::CloudController::Space.where(id: space_id).first
 
     space.has_member?(@user) || space.has_supporter?(@user) ||
-      @user.managed_organizations.map(&:id).include?(space.organization_id) ||
-      @user.audited_organizations.map(&:id).include?(space.organization_id)
+      @user.managed_organizations_dataset.where(id: space.organization_id).any? ||
```
will-gant, line 236: Same as my first comment - the `OrganizationManager` has `user_id` and `organization_id` columns, so if you use that you can cut organizations out entirely.

app/access/space_quota_definition_access.rb
will-gant: `managed organizations_dataset` produces this query:

```
SELECT  "organizations".*
FROM    "organizations"
       INNER JOIN  "organizations_managers"
               ON (  "organizations_managers" . "organization_id" = 
                    "organizations" . "id" )
WHERE  (  "organizations_managers" . "user_id" = 118 ) 
```

app/models/runtime/user.rb
```diff
@@ -82,13 +82,13 @@ def validate
     end
 
     def validate_organization(org)
-      unless org && organizations.include?(org)
+      unless org && organizations_dataset.where(id: org).any?
```
will-gant, line 85: I might be misunderstanding something here - and this logic already existed before you touched it - but is it actually possible to have a situation where `org` is true and `organizations_dataset.where(id: org).any?` is false?

app/presenters/v3/space_quota_presenter.rb
```diff
@@ -56,12 +56,12 @@ def space_quota
...
-      visible_spaces = if @all_spaces_visible
-                         space_quota.spaces
-                       else
-                         space_quota.spaces.select { |space| @visible_space_guids.include? space.guid }
-                       end
-      visible_spaces.map { |space| { guid: space.guid } }
+      visible_spaces_dataset = if @all_spaces_visible
+                                 space_quota.spaces_dataset
+                               else
+                                 space_quota.spaces_dataset.where(guid: @visible_space_guids)
```
will-gant, line 62: Ultimately `@visible_space_guids` comes from [`permissions.readable_space_guids`](https://github.com/cloudfoundry/cloud_controller_ng/blob/9b793086e472453382a451244eccebd49d9059ac/lib/cloud_controller/permissions.rb#L155-L157) but maybe we should instead that should be [`permissions.readable_space_guids_query`](https://github.com/cloudfoundry/cloud_controller_ng/blob/9b793086e472453382a451244eccebd49d9059ac/lib/cloud_controller/permissions.rb#L159-L165) (though we changed that recently to raise an error if the user reads globally..)

will-gant (changes requested):

PR #3043, opened 2022-11-01
Cut DB query complexity in permissions/visibility checks

will-gant:
* **A short explanation of the proposed change:**

Reduces the complexity of queries checking for user permissions and resource visibility, in particular by removing unnecessary table unions and joins, and by making greater use of role tables when retrieving space and organization ids.

This includes a reversal of changes to `membership.rb` introduced by #2502, which has resulted in inefficient unions of role tables that are unnecessary for most requests. Where an endpoint is authorized only for, say, space developers and org managers, our existing logic will union all eight role tables. This PR reverts to making a union only of the tables for roles that are actually authorized for the specific endpoint being hit.

n/a

n/a

  *_Run against the following cf-acceptance-test suites with the PR branch swapped into CAPI 1.140.0: apps, container_networking, detect, docker, internet_dependent, routing, service_instance_sharing, services, ssh, sso, tasks, v3_

### DIFF
--- a/app/models/runtime/organization.rb
+++ b/app/models/runtime/organization.rb
@@ -164,7 +164,7 @@ module VCAP::CloudController
 
     def self.user_visibility_filter(user)
       {
-        id: OrganizationRole.where(user_id: user.id).select(Sequel.as(:organization_id, :id))
+        id: user.membership_org_ids
       }
     end
 
--- a/app/models/runtime/organization.rb
+++ b/app/models/runtime/organization.rb
@@ -164,13 +164,7 @@ module VCAP::CloudController
 
     def self.user_visibility_filter(user)
       {
-        id: dataset.join_table(:inner, :organizations_managers, organization_id: :id, user_id: user.id).select(:organizations__id).union(
-          dataset.join_table(:inner, :organizations_users, organization_id: :id, user_id: user.id).select(:organizations__id)
-        ).union(
-          dataset.join_table(:inner, :organizations_billing_managers, organization_id: :id, user_id: user.id).select(:organizations__id)
-        ).union(
-          dataset.join_table(:inner, :organizations_auditors, organization_id: :id, user_id: user.id).select(:organizations__id)
-        ).select(:id)
+        id: OrganizationRole.where(user_id: user.id).select(Sequel.as(:organization_id, :id))
       }
     end
 
--- a/app/models/runtime/user.rb
+++ b/app/models/runtime/user.rb
@@ -173,38 +173,58 @@ module VCAP::CloudController
     end
 
     def membership_organizations
-      Organization.join(:organizations_users, organization_id: :id, user_id: id).select(:organizations__id).
-        union(
-          Organization.join(:organizations_auditors, organization_id: :id, user_id: id).select(:organizations__id)
-        ).
-        union(
-          Organization.join(:organizations_managers, organization_id: :id, user_id: id).select(:organizations__id)
-        ).
-        union(
-          Organization.join(:organizations_billing_managers, organization_id: :id, user_id: id).select(:organizations__id)
-        )
+      Organization.where(id: membership_org_ids).select(:id)
+    end
+
+    def membership_space_ids
+      SpaceRole.where(user_id: id).select_map(:space_id)
+    end
+
+    def membership_org_ids
+      OrganizationRole.where(user_id: id).select_map(:organization_id)
+    end
+
+    def org_user_org_ids
+      OrganizationUser.where(user_id: id).select(:organization_id)
+    end
+
+    def org_manager_org_ids
+      OrganizationManager.where(user_id: id).select(:organization_id)
+    end
+
+    def org_billing_manager_org_ids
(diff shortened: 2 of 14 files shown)


app/models/runtime/user.rb
```diff
@@ -163,48 +163,62 @@ def remove_spaces(space)
...
 
     def membership_spaces
-      Space.join(:spaces_developers, space_id: :id, user_id: id).select(:spaces__id).
-        union(
-          Space.join(:spaces_auditors, space_id: :id, user_id: id).select(:spaces__id)
-        ).
-        union(
-          Space.join(:spaces_managers, space_id: :id, user_id: id).select(:spaces__id)
-        )
+      Space.where(id: membership_space_ids).select(:id)
```
FloThinksPi, line 166: Also returns the space ID when the user is space_supporter. Dont know if this is right since the previous logic does not include space_supporters.

will-gant: Ok so on closer inspection it turns out that the user model's `membership_spaces` is called by (and only by) the service_plan model ([here](https://github.com/cloudfoundry/cloud_controller_ng/blob/8173f4677b350b79dc5c985a3005119b9644919a/app/models/services/service_plan.rb#L129-L131)). From what I can tell running through the unit tests with the debugger, only v2 code depends on it (in particular, via some indirection, the [v2 service instances controller](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/controllers/services/service_instances_controller.rb)). I've verified that service plans are visible to space supporters using the v3 API, and a complete run of v3 request tests with the debugger didn't result in `membership_spaces` being invoked.

I guess there are two options:
1. Since I'm here anyway and it's quick, change this into a union of the three space role tables that were previously being joined, perhaps with a comment pointing out that a refactor is needed if any future v3 use is made of it.
2. Just revert it back to the way it was before I turned up, since it's a v2-only thing and we're not really aiming to optimize that part of the codebase.

Any preference @FloThinksPi?

FloThinksPi: Puh whatever seems less work for you. Both are fine for me. Would probably go for a revert of this function since then the change is a no brainer and we know it works :)

will-gant: Cool, yeah I've done that. Rebased and force-pushed.

sethboyles:
Thanks for digging into this.  We've noticed issues with these queries as well and agree that essentially reverting https://github.com/cloudfoundry/cloud_controller_ng/pull/2502 is a best course of action for this.

app/models/runtime/organization.rb
```diff
@@ -164,13 +164,7 @@ def add_user(user)
...
     def self.user_visibility_filter(user)
       {
-        id: dataset.join_table(:inner, :organizations_managers, organization_id: :id, user_id: user.id).select(:organizations__id).union(
-          dataset.join_table(:inner, :organizations_users, organization_id: :id, user_id: user.id).select(:organizations__id)
-        ).union(
-          dataset.join_table(:inner, :organizations_billing_managers, organization_id: :id, user_id: user.id).select(:organizations__id)
-        ).union(
-          dataset.join_table(:inner, :organizations_auditors, organization_id: :id, user_id: user.id).select(:organizations__id)
-        ).select(:id)
+        id: OrganizationRole.where(user_id: user.id).select(Sequel.as(:organization_id, :id))
```
sethboyles, line 167: Did you find `OrganizationRole.where` to be faster than the joins?  In some of our testing we find UNIONs with the `where` call executed after to be slow, so curious if this doesn't follow that pattern.  I wrote more about it here: https://github.com/cloudfoundry/capi-release/issues/278#issuecomment-1303905237

lib/cloud_controller/membership.rb
```diff
@@ -16,77 +16,112 @@ def initialize(user)
...
-      org_roles(roles).any?
+    def space_role_sufficient?(roles)
+      roles_filter(roles, SPACE_ROLES).any?
     end
 
-    def space_ids_for_space_roles(roles)
-      if space_role?(roles)
-        SpaceRole.where(type: space_roles(roles), user_id: @user.id).select(:space_id)
-      end
+    def org_role_sufficient?(roles)
```
sethboyles, line 93: should these methods be called something like `contains_space_role?` and `contains_org_role?`

app/models/runtime/organization.rb
will-gant: Ahh right I hadn't been aware that you'd observed that being slower, but I guess it makes sense that `OrganizationRole` would result in a bunch of unnecessary role rows being fetched, only to be filtered out with the `where` at the higher level.

I've change this (here and in a couple of other places) to be a union of the individual role tables with separate `where` clauses at the 'lower' level.

lib/cloud_controller/membership.rb
will-gant: No problem - have adopted those names

will-gant:
> Thanks for digging into this. We've noticed issues with these queries as well and agree that essentially reverting #2502 is a best course of action for this.

No worries!

FloThinksPi (approved):
CATS succeded :)

PR #3052, opened 2022-11-08
Misc db optimizations

will-gant:

Tiny PR doing two things:
1. Use Sequel's `empty?` to avoid cumbersomely counting all the rows in a table when all we want to know is whether there are _any_ rows.
2. Where we want the guid of a service corresponding to a service_plan, use the latter's `service_guid` column instead of querying the services table.

n/a - minor optimizations for speed

n/a

### DIFF
--- a/app/actions/process_create.rb
+++ b/app/actions/process_create.rb
@@ -26,9 +26,7 @@ module VCAP::CloudController
       app.class.db.transaction do
         process = app.add_process(attrs)
         route_mappings = process.route_mappings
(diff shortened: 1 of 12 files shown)


philippthun:
@will-gant - Just a small remark. Instead of `!dataset.empty?` you could also use `dataset.any?` which might be nicer to read. Actually these two statements are not equivalent, but we are using a Sequel plugin that makes them behave the same. See #2856.

will-gant:
Doing that in #3053

PR #3058, opened 2022-11-10
support for asynchronous service binding creating in Space Manifests

sethboyles:
* Previously, the manifest apply job sent `accepts_incomplete: false` for any service binding, meaning manifest jobs would fail if trying to bind a service with only asynchronous binding support.  Now the manifest job will attempt a synchronous binding, but retry and send `accepts_incomplete: true` if the broker responds that it only supports async binding creation

Co-authored-by: David Alvarado <alvaradoda@vmware.com>
Co-authored-by: Alex Rocha <alexr1@vmware.com>
Co-authored-by: Michael Oleske <moleske@pivotal.io>

capi-release branch: https://github.com/cloudfoundry/capi-release/pull/281


app/actions/v3/service_binding_create.rb
```diff
@@ -72,6 +72,7 @@ def save_failed_state(binding, e)
             description: e.message,
           }
         )
+        binding
```
philippthun, line 75: `save_failed_state` is always called before an exception is raised, so this returned binding is unused.

app/models/services/service_binding_operation.rb
```diff
@@ -1,7 +1,7 @@
 module VCAP::CloudController
   class ServiceBindingOperation < Sequel::Model
     export_attributes :state, :description, :type, :updated_at, :created_at
-
+    many_to_one :service_binding
```
philippthun, line 4: What's the effect of this change?

app/actions/app_apply_manifest.rb
```diff
@@ -157,21 +162,41 @@ def create_service_bindings(manifest_service_bindings_message, app)
...
-            result = action.bind(binding, parameters: binding_message.parameters)
-            if result[:async]
-              raise ServiceBrokerRespondedAsyncWhenNotAllowed
-            end
-          rescue ServiceBrokerRespondedAsyncWhenNotAllowed,
-                 V3::ServiceBindingCreate::BindingNotRetrievable
-
-            raise ServiceBrokerRespondedAsyncWhenNotAllowed.new('The service broker responded asynchronously, but async bindings are not supported.')
+            binding = action.bind(binding, parameters: binding_message.parameters, accepts_incomplete: false)
+          rescue ServiceBrokerRespondedAsyncWhenNotAllowed
```
philippthun, line 166: `action.bind` is not raising `ServiceBrokerRespondedAsyncWhenNotAllowed`, or?

spec/unit/actions/app_apply_manifest_spec.rb
```diff
@@ -848,35 +849,182 @@ module VCAP::CloudController
...
+
+                  precursor_count = 0
+                  allow(service_cred_binding_create).to receive(:precursor) do
+                    precursor_count += 1
+                    if precursor_count == 1
+                      binding1
+                    else
+                      binding2
+                    end
                   end
```
philippthun, line 869: Isn't this the same as `allow(service_cred_binding_create).to receive(:precursor).and_return(binding1, binding2)`?

See [Specify different return values for multiple calls](https://relishapp.com/rspec/rspec-mocks/docs/configuring-responses/returning-a-value#specify-different-return-values-for-multiple-calls)

spec/unit/actions/app_apply_manifest_spec.rb
```diff
@@ -848,35 +849,182 @@ module VCAP::CloudController
...
+                  expect(orphan_mitigation_job).not_to be_nil
+                  expect(orphan_mitigation_job).to be_a_fully_wrapped_job_of Jobs::Services::DeleteOrphanedBinding
+                end
+
+                it 'has a maximum retry_after' do
+                  allow(service_cred_binding_create).to receive(:poll).and_return(V3::ServiceBindingCreate::ContinuePolling.call(24.hours),
+                                                                                  V3::ServiceBindingCreate::PollingFinished)
+                  expect(app_apply_manifest).to receive(:sleep).with(60)
+
+                  Timecop.scale(600) do
```
philippthun, line 975: Is this needed? What does it do?

spec/unit/actions/app_apply_manifest_spec.rb
```diff
@@ -848,35 +849,182 @@ module VCAP::CloudController
...
+                    precursor_count += 1
+                    if precursor_count == 1
+                      binding1
+                    else
+                      binding2
+                    end
+                  end
+
+                  count = 0
+                  allow(service_cred_binding_create).to receive(:bind).with(anything, parameters: anything, accepts_incomplete: true) do |args|
```
philippthun, line 914: `|args|` is unused.

spec/unit/actions/app_apply_manifest_spec.rb
```diff
@@ -848,35 +849,182 @@ module VCAP::CloudController
...
+
+                  expect(service_cred_binding_create).to receive(:poll).exactly(5).times
+                  expect(app_apply_manifest).to receive(:sleep).with(5).exactly(3).times
+
+                  app_apply_manifest.apply(app.guid, message)
+                end
+
+                it 'verifies exception is thrown if maximum polling duration is exceeded' do
+                  TestConfig.override(max_manifest_service_binding_poll_duration: 15)
+                  allow_any_instance_of(AppApplyManifest).to receive(:sleep) do |action, seconds|
```
philippthun, line 954: `action` is unused.

app/models/services/service_binding_operation.rb
sethboyles: allows us to use Factories to create Operations in tests like this: 
https://github.com/cloudfoundry/cloud_controller_ng/pull/3058/files#diff-f3725445b6d7d167e4273ed2a61ba2b541da978b8ea8f6e3b31a43f7e956ae9eR875

spec/unit/actions/app_apply_manifest_spec.rb
sethboyles: It makes time go faster by a factor of 600.  It's not actually needed for this test, so I'll remove

config/cloud_controller.yml
```diff
@@ -406,3 +406,5 @@ threadpool_size: 20
 webserver: thin
 default_app_lifecycle: buildpack
 custom_metric_tag_prefix_list: ["metric.tag.cloudfoundry.org"]
+
+max_manifest_service_binding_poll_duration: 60
```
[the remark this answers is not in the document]
sethboyles: seems like we add the unit for most other fields, so we will add it here

app/actions/app_apply_manifest.rb
sethboyles: We've replaced this error with the error we actually see get raised when testing with the OverviewBroker set to async only.

We're not sure if we need to handle this case:
```
result = action.bind(binding, parameters: binding_message.parameters)
if result[:async]
    raise ServiceBrokerRespondedAsyncWhenNotAllowed
end
```

According to the [OSBAPI spec](https://github.com/openservicebrokerapi/servicebroker/blob/v2.17/spec.md#asynchronous-operations), brokers should reject the request if `accepts_incomplete` is not set to true and they can not handle the request asynchronously.

spec/unit/actions/app_apply_manifest_spec.rb
sethboyles: It's not quite the same--because `binding1` and `binding2` are in `let` calls, using `and_return(binding1, binding2)` will evaluate them immediately and insert records in the database, when we don't want the records to be created until we actually call this method.

sethboyles: we may be able to use https://relishapp.com/rspec/rspec-mocks/v/3-12/docs/configuring-responses/mixed-responses if we prefer that syntax

app/models/services/service_binding_operation.rb
philippthun: Ah, thanks for the explanation.

spec/unit/actions/app_apply_manifest_spec.rb
philippthun: Ok, now I understand the difference. I did not know the "mixed responses" syntax, but I kind of like it. Maybe next time...

app/actions/app_apply_manifest.rb
philippthun: I'm unsure about this. On the one hand the OSBAPI spec is pretty clear (MUST) on the other hand it might be worth to keep this check as it has been there in the past...

What about moving the check [into the action](https://github.com/cloudfoundry/cloud_controller_ng/blob/3dcc660170c86c581b0d26ee4290028f38cbbd75/app/actions/v3/service_binding_create.rb#L27)? This would mean introducing a new error, e.g. `AsyncNotAllowed`, that needs to be handled in `AppApplyManifest`.

Two other remarks:
- The action (`ServiceBindingCreate`) also raises a `BindingNotRetrievable` error, which should be handled here as well.
- If the check is either removed or done within the action, it's not necessary to assign the binding variable.

sethboyles: Can you clarify why `BindingNotRetrievable` needs to be handled?  My understanding that if that is raised, it means that the Service Broker doesn't support fetching/polling for the Binding status, and so we should error out (i.e. keep the current behavior for those cases)

philippthun: @sethboyles You are right. Changing the error message in that case (from "The service broker responded asynchronously, but async bindings are not supported." to "The broker responded asynchronously but does not support fetching binding data") is correct.

sethboyles: @philippthun thanks for the feedback.  Regarding the `result[:async]`, I opted to readd it to AppApplyManifest, mostly because I don't have time at the moment to dig into how other parts of the code that use the `ServiceBindingCreate` action might be affected by that change.  But by keeping the check (and erroring out like before) it will hopefully be clearer when something goes wrong to users than the undefined behavior we would get if we removed the check.

philippthun (approved):

will-gant (commented):
@philippthun already granted approval and my observations are mostly stylistic, so feel free to disregard.

will-gant:
Alright, I've deployed this on a dev foundation using this branch and your one for capi-release, and can confirm it works as expected. The examples below all use a manifest specifying a service instance whose broker will only bind asynchronously:

**1. With a cloud controller that _isn't_ built from your branch (fails):**

```console
$ cf push -f manifest.yml
Pushing app cf-nodejs to org SAP / space cats-broker as provisioned_user_cf_admin...
Applying manifest file manifest.yml...

Updating with these attributes...
  ---
  applications:
  - name: cf-nodejs
    instances: 1
    memory: 512M
+   random-route: true
+   services:
+   - fake
For application 'cf-nodejs': For service 'fake': This service plan requires client support for asynchronous service operations.
FAILED
```

**2. A cloud controller that _is_ built from  your branch (succeeds):**

```console
$ cf push -f manifest.yml
Pushing app cf-nodejs to org SAP / space cats-broker as provisioned_user_cf_admin...
Applying manifest file manifest.yml...

Updating with these attributes...
  ---
  applications:
  - name: cf-nodejs
    instances: 1
    memory: 512M
+   random-route: true
+   services:
+   - fake

... 

Waiting for app cf-nodejs to start...

Instances starting...
Instances starting...
Instances starting...
Instances starting...

name:              cf-nodejs
requested state:   started
routes:            cf-nodejs-empathic-squirrel-dz.cfapps.aws-dev19.cfp.sapcloud.io
last uploaded:     Thu 01 Dec 09:04:52 UTC 2022
stack:             cflinuxfs3
buildpacks:
	name               version   detect output   buildpack name
	nodejs_buildpack   1.8.0     nodejs          nodejs

type:            web
sidecars:
instances:       1/1
memory usage:    512M
start command:   npm start
     state     since                  cpu    memory   disk     logging      details
#0   running   2022-12-01T09:05:11Z   0.0%   0 of 0   0 of 0   0/s of 0/s
```

**3. With your branch and a broker configured to say binding operation is permanently 'in progress' (fails)**

```console
$ cf push -f manifest.yml
Pushing app cf-nodejs to org SAP / space cats-broker as provisioned_user_cf_admin...
Applying manifest file manifest.yml...

Updating with these attributes...
  ---
  applications:
  - name: cf-nodejs
    instances: 1
    memory: 512M
+   random-route: true
+   services:
+   - fake
For application 'cf-nodejs': For service 'fake': Polling exceeded the maximum duration of 60
FAILED
```

sethboyles:
@will-gant thanks for testing this out!  I agree with both your stylistic comments, but am going to merge without changing them since this branch has been sitting here for a while.

PR #3064, opened 2022-11-14
PoC: Use Redis for rate limiting/request counts

johha:
When running cc with multiple puma workers the in-memory approach for rate limiting & concurrent request limiting (service brokers) is not working anymore as the worker processes are independent. Furthermore Puma distributes requests by load and thus there will be no equal request distribution.

This change replaces the in-memory parts with Redis. This also simplifies the logic in some parts as Redis keys have a TTL. Redis is supposed to be deployed on each VM and thus no HA setup, networking etc. is required.

This is only a poc!

Following things are missing or should be investigated
- Redis connection pooling
- Thread/Worker safety when interacting with Redis
- Redis integration into capi-release
- Error handling (skip rate limiting if redis is not available)


Gerg:
This sounds similar to the synchronization issue with [CC component metrics](https://docs.google.com/document/d/1PTHGMhayALG7Q9sBhBqiBcUm6_IiAIOiUm9KtdHwpME/edit) when running Puma. Would it be worthwhile to evaluate a similar solution here, namely using [the filesystem](https://github.com/prometheus/client_ruby#directfilestore-caveats-and-things-to-keep-in-mind), to sync state between workers?

cc @sethboyles

johha:
> This sounds similar to the synchronization issue with [CC component metrics](https://docs.google.com/document/d/1PTHGMhayALG7Q9sBhBqiBcUm6_IiAIOiUm9KtdHwpME/edit) when running Puma. Would it be worthwhile to evaluate a similar solution here, namely using [the filesystem](https://github.com/prometheus/client_ruby#directfilestore-caveats-and-things-to-keep-in-mind) to sync state between workers?

We discussed this in the capi open office hour last week. Implementing rate limiting and concurrent service broker request limiting (counting semaphore) with redis is rather simple. I haven't looked yet too much into `DirectFileStore` but I think this would require more effort. Plus redis could be easily co-deployed on the API VMs like [bosh](https://github.com/cloudfoundry/bosh/tree/stable-3184.1/release/jobs/redis) used to do.

Gerg:
I'm not inherently opposed to the idea, but I think we should also consider the other options, before deciding on Redis. Ease of implementation isn't the only factor we should consider.

My concerns about Redis:
1. It adds another dependency to capi-release that we need to keep patched/updated and another job that we need to maintain.
2. It adds another running process to api instances, competing for system resources.
3. It adds another failure point for capi-release. Redis crashes or bugs could make CC unavailable, depending on how we integrate with it.

If we are using a co-located Redis for each api instance, it's basically just a heavier-weight abstraction for the filesystem. Not saying that that's definitely a bad thing, but we should consider the trade offs.

@rkoster do you know the history behind why bosh removed its Redis dependency?

rkoster:
@cunnie Do you have more context on why bosh moved away from Redis, given that you were involved in the above-linked commits.

johha:
> I'm not inherently opposed to the idea, but I think we should also consider the other options, before deciding on Redis. Ease of implementation isn't the only factor we should consider.
> 
> My concerns about Redis:
> 
>     1. It adds another dependency to capi-release that we need to keep patched/updated and another bosh job that we need to maintain.
> 
>     2. It adds another running process to api instances, competing for system resources.
> 
>     3. It adds another failure point for capi-release. Redis crashes or bugs could make CC unavailable, depending on how we integrate with it.
> 
> 
> If we are using a co-located Redis for each api instance, it's basically just a heavier-weight abstraction for that instance's filesystem. Not saying that that's definitely a bad thing, but we should consider the trade-offs.


Instead of Redis we could also consider [Memcached](https://github.com/memcached/memcached) which is a simpler key/value store compared to Redis (plus it's in-memory).
Prometheus with DirectFileStore doesn't seem to be a good option as we would need a label for each user guid which is [not recommended](https://prometheus.io/docs/practices/naming/#labels):

> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values. 

Alternatives to Redis/Memcached:
- Move rate limiting to Nginx
- Implement a ruby process for managing request counts in-memory 

We made a list of possible solutions: https://sap-my.sharepoint.com/:w:/p/johannes_haass/EYtAtWAm6Y1MjROPwlimSjIB-ERmZMNwJMCdCgi8zWADsA?e=28WwiK

@sethboyles @Gerg @stephanme Maybe we can discuss this in the next capi open office hour (on Dec 14th) and try to come to a decision.

philippthun:
Just two small remarks.

- We would use Redis in a way that it does not persist data to the filesystem, so it would also be completely in-memory.
- Redis can also be used as Custom Data Store for metrics, so we could use it for both rate limiting and metrics. See https://github.com/gocardless/prometheus-client-ruby-data-stores-experiments/blob/master/data_stores/redis.rb

philippthun:
Work is continued in #3290.

PR #3067, opened 2022-11-17
Remove more inefficient counting

will-gant:

Small optimizations to a few DB queries related to counting.

Two small things:
1. In a few more places, switches to use `Sequel`'s `any?` and `empty?` methods instead of `count` when all we want to know is whether a non-zero number of matching records exist. Essentially a follow-up to #3052
2. In a few places we use formulations like `resource.associated_resources.map(&:other_associated_resources)`, which results in a single query to fetch `associated_resources` and then, for each one found, a separate query to count its `other_associated_resources`. This can be replaced with a single query with a join and `empty?`\\`any?`.

n/a

### DIFF
--- a/app/actions/service_offering_delete.rb
+++ b/app/actions/service_offering_delete.rb
@@ -3,7 +3,7 @@ module VCAP::CloudController
     class AssociationNotEmptyError < StandardError; end
 
     def delete(service_offering_model)
-      association_not_empty! unless service_offering_model.service_plans.empty?
+      association_not_empty! unless ServicePlan.where(service_id: service_offering_model.id).empty?
       service_offering_model.destroy
     end
 
--- a/app/actions/service_offering_delete.rb
+++ b/app/actions/service_offering_delete.rb
@@ -3,7 +3,7 @@ module VCAP::CloudController
     class AssociationNotEmptyError < StandardError; end
 
     def delete(service_offering_model)
-      association_not_empty! unless ServicePlan.where(service_id: service_offering_model.id).empty?
+      association_not_empty! if ServicePlan.where(service_id: service_offering_model.id).any?
       service_offering_model.destroy
     end
 
--- a/app/actions/space_delete.rb
+++ b/app/actions/space_delete.rb
@@ -82,7 +82,7 @@ module VCAP::CloudController
     def delete_service_brokers(space_model)
       broker_remover = service_broker_remover(@services_event_repository)
       private_service_brokers = space_model.service_brokers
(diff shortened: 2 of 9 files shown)


app/actions/service_offering_delete.rb
```diff
@@ -3,7 +3,7 @@ class ServiceOfferingDelete
     class AssociationNotEmptyError < StandardError; end
 
     def delete(service_offering_model)
-      association_not_empty! unless service_offering_model.service_plans.empty?
+      association_not_empty! unless ServicePlan.where(service_id: service_offering_model.id).empty?
```
philippthun, line 6: `if ... any?` instead of `unless empty?`

app/actions/space_delete.rb
```diff
@@ -82,7 +82,7 @@ def delete_apps(space_model)
     def delete_service_brokers(space_model)
       broker_remover = service_broker_remover(@services_event_repository)
       private_service_brokers = space_model.service_brokers
-      deletable_brokers = private_service_brokers.reject { |broker| broker.service_plans.map(&:service_instances).flatten.any? }
+      deletable_brokers = private_service_brokers.reject { |broker| broker.service_plans.map { |plan| plan.service_instances_dataset.any? }.flatten.any? }
```
philippthun, line 85: I actually had a look at this a few days ago and came up with the following:
```
deletable_brokers = private_service_brokers.reject do |broker|
  # Brokers with existing service instances are not deletable and thus rejected from the array.
  ServiceInstance.
    join(:service_plans, id: :service_instances__service_plan_id).
    join(:services, id: :service_plans__service_id).
    where(services__service_broker_id: broker.id).
    any?
end
```

Its (a) reducing the number of queries (m instead of m * n) and (b) reducing the amount of data fetched from the database (m boolean values instead of m service plans * n service instances).

lib/services/service_brokers/v2/catalog.rb
```diff
@@ -122,7 +122,7 @@ def taken_names(new_services_names)
     end
 
     def can_delete_service?(service)
-      service.service_plans_dataset.map(&:service_instances_dataset).map(&:count).all?(0)
+      service.service_plans_dataset.all? { |plan| plan.service_instances_dataset.empty? }
```
philippthun, line 125: I think this could also be combined into a single SELECT statement with JOINs.

Did you check the actual SQL statements being executed?

app/actions/service_offering_delete.rb
will-gant: Sure

philippthun (approved):

PR #3070, opened 2022-11-21
Use more efficient 'select' in place of 'select_map'

will-gant:

Fixes an anti-optimization introduced by mistake in #3043, whereby several methods that make DB queries were changed from `select` (which returns a `Sequel::Dataset` to be executed later) to `select_map` (which immediately executes a query and returns the result as an array).

```ruby
def foo1
  Foo.select(:id)
end

def foo2
  Foo.select_map(:id)
end
```

In the above example, `foo2` is an efficient way to get a list of ids when our ruby code needs these for something. But it turns out `foo1` is much more efficient when we want to use those ids as part of another Sequel query:

```ruby
Bar.where(foo_ids: foo1)
```

We noticed that (repeated) runs of the [cf-performance-tests](https://github.com/cloudfoundry/cf-performance-tests) showed the newly-released [CAPI 1.141.0](https://github.com/cloudfoundry/capi-release/releases/tag/1.141.0), which includes the changes from #3043, gave a much slower response to `GET /v3/domains` and `GET /v3/organizations/:guid/domains` (for the former, 8.51 seconds instead of 0.69). 

The tests made requests with a non-admin user and a DB seeded with 20k orgs (in which the user was an org manager in each), plus 400 private domains and 100 shared domains. These screenshots cut off the version numbers way below - but the sudden increase is from 1.141.0:
Screenshot 2022-11-21 at 15 43 46
Screenshot 2022-11-21 at 15 43 55

@philippthun correctly surmised that the root cause would be one of the places I'd switched to `select_map`. Here it was `membership.authorized_org_guids_subquery(ORG_ROLES_FOR_READING_DOMAINS_FROM_ORGS + SPACE_ROLES)`, which #3043 changed so it would call `member_org_ids`:

```ruby
def member_org_ids(roles)
  org_role_subqueries(roles).reduce do |query, subquery|
    query.union(subquery, from_self: false)
  end&.select_map(:organization_id)
end
```

```sql
SELECT     "guid"
FROM       "organizations"
INNER JOIN "organizations_private_domains"
ON         (
                      "organizations_private_domains"."organization_id" = "organizations"."id" )
WHERE      ( (
                                 "organizations_private_domains"."private_domain_id" = 1 )
           AND        (
                                 "guid" IN (xxx) )
           OR         (
                                 "id" IS NULL ) )) ) )
```

#3043 made `xxx` an array of guid strings, retrieved in a separate query, whereas prior to that PR (and in the current PR) it's a subquery that retrieves those guids.

Passing an array of guids was in fact a little faster _in the DB_ than passing a subquery (22.228 ms planning + 0.105 ms execution versus 2.973 ms and 30.199 ms planning before #3043), and in Kibana I observed less querytime overall. But this was outweighed by an enormous, silent gap in time between each query in the logs, leading to a slower request overall. Having played around with these queries locally and frozen my machine a few times, I think that's what comes from the host passing round arrays that hold tens of thousands of strings.

Since the overall request time is more relevant here, here are some speeds for `cf curl /v3/domains` timed from bash:

* c9a2ca978fef70cd2c634fc5415f4faaeade8d34 (pre-#⁠3043): 0.758 seconds
* 0b5e6e38e522da496799479591721a60fb7fef93 (#⁠3043): 8.512 seconds
* b14c6e5dc3d66f80ecbbe558920c4cc65a0c48cb (#⁠3070): 0.692 seconds

*Links to any other associated PRs*

#3043

### DIFF
--- a/lib/cloud_controller/membership.rb
+++ b/lib/cloud_controller/membership.rb
@@ -29,8 +29,8 @@ module VCAP::CloudController
     end
 
     def authorized_orgs_subquery(roles)
-      space_ids = member_space_ids(roles)
-      org_ids_from_space_roles = Space.where(id: space_ids).select(:organization_id) if space_ids
+      space_ids_query = member_space_ids(roles)
+      org_ids_from_space_roles = Space.where(id: space_ids_query).select(:organization_id) if space_ids_query
 
       if org_ids_from_space_roles
         Organization.where(id: member_org_ids(roles)).or(id: org_ids_from_space_roles)
@@ -59,13 +59,13 @@ module VCAP::CloudController
     def member_space_ids(roles)
       space_role_subqueries(roles).reduce do |query, subquery|
         query.union(subquery, from_self: false)
-      end&.select_map(:space_id)
+      end&.select(:space_id)
     end
 
     def member_org_ids(roles)
       org_role_subqueries(roles).reduce do |query, subquery|
         query.union(subquery, from_self: false)
-      end&.select_map(:organization_id)
+      end&.select(:organization_id)
     end
 
     private
@@ -73,13 +73,13 @@ module VCAP::CloudController
     def role_applies_to_space?(roles, space_id)
       return false unless space_id && contains_space_role?(roles)
 
-      member_space_ids(roles).include?(space_id)
+      member_space_ids(roles).select_map(:space_id).include?(space_id)
     end
 
     def role_applies_to_org?(roles, org_id)
       return false unless org_id && contains_org_role?(roles)
 
-      member_org_ids(roles).include?(org_id)
+      member_org_ids(roles).select_map(:organization_id).include?(org_id)
     end
 
     def roles_filter(roles, filter)
--- a/lib/cloud_controller/membership.rb
+++ b/lib/cloud_controller/membership.rb
@@ -73,13 +73,13 @@ module VCAP::CloudController
     def role_applies_to_space?(roles, space_id)
       return false unless space_id && contains_space_role?(roles)
 
-      member_space_ids(roles).select_map(:space_id).include?(space_id)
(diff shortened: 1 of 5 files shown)


lib/cloud_controller/membership.rb
```diff
@@ -59,27 +59,27 @@ def authorized_spaces_subquery(roles)
...
+      end&.select(:organization_id)
     end
 
     private
 
     def role_applies_to_space?(roles, space_id)
       return false unless space_id && contains_space_role?(roles)
 
-      member_space_ids(roles).include?(space_id)
+      member_space_ids(roles).select_map(:space_id).include?(space_id)
```
philippthun, line 76: The check could be moved to the database. The current statement selects all space ids and then checks if the given id is included in the list. This can be changed to something like:
```
member_space_ids(roles).where(space_id: space_id).any?
```

lib/cloud_controller/membership.rb
```diff
@@ -59,27 +59,27 @@ def authorized_spaces_subquery(roles)
...
     end
 
     def role_applies_to_org?(roles, org_id)
       return false unless org_id && contains_org_role?(roles)
 
-      member_org_ids(roles).include?(org_id)
+      member_org_ids(roles).select_map(:organization_id).include?(org_id)
```
philippthun, line 82: Same as above.

lib/cloud_controller/membership.rb
will-gant: Excellent point! That should massively cut down scans of role tables

will-gant: Done this, but refactored slightly so that the extra `where` clause is made in each of the subqueries on the role tables that we're doing a `union` on, as I gather this is faster than putting it at a 'higher' level.

PR #3077, opened 2022-11-25
Optimize space supporter filter for v2 calls

will-gant:
_**UPDATE: I've completely changed this optimization since first opening the PR so this description is out of date (I'm leaving it as a record of how this has progressed) - see comments below for latest details**_

Optimizes the checks that prevent users with only a space supporter role from accessing the v2 API, cutting around 7% off the overall querytime when there are large tables.

This PR uses Sequel's [`any?`](https://sequel.jeremyevans.net/rdoc-plugins/files/lib/sequel/extensions/any_not_empty_rb.html) method instead of the existing check, which retrieves [`.all`](https://www.rubydoc.info/github/jeremyevans/sequel/Sequel%2FDataset:all) of a given user's roles and then, in the CC's ruby code, counts them to see whether the user has exactly 1 space supporter role and zero of all other role types.

The method will also now return false right away, without querying the space supporters table, if the user has any non-supporter role.

An explanation of the use cases your change solves

The existing check currently adds a pretty significant delay to every v2 call on large foundations, and this change slightly mitigates that.

I've run the [old](https://gist.github.com/will-gant/2d38d02bd16b1e016e931cda60d0fdef) and the [new](https://gist.github.com/will-gant/b3dda4513ced76c45dc4176892c9199c) queries in postgres with `EXPLAIN ANALYZE` 100 times each on a foundation with about 75k users and 646k roles. Averages from the results are presented in the table below (times in ms):

|                | Old      | New      |
|----------------|----------|----------|
| Planning time  |  1.50904 |  1.06917 |
| Execution time | 36.42587 |   34.286 |
| Overall time   | 37.93491 | 35.35517 |

If the user lacks any non-supporter role, we then have to separately check the supporters table. That [little query](https://gist.github.com/will-gant/df8d92d7df34b565e5e26dd934201419) has a planning time of 0.186ms, and execution 1.366ms. So we're still a smidgen faster even for the minority of requests that also need to execute that.

So it's a pretty modest improvement, but given that this is middleware that gets executed for all v2 calls, I think cumulatively it should be a useful saving.

Links to any other associated PRs

n/a

### DIFF
--- a/middleware/block_v3_only_roles.rb
+++ b/middleware/block_v3_only_roles.rb
@@ -36,21 +36,21 @@ module CloudFoundry
       end
 
       def space_supporter_and_only_space_supporter?(current_user)
-        return false if VCAP::CloudController::User.
+        return false if VCAP::CloudController::User.db.select(1).
                         where do
                           Sequel.or(
                             [
-                              [VCAP::CloudController::SpaceDeveloper.select(1).where(user_id: current_user.id).exists, true],
-                              [VCAP::CloudController::OrganizationManager.select(1).where(user_id: current_user.id).exists, true],
-                              [VCAP::CloudController::SpaceManager.select(1).where(user_id: current_user.id).exists, true],
-                              [VCAP::CloudController::SpaceAuditor.select(1).where(user_id: current_user.id).exists, true],
-                              [VCAP::CloudController::OrganizationAuditor.select(1).where(user_id: current_user.id).exists, true],
-                              [VCAP::CloudController::OrganizationBillingManager.select(1).where(user_id: current_user.id).exists, true]
+                              [VCAP::CloudController::SpaceDeveloper.limit(1).select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::OrganizationManager.limit(1).select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::SpaceManager.limit(1).select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::SpaceAuditor.limit(1).select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::OrganizationAuditor.limit(1).select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::OrganizationBillingManager.limit(1).select(1).where(user_id: current_user.id).exists, true]
                             ]
                           )
                         end.any?
 
-        VCAP::CloudController::SpaceSupporter.select(1).where(user_id: current_user.id).any?
+        VCAP::CloudController::SpaceSupporter.where(user_id: current_user.id).any?
       end
 
       def roles
--- a/middleware/block_v3_only_roles.rb
+++ b/middleware/block_v3_only_roles.rb
@@ -36,15 +36,21 @@ module CloudFoundry
       end
 
       def space_supporter_and_only_space_supporter?(current_user)
-        return false unless current_user.org_manager_org_ids.limit(1).
-                            union(current_user.org_billing_manager_org_ids.limit(1), from_self: false).
-                            union(current_user.org_auditor_org_ids.limit(1), from_self: false).
-                            union(current_user.space_manager_space_ids.limit(1), from_self: false).
-                            union(current_user.space_auditor_space_ids.limit(1), from_self: false).
-                            union(current_user.space_developer_space_ids.limit(1), from_self: false).
-                            empty?
+        return false if VCAP::CloudController::User.
+                        where do
+                          Sequel.or(
+                            [
+                              [VCAP::CloudController::SpaceDeveloper.select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::OrganizationManager.select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::SpaceManager.select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::SpaceAuditor.select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::OrganizationAuditor.select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::OrganizationBillingManager.select(1).where(user_id: current_user.id).exists, true]
+                            ]
+                          )
+                        end.any?
 
-        current_user.space_supporter_space_ids.limit(1).any?
+        VCAP::CloudController::SpaceSupporter.select(1).where(user_id: current_user.id).any?
       end
 
       def roles
--- a/middleware/block_v3_only_roles.rb
+++ b/middleware/block_v3_only_roles.rb
@@ -36,20 +36,15 @@ module CloudFoundry
       end
 
       def space_supporter_and_only_space_supporter?(current_user)
-        user_roles =
-          VCAP::CloudController::SpaceManager.select(Sequel.as(VCAP::CloudController::RoleTypes::SPACE_MANAGER, :type)).limit(1).where(user_id: current_user.id).
-          union(VCAP::CloudController::SpaceDeveloper.select(Sequel.as(VCAP::CloudController::RoleTypes::SPACE_DEVELOPER, :type)).limit(1).where(user_id: current_user.id)).
-          union(VCAP::CloudController::SpaceAuditor.select(Sequel.as(VCAP::CloudController::RoleTypes::SPACE_AUDITOR, :type)).limit(1).where(user_id: current_user.id)).
-          union(VCAP::CloudController::SpaceSupporter.select(Sequel.as(VCAP::CloudController::RoleTypes::SPACE_SUPPORTER, :type)).limit(1).where(user_id: current_user.id)).
-          union(VCAP::CloudController::OrganizationManager.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_MANAGER,
-                                                                            :type)).limit(1).where(user_id: current_user.id)).
-          union(VCAP::CloudController::OrganizationBillingManager.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_BILLING_MANAGER,
-                                                                                   :type)).limit(1).where(user_id: current_user.id)).
-          union(VCAP::CloudController::OrganizationAuditor.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_AUDITOR,
-                                                                            :type)).limit(1).where(user_id: current_user.id)).
-          all
+        return false unless current_user.org_manager_org_ids.limit(1).
+                            union(current_user.org_billing_manager_org_ids.limit(1), from_self: false).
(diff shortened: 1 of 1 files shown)


philippthun (commented):
The main difference between your query and the original one is the `from_self: false` parameter to the `union` method invocations. That's why the original query is nested (e.g. `SELECT * FROM ( SELECT * FROM ...`) and yours is not.

I don't think that `.all` vs. `.any?` makes a huge difference here, as the returned list from the old query consists of 0 to 7 type strings. That's why I would be in favor of keeping the original single query instead of having (up to) two queries.

Of course, adding the `from_self: false` to the original query would still make sense.

will-gant:
> I don't think that .all vs. .any? makes a huge difference here, as the returned list from the old query consists of 0 to 7 type strings. That's why I would be in favor of keeping the original single query instead of having (up to) two queries.

So I think the difference is that with `.any?` you get a `SELECT 1` - i.e. check whether the row exists, don't retrieve any values - which is cheaper than the `SELECT *` of `.all`. I do agree it's a teeny tiny thing, and admittedly only at the top level, though it still seems we might as well take it. But I could improve it by getting rid of those unnecessary selects at the lower level.

But the main thing is that while space supporter roles can apparently be quite common (almost 10% of all roles on this foundation), users who _only_  have a space supporter role seem to be very rare. In fact, this foundation has just _one_ space supporter role (out of 60k) that belongs to a user who has no other role (not counting organization users here, since any user with a space role has that too). That means in, like, 99% of cases we'll avoid that separate scan of the space supporters table. And the numbers show that when we actually do run both queries the combined querytime is about the same as it is for a single, bigger query scanning every role table. So we don't seem to lose even then.

will-gant:
@philippthun Your feedback made me ponder this a bit more deeply, and I think I've come up with a much better solution which I hope will turn your frown upside down. Although it still retains the separate query for space supporters.

My latest commit produces a SQL query that looks like this:

```sql
EXPLAIN ANALYZE SELECT 1 AS "one"
FROM   "users"
WHERE  ( ( ( EXISTS (SELECT 1
                     FROM   "spaces_developers"
                     WHERE  ( "user_id" = 2023961 )) ) IS TRUE )
          OR ( ( EXISTS (SELECT 1
                         FROM   "organizations_managers"
                         WHERE  ( "user_id" = 2023961 )) ) IS TRUE )
          OR ( ( EXISTS (SELECT 1
                         FROM   "spaces_managers"
                         WHERE  ( "user_id" = 2023961 )) ) IS TRUE )
          OR ( ( EXISTS (SELECT 1
                         FROM   "spaces_auditors"
                         WHERE  ( "user_id" = 2023961 )) ) IS TRUE )
          OR ( ( EXISTS (SELECT 1
                         FROM   "organizations_auditors"
                         WHERE  ( "user_id" = 2023961 )) ) IS TRUE )
          OR ( ( EXISTS (SELECT 1
                         FROM   "organizations_billing_managers"
                         WHERE  ( "user_id" = 2023961 )) ) IS TRUE ) )
LIMIT  1;
```

The cool thing about it is that, unlike with a `UNION`, it will short-circuit as soon as a match is found. That makes it almost 20x faster if the user has a role in the first table checked, about 2x faster if they're in the second, and for the others it's comparable to the old query (slightly faster even for those). I made space developers first and org managers second because they're the first and second most-used roles on a lot of foundations I've seen, though we've no way of knowing if they make the most _requests_.

For each of the tables below, I retrieved a sample user_id for a user who had _that role only_ (except OrganizationUser) and then ran the query 100 times for that user id. So the numbers are all averages.

Here's for the old query:
| Old                 | Rows count | Planning | Execution | Total    |
|---------------------|------------|----------|-----------|----------|
| Space Developer     |     174774 |  1.45321 |   20.2723 | 21.72551 |
| Org Manager         |     118982 |  1.50626 |  32.64812 | 34.15438 |
| Space Manager       |     100263 |  1.48238 |  33.85913 | 35.34151 |
| Space Auditor       |     105297 |  1.54372 |  26.27451 | 27.81823 |
| Org Auditor         |      53303 |  1.55326 |  35.10541 | 36.65867 |
| Org Billing Manager |        175 |  1.51813 |  33.70744 | 35.22557 |

And here's for the new one that I committed in 2db8b34a59bcb2a2e7f76fc1dd7debdaddec0a35:

| New                 | Rows count | Planning | Execution | Total    | % speed of the old |
|---------------------|------------|----------|-----------|----------|--------------------|
| Space Developer     |     174774 |  1.25895 |   0.10354 |  1.36249 |              6.27% |
| Org Manager         |     118982 |  1.26675 |  14.28919 | 15.55594 |             45.55% |
| Space Manager       |     100263 |  1.25602 |  32.60624 | 33.86226 |             95.81% |
| Space Auditor       |     105297 |  1.28501 |  23.51924 | 24.80425 |             89.17% |
| Org Auditor         |      53303 |  1.25375 |  33.34759 | 34.60134 |             94.39% |
| Org Billing Manager |        175 |  1.32794 |  32.54151 | 33.86945 |             96.15% |

middleware/block_v3_only_roles.rb
```diff
@@ -36,20 +36,21 @@ def globally_authenticated?
...
-          union(VCAP::CloudController::OrganizationAuditor.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_AUDITOR,
-                                                                            :type)).limit(1).where(user_id: current_user.id)).
-          all
-
-        user_roles.count == 1 && user_roles[0][:type] == 'space_supporter'
+        return false if VCAP::CloudController::User.
+                        where do
+                          Sequel.or(
+                            [
+                              [VCAP::CloudController::SpaceDeveloper.select(1).where(user_id: current_user.id).exists, true],
```
philippthun, line 43: You can also add `.limit(1)` to all the sub-selects.

middleware/block_v3_only_roles.rb
```diff
@@ -36,20 +36,21 @@ def globally_authenticated?
...
+                              [VCAP::CloudController::OrganizationManager.select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::SpaceManager.select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::SpaceAuditor.select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::OrganizationAuditor.select(1).where(user_id: current_user.id).exists, true],
+                              [VCAP::CloudController::OrganizationBillingManager.select(1).where(user_id: current_user.id).exists, true]
+                            ]
+                          )
+                        end.any?
+
+        VCAP::CloudController::SpaceSupporter.select(1).where(user_id: current_user.id).any?
```
philippthun, line 53: Here the `.select(1)` is not needed. It's added by the `.any?`.

will-gant: Good point

middleware/block_v3_only_roles.rb
will-gant: Nice idea!

middleware/block_v3_only_roles.rb
```diff
@@ -36,20 +36,21 @@ def globally_authenticated?
...
-          union(VCAP::CloudController::OrganizationManager.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_MANAGER,
-                                                                            :type)).limit(1).where(user_id: current_user.id)).
-          union(VCAP::CloudController::OrganizationBillingManager.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_BILLING_MANAGER,
-                                                                                   :type)).limit(1).where(user_id: current_user.id)).
-          union(VCAP::CloudController::OrganizationAuditor.select(Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_AUDITOR,
-                                                                            :type)).limit(1).where(user_id: current_user.id)).
-          all
-
-        user_roles.count == 1 && user_roles[0][:type] == 'space_supporter'
+        return false if VCAP::CloudController::User.
```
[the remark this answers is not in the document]
will-gant: Ahh nice one! I spent ages trying to figure out how to get Sequel to produce that. Thanks.

philippthun (approved):

philippthun:
It looks like there is something that old MySQL servers don't like:
```
--- Caused by: ---
Mysql2::Error:
  You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'WHERE (((EXISTS (SELECT 1 FROM `spaces_developers` WHERE (`user_id` = 489) LIMIT' at line 1
```

I hope you have the correct manual at hand ;-)

will-gant:
> It looks like there is something that old MySQL servers don't like:
> 
> ```
> --- Caused by: ---
> Mysql2::Error:
>   You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'WHERE (((EXISTS (SELECT 1 FROM `spaces_developers` WHERE (`user_id` = 489) LIMIT' at line 1
> ```
> 
> I hope you have the correct manual at hand ;-)

Whaaaat

will-gant:
Alright this seems to be a bug in Sequel, which generates a [query](https://gist.github.com/will-gant/1fe8a359dca3eeed136eec10caf024a9) that is invalid for mysql 5.7. Though the reference manuals for mysql [5.7](https://dev.mysql.com/doc/refman/5.7/en/select.html) and [8.0](https://dev.mysql.com/doc/refman/8.0/en/select.html) don't hint at any such change in behaviour, the former will explode if you use a `WHERE` without a `FROM`:

```mysql
mysql> SELECT
    ->   1
    -> WHERE
    ->   (
    ->     EXISTS (
    ->       SELECT
    ->         1
    ->       FROM
    ->         spaces_developers
    ->       WHERE
    ->         'user_id' = 1
    ->       LIMIT
    ->         1
    ->     )
    ->   );
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'WHERE
  (
    EXISTS (
      SELECT
        1
      FROM
        spaces_deve' at line 3
mysql> SELECT
    ->   1 FROM users
    -> WHERE
    ->   (
    ->     EXISTS (
    ->       SELECT
    ->         1
    ->       FROM
    ->         spaces_developers
    ->       WHERE
    ->         'user_id' = 1
    ->       LIMIT
    ->         1
    ->     )
    ->   );
ERROR 1046 (3D000): No database selected
```

With this added it's happy:
```mysql
mysql> SELECT
    ->   1 FROM users
    -> WHERE
    ->   (
    ->     EXISTS (
    ->       SELECT
    ->         1
    ->       FROM
    ->         spaces_developers
    ->       WHERE
    ->         'user_id' = 1
    ->       LIMIT
    ->         1
    ->     )
    ->   );
Empty set, 1 warning (0.00 sec)
```

@philippthun I'll open an issue with [jeremyevans/sequel](https://github.com/jeremyevans/sequel), but perhaps we can just merge this with my original `SELECT 1 FROM users WHERE...`, which is still quite a bit faster than the status quo?

philippthun:
> Perhaps we could revert to my original `SELECT 1 FROM users WHERE...`, given that that's still quite a bit faster than the status quo?

Yes, let's do this.

PR #3106, opened 2022-12-14
Use space subqueries for fetchting service credential bindings

johha:
As a regular user the select query for service credential bindings is filtered by a list of space guids (spaces the user has access to).
This can produce very long db queries. Instead we can use subqueries to filter the spaces directly in the db. The same approach is used in service lists, see #2580.


app/controllers/v3/service_credential_bindings_controller.rb
```diff
@@ -328,14 +328,14 @@ def fetch_credentials_value(name)
...
   end
 
-  def space_guids
+  def spaces_query
     if permission_queryer.can_read_globally?
-      :all
+      nil
     else
-      permission_queryer.readable_space_guids
+      permission_queryer.readable_space_guids_query
```
philippthun, line 338: I would create a new method in `VCAP::CloudController::Permissions` named `readable_spaces_query` that uses `membership.authorized_spaces_subquery` instead of `membership.authorized_space_guids_subquery`.

app/models/services/service_credential_binding_view.rb
```diff
@@ -10,7 +10,7 @@ module Types
         Sequel.as(:service_keys__id, :id),
         Sequel.as(:service_keys__guid, :guid),
         Sequel.as(Types::SERVICE_KEY, :type),
-        Sequel.as(:spaces__guid, :space_guid),
+        Sequel.as(:spaces__id, :space_id),
```
philippthun, line 13: I think you can change the alias to `Sequel.as(:service_instances__space_id, :space_id)` and remove [this JOIN](https://github.com/cloudfoundry/cloud_controller_ng/blob/ecbc74972ad27cf1e0055d6760ab35cb185ff170/app/models/services/service_credential_binding_view.rb#L40).

app/models/services/service_credential_binding_view.rb
```diff
@@ -36,15 +36,13 @@ module Types
...
         :services, id: Sequel[:service_plans][:service_id]
-      ).join(
-        :spaces, id: Sequel[:service_instances][:space_id]
       ).freeze
 
       SERVICE_BINDING_VIEW = Sequel::Model(:service_bindings).select(
         Sequel.as(:service_bindings__id, :id),
         Sequel.as(:service_bindings__guid, :guid),
         Sequel.as(Types::SERVICE_BINDING, :type),
-        Sequel.as(:spaces__guid, :space_guid),
```
philippthun, line 47: I think for service bindings the JOIN is still required.

philippthun (approved):

PR #3107, opened 2022-12-14
Update metric tags on rename

acrmp:

**⚠️ Requires Diego changes to be completed, merged and released ⚠️** 

Modifies the cloud controller to send a DesiredLRPUpdate with updated metric tags to Diego when an application is renamed.

* ~~The diego client will be using the new `/v1/desired_lrp/update.r1` BBS endpoint.~~
* The protobuf DesiredLRPUpdate message now includes the metric tags.
* If the app is started we issue a DesiredLRPUpdate message for each process on the app.
* If the app is staging we don't issue the DesiredLRPUpdate. We hadn't thought too hard about this so let us know if this makes sense or not.
* We don't issue the DesiredLRPUpdate when the v2 API is used. The sync should take care of this with a small delay.

We noticed that the CAPI pipeline uses [protoc v3.6.1](https://github.com/protocolbuffers/protobuf/releases/tag/v3.6.1) which is quite an old version. We used that version to regenerate the protobuf client code in this PR but it may be worth considering upgrading. We also wondered if the [associated concourse task script](https://github.com/cloudfoundry/capi-ci/blob/39c9b5fe491baa47882779664fd61bc40e0ba24e/ci/protos/bump_bbs_protos.sh) should include `set -e`.

Pending changes to Diego, this should allow renames to take effect without requiring an application restart.

https://github.com/cloudfoundry/diego-release/issues/662

### DIFF
--- a/app/actions/app_update.rb
+++ b/app/actions/app_update.rb
@@ -33,14 +33,6 @@ module VCAP::CloudController
 
         raise CloudController::Errors::ApiError.new_from_details('CustomBuildpacksDisabled') if using_disabled_custom_buildpack?(app)
 
-        if message.requested?(:name)
-          app.processes.each do |process|
-            @runners.runner_for_process(process).update_metric_tags if process.state == ProcessModel::STARTED
-          rescue Diego::Runner::CannotCommunicateWithDiegoError => e
-            @logger.error("failed communicating with diego backend: #{e.message}")
-          end
-        end
-
         Repositories::AppEventRepository.new.record_app_update(
           app,
           app.space,
@@ -48,6 +40,14 @@ module VCAP::CloudController
           message.audit_hash,
           manifest_triggered: @manifest_triggered
         )
+      end
+
+      if message.requested?(:name)
+        app.processes.each do |process|
+          @runners.runner_for_process(process).update_metric_tags if process.state == ProcessModel::STARTED
+        rescue Diego::Runner::CannotCommunicateWithDiegoError => e
+          @logger.error("failed communicating with diego backend: #{e.message}")
+        end
       end
 
       app
--- a/app/actions/app_update.rb
+++ b/app/actions/app_update.rb
@@ -7,10 +7,11 @@ module VCAP::CloudController
     class DropletNotFound < StandardError; end
     class InvalidApp < StandardError; end
 
-    def initialize(user_audit_info, manifest_triggered: false)
+    def initialize(user_audit_info, manifest_triggered: false, runners: nil)
       @user_audit_info = user_audit_info
       @logger = Steno.logger('cc.action.app_update')
       @manifest_triggered = manifest_triggered
+      @runners = runners || CloudController::DependencyLocator.instance.runners
     end
 
     def update(app, message, lifecycle)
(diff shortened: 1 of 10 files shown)


Gerg:
Without looking at this PR in much detail, it appears y'all implemented this differently than I would have.

I would have expected the change to go through the `ProcessObserver`. For more (overwhelming) detail, see [this wiki page](https://github.com/cloudfoundry/cloud_controller_ng/wiki/When-and-How-CC-Updates-the-Runtime).

acrmp:
@Gerg Thanks. We chatted about this with @sethboyles and @dalvarado and agreed to modify the action rather than changing process observer. @sethboyles liked having the logic in the action and it wasn't clear how to access the app name change from process observer.

acrmp:
@sethboyles @dalvarado Thanks for the feedback yesterday. We pushed an additional commit d9539459fe443eeb130 to move the Desired LRP Update outside of the database transaction.

acrmp:
@sethboyles @dalvarado We pushed another commit 1afdda2a14f9acbadf4d18ef41b36cf74cd258f5 to not refer to a new revision of the desired lrp update endpoint. We agreed with @winkingturtle-vmw that introducing a new revision wasn't necessary.

acrmp:
@sethboyles We saw that the changes in this PR currently aren't sufficient to ensure that the metric tags are updated when the initial call from CAPI to the BBS fails.

The processes sync job [uses the process updated_at timestamp to determine whether it needs to update the diego process](https://github.com/cloudfoundry/cloud_controller_ng/blob/39b4dfed2a5c19e1a7c8798e82366b85a9642588/lib/cloud_controller/diego/processes_sync.rb#L36), and this hasn't changed when an app is renamed.

1. One alternative is to update the timestamp on associated processes when renaming the app so that the convergence loop will notice and synchronise with the BBS. This can be as simple as calling save on each process, though it seems strange to call it just for the side effect of updating the updated at timestamp. Looking into testing this approach and asserting that the process timestamp is updated in the database we also saw that it doesn't play nicely with Timecop because we have a sequel current_timestamp extension that will ignore the Timecop clock.

1. Another alternative might be to change the processes sync job to also refer to the app updated_at when determining whether to update the process at the BBS. This wouldn't require the process timestamps to be touched but we're not sure if there are other app updates that we might want not to cause updates to the BBS.

Do y'all have a preferred option?

Gerg:
If you do end up changing the sync logic, I'll draw your attention to https://github.com/cloudfoundry/sync-integration-tests.

acrmp:
> Another alternative might be to change the processes sync job to also refer to the app updated_at when determining whether to update the process at the BBS.
Though thinking about it that might require saying where annotation < app updated_at which means we'd be going from treating the annotation as an opaque field to requiring it to be an epoch time. I don't think I like that.

acrmp:
We pushed a commit that updates the process timestamps when renaming an app to ensure that convergence takes place.

We're not entirely happy with the sleep added to the tests. We had trouble getting Timecop to play nicely with the Sequel current_datetime_timestamp extension, if you have an alternative approach to implementing this that would be great too.

acrmp:
We pushed another commit with a different testing approach using doubles instead. We get to remove the sleep but the test is possibly not quite as good.

evanfarrar:
It sounds like the strategy from earlier guidance has led y'all down a fine path.

LGTM, for now

ctlong:
Please feel free to squash merge this PR if y'all want to keep your history more clean.

That would get rid of some of the back and forth commits that undo each other lol.

PR #3068, opened 2022-11-18
Optimize Role model/dataset

philippthun:
The `Role` model is not backed by a dedicated table, but a dataset that is a UNION of eight individual roles tables. When filtering for roles, this was done by adding WHERE conditions to the overall dataset, which resulted in slow query executions. A more efficient approach is to move the WHERE conditions into the individual datasets per role.

This commit customizes the `Role` model/dataset. The creation of the UNIONed dataset is moved into a dedicated class (`RoleDataset`) that supports building a dataset with filters (i.e. WHERE conditions) per role (i.e. individual table). The `where` method of the `Role` dataset is overridden and uses `Dataset.from` ([1]) to change the dataset's source. It replaces the source with a new dataset built with filters per role.

As `where` calls can be chained, the filters need to be kept in the dataset instance (`cache`) and the source dataset needs to be built again on each `where` invocation with all the filters, i.e. previous and new WHERE conditions.

Furthermore fully qualified identifiers (table name and column name) have to be adapted; as the WHERE conditions are applied on the individual roles tables, no table name prefix is required and thus removed from the identifiers.

Some filters can now be handled in a more efficient way. When filtering for a specific role type, the corresponding dataset gets a `WHERE 1 = 1`, whereas all the other datasets get a `WHERE 1 = 0`, meaning they never return any rows. Something similar is done when filtering for organizations or spaces; the opposite role types get a `WHERE 1 = 0` as they don't have the requested `id` field.

As there might be filters that still should be applied to the overall dataset, `where` invocations with a block (aka. virtual row block) are treated as before, i.e. applied at the end.

Drawbacks:
- The current implementation of the custom `Role` model/dataset is rather strict with regards to allowed filter expressions. This means that if some functionality is added (e.g. new or different filter), the custom `Role` model/dataset needs to be adapted. This was done by intention to ensure that every code path is tested.
- To limit the number of different filter expressions, invocations of `first(cond)` and `find(cond)` on the custom `Role` model/dataset have been replaced with `where(cond).first` which is semantically identical.

[1] https://sequel.jeremyevans.net/rdoc/classes/Sequel/Dataset.html#method-i-from

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

[...]

PR #3153, opened 2023-01-20
WIP/RFC: parameterize roles by org

evanfarrar:

If `/v3/roles` is filtered by `organization_guids`, the role join models (`organizations_managers`, `spaces_developers` etc) are all `UNION`ed into a single unfiltered virtual table and a filter is applied. If the role tables are filtered by the organization *first* then it was found to speed up the query by about 2 orders of magnitude (probably depends on the user base of the org vs the user base of _all_ orgs in the Foundry install).

Thanks to @pivotal-Josh-Gainey for finding this optimization (he had a pair too but I can't find his github).

In many of VMware's usages of the roles endpoint, we are allowing a user to edit the roles for a space or an org in a GUI. It ends up calling `/v3/roles` with an `organization_guids` or `space_guids` filter for the organization or space that the user is editing. This is a WIP that only solves for organization_guids and has no tests, but I also plan to extend it to `space_guids` and look for other datasets that `UNION` these join models because they also might need this optimization.

This is particularly important to Cloud Foundry installs with many users given access to many orgs/spaces. For example, given a CC database on mysql with 50 orgs, 50 spaces in each org, and 25 users who all have access to every org and space the `Role` model will have 255000 records. We then approximated the roles endpoint as such and timed it:
```ruby
rlm = RolesListMessage.from_params({organization_guids: [org.guid], page: 1, user_guids: org.users[0..50].map(&:guid), per_page: 50})
RoleListFetcher.fetch(rlm, Role, eager_loaded_associations: Presenters::V3::RolePresenter.associated_resources).all
```
Before this change : **0.9s**
After this change: **0.004s**

Don't know of any PRs but I am sure `Role` has some optimizations recently that are related but didn't have this impact on this scale of data.

(I have signed a great many of these CF CLAs but I probably need to sign the latest one)

* [x] I have run all the unit tests using `bundle exec rake` (currently failing on RoleListFetcher test, will resolve)

### DIFF
--- a/app/models/runtime/role.rb
+++ b/app/models/runtime/role.rb
@@ -5,98 +5,198 @@ module VCAP::CloudController
 
   # Sequel allows to create models based on datasets. The following is a dataset that unions all the individual roles
   # tables and labels each row with a `type` column based on which table it came from
-  class Role < Sequel::Model(
-    OrganizationUser.select(
-      Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_USER, :type),
-      Sequel.as(:role_guid, :guid),
-      :user_id,
-      :organization_id,
-      Sequel.as(SPACE_OR_ORGANIZATION_NOT_SPECIFIED, :space_id),
-      :created_at,
-      :updated_at
-    ).union(
-      OrganizationManager.select(
-        Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_MANAGER, :type),
+
+  def self.role_dataset(organization_id=false)
+    if organization_id
+      spaces_in_org = Space.where(organization_id: organization_id).map(&:id)
+      OrganizationUser.select(
+        Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_USER, :type),
         Sequel.as(:role_guid, :guid),
         :user_id,
         :organization_id,
         Sequel.as(SPACE_OR_ORGANIZATION_NOT_SPECIFIED, :space_id),
         :created_at,
-        :updated_at),
-      all: true,
-      from_self: false
-    ).union(
-      OrganizationBillingManager.select(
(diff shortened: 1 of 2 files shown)


evanfarrar:
@sethboyles hoping for feedback if you have any, particularly on how we might parameterize this giant dataset in the least bad way possible, module method and 76 line copy/paste seem like the worst way possible.

app/models/runtime/role.rb
```diff
@@ -5,99 +5,199 @@ module VCAP::CloudController
...
-        :updated_at),
-      all: true,
-      from_self: false
-    ).union(
-      OrganizationAuditor.select(
-        Sequel.as(VCAP::CloudController::RoleTypes::ORGANIZATION_AUDITOR, :type),
+
+  def self.role_dataset(organization_id=false)
+    if organization_id
+      spaces_in_org = Space.where(organization_id: organization_id).map(&:id)
```
sethboyles, line 11: is it necessary to pull the `id`s into memory?  Usually pulling IDs into memory is slower than constructing a dataset (see https://github.com/cloudfoundry/cloud_controller_ng/pull/3070 for an example).

philippthun:
I think this PR is trying to achieve something similar as #3068. The "Optimize Role model/dataset" PR allows filtering by arbitrary attributes and also combinations of attributes. I'm currently working on adding some [performance tests](https://github.com/cloudfoundry/cf-performance-tests) for the `/v3/roles` endpoint and marking the PR as "Ready for review" afterwards.

app/models/runtime/role.rb
sethboyles: I wonder if it is better in this case because it guarantees we execute the query once, instead of this being evaluated as a sub-query each time.  Not sure if the DB will cache the results for multiple identical subqueries or something.  Might be worth testing.

evanfarrar:
Closing this, going to work with CCNG maintainers on the team to merge #3068

PR #3170, opened 2023-02-01
Use Net::HTTP::Persistent in Diego client

philippthun:
This replaces `HTTPClient` with `Net::HTTP::Persistent` in the `Diego::Client` class.

`Net::HTTP::Persistent` is based on `Net::HTTP` which automatically triggers a failover to another IP address after reaching the given connect timeout.

Furthermore `Net::HTTP::Persistent` adds persistent HTTP connections and thread safety. It is configured to use a pool of up to 50 connections. This equals the size of `WorkPool`s used by various background processes (e.g. deployment updater).

### Background information

:information_source: **Use case**: Mitigates the risk of `Runner is unavailable` errors during a cf push when the client is unable to reach a bbs instance.

:x: First, **discarded** solution proposal: #3002

:heavy_check_mark: Second, **merged** solution proposal: #3048

:boom: Issues caused by #3048
- #3109
- #3112
- sporadically failing `sync-integration-tests`

:arrows_counterclockwise: Revert PR: #3113

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)

PR #3171, opened 2023-02-01
💩💩💩 Revert "Use Net::HTTP::Persistent in Diego client"

philippthun:
Reverts cloudfoundry/cloud_controller_ng#3170

This only works with Ruby 3.1.2! Ruby 3.1.3 (which is used to run the unit tests on concourse) includes a newer Net::HTTP version that reverted the usage of the `Socket.tcp`'s `connect_timeout` option (see https://github.com/ruby/net-http/pull/74).


johha (approved):
😿

PR #3175, opened 2023-02-06
Remove ruby-debug-ide and debase

sethboyles:
This PR removes `ruby-debug-ide`.  We face an issue where, when bundling for the first time on a new env, the first attempt at installing `ruby-debug-ide` fails and bundling a second time magically works.

We're opening this PR because we aren't sure anyone is using it--we think IDEs like RubyMine should work fine without this gem.  As far as we can gather, we think this was added to debug running instances of CCNG.  We think breakpoints in tests run from RubyMine should work.

Alternatively we can add this to a separate gem group (e.g. `ide`) and exclude that when installing if others are using this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -96,10 +96,8 @@ end
 
 group :development do
   gem 'byebug'
-  gem 'debase', '>= 0.2.2.beta14'
   gem 'listen'
   gem 'roodi'
-  gem 'ruby-debug-ide', '>= 0.7.0.beta4'
   gem 'solargraph'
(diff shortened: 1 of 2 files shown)


moleske (approved):
Approval assuming we don't get feedback saying lots of folks use this gem

johha:
I also had issues with `ruby-debug-ide` in IntelliJ which I could fix by removing it from the Gemfile locally.
There are multiple debug gems which are optimized for different IDEs so it make sense to remove it from the the Gemfile.
Example: @will-gant used [`debug`](https://github.com/ruby/debug) with VSCode and it worked fine but caused some issues with IntelliJ/RubyMine.

johha (approved):

PR #3187, opened 2023-02-13
[DOCS] Adjust docs for query parameter encoding

svkrieger:
Adjusted the example for double encoded query param to actually match that of a `,`.


philippthun (approved):

moleske (approved):
I spent way too much time checking this cause I kept getting confused at what did or did not need encoding.  Good catch on this!

PR #3117, opened 2022-12-19
Exclude events table from using window function for pagination

kathap:
The events table is one of the biggest tables. We saw slow queries on big landscapes like:
```
cf curl "/v3/audit_events?types=audit.user.organization_user_add%2Caudit.user.organization_billing_manager_remove%2Caudit.user.space_developer_add%2Caudit.user.organization_auditor_remove%2Caudit.user.space_developer_remove%2Caudit.user.space_auditor_remove%2Caudit.user.space_auditor_add%2Caudit.user.space_manager_add%2Caudit.user.organization_billing_manager_add%2Caudit.user.organization_manager_remove%2Caudit.user.organization_auditor_add%2Caudit.user.space_manager_remove%2Caudit.user.organization_manager_add%2Caudit.user.organization_user_remove&per_page=500&order_by=-created_at" 
```
We experimented and discovered using v2 for audit_events is much faster. The difference between v2 and v3 is that for v3 pagination is done via window function. 
```
SELECT *, count(*) OVER () AS "pagination_total_results" FROM "events" WHERE ("type" IN ('audit.user.organization_user_add', 'audit.user.organization_billing_manager_remove', 'audit.user.space_developer_add', 'audit.user.organization_auditor_remove', 'audit.user.space_developer_remove', 'audit.user.space_auditor_remove', 'audit.user.space_auditor_add', 'audit.user.space_manager_add', 'audit.user.organization_billing_manager_add', 'audit.user.organization_manager_remove', 'audit.user.organization_auditor_add', 'audit.user.space_manager_remove', 'audit.user.organization_manager_add', 'audit.user.organization_user_remove')) ORDER BY "events"."created_at" DESC, "events"."guid" ASC LIMIT 500 OFFSET 0
```
In v2 it was done with 2 queries.
```
SELECT * FROM "events" WHERE (type  IN  ('audit.user.organization_user_add', 'audit.user.organization_billing_manager_remove', 'audit.user.space_developer_add', 'audit.user.organization_auditor_remove', 'audit.user.space_developer_remove', 'audit.user.space_auditor_remove', 'audit.user.space_auditor_add', 'audit.user.space_manager_add', 'audit.user.organization_billing_manager_add', 'audit.user.organization_manager_remove', 'audit.user.organization_auditor_add', 'audit.user.space_manager_remove', 'audit.user.organization_manager_add', 'audit.user.organization_user_remove')) ORDER BY "timestamp" DESC, "id" DESC LIMIT 500 OFFSET 0

[...]

PR #3191, opened 2023-02-14
Exclude app usage events table from using window function

sethboyles:
Similar to the issue reported in https://github.com/cloudfoundry/cloud_controller_ng/pull/3117, we have noticed performance degradation when using the OVER window function with `app_usage_events` when the tables has a large-ish number of row (around 150,000).  This PR disables use of the window function for `app_usage_events`, just as we have disabled it for `events`.

@kathap @philippthun , since you both worked on the previous PR--are you concerned that this might be a trend with several of our tables?  We don't have a lot of context on the performance benefits of the use of the OVER window function, and it feels like its use may be high risk.  We were wondering if it might be worth disabling the OVER window for more models, or even all of them.  Perhaps a configuration value in capi-release?

### DIFF
--- a/app/models/runtime/app_usage_event.rb
+++ b/app/models/runtime/app_usage_event.rb
@@ -9,5 +9,10 @@ module VCAP::CloudController
       :buildpack_guid, :buildpack_name,
       :package_state, :previous_package_state, :parent_app_guid,
       :parent_app_name, :process_type, :task_name, :task_guid
+    AppUsageEvent.dataset_module do
+      def supports_window_functions?
+        false
+      end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


philippthun (approved):

philippthun:
I had a look at some database/query statistics/performance dashboards for our largest foundation and can confirm that queries on the `app_usage_events` table using the `OVER` window function are relatively slow. It does not seem to be as bad as for the `events` table, but excluding this table makes sense.

I agree that we should revisit the use of the `OVER` window function. It might be the case that - due to other recent improvements - its benefit is now neglectable. But in order to judge this, we would need to perform some performance tests with reasonable table sizes.

I'm also open to provide a means to disable the `OVER` window function by configuration (this might be even helpful to do the performance tests).

PR #3208, opened 2023-03-01
Keep original VCAP-Request-ID for re-enqueued ReoccurringJobs

philippthun:
All background jobs are wrapped in a `LoggingContextJob` that stores the original VCAP-Request-ID and ensures that the same ID is used (e.g. for logs) while performing the job.

For `ReoccurringJob`s, this capability got lost when the job was re-enqueued. Adding the next job to the queue did not happen as part of the `perform` method, but in the `success` hook. So by using the `with_request_id_set` method also for the `success` hock, the re-enqueued job keeps the original VCAP-Request-ID.

The effect can be seen when looking at logs with source `cc.service_broker.v2.http_client`. When polling for the last operation state of a provisioned service instance, the VCAP-Request-ID was only set for the initial 'fetch' request (which was done in the same job as the 'provision' request). With this change, all subsequent state polls are also logged with `"data":{"request_guid":"original-guid",...}`.

### DIFF
--- a/app/jobs/logging_context_job.rb
+++ b/app/jobs/logging_context_job.rb
@@ -20,15 +20,23 @@ module VCAP::CloudController
         raise CloudController::Errors::ApiError.new_from_details('BlobstoreError', e.message)
       end
 
+      def success(job)
+        with_request_id_set do
+          super(job)
+        end
+      end
+
       def error(job, e)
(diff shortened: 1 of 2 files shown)


johha (approved):
lgtm 👍

PR #3154, opened 2023-01-20
Truncate job errors 2

MerricdeLauney:
Removed the migration from #3145 so we can consider the migration separately

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [ ] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)

PR #3145, opened 2023-01-17
Truncate Large Service Broker Errors

moleske:
Thanks for contributing to cloud_controller_ng. To speed up the process of reviewing your pull request please provide us with:

* A short explanation of the proposed change:
Truncate error descriptions from service brokers that are too big for our database.

* An explanation of the use cases your change solves
An operator was noticing that cloud controller worker was consistently restarting every few minutes because it aggressively kept retrying a failing create-service job.  It was consistently retrying this job because it could never save the failed attempt due to the error message being too big for the delayed_jobs last_api_error field.  This truncates the error message when cloud controller gets the initial response from the service broker.

We left our previous commits to try and show how we concluded on doing it here rather than close to where the database save calls are made.  We also tried to truncate to fit the `text` column size, but another table, `jobs` also saves the error, but in a smaller column of type `varchar(16000)`.  This led to the user of the command `cf create-service` to not get any error message back as the column would have no data as it was too big for that as well.  That is how we settled on anything bigger than 2^14 bytes being truncated.

We should squash the commits on merge to make it easier to see history, this pr will then be around to provide future folks context on this change.

* Links to any other associated PRs

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)

PR #3231, opened 2023-03-21
Truncate job errors 3

sethboyles:
Remake of https://github.com/cloudfoundry/cloud_controller_ng/pull/3145, done via cherry-picking instead of rebasing.  This was broken out so we could consider/talk about changing the type of the column without rejecting the other validations added in https://github.com/cloudfoundry/cloud_controller_ng/pull/3154

copied from the other PR:

> An operator was noticing that cloud controller worker was consistently restarting every few minutes because it aggressively kept retrying a failing create-service job. It was consistently retrying this job because it could never save the failed attempt due to the error message being too big for the delayed_jobs last_api_error field. This truncates the error message when cloud controller gets the initial response from the service broker.

> We left our previous commits to try and show how we concluded on doing it here rather than close to where the database save calls are made. We also tried to truncate to fit the text column size, but another table, jobs also saves the error, but in a smaller column of type varchar(16000). This led to the user of the command cf create-service to not get any error message back as the column would have no data as it was too big for that as well. That is how we settled on anything bigger than 2^14 bytes being truncated.

### DIFF
--- a/db/migrations/20230118223008_convert_cf_api_error_to_text.rb
+++ b/db/migrations/20230118223008_convert_cf_api_error_to_text.rb
@@ -1,0 +1,7 @@
+Sequel.migration do
+  change do
+    alter_table :jobs do
+      set_column_type(:cf_api_error, 'text')
+    end
+  end
+end
--- a/spec/unit/jobs/pollable_job_wrapper_spec.rb
+++ b/spec/unit/jobs/pollable_job_wrapper_spec.rb
@@ -181,26 +181,7 @@ module VCAP::CloudController::Jobs
           expect(errors.size).to eq(1)
           error = errors[0]['test_mode_info']
           expect(error['detail']).to eq('VCAP::CloudController::Jobs::BigException')
-          expect(error['backtrace'].size).to be == 8
-        end
-      end
-
-      context 'with a big message' do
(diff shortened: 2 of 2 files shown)


FloThinksPi:
I think this is still important however we should safeguard this and abort the keep retring without saving the full error message since text is unlimited and someone could construct a broker answering with a gigabyte long error message yikes :>

sethboyles:
@FloThinksPi I think that part was already merged? https://github.com/cloudfoundry/cloud_controller_ng/pull/3154/files I don't actually have a lot of context on why we might want to have the migration as well.  I haven't worked on this track, just cherry-picked the migration out.  Maybe @MerricdeLauney @moleske or anyone else who is interested can pick up this PR

moleske:
Yes part of this was merged.  It is confusing cause there are three pull requests - this one and the two Seth mention in the description of this.  We broke the original down so we could unblock the customer issue we saw with large error messages and decide later about changing the column.  Changing the column type would potentially help the fact that pollable job was causing issues (see commit messages in the original pr)

I think at this point we've kinda forgotten any detailed context.  This might be worth revisiting but I think a new pr would be better.  I'll close this for now cause we're not really looking into this anymore

PR #3232, opened 2023-03-22
Add Ports section to V3 upgrade guide conceptual changes

sethboyles:

Ports works differently enough in V3 compared to V2 that it probably warrants its own section in the docs, as some have found the migration confusing.  Hoping this clears it up.


sethboyles:
cc @jpalermo

moleske (approved):
looks good to me.  This should help folks migrating

PR #3238, opened 2023-04-02
Implement super secure HA blob storage

0401-7r011:

My employer, Bad Development Practices Inc, has strict policies about secret storage. All secrets must be stored within the codebase so we never lose them. (Fool me once ransomware hackers, fool me once...)

The problem is, we don't really feel like the blobstores offered in Cloud Foundry are secure enough for our secrets. They just don't SOUND very secure. "webdav", "fog"?

You know what sounds secure? Credhub. That's where secrets belong, and if my blobs have my secrets, that's where my blobs belong too.

Let's just put a pin in that

* [⛔️] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [🚳] I have viewed, signed, and submitted the Contributor License Agreement

* [🤔] I have made this pull request to the `main` branch

* [🚱] I have run all the unit tests using `bundle exec rake`

* [🚷] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)

* A list of items that my help in our rigorous testing...
  - This reuses the existing Credhub credentials that CCNG already has access to, unfortunately those only have read access, so you'll need to bump the permissions up to full admin access to all credhub paths...
  - The worker also needs access to these super credentials, so make sure you modify the worker `cloud_controller_ng.yml.erb` and `spec` to pass them in there too
  - Installing buildpacks during post-start now requires Credhub to be up. Shouldn't be a problem, unless you co-locate Cloud Controller and Credhub because apparently "started" doesn't really mean "started" for Credhub. No worries, adding some sleep to the `post-start` easily solves this
  - There are some "limitations" in Credhub blob storage. It seems to be limited to around 50-60k per credentials, which makes storing ~~large~~ blobs in a single credential quite difficult. Striping the blob across multiple credentials is an easy solution though
  - Unfortunately reconstructing blobs can take some time, and sometimes diego isn't super patient. I'd recommend using the binary buildpack and smallish apps
  - Oh, there's a `/v3/blobs` resource now, and no security, so maybe add some of that when you get a chance

### DIFF
--- a/app/controllers/v3/blobs_controller.rb
+++ b/app/controllers/v3/blobs_controller.rb
@@ -1,0 +1,24 @@
+class BlobsController < ApplicationController
+
+  def show
+    client = CloudController::DependencyLocator.instance.credhub_client
+
+    response.headers["ETag"] = 'version1'
+    response.headers["Last-Modified"] = 'Sat, 1 Apr 2023 00:00:00 GMT'
+    send_data(decode_file(client.get_chunked_credential_by_name(hashed_params[:key])), filename: 'file')
+  end
+
+  private
+
+  def decode_file(data)
+    Base64.strict_decode64(data).force_encoding('BINARY')
+  end
+
+  def enforce_read_scope?
+    false
+  end
+
+  def enforce_authentication?
+    false
+  end
+end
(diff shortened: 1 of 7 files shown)


moleske:
ah my favorite once-a-year april fools pr maker.  Thanks for the pr but I think we'll politely let Bad Development Practices Inc know that we're not interested in this workflow, but look forward to Bad Development Practices Inc competing cloud foundry distribution

Gerg:
How will this integrate with Bits Service?

PR #3256, opened 2023-04-13
Fix sequel 5.67.0

moleske:

Sequel 5.67.0 no longer allows us to set class/instance variables on Sequel objects. More generally, it deprecates registering database extensions through means other than a module, so we have broken out this code into a sequel extension.

@philippthun - we noticed you originally added this code.  We have concerns that the solution we provide here will cause many more database queries.  This is because the extension will call db.fetch every time instead of the previous way that saved it to the class so it was only called once

broken #3243 bump


philippthun:
Thanks @moleske and @tjvman for looking into why the sequel version bump fails. Moving the code into an extension is definitely a good thing - I should have done this right from the start.

I share your concern about the increasing number of database queries and thought about alternative means to store the result. I came up with the following 'mechanism' that lazily initializes a constant. What do you think?

``` ruby
module Sequel::UniqueColumnSubquery
  def requires_unique_column_names_in_subquery_select_list?
    unless Sequel::UniqueColumnSubquery.const_defined?(:UNIQUE_NAMES_IN_SUBQUERY)
      Sequel::UniqueColumnSubquery.const_set(:UNIQUE_NAMES_IN_SUBQUERY, unique_names_in_subquery?)
    end

    Sequel::UniqueColumnSubquery.const_get(:UNIQUE_NAMES_IN_SUBQUERY)
  end

  private

  def unique_names_in_subquery?
    self.db.fetch('SELECT * FROM (SELECT 1 AS a, 1 AS a) AS t1').all
    false
  rescue Sequel::DatabaseError
    true
  end
end

Sequel::Dataset.register_extension(:requires_unique_column_names_in_subquery, Sequel::UniqueColumnSubquery)
```

moleske:
It seemed reasonable to me, so I went ahead and made a commit and ran unit tests on my machine.  I'll let @tjvman chime in later since I didn't get a chance to bother him

philippthun (approved):

PR #3280, opened 2023-05-10
Prevent duplicates in labels and annotations tables

kathap:
We observed duplicated entries in the service_instance_labels table, where key_name and resource_guid were the same. This should not be possible. This leads to errors like e.g. if you are using cf metadata labels as mechanism for mta discovery, you can observe strange results when executing the following query: `cf curl 'v3/service_instances?fields[service_plan]=guid,name,relationships.service_offering&fields[service_plan.service_offering]=guid,name&space_guids=123&label_selector=some_key=some_value'` The result contains both entries which have the duplicated resource_guid-key_name pair, but expected is only one entry in the result.

There should only be one entry per resource_guid-key_name-key_prefix pair. To prevent duplicate entries in the xxx_labels and xxx_annotations tables, we propose the following:

1. remove duplicate service_instance_labels.key_name + key_prefix with same resource_guid (same for annotations), leaving the one with the highest id, in order to prepare for adding the new column and unique index on the column.

2. add a new column (column `resource_guid_key_prefix_key_name_hash` for labels and column `resource_guid_key_prefix_key_hash` for annotations) 

3. update the column value to an hash (SHA256) of resource_guid, key_prefix and key_name (key for annotations). 

4. add an unique index on the new column

5. create two triggers, one on insert and one on update statements, in order to get the hash, and the new column filled with that hash, also for new entries (`<tablename abv>_add_unique_hash_insert_trigger` for annotations and 
`<tablename abv>_add_unique_hash_update_trigger` for labels). 

The trigger creates the hash and the entry in the new column, which will fail, if an entry with the same values already exists, due to to the unique index on the new column.

After the migration is applied, such entries with duplicated resource_guid-key_name-key_prefix pair for the label tables will be not possible any more (same for annotations), due to the index `<tablename abv>_resource_guid_key_prefix_key_hash_idx` for annotations and `<tablename abv>_resource_guid_key_prefix_key_name_hash_idx` for labels.


kathap:
retest this please

sethboyles:
We talked this over and agreed that we didn't see major issues with merging in this PR.  

We were wondering if you could elaborate on the specifics of how removing postgres 10 support would (or would not) simplify the solution here.

sethboyles (approved):

kathap:
> We talked this over and agreed that we didn't see major issues with merging in this PR.
> 
> We were wondering if you could elaborate on the specifics of how removing postgres 10 support would (or would not) simplify the solution here.

We had one approach where we added the new column for the hash value as generated column, code looked ~ like following:
```alter_table table do
        if dbtype == 'mysql'
          add_column :resource_guid_key_prefix_key_name_hash, Sequel.lit("varchar(70) generated always as ( SHA2(CONCAT(COALESCE(resource_guid, '' ) , COALESCE(key_prefix , '')  , COALESCE(key_name , '')),256)) stored;")
        elsif dbtype == 'postgres'
          add_column :resource_guid_key_prefix_key_name_hash, Sequel.lit("varchar(70) generated always as ( encode(digest(COALESCE(\"resource_guid\" , '' ) || COALESCE(\"key_prefix\" , '')  || COALESCE(\"key_name\" , ''), 'sha256'),'hex')) stored;")
        end
      end
```

But the generated column is [only supported by postgres 12 and higher](https://www.postgresql.org/docs/current/ddl-generated-columns.html#:~:text=Supported%20Versions%3A%20Current%20(15)%20/%2014%20/%2013%20/%2012). But there were random issues with failing tests because of column updates:
```
V3 service instances PATCH /v3/service_instances/:guid user-provided service instance when the request is invalid is rejected
      Failure/Error: super

      Sequel::DatabaseError:
        PG::GeneratedAlways: ERROR:  column "resource_guid_key_prefix_key_hash" can only be updated to DEFAULT
        DETAIL:  Column "resource_guid_key_prefix_key_hash" is a generated column.
```
but the documentation says[ A generated column cannot have a column default](https://www.postgresql.org/docs/current/ddl-generated-columns.html#:~:text=A%20generated%20column%20cannot%20have%20a%20column%20default).
Things we do not have to deal with in the current implementation.

philippthun:
New PR: #3394

PR #3289, opened 2023-05-16
Refactor rate limiters

### LINKED ISSUE #2544: Retry-After header in 429 Too Many Request responses should be time until rate limit reset
## Issue
The `Retry-After` header should be time until reset rather than the current implementation of Unix time at which the rate limit reset 

## Context
The Mozilla docs cover this fairly well https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After. We currently just set the value to be the same as the `X-RateLimit-Reset` header value

## Steps to Reproduce
1. Query a CC API instance with a rate limit enabled
2. Make enough requests to exhaust limit
3. Look at headers on 429 response

## Expected result
1. `X-RateLimit-Reset` header contains timestamp at which limit resets
2. `Retry-After` header contains number of seconds until limit resets

## Current result
1. `X-RateLimit-Reset` header contains timestamp at which limit resets
2. `Retry-After` header contains timestamp at which limit resets

## Possible Fix
Calculate the diff between the `X-RateLimit-Reset` and `Time.now` for the `Retry-After` header

philippthun:
To prepare for the introduction of a Redis-based rate limiting (required for Puma), the current implementation has been refactored.

- `ExpiringRequestCounter`
-- Renamed from `RequestCounter`.
-- Single `increment` method (combines old `get` + `increment` methods).
-- Returns _expires_in_ seconds instead of _valid_until_ timestamp.

- `ConcurrentRequestCounter`
-- Renamed from `ServiceBrokerRequestCounter`.
-- Not a singleton anymore.
-- Renamed methods from `try_acquire?`/`release` to `try_increment?`/`decrement`.

- `UserResetInterval` returns next _expires_in_ seconds instead of _reset_interval_ timestamp.

- `Retry-After` header returns "delay-seconds" instead of epoch seconds (fixes #2544).

- Tests cleanup.

### DIFF
--- a/lib/cloud_controller/rack_app_builder.rb
+++ b/lib/cloud_controller/rack_app_builder.rb
@@ -40,9 +40,9 @@ module VCAP::CloudController
           }
         end
         if config.get(:max_concurrent_service_broker_requests) > 0
-          CloudFoundry::Middleware::ServiceBrokerRequestCounter.instance.limit = config.get(:max_concurrent_service_broker_requests)
           use CloudFoundry::Middleware::ServiceBrokerRateLimiter, {
             logger: Steno.logger('cc.service_broker_rate_limiter'),
+            max_concurrent_requests: config.get(:max_concurrent_service_broker_requests),
             broker_timeout_seconds: config.get(:broker_client_timeout_seconds),
           }
         end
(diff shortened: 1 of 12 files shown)


sethboyles:
The Mozilla docs state that both HTTP Dates and seconds are valid for the `Retry-After` header, so I'm curious about the impetus for that change.  Are we sending a Date, but still not the right format?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

Any concern that some clients have code expecting `Retry-After` to be a date and this will be a behavior change?

sethboyles (approved):

PR #3305, opened 2023-05-31
Pass X-Vcap-Request-Id header in calls to BBS

mariash:

To support distributed tracing request id received from the Gorouter is passed to Diego BBS so that the same request can be traced along multiple components.

This is part of distributed tracing feature - the ability to trace the same request across multiple components, e.g. stopping an app instance first will come to Gorouter, which will generate request ID and pass it down to Cloud Controller, which will pass it to Diego BBS, which will pass it to downstream components like Auctioneer, Rep and Garden.

### DIFF
--- a/lib/diego/client.rb
+++ b/lib/diego/client.rb
@@ -5,8 +5,6 @@ require 'http/httpclient'
 
 module Diego
   class Client
-    PROTOBUF_HEADER = { 'Content-Type'.freeze => 'application/x-protobuf'.freeze }.freeze
-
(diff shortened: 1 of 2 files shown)


sethboyles:
We were curious why there was an extra uuid appended to the original vcap request id, and it seems like it was carrying over behavior explained in this comment? 

https://github.com/cloudfoundry/cloud_controller_ng/blob/0a410b1052f4da770c79c6d20fa96bced6454a46/lib/sinatra/vcap.rb#L103-L110

sethboyles (approved):

PR #3312, opened 2023-06-07
Fix duplicate routes sent to Diego

danielfor:
When having multiple applications using the same route, during app push the routes sent to Diego will be not only from the application being deployed but also from the other apps using the same route, making duplicate entries in the grpc payload.


moleske (approved):
Nice find!

PR #3316, opened 2023-06-16
Clarify that process stats returns one stats object per instance

sethboyles:
I thought it could be clarified that process stats returns one stats object per instance (as opposed to some form of aggregate)


moleske (approved):
Seems reasonable to clarify.  Those `[]` in the example are easy to miss as the indication of a list

PR #3346, opened 2023-07-11
More aggressive cleanup of failed delayed jobs

kathap:
The idea was to have an additional absolute limit of failed jobs so that they get deleted even before the 14d failed_jobs.cutoff_age_in_days. The limit should be configurable.

Therefore we implement a new parameter for maximum number of failed delayed_jobs that should be in the delayed_jobs table. If the number gets exceeded, entries will be deleted keeping the ones with the higher ids. The failed_jobs cleanup job will now run more often than once per day.

Currently, failed delayed_jobs are deleted after 14d (configurable) to keep some info about failed jobs that helps debugging: https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/jobs/runtime/failed_jobs_cleanup.rb. 
This can still lead to very large number of delayed_jobs records that slow down DB queries working on delayed_jobs (also addressed by an index, ccng #3324).

Related to https://github.com/cloudfoundry/capi-release/pull/328

### DIFF
--- a/app/jobs/runtime/failed_jobs_cleanup.rb
+++ b/app/jobs/runtime/failed_jobs_cleanup.rb
@@ -21,20 +21,19 @@ module VCAP::CloudController
           old_delayed_jobs.delete
 
           unless max_number_of_failed_delayed_jobs.nil?
-            ids_from_failed_jobs_that_are_too_much = Delayed::Job.
-                                                     where(Sequel.lit('failed_at is not null')).
-                                                     order(Sequel.desc(:id)).
-                                                     offset(max_number_of_failed_delayed_jobs.to_i).
-                                                     # Mysql handles offset differently, therefore using a large
-                                                     # number to mimic unlimited offset
-                                                     limit(1844674407370955161).
-                                                     select(:id)
+            ids_exceeding_limit = Delayed::Job.
+                                  where(Sequel.lit('failed_at is not null')).
+                                  order(Sequel.desc(:id)).
+                                  offset(max_number_of_failed_delayed_jobs.to_i).
+                                  # Mysql handles offset differently, therefore using a large
+                                  # number to mimic unlimited offset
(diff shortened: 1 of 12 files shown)


app/jobs/runtime/failed_jobs_cleanup.rb
```diff
@@ -15,9 +16,25 @@ def perform
                              where(Sequel.lit("run_at < CURRENT_TIMESTAMP - INTERVAL '?' DAY", cutoff_age_in_days.to_i))
 
           logger = Steno.logger('cc.background')
-          logger.info("Cleaning up #{old_delayed_jobs.count} Failed Delayed Jobs")
+          logger.info("Cleaning up #{old_delayed_jobs.count} old Failed Delayed Jobs")
 
           old_delayed_jobs.delete
+
+          unless max_number_of_failed_delayed_jobs.nil?
+            ids_from_failed_jobs_that_are_too_much = Delayed::Job.
```
philippthun, line 24: `ids_from_failed_jobs_that_are_too_much` sounds a little bit strange to me...

app/jobs/runtime/failed_jobs_cleanup.rb
```diff
@@ -15,9 +16,25 @@ def perform
...
+                                                     # Mysql handles offset differently, therefore using a large
+                                                     # number to mimic unlimited offset
+                                                     limit(1844674407370955161).
+                                                     select(:id)
+
+            logger.info("Cleaning up #{ids_from_failed_jobs_that_are_too_much.count} Failed Delayed Jobs because they exceed max_number_of_failed_delayed_jobs")
+
+            Delayed::Job.where(id: ids_from_failed_jobs_that_are_too_much).delete
+          end
+          logger.info('Cleaning up ')
```
philippthun, line 37: Looks like a leftover...

johha (approved):

PR #3352, opened 2023-07-21
Improve SQL Performance by Eliminating Unnecessary Order 

FloThinksPi:
This commit corrects and improves SQL queries by avoiding redundant orders. It was previously observed that SQL queries were being ordered both by ID and GUID when the default order was applied. The paginator was also adding a secondary sort, introduced to offset issues encountered when using 'created_at' or 'updated_at' as sorting parameters, as the order was not deterministic. However, for default sorting by ID, this was superfluous.

With this change, the paginator now adds the secondary sort to the SQL query only if the order by parameter is not the ID column, in order to ensure fixed order for all column values. This alteration aims to enhance query performance across all listing endpoints.

The change involves a schema update that aligns all related indices in the database. A combined index with the GUID column is created with the understanding that databases can use such combined indices as a replacement for the most left column in the index. For instance, an index over (updated_at, GUID) can be utilized by the database, providing enhanced performance either when querying this table with both fields in the where clause, or when ordering by both columns.

The primary and secondary ordering now use the same direction, an improvement from the previous method where order was only applied to primary sorting. This prevents the need for separate indices for ascending and descending orders, resulting in a system where descending order is the exact opposite of ascending order.

Performance for end users and platform stability by being able to handle more load.

### DIFF
--- a/spec/unit/lib/cloud_controller/paging/sequel_paginator_spec.rb
+++ b/spec/unit/lib/cloud_controller/paging/sequel_paginator_spec.rb
@@ -11,6 +11,8 @@ module VCAP::CloudController
       let!(:space) { Space.make }
       let!(:app_model1) { AppModel.make(space: space) }
       let!(:app_model2) { AppModel.make }
+      let!(:app_model3) { AppModel.make }
+      let!(:app_model4) { AppModel.make }
       let(:page) { 1 }
       let(:per_page) { 1 }
 
@@ -78,7 +80,7 @@ module VCAP::CloudController
         expect {
           paginated_result = paginator.get_page(eager_dataset, pagination_options)
         }.to have_queried_db_times(/select/i, paginator.can_paginate_with_window_function?(dataset) ? 2 : 3)
-        expect(paginated_result.total).to eq(2)
+        expect(paginated_result.total).to eq(4)
         expect(paginated_result.records[0].associations[:space].name).to eq(space.name)
       end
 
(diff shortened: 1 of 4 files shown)


johha (approved):

spec/unit/lib/cloud_controller/paging/sequel_paginator_spec.rb
```diff
@@ -117,6 +119,38 @@ module VCAP::CloudController
         expect(paginated_result.total).to be 1
       end
 
+      it 'does not order by GUID when the primary order is by ID' do
+        options = { page: page, per_page: 2, order_by: 'id', order_direction: 'asc' }
+
+        pagination_options = PaginationOptions.new(options)
+        expect {
+          paginator.get_page(dataset, pagination_options)
+        }.to have_queried_db_times(/.*ORDER BY `apps`.`id` ASC, `apps`.`guid` ASC.*/i, 0)
```
philippthun, line 128: I would add a positive test
```
... have_queried_db_times(/ORDER BY `apps`.`id` ASC/i, 1)
```
to ensure that the SQL string is correct for all tested databases. I remember that formulating a regex that matches all databases was difficult...

philippthun:
> Additionally when sorting by created_at we can also just sort by id sice this would sort by created_at automatically and maintain a fixed order since ids are uniqe.

That's not part of the PR (anymore). Can you adapt the commit message as well as the PR description?

PR #3354, opened 2023-07-21
Readiness health checks story 1: adding properties to app manifest

ameowlia:
This PR is part of a larger effort detailed [here in this Cloud Controller PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/3351) and here in [this CFF RFC](https://github.com/cloudfoundry/community/blob/rfc-readiness-healthchecks/toc/rfc/rfc-draft-readiness-healthchecks.md).

In this story users can set 3 new app manifest properties. When they are set to valid properties, nothing happens. When they are set to invalid properties, it responds with an error.

### New App Manifest Properties
```
applications:
- name: test-app
  processes:
  - type: web
    health-check-http-endpoint: /health
    health-check-invocation-timeout: 2
    health-check-type: http
    timeout: 80
    readiness-health-check-http-endpoint: /ready       # 👈 new property
    readiness-health-check-invocation-timeout: 2       # 👈 new property
    readiness-health-check-type: http                  # 👈 new property
```

### Error Cases
* None of these are required.
* readiness-health-check-type needs to be validated as one of: process, http or port [similar to health-check-type](https://github.com/cloudfoundry/cloud_controller_ng/blob/021b1633cd5c65e50d29633905672e3cd68ad446/app/messages/manifest_process_update_message.rb#L20-L25).
* readiness-health-check-invocation-timeout - validated as an integer > 1 [similar to health-check-invocation-timeout](https://github.com/cloudfoundry/cloud_controller_ng/blob/021b1633cd5c65e50d29633905672e3cd68ad446/app/messages/manifest_process_update_message.rb#L34-L37)
* readiness-health-check-endpoint is validated as a URL [similar to health-check-endpoint](https://github.com/cloudfoundry/cloud_controller_ng/blob/021b1633cd5c65e50d29633905672e3cd68ad446/app/messages/manifest_process_update_message.rb#L29C16-L32)
* readiness-health-check-endpoint can only be set when readiness-health-check-type is set to http.

PR #3351, opened 2023-07-19
Readiness health checks

ameowlia:
### Context
This PR is part of the effort to add readiness health checks to Cloud Foundry. Deeper discussion of why can be read here in [this CFF RFC](https://github.com/cloudfoundry/community/blob/rfc-readiness-healthchecks/toc/rfc/rfc-draft-readiness-healthchecks.md). 

### Design
Users can set these 3 new properties on the app manifest. Error cases for invalid values etc are handled the same way as the other aready existing health-check properties.
```
applications:
- name: test-app
  processes:
  - type: web
    health-check-http-endpoint: /health
    health-check-invocation-timeout: 2
    health-check-type: http
    timeout: 80
    readiness-health-check-http-endpoint: /ready       # 👈 new property
    readiness-health-check-invocation-timeout: 2       # 👈 new property
    readiness-health-check-type: http                  # 👈 new property
```

### Acceptance
**As** an App Dev
**When** I set the readiness health check properties in my app manifest 
**Then** I see those properties set on my app process
**And then** those properties are sent to bbs on the LRP

### This change includes
* [x] Values can be set on the app manifest
* [x] Values set on app manifest are validated
* [x] These properties are stored in the DB
* [x] /v3/apps/:guid/manifest returns these properties
* [x] Docs for the /v3/apps/:guid/manifest are updated
* [x] /v3/processes returns these properties on the process 
* [x] Docs for the /v3/processes are updated
* [x] These processes are passed to the LRP
* [x] When these changes are run with a new diego-release then everything works.

### Contribution Steps
* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)
* [x] I have viewed, signed, and submitted the Contributor License Agreement
* [x] I have made this pull request to the `main` branch
* [x] I have run all the unit tests using `bundle exec rake`
* [x] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)
* [x] Remove all TODOs from code

### Related PRs
* https://github.com/cloudfoundry/cloud_controller_ng/pull/3354

PR #3361, opened 2023-07-26
Readiness health checks story 2 in db

ameowlia:
This PR is part of a larger effort detailed [here in this Cloud Controller PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/3351) and here in [this CFF RFC](https://github.com/cloudfoundry/community/blob/rfc-readiness-healthchecks/toc/rfc/rfc-draft-readiness-healthchecks.md).

In this story, the 3 new app manifest properties ([created in this PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/3354)), are set in the db.

### New App Manifest Properties
```
applications:
- name: test-app
  processes:
  - type: web
    health-check-http-endpoint: /health
    health-check-invocation-timeout: 2
    health-check-type: http
    timeout: 80
    readiness-health-check-http-endpoint: /ready       # 👈 new property
    readiness-health-check-invocation-timeout: 2       # 👈 new property
    readiness-health-check-type: http                  # 👈 new property
```

### Acceptance Steps
1. When I provide valid properties they are saved on the db. Below are examples of valid configs.
  ```
   # nothing! will default to process in the db
  ```

  ```
    readiness-health-check-http-endpoint: /ready
    readiness-health-check-invocation-timeout: 2
    readiness-health-check-type: http
  ```
  
  ```
    readiness-health-check-type: http
  ```
  
  ```
    readiness-health-check-type: port
  ```

  ```
    readiness-health-check-type: process
  ```
  
2. When I provide invalid properties, I get an error, and they are not saved in the db. Below are examples of invalid configs:
  ```
    readiness-health-check-http-endpoint: ready
    readiness-health-check-invocation-timeout: -10
    readiness-health-check-type: http
  ```
  
  ```
    readiness-health-check-http-endpoint: /ready
    readiness-health-check-type: port
  ```
  
  ```
    readiness-health-check-type: meow
  ```

### Previous PRs
* https://github.com/cloudfoundry/cloud_controller_ng/pull/3354


app/models/runtime/constraints/health_check_policy.rb
```diff
@@ -1,14 +1,29 @@
+require 'models/helpers/health_check_types'
+require 'utils/uri_utils'
+
 class HealthCheckPolicy
-  def initialize(process, health_check_timeout, health_check_invocation_timeout)
+  def initialize(process, health_check_timeout, health_check_invocation_timeout, health_check_type, health_check_http_endpoint)
```
ameowlia, line 5: We refactored the health check policy, to contain all of the checks for health checks, and removed some checks that were in the process model. We think this consolidates these checks more. 

Also we refactored so that we could make a ReadinessHealthCheckPolicy that inherits from this class.

app/models/helpers/health_check_types.rb
```diff
@@ -4,5 +4,14 @@ class HealthCheckTypes
     PROCESS = 'process'.freeze
     HTTP = 'http'.freeze
     NONE = 'none'.freeze
+
+    def self.constants_to_array
+      [
+        HTTP,
+        NONE,
```
sethboyles, line 11: I don't understand why we have `NONE` at all, do you?  Trying to set a health check to none via the API results in the following error:

```
± sb |readiness-health-checks-story-2-in-db → origin {3} ?:1 ✗| → cf curl -X PATCH v3/processes/c6493e8f-84b0-4e84-be1f-7156b7704039 -d '{"health_check": {"type": "none","data":  {  }}}'
{
   "errors": [
      {
         "detail": "Health check type must be \"port\", \"process\", or \"http\"",
         "title": "CF-UnprocessableEntity",
         "code": 10008
      }
   ]
}
```

app/models/helpers/health_check_types.rb
```diff
@@ -4,5 +4,14 @@ class HealthCheckTypes
     PROCESS = 'process'.freeze
     HTTP = 'http'.freeze
     NONE = 'none'.freeze
+
+    def self.constants_to_array
```
sethboyles, line 8: maybe `all_types` would be a clearer name? You could also have a helper method `readiness_types` that does the subtraction in the ReadinessHealthCheckPolicy

https://github.com/cloudfoundry/cloud_controller_ng/pull/3361/files#diff-b390b8806efca835a9f8d37a91c6cf6d9d0aef019708a78d6c5086272b18d3abR4

app/models/helpers/health_check_types.rb
ameowlia: It looks like it used to be an option, but no longer is.[ It looks like for awhile there was backwards compatible logic where it would map none -> process](https://github.com/cloudfoundry/cloud_controller_ng/blob/f47b8ebaaa81904d8e03ff92b978afa73815561c/app/messages/app_manifest_message.rb#L41). I suspect it is dead code now and could be ripped out. If you want we can do it in another story.

sethboyles: It'd be nice to clean up, but `NONE`'s presence doesn't seem too onerous right now.  I'll leave it to you to decide if you want to do it, and if not I'll create a story/github issue to track it.

sethboyles (approved):
LGTM!  The only major thing I would ask you to consider changing is the `ReadinessHealthCheckPolicy < HealthCheckPolicy` inheritance pattern. I think having a `BaseHealthCheckPolicy` that both policies inherit from would avoid any surprises that might come from changing `HealthCheckPolicy` and seeing the effects in `ReadinessHealthCheckPolicy`

app/models/helpers/health_check_types.rb
ameowlia: I added a story in our epic to do it.

app/models/helpers/health_check_types.rb
ameowlia: ✅ done!

ameowlia:
@sethboyles The latest commit adds BaseHealthCheckPolicy class for both the ReadinessHealthCheckPolicy and HealthCheckPolicy classes to inherit from. Both the ReadinessHealthCheckPolicy and HealthCheckPolicy are very slimmed down now. I wasn't sure about the best way to test all these components.

**Options for testing health check policy classes**
1. Test all the main logic once in base_health_check_policy_spec. Then only test the differences on each of the inheriting classes.
2. Don't test BaseHealthCheckPolicy at all since nothing uses it explicitly. Put all tests in on ReadinessHealthCheckPolicy and HealthCheckPolicy.
3. Test everywhere! Put tests on BaseHealthCheckPolicy, ReadinessHealthCheckPolicy, and HealthCheckPolicy. 👈 This is the current state of the world.

I chose option 3. I am happy to change the testing pattern if you would prefer a different option.

sethboyles:
LGTM!

PR #3362, opened 2023-07-27
Readiness health checks story 3 add to lrps

ameowlia:
This PR is part of a larger effort detailed [here in this Cloud Controller PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/3351) and here in [this CFF RFC](https://github.com/cloudfoundry/community/blob/rfc-readiness-healthchecks/toc/rfc/rfc-draft-readiness-healthchecks.md).

In this story, the 3 new app manifest properties ([created in this PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/3354)), are set in the db.

### New App Manifest Properties
```
applications:
- name: test-app
  processes:
  - type: web
    health-check-http-endpoint: /health
    health-check-invocation-timeout: 2
    health-check-type: http
    timeout: 80
    readiness-health-check-http-endpoint: /ready       # 👈 new property
    readiness-health-check-invocation-timeout: 2       # 👈 new property
    readiness-health-check-type: http                  # 👈 new property
```

### Acceptance Setup
1. Deploy with diego-release develop. 
2. Get on the VM with the bbs and run `cfdot desired-lrps | jq .check_definition`

 ### Acceptance Steps
1. When I do not set any readiness health check properties, then NO readiness checks are set on the LRP.
1. When I provide valid properties for http readiness health checks they are set on the LRP.
  ```
    readiness-health-check-http-endpoint: /ready
    readiness-health-check-invocation-timeout: 2
    readiness-health-check-type: http
  ```

1. When I provide valid properties for port readiness health checks they are set on the LRP.
  ```
    readiness-health-check-type: port
  ```
  
1. When I provide valid properties for process readiness health checks they are NOT set on the LRP.
  ```
    readiness-health-check-type: process
  ```

### Previous PRs
* https://github.com/cloudfoundry/cloud_controller_ng/pull/3354
* https://github.com/cloudfoundry/cloud_controller_ng/pull/3361

### DIFF
--- a/lib/cloud_controller/diego/app_recipe_builder.rb
+++ b/lib/cloud_controller/diego/app_recipe_builder.rb
@@ -196,15 +196,10 @@ module VCAP::CloudController
         checks = generate_liveness_and_startup_health_check_defintion(lrp_builder)
         readiness_checks = generate_readiness_health_check_definition(lrp_builder)
 
-        if checks.empty? && readiness_checks.empty?
-          return
-        elsif checks.empty?
-          return  ::Diego::Bbs::Models::CheckDefinition.new(readiness_checks: readiness_checks)
-        elsif readiness_checks.empty?
-          return  ::Diego::Bbs::Models::CheckDefinition.new(checks: checks)
-        else
-          return  ::Diego::Bbs::Models::CheckDefinition.new(checks: checks, readiness_checks: readiness_checks)
-        end
+        params = {}
+        params[:checks] = checks unless checks.empty?
+        params[:readiness_checks] = readiness_checks unless readiness_checks.empty?
+        ::Diego::Bbs::Models::CheckDefinition.new(**params) unless params.empty?
       end
 
       def generate_readiness_health_check_definition(lrp_builder)
--- a/lib/cloud_controller/diego/app_recipe_builder.rb
+++ b/lib/cloud_controller/diego/app_recipe_builder.rb
@@ -17,6 +17,7 @@ module VCAP::CloudController
       METRIC_TAG_VALUE = ::Diego::Bbs::Models::MetricTagValue
 
       MONITORED_HEALTH_CHECK_TYPES = [HealthCheckTypes::PORT, HealthCheckTypes::HTTP, ''].map(&:freeze).freeze
+      MONITORED_READINESS_HEALTH_CHECK_TYPES = [HealthCheckTypes::PORT, HealthCheckTypes::HTTP].map(&:freeze).freeze
 
       def initialize(config:, process:, ssh_key: SSHKey.new)
         @config  = config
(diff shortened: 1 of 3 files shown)


lib/cloud_controller/diego/app_recipe_builder.rb
```diff
@@ -192,15 +193,62 @@ def generate_volume_mounts
...
+
+        if checks.empty? && readiness_checks.empty?
+          return
+        elsif checks.empty?
+          return  ::Diego::Bbs::Models::CheckDefinition.new(readiness_checks: readiness_checks)
+        elsif readiness_checks.empty?
+          return  ::Diego::Bbs::Models::CheckDefinition.new(checks: checks)
+        else
+          return  ::Diego::Bbs::Models::CheckDefinition.new(checks: checks, readiness_checks: readiness_checks)
+        end
```
sethboyles, line 207: I think you can simplify this:

```
params = {}
params[:checks] = checks unless checks.empty?
params[:readiness_checks] = readiness_checks unless readiness_checks.empty?
::Diego::Bbs::Models::CheckDefinition.new(**params) unless !params.empty?
```

(you also don't need explicit returns in ruby)

ameowlia: Done!

spec/unit/lib/cloud_controller/diego/app_recipe_builder_spec.rb
```diff
@@ -570,6 +570,94 @@ module Diego
             end
           end
 
+          context 'readiness health check' do
+            context 'when the readiness health check type is not set' do
+              before do
+                process.readiness_health_check_type = ''
```
sethboyles, line 576: Is this case possible?  I thought the migration that adds a default to the column would mean that all processes would initially have a `readiness_health_check_type` of process

`add_column :readiness_health_check_type, String, default: 'process'`

https://github.com/cloudfoundry/cloud_controller_ng/pull/3361/files#diff-bc587143ec711deb3455361c0dfd4e832be04b77dd46edc603f731e53253481aR6

ameowlia: I don't _think_ so. There is logic in the "normal" health check, where an empty health check type defaults to "port" (I don't even know if this is possible anymore? Perhaps this is part of the remnants of when "none" was allowed?). So I did the same, but for readiness health check types and defaulting to process. I will remove it.

ameowlia: You are right. This test should be about when the readiness health check defaults to process. Good catch. Fixed.

sethboyles (approved):

PR #3393, opened 2023-08-17
Refactor presenter's handling of metadata prefixes

FloThinksPi:

This commit updates how the presenter handles metadata (labels and annotations), treating nil and an empty string as an empty prefix. Earlier behavior would render an empty string as "/key=value" due to the prefix's non-nil status. The correct format in the absence of a prefix should be "key=value"; this commit rectifies this. This adjustment was made anticipating a subsequent change that will alter the default datatype of the prefix column in the database from NULL to an empty string. This step is essential to ensure the proper functioning of a unique key constraint, as multiple NULL values are considered distinct, hence, the UNIQUE constraints in the DB fail to work with NULL values in the columns.

Preparation for https://github.com/cloudfoundry/cloud_controller_ng/pull/3394

### DIFF
--- a/app/actions/annotations_update.rb
+++ b/app/actions/annotations_update.rb
@@ -16,14 +16,25 @@ module VCAP::CloudController
             annotation_klass.find(resource_guid: resource.guid, key: key)&.destroy
           end
 
-          annotation = annotation_klass.find_or_create(resource_guid: resource.guid, key: key_name, key_prefix: prefix)
-
-          if value.nil? && destroy_nil # this is delete
-            annotation.try(:destroy)
+          if value.nil? && destroy_nil # Delete Annotation
+            annotation_klass.where(resource_guid: resource.guid, key: key_name).where(Sequel.or([[:key_prefix, prefix], [:key_prefix, prefix.to_s]])).try(:destroy)
             next
           end
 
-          annotation.update(value: value)
+          begin
+            tries ||= 2
+            annotation_klass.db.transaction(savepoint: true) do
+              annotation = annotation_klass.where(resource_guid: resource.guid, key: key_name).where(Sequel.or([[:key_prefix, prefix], [:key_prefix, prefix.to_s]])).first
+              annotation ||= annotation_klass.create(resource_guid: resource.guid, key: key_name.to_s, key_prefix: prefix.to_s)
+              annotation.update(value: value)
+            end
+          rescue Sequel::UniqueConstraintViolation => e
+            if (tries -= 1).positive?
+              retry
+            else
+              v3_api_error!(:UniquenessError, e.message)
+            end
+          end
         end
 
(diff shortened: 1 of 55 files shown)


johha (approved):
lgtm 👍

app/actions/annotations_update.rb
```diff
@@ -16,14 +16,27 @@ def update(resource, annotations, annotation_klass, destroy_nil: true)
             annotation_klass.find(resource_guid: resource.guid, key: key)&.destroy
           end
 
-          annotation = annotation_klass.find_or_create(resource_guid: resource.guid, key: key_name, key_prefix: prefix)
-
-          if value.nil? && destroy_nil # this is delete
-            annotation.try(:destroy)
+          if value.nil? && destroy_nil # Delete Annotation
+            annotation_klass.where(resource_guid: resource.guid, key: key_name).
+              where(Sequel.|({ key_prefix: prefix }, { key_prefix: prefix.to_s })).try(:destroy)
```
johha, line 21: Maybe `Sequel.or` is more readable

app/models/runtime/stack_label_model.rb
```diff
@@ -1,9 +1,11 @@
 module VCAP::CloudController
   class StackLabelModel < Sequel::Model(:stack_labels)
     many_to_one :stack,
-      class: 'VCAP::CloudController::Stack',
-      primary_key: :guid,
-      key: :resource_guid,
-      without_guid_generation: true
+                class: 'VCAP::CloudController::Stack',
```
johha, line 4: weird formatting

app/actions/annotations_update.rb
```diff
@@ -16,14 +16,27 @@ def update(resource, annotations, annotation_klass, destroy_nil: true)
...
           end
 
-          annotation.update(value: value)
+          begin
+            tries ||= 2
+            annotation_klass.db.transaction(savepoint: true) do
+              annotation = annotation_klass.where(resource_guid: resource.guid, key: key_name.to_s).
+                           where(Sequel.|({ key_prefix: prefix }, { key_prefix: prefix.to_s })).first
+              annotation ||= annotation_klass.create(resource_guid: resource.guid, key: key_name.to_s, key_prefix: prefix.to_s)
+              annotation.update(value: value)
```
johha, line 31: Guess we can clean this up at some point? Then a comment/reminder would be good

FloThinksPi: It will be part of the linked PR and is part of the same BL think its ok.

app/models/runtime/stack_label_model.rb
FloThinksPi: Fixed, somehow rubocop dit this strange

app/actions/annotations_update.rb
FloThinksPi: fixed

johha (approved):

PR #3394, opened 2023-08-17
Ensure uniqueness of labels and annotations

FloThinksPi:

Ensure uniqueness of labels and annotations
This commit includes both code change and a database migration, contributing to ensuring labels and annotations' uniqueness.
The code update inserts an empty string as prefix into the database when the prefix is empty. It also addresses the race condition issue where two concurrent updates on an object’s labels could result in the label/annotation being created twice in the DB. A uniqueness constraint error is now caught, and the find_or_create operation is retried to handle this issue.
The database migration:

- Modifies the default value of the prefix column to an empty string.
- Converts all NULL values in the table to an empty string.
- Restricts the column from accepting NULL values.
- Reduces the length of the 'key' column in annotation tables to a maximum of 63 characters,
  in accordance with the documentation and current API input sanitation practices.
- Identifies and removes all duplicates based on identical values of resource_id, key_prefix, and
  key (annotation tables)/key_name (label tables).
- Implements a unique key constraint on resource_id, key_prefix, and
  key (annotation tables)/key_name (label tables) to prevent new duplicates.

Note that this change heavily utilizes table locking and transactions to ensure the change is carried out consistently and with resilience.
Also the change in the datatype of the key_prefix column is essential as
as for a unique constraint NULL vales are allways different as NULL !=
NULL in SQL99. A unique constraint containing NULL values would be not
of any use since it wont provide proper uniqueness for all columns it
contains.

See also this ADR for more Information: https://github.com/cloudfoundry/cloud_controller_ng/blob/main/decisions/0011-make-labels-and-annotations-unique.md

Solves a race condtition where it was possible to have labels with different values for the same resource persisted in the database.

Needs https://github.com/cloudfoundry/cloud_controller_ng/pull/3393 to be a compatible change on update.
Without it the presenters render the metadata section incorrectly between the time the migration was executed and the new CCs are deployed.

### DIFF
--- a/.rubocop_cc.yml
+++ b/.rubocop_cc.yml
@@ -126,12 +126,8 @@ Style/ExpandPathArguments:
   Exclude:
     - 'db/migrations/20130911111938_encrypt_app_env_json.rb'
     - 'db/migrations/201805*'
-Rails/DangerousColumnNames: # Disabled for old migrations as we cannot change content of those
-  Enabled: true
-  Exclude:
-    - !ruby/regexp /db/migrations/201([0-9]).+\.rb$/
-    - !ruby/regexp /db/migrations/202([0-2]).+\.rb$/
-    - db/migrations/20221125134500_add_request_count_table.rb
+Rails/DangerousColumnNames: # Disabled, in comparison to active_record we need to add the id column manually in sequel
+  Enabled: false
 Rails/SkipsModelValidations: # We don`t want any model at all in migrations and migration specs
   Enabled: true
   Exclude:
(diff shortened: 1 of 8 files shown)


FloThinksPi:
Did a performance test of this migration. Especially the part that finds and deletes duplicates. For a table with 100k entries(30% & 0% duplicates) the migration ran in ~2 Seconds for that table with both rates of duplicates.
For 500k entries and also 30% & 0% duplicates it ran in 15 seconds for that table. But could also be a memory issue of my local postgres. In summary we dont expect the migration to have a shared exclusive lock on a tables for more than 5s which is perfectly fine as requests will wait until the lock is released and are delayed a bit while the migration runs.

PR #3402, opened 2023-08-22
Prepare Annotation Key Column Rename

FloThinksPi:
This change introduces views on all annotation tables where the column
key has the alias key_name. The coding then uses these Views and all
occurences of 'key' are replaced by 'key_name' in the CCs code.

Later on on can then change the column name of the 'key' column in all
annotations tabels to 'key_name' transparently without downtime and
remove the view in a third step.

This change is a refactoring to streamline annotations and labels
and finish tasks proposed in
/decitions/0004-adding-key-prefix-to-annotations.md

Also see this ADR for more information https://github.com/cloudfoundry/cloud_controller_ng/blob/main/decisions/0011-make-labels-and-annotations-unique.md


app/models/runtime/app_label_model.rb
johha, line 1: Can be removed

app/models/services/service_broker_update_request_label_model.rb
```diff
@@ -5,5 +5,6 @@ class ServiceBrokerUpdateRequestLabelModel < Sequel::Model(:service_broker_updat
       key: :resource_guid,
       without_guid_generation: true
     include MetadataModelMixin
+    include MetadataModelMixin
```
johha, line 8: Twice

johha (approved):

PR #3407, opened 2023-08-24
Introduce MigrationMixin and document Sequel migration tests

FloThinksPi:
This commit introduces the MigrationMixin module, a valuable utility for Sequel migration testing. This new module reverts the database schema to its version prior to a specific migration, provides a directory with the migration file for running a given migration in a test, and restores the schema to its full state after test completion to avoid half-migrated database scenarios. The accompanying updates to the documentation elaborate on the distinctive operating nature of Sequel migration tests, which involve invoking the Down migration and restoring the database state before executing the test-specific migration. Constraints and restrictions on Sequel migration test writing have also been detailed, stressing the need to exclude Cloud Controller (CC) code in migration tests to maintain consistency in test outcomes and behaviors. The usage section illustrates how the MigrationMixin accesses and executes a specific migration while managing other aspects of migration. An example test for a migration involving the isolation_segments table has been included. This test verifies that the original name and guid are persisted even after the migration. It highlights the use of Sequel for direct database interactions, ensuring that tests continue to work effectively even with future modifications to the table. These revisions aim to offer clearer guidelines on Sequel migration testing, enhanced by the introduction of MigrationMixin, which streamlines the process and handles complexities of migrations.


philippthun:
Here is the existing [CAPI-Migration-Style-Guide](https://github.com/cloudfoundry/cloud_controller_ng/wiki/CAPI-Migration-Style-Guide). We should merge the contents and decide for a single location (and reference it from the other location).

spec/migrations/helpers/migration_mixin.rb
```diff
@@ -0,0 +1,35 @@
+module MigrationMixin
```
philippthun, line 1: This mixin looks _interesting_... I've checked out the code and turned it into a **shared context**. Looks less hacky, I would say: https://github.com/sap-contributions/cloud_controller_ng/commit/ad8441b3225449a94fbc9879b8474324a7adb058

spec/migrations/Readme.md
```diff
@@ -0,0 +1,107 @@
...
+1. Separate data-changing migrations from schema-changing DDL statements to leverage at least the database rollback function for DML Statements. Explicitly wrap data changes in a transaction that doesn't include DDL statements. Avoid including DDL in that transaction otherwise MySql will autocommit data changes and a rollback is impossible.
+1. Try to handle DML (Data Manipulation Language) statements entirely within the DB, avoiding any looping over returned data in the migration process. This helps to reduce the time a table is locked. For example, instead of finding all duplicates with a select and then iterating over the results in Ruby and deleting the rows, use a subquery to let the DB perform the heavy lifting. Since these operations should be performed with a table lock to prevent new duplicates, it is advantageous to minimize runtime as much as possible. Involving Ruby and dealing with large tables could increase runtimes to several minutes, leading to locked API requests and potential unavailability.
+   ```ruby
+   min_ids_subquery = self.from("mytable__subquery".to_sym).
```
philippthun, line 33: I know, this is only an example, but shouldn't both queries be done on the same table?
```suggestion
   min_ids_subquery = self[table.to_sym].
```

spec/migrations/Readme.md
```diff
@@ -0,0 +1,107 @@
...
+            ...
+            done
+         end
+       end
+     end
+   end
+   
+   ```
+1. If you're writing a uniqueness constraint where some of the values can be null, remember that `null != null`. For instance, the values `[1, 1, null]` and `[1, 1, null]` are considered unique. Uniqueness constraints only work on columns that do not allow `NULL` as a value. If this is the case, change the column to disallow `NULL` and set the default to an empty string instead.
+1. If you need to execute different operations for MySQL and PostgreSQL, you can check the database type as follows: `... if database_type == :postgres` or `... if database_type == :mysql`.
```
philippthun, line 71:
```suggestion
1. If you need to execute different operations for MySQL and Postgres, you can check the database type as follows: `... if database_type == :postgres` or `... if database_type == :mysql`.
```

spec/migrations/Readme.md
```diff
@@ -0,0 +1,107 @@
...
+1. It's recommended to use the MigrationMixin, as it ensures that the database schema first reverts to the version before the migration you aim to test. This mixin also provides a directory containing a single migration for running a particular migration within a test. When the test is done, this mixin makes sure to restore the correct schema by running the migrations that post-date the one being tested. Thus, avoiding cases of a half-migrated database that could result in random test failures. It also makes sure to test not every migration that comes after the test migration, but just a single migration is executed and then the expected behavior is evaluated.
+
+### Usage
+
+Here’s a guide on how to write a migration spec. The `MigrationMixin` uses the `migration_file` variable, containing the filename of the migration under review. This mixin also provides a `migration_to_test` variable to perform the specific migration. All other processes, including migrating one version before the test migration and fully migrating the table after the test has finished, are handled automatically by the Mixin.
+
+```ruby
```
philippthun, line 84: Would need to be adjusted to the `shared_context`.

spec/migrations/Readme.md
```diff
@@ -0,0 +1,107 @@
...
+1. All constraints (views, indices, etc.) must be named. If not, MySQL and PostgreSQL default to different constraint names, complicating future constraint editing.
+1. While Postgres allows indices/constraints to incorporate unlimited-sized columns, MySQL doesn't. An index/constraint can hold columns (all combined) with a size of only [3072 bytes for InnoDB](https://dev.mysql.com/doc/refman/8.1/en/column-indexes.html). Plan accordingly for column sizes and their potential future use in a combined index/constraint. Reducing column sizes later is generally not feasible.
+1. A size should always be specified for a string (and text should not be used). Postgres and MySQL have different size limits on String and TEXT fields. To ensure data can be migrated across the two databases (and others in future) without issues, a maximum size should always be specified. A rubocop linter has enforced this since 2017-07-30.
+   1. For MySQL, a string is varchar(255), String, text: true has a max size of 16_000 for UTF-8-encoded DBs.
```
philippthun, line 18:
```suggestion
   1. For MySQL, `String` is `varchar(255)`, `String, text: true` has a max size of 16_000 for UTF-8-encoded DBs.
```

spec/migrations/Readme.md
```diff
@@ -0,0 +1,107 @@
...
+   1. For Postgres, both String and String, text: true are TEXT and have a max size of ~1GB.
```
philippthun, line 19:
```suggestion
   1. For Postgres, both `String` and `String, text: true` are TEXT and have a max size of ~1GB.
```

spec/migrations/Readme.md
FloThinksPi: the important thing is the double underscore here which is a magic funcitonality in sequel to alias the subquery which is required for locking the table since locks must be on subquery aliases in mysql

PR #3329, opened 2023-06-29
Add validation for resource match mode

philippthun:
If provided, the `mode` field (resource match object) must be a string. Without this fix a request to `/v3/resource_matches` with
```
  { .. "mode": 644 .. }
```
resulted in an `UnknownError`. Now the error message
```
  "array contains at least one resource with a non-string mode"
```
is returned instead.

Furthermore the format is checked and if the provided value is not a _POSIX file permissions_ string, the following error message is returned:
```
  "array contains at least one resource with an incorrect mode"
```

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)

PR #3409, opened 2023-08-24
Revert "Add validation for resource match mode"

FloThinksPi:
This reverts commit 22e21222a4bf995613f014bdd9f4afdb0b37cb00 due to the mode limitation to 3-4 characters beeing against the POSIX Standard as it defines 9 permission bits and 3 file bits with no limit for more bits. As the documentation is not specific on the further limitations the currently implemented validation is incorrect.

See also:
https://en.wikipedia.org/wiki/Unix_file_types#Numeric 
https://pubs.opengroup.org/onlinepubs/9699919799/

#3329 

### DIFF
--- a/app/messages/resource_match_create_message.rb
+++ b/app/messages/resource_match_create_message.rb
@@ -22,71 +22,53 @@ module VCAP::CloudController
 
     private
 
-    PERMISSIONS_REGEX = /^[0-7]{3,4}$/
-
     def each_resource
       if resources.is_a?(Array)
         resources.each do |r|
           checksum_validator(r[:checksum])
           size_validator(r[:size_in_bytes])
-          mode_validator(r[:mode])
         end
       end
     end
 
-    RESOURCE_ERROR_PREAMBLE = 'array contains at least one resource with'.freeze
+    RESOURCE_ERROR_PREAMBLE = 'array contains at least one resource with a'.freeze
 
     def checksum_validator(checksum)
       unless checksum.is_a?(Hash)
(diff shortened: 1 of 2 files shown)


FloThinksPi:
Findings on the API:

The API Docs specify the mode shall be a POSIX file mode(could be more strict enforced in the docs)[[1]](https://v3-apidocs.cloudfoundry.org/version/3.100.0/#the-resource-match-object)
The POSIX standard does not define a maximum length of file permission bits. However it defines the type mode must at least be 12 bits of permissions and 4 bits for the file type which makes it 16 bits(6 octal digits). Writing can be shortened to smaller than 3 digits too. [[2]](https://pubs.opengroup.org/onlinepubs/009604499/basedefs/sys/stat.h.html) [[3]](https://en.wikipedia.org/wiki/Unix_file_types#Numeric)
That means e.g. POSXI Permissions 100700 = 0700 = 07.
That in return means the implementation of the API is incorrect and the Java Client is correct in supplying 16bit mode flag.
At least when it comes to the CF API Docs and one follows them strictly. It doesnt make sense of course since the first 4 bits define the File Type which doesn't make sense here but since the API doc just says POSIX compliant mode there is nothing wrong with supplying a 16bit mode and letting the api ignore the 4 file type bits. Also suppling 2 digit long file permissions would fail now but looks like those are also valid POSIX permission values for mode.

FloThinksPi:
Closed in favour of #3414

FloThinksPi:
Lets do a revert for the short term fix, we are not sure how exactly the forward code fix should look like, likely needs more time and this should be gone in the next capi release for sure.

svkrieger (approved):
Makes sense to revert it for now and implement a more loose validation, to not break officially supported clients.

PR #3424, opened 2023-09-05
Fix migration deadlock on busy systems

fearoffish:
In #3417 there was a change to fix deadlocks created by table joins. However, on our busy system at gov.uk a deadlock still occurred and our deployment attempts in production failed repeatedly.

We fixed this and successfully deployed by forcing a `no_transaction` outside the `up` and `down` but not inside the index migration. I'm not entirely sure why this worked, but our assumption was that the `up` created a transaction and then we end up with sub-transactions for each migration.

Either way, this is the only way we could deploy on a busy system.

I'm open to a discussion on why if anyone can explain, and maybe tell me why this was the wrong solution.

### DIFF
--- a/db/migrations/20230725110800_restructure_index_for_improved_pagination_order.rb
+++ b/db/migrations/20230725110800_restructure_index_for_improved_pagination_order.rb
@@ -547,6 +547,7 @@ Sequel.migration do
     },
   ]
 
+  no_transaction
   up do
     index_migration_data.each do |index_migration|
       transaction do


FloThinksPi:
Also woth mentioning this has happened before and the solution they found was splitting each table into an own migration file: https://github.com/cloudfoundry/cloud_controller_ng/commit/00321c70a8fca52a9a5e393f0c30d12b25d35a8a

FloThinksPi:
@philippthun has tested the behaviour of `no_transaction` as well as the `transaction do` blocks inside Sequel migration by looking at the RAW SQL Logs of CC. He found out Sequel behaves as following:

1. The  `transaction do` block inside a migration block (up/down/change do) has no effect at all by default. In the logged SQL statements there is only one ocurence of `BEGIN` for the whole migration. If removing the `transaction do` block still only one ocurence of `BEGIN` can be seen.
2. When adding `no_transaction` before the migration blocks and additionally removing the `transaction do` block there are no `BEGIN` statements logged at all.
3. When adding  `no_transaction` before the migration blocks and having the `transaction do` block still in place, 24 `BEGIN` statements are logged. As wanted now every loop run over the tables opens an own transaction and closes before starting with the next table.

This is not so well documented in Sequel, or i just did`t find it. This in return means when migrating multiple tables not only pack the contents of each table in an own `transaction do` block, now one also is absolutely required to add `no_transaction` before the migration blocks.

Will add this to the docs i prepared as one can do so many things wrong in Sequel migrations as the framework does a lot of magic and psql and mysql behave quite differently.
Also i will open another PR to introduce this property also for other migrations that suffer from the same problem.

PR #3434, opened 2023-09-12
add script to start docker container with db for unit tests

ameowlia:
I don't work on a machine that has a local DB. When I started working on readiness healthcheck stuff I spent 2+ hours just trying to get one setup. In the end, I made this script from a similar one in cf-networking-release. I don't promise that it works for every CAPI use case (it only works with postgres for example), but it works great for when a new person wants to run some tests and iterate quickly.

To use, run: `./scripts/docker-shell-with-started-db`

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,14 @@ keep it up to date, documenting the purpose of the various types of tests.
 
 By default `rspec` will randomly pick between postgres and mysql.
 
+#### Running the Units Tests in Docker
+
+If you don't want to set up a up a test DB locally, you can run
+`./scripts/start-db-in-docker.sh`. This script will start a docker container
+with a running postgres. Then you can run the tests from the docker container
+by running `rspec`.
+
+#### Running the Test DB Locally
 If postgres is not running on your OSX machine, you can start up a server by doing the following:
 ```
 brew services start postgresql
--- a/scripts/start-db-in-docker.sh
+++ b/scripts/start-db-in-docker.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+set -e -u
+
+function bootDB {
+  db="$1"
+   
+  launchDB="(/postgres-entrypoint.sh postgres &> /var/log/postgres-boot.log) &"
+  testConnection="psql -h localhost -U postgres -c '\conninfo'"
+  createTestDB="psql -h localhost -U postgres -c 'create database cc_test'"
+
+  echo -n "booting ${db}"
+  eval "$launchDB"
+  for _ in $(seq 1 60); do
+    if eval "${testConnection}" &> /dev/null; then
+      break
+    fi
+    echo -n "."
+    sleep 1
+  done
+
+  if eval "${testConnection}" &> /dev/null; then
+    echo "connection established to ${db}"
+  else 
+    echo "unable to connect to ${db}"
+    exit 1
+  fi
+}
+
+function moreSetup {
+  apt-get update
+  apt-get install -y libpq-dev default-libmysqlclient-dev
+  bundle install
+  rake db:create
+}
+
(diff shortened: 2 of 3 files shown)


philippthun:
Thanks @ameowlia for your PR. It's always good to lower the entry barrier for new contributors.

Why did you decide to run `rake db:create` (and also the tests) inside the container? I would prefer an even simpler setup, i.e. docker container with database and everything else done from the outside.

We also have a local setup with multiple containers that @FloThinksPi converted into a Codespace (see #2881). Did you had a chance to try out this setup?

ameowlia:
👋 Hi @philippthun, thanks for the response.

> Why did you decide to run rake db:create (and also the tests) inside the container? I would prefer an even simpler setup, i.e. docker container with database and everything else done from the outside.

This is the pattern we follow for the runtime components. That way someone can run the tests without having anything but docker setup on their machine. Though I can understand your preference. 

> We also have a local setup with multiple containers that @FloThinksPi converted into a Codespace (see https://github.com/cloudfoundry/cloud_controller_ng/pull/2881). Did you had a chance to try out this setup?

Thank you for the pointer to this. I didn't see this before. I've never used codespaces before. I am curious to see if it works without an IDE since my team develops entirely in vim on gcp vms. 

Feel free to close this issue if this testing patterns isn't one you want to endorse.

FloThinksPi:
@ameowlia The setup we added work both as a github codespace and also localy by calling e.g. `./devenv.sh create` on your machine. Prerequsits and more details are available in the Readme: https://github.com/cloudfoundry/cloud_controller_ng#predefined-development-environment :) 

Happy to hear if this setup supports you working mode :)

README.md
```diff
@@ -116,6 +116,14 @@ keep it up to date, documenting the purpose of the various types of tests.
 
 By default `rspec` will randomly pick between postgres and mysql.
 
+#### Running the Units Tests in Docker
+
+If you don't want to set up a up a test DB locally, you can run
```
philippthun, line 121:
```suggestion
If you don't want to set up a test DB locally, you can run
```

scripts/start-db-in-docker.sh
```diff
@@ -0,0 +1,50 @@
...
+
+SCRIPT_PATH="$(cd "$(dirname "${0}")" && pwd)"
+
+function bootDB {
+  db="$1"
+
+  if [ "${db}" = "postgres" ]; then
+    launchDB="(/postgres-entrypoint.sh postgres &> /var/log/postgres-boot.log) &"
+    testConnection="psql -h localhost -U postgres -c '\conninfo'"
+  elif [ "${db}" = "mysql" ]  || [ "${db}" = "mysql-5.6" ] || [ "${db}" = "mysql8" ]; then
```
philippthun, line 13: In `scripts/docker-shell-with-started-db` the `db` variable will be set to `mysql` if e.g. `DB=mysql8`; so the two `OR`s are not needed.

scripts/start-db-in-docker.sh
```diff
@@ -0,0 +1,50 @@
...
+
+function moreSetup {
+  apt-get update
+  apt-get install -y libpq-dev default-libmysqlclient-dev
+  bundle install
+  rake db:create
+}
+
+cd /cloud_controller_ng
+export GOPATH=$PWD
```
philippthun, line 44: Is this needed? We don't have Golang code...

README.md
```diff
@@ -116,6 +116,14 @@ keep it up to date, documenting the purpose of the various types of tests.
...
+`./scripts/start-db-in-docker.sh`. This script will start a docker container
```
philippthun, line 122: You run `scripts/docker-shell-with-started-db` from the command line, right?

PR #3442, opened 2023-09-19
Throw different error when route binding status is delete_failed or delete_in_progess / Treat route binding in status create_failed as non-existent

kathap:
This is a fix as part of #2637. It contains changes for: 
POST /v3/service_route_bindings

Two behaviours change with this PR:
1: When a service route binding is in state delete_in_progess or delete_failed, we want an error different from "Route is already bound to the service instance" to be thrown. It will now throw an error with "The binding is getting deleted or its deletion failed". 2: When a service route binding is in state create_failed, it should recreate the binding and not reject the request with "Route is already bound to the service instance".

With regard to point 1: it's important to note that the initial statement may not be entirely accurate, as there's a possibility that orphan mitigation processes may have already removed the binding in question. The Cloud Foundry Command Line Interface (CF CLI) correctly interprets the situation as a success. This is because the message conveyed to the CF CLI indicates that the task it was instructed to perform as already been completed. As a result, if the CF CLI is attempting to create a service route binding but finds that a binding already exists in a state where it's either failed deletion or is pending deletion, the CF CLI will consider this a failure.

#2657
#2636


app/actions/service_route_binding_create.rb
```diff
@@ -43,19 +46,25 @@ def event_repository
...
-        already_exists! if route.service_instance == service_instance
-        already_bound! if route.service_instance
+        already_bound! if route.service_instance && route.service_instance != service_instance
         return unless service_instance.managed_instance?
 
         service_instance_not_found! if service_instance.create_failed?
         operation_in_progress! if service_instance.operation_in_progress?
       end
 
+      def validate_binding!(binding, _route, _service_instance)
```
philippthun, line 61: You should remove the unneeded parameters.

philippthun:
@kathap - I've added one small comment, otherwise the changes look good! Could you please squash your commits, then I will approve and merge the PR.

app/actions/service_route_binding_create.rb
kathap: Solved in #3469

kathap:
Closing this one, moved it to #3469

kathap:
moved it to https://github.com/cloudfoundry/cloud_controller_ng/pull/3469 (get clean commits without main merge)

PR #3443, opened 2023-09-19
Don't show service route binding parameters not in state create succeeded

kathap:
This is a fix as part of https://github.com/cloudfoundry/cloud_controller_ng/issues/2637. It contains changes in behaviour for:
- GET /v3/service_route_bindings/:guid/parameters

For GET /v3/service_route_bindings/:guid/parameters, if the state is not create succeeded, it should not return parameters. It should raise an error when GET /v3/service_route_bindings/:guid/parameters is called instead of showing empty data.

#2636
#2679


app/controllers/v3/service_route_bindings_controller.rb
```diff
@@ -85,6 +85,9 @@ def destroy
 
   def parameters
     route_binding_not_found! unless @route_binding && can_read_from_space?(@route_binding.route.space)
+    if @route_binding.create_initial? || @route_binding.create_failed?
```
philippthun, line 88: The check in the service credential bindings controller is `unless create_succeeded?` Thus I would prefer the same check here.

app/controllers/v3/service_route_bindings_controller.rb
```diff
@@ -248,6 +251,12 @@ def route_binding_not_found!
     resource_not_found!(:service_route_binding)
   end
 
+  def route_binding_not_found_with_message!(service_route_binding)
```
philippthun, line 254: The name could be shortened to `not_found_with_message!`.

app/controllers/v3/service_route_bindings_controller.rb
```diff
@@ -248,6 +251,12 @@ def route_binding_not_found!
...
+    type = service_route_binding.is_a?(RouteBinding) ? 'route' : 'binding'
```
philippthun, line 255: There is no different type, so this line is not needed.

app/controllers/v3/service_route_bindings_controller.rb
```diff
@@ -248,6 +251,12 @@ def route_binding_not_found!
...
+    state = service_route_binding.last_operation.state
+    resource_not_found_with_message!("Creation of binding #{type} #{state}")
```
philippthun, line 257: Similar to the service credential bindings controller, this error should be returned for "Creation" and "Deletion", thus the `last_operation.type` needs to be taken into account.

philippthun, line 257: The message should be `#{operation} of route binding #{state}`.

spec/request/service_route_bindings_spec.rb
```diff
@@ -1618,7 +1618,7 @@
         end
       end
 
-      context 'when there is an operation in progress' do
+      context "when last binding operation is in 'create in progress' state" do
```
philippthun, line 1621: I would add tests for the following states:
- `create succeeded`
- `delete in progress`
- `delete failed`

The `create initial` test should not be needed as it is handled in the exact same way as `create in progress`.

philippthun:
@kathap - Changes look good! Could you please squash your commits, then I will approve and merge the PR.

moleske (approved):
looks good and since @philippthun also said it looks good I'm just using github's squash and merge to get the changes in one commit

PR #3445, opened 2023-09-21
Prometheus on Thin

svkrieger:

Adjust Prometheus endpoint:
- Remove deprecated metrics and metrics which have been found not useful according to discussions in the community
- Make use of the DependencyLocator for retrieving a singleton of the PrometheusUpdater and PeriodicUpdater
- Change vitals_uptime to vitals_started_at
- Emit cc_staging_requests_total metric
- Apply prometheus best practices like naming, base units, using labels, initialising metrics for discoverability
- Use counter metrics for metrics which do not decrease
- Remove metrics, which are emitted on the scheduler VM. Those metrics currently cannot be collected and will be still emitted via statsd

This implementation is meant for experimental usage. It is difficult to decide on the best metric type or bucket size initially, without having it used in productive environments. The metrics emitted via Prometheus can be collected and displayed in dashboards while still using the statsd metrics for alert/real monitoring. We should communicate that breaking changes are likely, as long as this is treated as a experimental feature.

### DIFF
--- a/app/controllers/internal/metrics_controller.rb
+++ b/app/controllers/internal/metrics_controller.rb
@@ -8,6 +8,7 @@ module VCAP::CloudController
       get '/internal/v4/metrics', :index
 
       def index
+        CloudController::DependencyLocator.instance.periodic_updater.update!
         [200, Prometheus::Client::Formats::Text.marshal(Prometheus::Client.registry)]
       end
     end
--- a/app/controllers/internal/metrics_controller.rb
+++ b/app/controllers/internal/metrics_controller.rb
@@ -1,6 +1,5 @@
 require 'prometheus/client'
 require 'prometheus/client/formats/text'
-require 'cloud_controller/metrics/prometheus_updater'
 
 module VCAP::CloudController
   module Internal
@@ -9,16 +8,6 @@ module VCAP::CloudController
       get '/internal/v4/metrics', :index
 
       def index
-        periodic_updater = VCAP::CloudController::Metrics::PeriodicUpdater.new(
-          Time.now.utc,
-          Steno::Sink::Counter.new,
-          Steno.logger('cc.api'),
-          [
-            VCAP::CloudController::Metrics::StatsdUpdater.new,
-            VCAP::CloudController::Metrics::PrometheusUpdater.new
-          ]
-        )
-        periodic_updater.update!
         [200, Prometheus::Client::Formats::Text.marshal(Prometheus::Client.registry)]
       end
     end
--- a/app/jobs/diego/sync.rb
+++ b/app/jobs/diego/sync.rb
@@ -7,9 +7,8 @@ module VCAP::CloudController
   module Jobs
     module Diego
       class Sync < VCAP::CloudController::Jobs::CCJob
-        def initialize(statsd=Statsd.new, prometheus_updater=VCAP::CloudController::Metrics::PrometheusUpdater.new)
+        def initialize(statsd=Statsd.new)
           @statsd = statsd
-          @prometheus_updater = prometheus_updater
         end
 
         def perform
@@ -28,7 +27,6 @@ module VCAP::CloudController
             elapsed_ms = ((finish - start) * 1000).round
 
             @statsd.timing('cc.diego_sync.duration', elapsed_ms)
-            @prometheus_updater.report_diego_cell_sync_duration((finish - start))
           end
         end
 
--- a/app/jobs/diego/sync.rb
+++ b/app/jobs/diego/sync.rb
@@ -28,7 +28,7 @@ module VCAP::CloudController
             elapsed_ms = ((finish - start) * 1000).round
 
             @statsd.timing('cc.diego_sync.duration', elapsed_ms)
-            @prometheus_updater.report_diego_cell_sync_duration(elapsed_ms)
+            @prometheus_updater.report_diego_cell_sync_duration((finish - start))
           end
         end
 
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -64,16 +64,23 @@ module CloudController
         register(:periodic_updater,
                  VCAP::CloudController::Metrics::PeriodicUpdater.new(
                    Time.now.utc,
-                   Steno::Sink::Counter.new,
+                   log_counter,
                    Steno.logger('cc.api'),
-                   VCAP::CloudController::Metrics::StatsdUpdater.new(statsd_client),
+                   statsd_updater,
                    prometheus_updater
                  ))
     end
 
     def prometheus_updater
-      register(:prometheus_updater, VCAP::CloudController::Metrics::PrometheusUpdater.new) unless @dependencies[:prometheus_updater]
-      @dependencies[:prometheus_updater]
+      @dependencies[:prometheus_updater] || register(:prometheus_updater, VCAP::CloudController::Metrics::PrometheusUpdater.new)
+    end
+
+    def statsd_updater
+      @dependencies[:statsd_updater] || register(:statsd_updater, VCAP::CloudController::Metrics::StatsdUpdater.new(statsd_client))
+    end
+
+    def log_counter
+      @dependencies[:log_counter] || register(:log_counter, Steno::Sink::Counter.new)
     end
 
     def stagers
@@ -344,8 +351,12 @@ module CloudController
     end
 
     def statsd_client
-      @dependencies[:statsd_client] ||
+      if @dependencies[:statsd_client]
+        @dependencies[:statsd_client]
+      else
+        Statsd.logger = Steno.logger('statsd.client')
         register(:statsd_client, Statsd.new(config.get(:statsd_host), config.get(:statsd_port)))
+      end
     end
 
     private
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -68,8 +68,7 @@ module CloudController
                    Steno.logger('cc.api'),
                    VCAP::CloudController::Metrics::StatsdUpdater.new(statsd_client),
                    prometheus_updater
-                 )
-        )
+                 ))
     end
 
     def prometheus_updater
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -59,6 +59,19 @@ module CloudController
       @dependencies[:runners] || register(:runners, VCAP::CloudController::Runners.new(config))
     end
 
+    def periodic_updater
+      @dependencies[:periodic_updater] ||
+        register(:periodic_updater,
+                 VCAP::CloudController::Metrics::PeriodicUpdater.new(
+                   Time.now.utc,
+                   Steno::Sink::Counter.new,
+                   Steno.logger('cc.api'),
+                   VCAP::CloudController::Metrics::StatsdUpdater.new(statsd_client),
+                   prometheus_updater
+                 )
+        )
+    end
+
     def prometheus_updater
       register(:prometheus_updater, VCAP::CloudController::Metrics::PrometheusUpdater.new) unless @dependencies[:prometheus_updater]
       @dependencies[:prometheus_updater]
--- a/lib/cloud_controller/deployment_updater/scheduler.rb
+++ b/lib/cloud_controller/deployment_updater/scheduler.rb
@@ -10,13 +10,11 @@ module VCAP::CloudController
           with_error_logging('cc.deployment_updater') do
             config = CloudController::DependencyLocator.instance.config
             statsd_client = CloudController::DependencyLocator.instance.statsd_client
-            prometheus_updater = CloudController::DependencyLocator.instance.prometheus_updater
 
             update_step = proc {
               update(
                 update_frequency: config.get(:deployment_updater, :update_frequency_in_seconds),
-                statsd_client: statsd_client,
-                prometheus_updater: prometheus_updater
+                statsd_client: statsd_client
               )
             }
 
@@ -42,7 +40,7 @@ module VCAP::CloudController
 
         private
 
-        def update(update_frequency:, statsd_client:, prometheus_updater:)
+        def update(update_frequency:, statsd_client:)
           logger = Steno.logger('cc.deployment_updater.scheduler')
 
           update_start_time = Time.now
@@ -54,7 +52,6 @@ module VCAP::CloudController
           ##       so feed in the entire value!
           update_duration_ms = update_duration * 1000
           statsd_client.timing('cc.deployments.update.duration', update_duration_ms)
-          prometheus_updater.report_deployment_duration(update_duration)
 
           logger.info("Update loop took #{update_duration}s")
 
--- a/lib/cloud_controller/deployment_updater/scheduler.rb
+++ b/lib/cloud_controller/deployment_updater/scheduler.rb
@@ -54,7 +54,7 @@ module VCAP::CloudController
           ##       so feed in the entire value!
           update_duration_ms = update_duration * 1000
           statsd_client.timing('cc.deployments.update.duration', update_duration_ms)
-          prometheus_updater.report_deployment_duration(update_duration_ms)
+          prometheus_updater.report_deployment_duration(update_duration)
 
           logger.info("Update loop took #{update_duration}s")
 
--- a/lib/cloud_controller/metrics/periodic_updater.rb
+++ b/lib/cloud_controller/metrics/periodic_updater.rb
@@ -46,9 +46,9 @@ module VCAP::CloudController::Metrics
     def update_task_stats
       running_tasks = VCAP::CloudController::TaskModel.where(state: VCAP::CloudController::TaskModel::RUNNING_STATE)
       running_task_count = running_tasks.count
-      running_task_memory = running_tasks.sum(:memory_in_mb)
-      running_task_memory = 0 if running_task_memory.nil?
-      [@statsd_updater, @prometheus_updater].each { |u| u.update_task_stats(running_task_count, running_task_memory) }
+      running_task_memory = running_task_memory.nil? ? 0 : running_tasks.sum(:memory_in_mb)
+      @statsd_updater.update_task_stats(running_task_count, running_task_memory)
+      @prometheus_updater.update_task_stats(running_task_count, running_task_memory * 1000 * 1000)
     end
 
     def update_log_counts
@@ -85,7 +85,8 @@ module VCAP::CloudController::Metrics
       end
 
       pending_job_count_by_queue.reverse_merge!(@known_job_queues)
-      [@statsd_updater, @prometheus_updater].each { |u| u.update_job_queue_length(pending_job_count_by_queue, total) }
+      @statsd_updater.update_job_queue_length(pending_job_count_by_queue, total)
+      @prometheus_updater.update_job_queue_length(pending_job_count_by_queue)
     end
 
     def update_thread_info
@@ -105,7 +106,8 @@ module VCAP::CloudController::Metrics
       end
 
       failed_jobs_by_queue.reverse_merge!(@known_job_queues)
-      [@statsd_updater, @prometheus_updater].each { |u| u.update_failed_job_count(failed_jobs_by_queue, total) }
+      @statsd_updater.update_failed_job_count(failed_jobs_by_queue, total)
+      @prometheus_updater.update_failed_job_count(failed_jobs_by_queue)
     end
 
     def update_vitals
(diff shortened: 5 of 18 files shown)


app/controllers/internal/metrics_controller.rb
```diff
@@ -9,16 +8,7 @@ class MetricsController < RestController::BaseController
...
-          Time.now.utc,
-          Steno::Sink::Counter.new,
-          Steno.logger('cc.api'),
-          [
-            VCAP::CloudController::Metrics::StatsdUpdater.new,
-            VCAP::CloudController::Metrics::PrometheusUpdater.new
-          ]
-        )
-        periodic_updater.update!
+        CloudController::DependencyLocator.instance.periodic_updater.update!
```
svkrieger, line 11: Currently, whenever the metrics endpoint is being called, we will collect fresh metrics via PeriodicUpdater.update!. As for now there is only one scraper, which will call this endpoint, there won't be a performance problem with that. If in the future there might be several or even an unknown number of scrapers, which could access this endpoint, we might want to remove this update and serve the metrics, which have been periodically collected. TODO: If we update the metrics, each time the endpoint is being called, then the PeriodicUpdater does actually not need to collect metrics periodically for the Prometheus endpoint.

app/jobs/diego/sync.rb
```diff
@@ -28,7 +27,6 @@ def perform
             elapsed_ms = ((finish - start) * 1000).round
 
             @statsd.timing('cc.diego_sync.duration', elapsed_ms)
-            @prometheus_updater.report_diego_cell_sync_duration(elapsed_ms)
```
svkrieger, line 31: This is being emitted on the scheduler VM. Therefore it can't be collected because there is no web server running.

lib/cloud_controller/deployment_updater/scheduler.rb
```diff
@@ -54,7 +52,6 @@ def update(update_frequency:, statsd_client:, prometheus_updater:)
           ##       so feed in the entire value!
           update_duration_ms = update_duration * 1000
           statsd_client.timing('cc.deployments.update.duration', update_duration_ms)
-          prometheus_updater.report_deployment_duration(update_duration_ms)
```
svkrieger, line 57: This is being emitted on the scheduler VM. Therefore it can't be collected because there is no web server running.

lib/cloud_controller/metrics/periodic_updater.rb
```diff
@@ -45,9 +46,9 @@ def catch_error
     def update_task_stats
       running_tasks = VCAP::CloudController::TaskModel.where(state: VCAP::CloudController::TaskModel::RUNNING_STATE)
       running_task_count = running_tasks.count
-      running_task_memory = running_tasks.sum(:memory_in_mb)
-      running_task_memory = 0 if running_task_memory.nil?
-      @updaters.each { |u| u.update_task_stats(running_task_count, running_task_memory) }
+      running_task_memory = running_tasks.sum(:memory_in_mb) || 0
+      @statsd_updater.update_task_stats(running_task_count, running_task_memory)
+      @prometheus_updater.update_task_stats(running_task_count, running_task_memory * 1000 * 1000)
```
svkrieger, line 51: Prometheus best practice is to [emit in base units](https://prometheus.io/docs/practices/naming/#base-units). Therefore, emitted in bytes and not MB.

lib/cloud_controller/metrics/periodic_updater.rb
```diff
@@ -120,7 +123,12 @@ def update_vitals
         num_cores: VCAP::HostSystem.new.num_cores
       }
 
-      @updaters.each { |u| u.update_vitals(vitals) }
+      @statsd_updater.update_vitals(vitals)
+
+      prom_vitals = vitals.clone
+      prom_vitals.delete(:uptime)
+      prom_vitals[:started_at] = @start_time.to_i
+      @prometheus_updater.update_vitals(prom_vitals)
```
svkrieger, line 131: Prometheus best practice is to [emit the timestamp when something happened and not the time since something happened](https://prometheus.io/docs/practices/instrumentation/#timestamps-not-time-since). The uptime can then be calculated by the monitoring system.

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -4,11 +4,30 @@ module VCAP::CloudController::Metrics
   class PrometheusUpdater
     def initialize(registry=Prometheus::Client.registry)
       @registry = registry
+
+      # Register all metrics, to initialize them for discoverability
```
svkrieger, line 8: Prometheus best practice is to [initialise all time series to prevent missing metrics](https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics).

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -62,70 +78,46 @@ def update_thread_info(thread_info)
       update_gauge_metric(:cc_thread_info_event_machine_resultqueue_num_waiting, thread_info[:event_machine][:resultqueue][:num_waiting], 'EventMachine requests waiting in queue')
     end
 
-    def update_failed_job_count(failed_jobs_by_queue, total)
+    def update_failed_job_count(failed_jobs_by_queue)
       failed_jobs_by_queue.each do |key, value|
-        metric_key = :"cc_failed_job_count_#{key.to_s.underscore}"
-        update_gauge_metric(metric_key, value, "Failed jobs for worker #{key}")
+        update_gauge_metric(:cc_failed_jobs_total, value, "Failed jobs for worker #{key}", labels: { queue: key.to_s.underscore })
```
svkrieger, line 83: Using labels for the different queues enables us to display each queue and also show the total number of failed jobs in the monitoring system, using a single metric. (https://prometheus.io/docs/practices/instrumentation/#use-labels)

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -62,70 +78,46 @@ def update_thread_info(thread_info)
...
+      update_gauge_metric(:cc_running_tasks_memory_bytes, total_memory_in_bytes, 'Total memory consumed by running tasks')
     end
 
     def start_staging_request_received
-      increment_counter_metric(:cc_staging_requested, 'Number of staging requests')
+      increment_counter_metric(:cc_staging_requested_total, 'Number of staging requests')
     end
 
     def report_staging_success_metrics(duration_ns)
-      increment_counter_metric(:cc_staging_succeeded, 'Number of successful staging events')
```
svkrieger, line 102: Histogram metrics also emit one time series for `_count` and a bucket `+Inf`, which both show the total number. There is no need for another counter metric.

spec/unit/lib/cloud_controller/metrics/prometheus_updater_spec.rb
```diff
@@ -1,46 +1,12 @@
...
-      @registry = Prometheus::Client::Registry.new
-    end
-  end
-end
-
 module VCAP::CloudController::Metrics
   RSpec.describe PrometheusUpdater do
     let(:updater) { PrometheusUpdater.new(prom_client) }
-    let(:prom_client) { Prometheus::Client.registry }
+    let(:prom_client) { Prometheus::Client::Registry.new }
```
svkrieger, line 7: Am I missing something, or does this do the same job as the monkey patch, I deleted above?

lib/cloud_controller/dependency_locator.rb
```diff
@@ -59,6 +59,18 @@ def runners
...
     end
 
+    def periodic_updater
+      @dependencies[:periodic_updater] ||
+        register(:periodic_updater,
+                 VCAP::CloudController::Metrics::PeriodicUpdater.new(
+                   Time.now.utc,
+                   Steno::Sink::Counter.new,
+                   Steno.logger('cc.api'),
+                   VCAP::CloudController::Metrics::StatsdUpdater.new(statsd_client),
```
johha, line 69: Should we also have a singleton `StatsdUpdater` like  `prometheus_updater`?

svkrieger: Maybe. But I would not add this within this PR as it is quite big already and not required.

spec/unit/lib/cloud_controller/metrics/prometheus_updater_spec.rb
```diff
@@ -206,144 +160,64 @@ module VCAP::CloudController::Metrics
...
         # expected buckets for duration, in millis : 10000, 15000, 20000, 25000, 30000
-        expect(metric.get).to eq({ '10000.0' => 0, '15000.0' => 0, '20000.0' => 1, '25000.0' => 1, '30000.0' => 1, 'sum' => 20_000, '+Inf' => 1 })
+        expect(metric.get).to eq({ '+Inf' => 1.0, '10' => 0.0, '30' => 1.0, '300' => 1.0, '5' => 0.0, '60' => 1.0, '600' => 1.0, '890' => 1.0, 'sum' => 20.0 })
       end
     end
 
     describe '#report_staging_failure_metrics' do
       it 'emits staging failure metrics' do
-        duration_ns = 20 * 1e9
+        # 20 seconds
```
philippthun, line 213: 900 seconds...

lib/cloud_controller/metrics/prometheus_updater.rb
svkrieger: So far I only initialised the metrics, which are not collected by the PeriodicUpdater, because those will be initialised on startup. I don't necessarily like how this is solved currently and might give it some more thoughts.

lib/cloud_controller/runner.rb
```diff
@@ -167,10 +167,10 @@ def periodic_updater
         Time.now.utc,
         @log_counter,
         Steno.logger('cc.api'),
-        [
-          VCAP::CloudController::Metrics::StatsdUpdater.new(statsd_client)
-        ]
+        VCAP::CloudController::Metrics::StatsdUpdater.new(statsd_client),
+        prometheus_updater
       )
+      CloudController::DependencyLocator.instance.register(:periodic_updater, @periodic_updater)
```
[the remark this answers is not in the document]
svkrieger: I did that because the `statsd_client`, passed to the constructor of the `PeriodicUpdater` has another configuration than the one in `DependencyLocator`. And also other instances like `log_counter` are passed. Using parameters for  `DependencyLocator.instance.periodic_updater` to configure the `PeriodicUpdater` feels not right to me. So I thought it's okay to initialise it here and then register it with the `DependencyLocator` for the rest of the code. I think it would make sense to look at all those metric/log clients, which seem to get initialised all over the code and check whether it would make sense to store a singleton in the `DependencyLocator` for all of those. But that's probably then another PR.

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -4,11 +4,30 @@ module VCAP::CloudController::Metrics
   class PrometheusUpdater
     def initialize(registry=Prometheus::Client.registry)
       @registry = registry
+
+      # Register all metrics, to initialize them for discoverability
+      @registry.gauge(:cc_job_queues_length_total, docstring: 'Job queues length of worker processes', labels: [:queue]) unless @registry.exist?(:cc_job_queues_length_total)
+      @registry.gauge(:cc_failed_jobs_total, docstring: 'Number of failed jobs of worker processes', labels: [:queue]) unless @registry.exist?(:cc_failed_jobs_total)
```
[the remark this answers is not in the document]
svkrieger: Good question. Currently we read the failed jobs from the DB. As old jobs get cleaned up by a background job, a lower number will be reported. I think this also makes more sense than keeping an ever increasing counter, which will loose quickly any meaning, except people track the rate with which the counter increases.
Another idea would be to keep the number of failed jobs in memory, so that when a worker gets restarted it starts at 0 again. (then we need a prom endpoint on worker VMs though).

lib/cloud_controller/metrics/periodic_updater.rb
```diff
@@ -84,13 +85,14 @@ def update_job_queue_length
...
-      @updaters.each { |u| u.update_job_queue_length(pending_job_count_by_queue, total) }
+      @statsd_updater.update_job_queue_length(pending_job_count_by_queue, total)
+      @prometheus_updater.update_job_queue_length(pending_job_count_by_queue)
     end
 
     def update_thread_info
       local_thread_info = thread_info
 
-      @updaters.each { |u| u.update_thread_info(local_thread_info) }
+      [@statsd_updater, @prometheus_updater].each { |u| u.update_thread_info(local_thread_info) }
```
philippthun, line 95: I'm not sure if we should add this "thread info" thing to the prometheus world. This is very Thin/EventMachine specific and our goal should be to make the `prometheus_updater` compatible with Puma right from the start.

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -62,70 +78,46 @@ def update_thread_info(thread_info)
...
       end
-
-      update_gauge_metric(:cc_failed_job_count_total, total, 'Total failed jobs')
     end
 
     def update_vitals(vitals)
       vitals.each do |key, value|
+        next if key.to_s.underscore == 'cpu'
+
         metric_key = :"cc_vitals_#{key.to_s.underscore}"
```
philippthun, line 91: I'm wondering how these metrics would look like for Puma with multiple workers (processes)... If we want to emit memory metrics on our own, the `periodic_updater` needs to run for every process. For the other metrics, that are fetched from the database, we only need one `periodic_updater` for an entire VM.

philippthun: We could try to keep the memory metric as a sum of all Puma subprocesses...

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -4,11 +4,30 @@ module VCAP::CloudController::Metrics
...
+                            docstring: 'Durations of successful staging events',
+                            buckets: duration_buckets)
+      end
+      unless @registry.exist?(:cc_staging_failed_duration_seconds)
+        @registry.histogram(:cc_staging_failed_duration_seconds,
+                            docstring: 'Durations of failed staging events',
+                            buckets: duration_buckets)
+      end
+
+      @registry.gauge(:cc_requests_outstanding_gauge, docstring: 'Requests Outstanding Gauge') unless @registry.exist?(:cc_requests_outstanding_gauge)
```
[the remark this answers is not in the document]
svkrieger: Good catch. Seems like I missed to rename some of the metrics and make them "prometheus conform".

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -62,70 +78,46 @@ def update_thread_info(thread_info)
...
-        update_gauge_metric(metric_key, value, "Failed jobs for worker #{key}")
+        update_gauge_metric(:cc_failed_jobs_total, value, "Failed jobs for worker #{key}", labels: { queue: key.to_s.underscore })
       end
-
-      update_gauge_metric(:cc_failed_job_count_total, total, 'Total failed jobs')
     end
 
     def update_vitals(vitals)
       vitals.each do |key, value|
+        next if key.to_s.underscore == 'cpu'
```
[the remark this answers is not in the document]
svkrieger: Totally!

spec/unit/lib/cloud_controller/metrics/prometheus_updater_spec.rb
svkrieger: oops

lib/cloud_controller/metrics/prometheus_updater.rb
svkrieger: The code changes in this PR do not consider Puma at all. But that's definitively something we have to think about next.

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -37,20 +56,17 @@ def update_summary_metric(metric, value, message)
...
-      update_gauge_metric(:cc_total_users, user_count, 'Number of users')
+      update_gauge_metric(:cc_users_total, user_count, 'Number of users')
     end
 
-    def update_job_queue_length(pending_job_count_by_queue, total)
+    def update_job_queue_length(pending_job_count_by_queue)
       pending_job_count_by_queue.each do |key, value|
-        metric_key = :"cc_job_queue_length_#{key.to_s.underscore}"
-        update_gauge_metric(metric_key, value, docstring: "Job queue length for worker #{key}")
+        update_gauge_metric(:cc_job_queues_length_total, value, "Job queue length for worker #{key}", labels: { queue: key.to_s.underscore })
```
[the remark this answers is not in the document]
svkrieger: No it doesn't. I also don't like that we have to pass the docstring all the time when emitting metrics. I refactored that so that all metrics get registered on creation of the PrometheusUpdater.

lib/cloud_controller/metrics/prometheus_updater.rb
svkrieger: Now all metrics are initialised on creation of the PrometheusUpdater object.

app/controllers/internal/metrics_controller.rb
philippthun: I think I would simply remove this `update!` from here and return the values that are already periodically fetched.

lib/cloud_controller/metrics/periodic_updater.rb
philippthun: Did you find any arguments pro/contra 1000/1024?

svkrieger: Not really. As you said from an academic point of view 1000 is correct, but if 1024 is used in the CF CLI, I guess we should use that for consistency.

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -4,128 +4,139 @@ module VCAP::CloudController::Metrics
...
-    def report_diego_cell_sync_duration(duration_ms)
-      update_summary_metric(:cc_diego_sync_duration, duration_ms, 'Diego cell sync duration')
-      update_gauge_metric(:cc_diego_sync_duration_gauge, duration_ms, 'Diego cell sync duration (gauge metric)')
-    end
+    private
 
-    def report_deployment_duration(duration_ms)
-      update_summary_metric(:cc_deployments_update_duration, duration_ms, 'Deployment duration')
-      update_gauge_metric(:cc_deployments_update_duration_gauge, duration_ms, 'Deployment duration (gauge metric)')
+    def metrics
```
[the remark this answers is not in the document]
svkrieger: Good point. Done!

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -4,128 +4,139 @@ module VCAP::CloudController::Metrics
...
+        { type: :gauge, name: :cc_users_total, docstring: 'Number of users' },
+        { type: :gauge, name: :cc_deployments_in_progress_total, docstring: 'Number of in progress deployments' }
+      ]
     end
 
-    private
-
     def duration_buckets
-      Prometheus::Client::Histogram.linear_buckets(start: 10_000, width: 5000, count: 5)
+      [5, 10, 30, 60, 300, 600, 890]
```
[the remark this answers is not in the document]
svkrieger: Done!

app/controllers/internal/metrics_controller.rb
svkrieger: While working on visualising the new metrics I realised that we loose precision compared to the statsd stack due to the pull mechanism. I think we should leave the `update!` so people can choose the level of precision through changing the pull interval of the prom scraper.

lib/cloud_controller/runner.rb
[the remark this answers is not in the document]
svkrieger: Good point, I adjusted it accordingly.

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -4,128 +4,139 @@ module VCAP::CloudController::Metrics
...
+    private
 
-    def report_deployment_duration(duration_ms)
-      update_summary_metric(:cc_deployments_update_duration, duration_ms, 'Deployment duration')
-      update_gauge_metric(:cc_deployments_update_duration_gauge, duration_ms, 'Deployment duration (gauge metric)')
+    def metrics
+      [
+        { type: :gauge, name: :cc_job_queues_length_total, docstring: 'Job queues length of worker processes', labels: [:queue] },
+        { type: :gauge, name: :cc_failed_jobs_total, docstring: 'Number of failed jobs of worker processes', labels: [:queue] },
+        { type: :counter, name: :cc_staging_requested_total, docstring: 'Number of staging requests' },
```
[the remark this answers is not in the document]
svkrieger: `cc_staging_requests_total` sounds more smooth!

PR #3450, opened 2023-09-28
Require cloud_controller.read access(or equivalent) to access list endpoints.

Benjamintf1:

Require cloud_controller.read access(or equivalent) to access list endpoints.

### DIFF
--- a/app/access/app_event_access.rb
+++ b/app/access/app_event_access.rb
@@ -31,8 +31,7 @@ module VCAP::CloudController
     end
 
     def index?(_object_class, _params=nil)
-      # This can return true because the index endpoints filter objects based on user visibilities
-      true
+      admin_user? || admin_read_only_user? || has_read_scope? || global_auditor?
     end
 
     # These methods should be called first to determine if the user's token has the appropriate scope for the operation
@@ -66,8 +65,7 @@ module VCAP::CloudController
     end
 
     def index_with_token?(_)
-      # This can return true because the index endpoints filter objects based on user visibilities
-      true
+      admin_user? || admin_read_only_user? || has_read_scope? || global_auditor?
     end
   end
 end
--- a/app/access/service_access.rb
+++ b/app/access/service_access.rb
@@ -27,7 +27,7 @@ module VCAP::CloudController
     end
 
     def index?(_object_class, _params=nil)
-      admin_user? || admin_read_only_user? || has_read_scope? || global_auditor?
+      admin_user? || admin_read_only_user? || has_read_scope? || global_auditor? || !VCAP::CloudController::FeatureFlag.enabled?(:hide_marketplace_from_unauthenticated_users)
     end
 
     # These methods should be called first to determine if the user's token has the appropriate scope for the operation
@@ -61,7 +61,7 @@ module VCAP::CloudController
     end
 
     def index_with_token?(_)
-      admin_user? || admin_read_only_user? || has_read_scope? || global_auditor?
+      admin_user? || admin_read_only_user? || has_read_scope? || global_auditor? || !VCAP::CloudController::FeatureFlag.enabled?(:hide_marketplace_from_unauthenticated_users)
     end
 
     def delete?(service, _=nil)
--- a/app/access/service_access.rb
+++ b/app/access/service_access.rb
@@ -27,8 +27,7 @@ module VCAP::CloudController
     end
 
     def index?(_object_class, _params=nil)
-      # This can return true because the index endpoints filter objects based on user visibilities
-      true
+      admin_user? || admin_read_only_user? || has_read_scope? || global_auditor?
     end
 
     # These methods should be called first to determine if the user's token has the appropriate scope for the operation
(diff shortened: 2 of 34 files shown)


Benjamintf1:
Two things to note:

1) users with write permissions but not read permissions can no longer read as well in this pr. 

2) unauthenticated users can access marketplace without permissions if the toggle for the marketplace visibility is turned off, unlike other endpoints. 
- Now it seems to me that this is possibly an oversight brought by interaction between the list endpoint logic and the filtering logic. We have one layer of filters that effects permissions that looks at tokens and one that looks at users. The token logic isn't implemented because we assume that any filtering will happen at the user logic level, and the user logic filtering level doesn't actually effect "public" marketplace values. This means that unauthenticated users can access marketplace values. - Now If I had to guess, we "papered over" this problem with the "allow unauthenticated users" toggle. Given it exists, it seems resonable to honor that contract, but at some point it seems like we should possibly eol that toggle and just make sure that only cf users can access "public"(aka, for all orgs and spaces) information always.

Gerg:
Looking into this sent me down a rabbit hole about the `*_with_token` methods on the access classes.

In summary, they were first introduced by https://github.com/cloudfoundry/cloud_controller_ng/pull/221 for [this story](https://www.pivotaltracker.com/n/projects/969486/stories/68037500).

It appears the main difference between the two methods is what error is returned: `InsufficientScope` vs `NotAuthorized`.

app/access/app_event_access.rb
```diff
@@ -31,8 +31,7 @@ def delete?(_object)
     end
 
     def index?(_object_class, _params=nil)
```
Gerg, line 33: In general, is it necessary to fill in this logic in both `index?` and `index_with_token?` for all these access classes? From looking at https://github.com/cloudfoundry/cloud_controller_ng/blob/7b1848e6b3d33ecc486adb55fbbe35fe44fb2cb3/app/controllers/base/model_controller.rb#L239-L251 the `index_with_token?` is called first and the request will never reach `index?`. Also the `index_with_token?` error seems more appropriate for this case (though, we aren't very consistent with which access method we use, in general).

app/access/service_access.rb
```diff
@@ -27,8 +27,7 @@ def update?(_object, _params=nil)
     end
 
     def index?(_object_class, _params=nil)
-      # This can return true because the index endpoints filter objects based on user visibilities
-      true
+      admin_user? || admin_read_only_user? || has_read_scope? || global_auditor? || !VCAP::CloudController::FeatureFlag.enabled?(:hide_marketplace_from_unauthenticated_users)
```
Gerg, line 30: I'm not super excited about including business logic here, but there is some prior art for checking feature flags in the v2 access classes: https://github.com/search?q=repo%3Acloudfoundry%2Fcloud_controller_ng+path%3A%2F%5Eapp%5C%2Faccess%5C%2F%2F+FeatureFlag&type=code

If this logic needs to go here, can/should we remove the corresponding logic in the controller?https://github.com/cloudfoundry/cloud_controller_ng/blob/7b1848e6b3d33ecc486adb55fbbe35fe44fb2cb3/app/controllers/services/services_controller.rb#L28-L39

Gerg (changes requested):

app/access/app_event_access.rb
Benjamintf1: Broadly speaking, it seems to me both safer, and moreso, consistant for us to double up on both. 

It doesn't really cost anything for us to have the same thing both places, and even before the change, the same comment about how "filtering happens later" in the old code was often in both places anyways, so I just went with both.

app/access/service_access.rb
Benjamintf1: > If this logic needs to go here, can/should we remove the corresponding logic in the controller?

possibly? Do you want me to?

I'd rather there not be that code in the access controller at all tbh, but i would have done that by just, never anymore allowing unauthorized users to access the marketplace endpoint. Given we have what I'd, honestly consider a "compatibility flag" more then a "feature flag"(as I mentioned in my comment), I figured I should try to respect that flag unless we're thinking we're ok with dropping it.

app/access/app_event_access.rb
Gerg: My main concern is adding code that isn't necessary to achieve the desired behavior and may never execute.

I think the "original sin" was https://github.com/cloudfoundry/cloud_controller_ng/commit/1d6b036fbf4c5c84e092f12b7d3d62996a5a9fb8, which inlined all these methods (not disparaging; it made sense at the time). These methods made sense as base class defaults, but not necessarily included in each access class.

That said, especially since this is v2 only, it's also reasonable to just continue the current pattern, messy as it is. I'm not going to block on this.

app/access/service_access.rb
Gerg: Is it ever possible for the services controller code to execute? If the `InvalidAuthToken` and `InvalidAuthToken` cases are already handled elsewhere, then I think we should delete the vestigial code.

Benjamintf1: I'll try and look into this and remove the redundancy.

Benjamintf1: hmmm. I did "quick" checks. I'll do more tomorrow. It appears to be not possible to quickly remove.

Benjamintf1: I'm doing research, It seems like service and service_plan access effect /v2/service calls, but weirdly enough not /v2/service_plan calls?

Gerg (approved):

PR #3461, opened 2023-10-02
fix app revision pruning log to only include count of rows deleted

Benjamintf1:
- ends up becoming excessively long and original code looks like it intended to be number of rows, not exact rows

fix app rivision pruning log to only include count of rows deleted

- ends up becoming excessively long and original code looks like it intended to be number of rows, not exact rows, this should reduce overly verbose logs. 


moleske (approved):
seems good to me.  this will help folks who are having their logs fill up with the entire object rather than just a count like the message implies

PR #3462, opened 2023-10-05
Restructure runner spec

svkrieger:
Put tests in correct context

### DIFF
--- a/spec/unit/lib/cloud_controller/runner_spec.rb
+++ b/spec/unit/lib/cloud_controller/runner_spec.rb
@@ -26,11 +26,10 @@ module VCAP::CloudController
     end
 
     describe '#run!' do
-      let(:runner) { Runner.new(argv + ['-c', config_file.path]) }
-
       it 'creates a pidfile' do
         expect(VCAP::PidFile).to receive(:new).with('/tmp/cloud_controller.pid')
-        runner.run!
+
+        subject.run!
       end
 
       it 'starts the web server' do
@@ -38,7 +37,7 @@ module VCAP::CloudController
         allow(server).to receive(:start!)
         expect(VCAP::CloudController::ThinRunner).to receive(:new).and_return(server)
 
-        runner.run!
+        subject.run!
         expect(server).to have_received(:start!).once
       end
     end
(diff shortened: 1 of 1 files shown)


spec/unit/lib/cloud_controller/runner_spec.rb
```diff
@@ -120,17 +158,15 @@ module VCAP::CloudController
         allow(StenoConfigurer).to receive(:new).and_return(steno_configurer)
         allow(steno_configurer).to receive(:configure)
 
-        subject.run!
-        subject.run!
+        subject
```
philippthun, line 161: The test is called `only sets up logging once` and there were two invocations of `subject.run!`. So I guess there should be two invocations of `subject` now.

spec/unit/lib/cloud_controller/runner_spec.rb
```diff
@@ -24,9 +26,45 @@ module VCAP::CloudController
     end
 
     describe '#run!' do
+      let(:runner) { Runner.new(argv + ['-c', config_file.path]) }
+
       it 'creates a pidfile' do
         expect(VCAP::PidFile).to receive(:new).with('/tmp/cloud_controller.pid')
-        subject.run!
+        runner.run!
```
philippthun, line 33: What's the reason for defining this `runner` `let` and not using `subject` here? Should have the same result, or?

spec/unit/lib/cloud_controller/runner_spec.rb
```diff
@@ -120,17 +158,15 @@ module VCAP::CloudController
...
 
         expect(steno_configurer).to have_received(:configure).once
       end
 
       it 'sets up telemetry logging once' do
         allow(TelemetryLogger).to receive(:init)
 
-        subject.run!
-        subject.run!
+        subject
```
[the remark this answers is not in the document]
svkrieger: Good point. I thought with `let` I could pull out the initialisation of the runner out of the test context, so that if I would test for something which happens in the constructor, the test would fail. Unfortunately that does not work. I changed it back to use `subject.run!` again.

spec/unit/lib/cloud_controller/runner_spec.rb
svkrieger: Good point. I thought with let I could pull out the initialisation of the runner out of the test context, so that if I would test for something which happens in the constructor, the test would fail. Unfortunately that does not work. I changed it back to use subject.run! again.

philippthun (approved):

Issue #3304, opened 2023-05-31
Docker's registry password


krismarc:
Dear Community,

just today, I've learnt the docker image is pulled each time the app is getting restarted. Somehow I though it works like with droplets and those are internally stored within CF until the next app deployment.

It means, if the password is changed the app can't be successfully restarted anymore.

I've also learnt about your recommendation about using 2 passwords. I can imagine this while using registry protected by secrets like in Azure or so. https://docs.cloudfoundry.org/devguide/deploy-apps/push-docker.html#private-repo

However, our internal registry is AD protected and there's no option to configure secondary password for the same user.

Imagine, we have hundreds of applications deployed this way and all of them are using the same user to access the registry. If the password would be changed we are in trouble.

Is there any option to address this? A magic way to update the password on existing app without re-pushing it? I couldn't find such an option using cf cli nor v3 api. I assume it's somehow internally stored within CF because initial password is passed only once when pushing the app as a variable CF_DOCKER_PASSWORD=YOUR-PASSWORD

eg. a direct update on the database level?

Looking forward for your support. Thank you.

Best regards,
K.M.

johha:
Just a quick thought on this, have you tried updating the package? The [package object](https://v3-apidocs.cloudfoundry.org/version/3.138.0/index.html#the-package-object) has a field `data.password` for docker images. If you [update the password of the corresponding app package](https://v3-apidocs.cloudfoundry.org/version/3.138.0/index.html#update-a-package) the new password should be used during the next restart. The endpoint can be called e.g. with `cf curl`.

krismarc:
Unfortunately, seems to be not supported.
```
{
   "errors": [
      {
         "detail": "Unknown field(s): 'data'",
         "title": "CF-UnprocessableEntity",
         "code": 10008
      }
   ]
}
```
even in api documentation optional parameters are just about metadata keys

> Optional parameters
> Name 	Type 	Description
> metadata.labels 	[label object](https://v3-apidocs.cloudfoundry.org/version/3.138.0/index.html#labels) 	Labels applied to the package
> metadata.annotations 	[annotation object](https://v3-apidocs.cloudfoundry.org/version/3.138.0/index.html#annotations) 	Annotations applied to the package

krismarc:
I'm not Ruby expert but I assume having this supported shouldn't be a big deal if there are no 'conceptional' blockers.

[...]

PR #3467, opened 2023-10-10
v3 should allow to update docker registry credentials

kathap:
CAPI issue: https://github.com/cloudfoundry/cloud_controller_ng/issues/3304 Missing v3 feature parity.

PATCH /v3/packages/:guid should allow to update docker registry credentials.

This change adds the possibility to update the docker credentials via PATCH /v3/packages/:guid. After the PATCH the user has to restage the app in order to compile a new droplet with the new credentials.

https://github.com/cloudfoundry/cloud_controller_ng/issues/3304
https://github.com/cloudfoundry/docs-dev-guide/pull/494

### DIFF
--- a/app/actions/package_update.rb
+++ b/app/actions/package_update.rb
@@ -9,9 +9,14 @@ module VCAP::CloudController
 
     def update(package, message)
       package.db.transaction do
+        if package.type == PackageModel::DOCKER_TYPE && (message.username || message.password)
+          package.docker_username = message.username unless message.username.nil?
+          package.docker_password = message.password unless message.password.nil?
+          package.save
+        end
         MetadataUpdate.update(package, message)
       end
-      @logger.info("Finished updating metadata on package #{package.guid}")
+      @logger.info("Finished updating package #{package.guid}")
       package
     rescue Sequel::ValidationFailed => e
       raise InvalidPackage.new(e.message)
--- a/app/controllers/v3/packages_controller.rb
+++ b/app/controllers/v3/packages_controller.rb
@@ -125,6 +125,7 @@ class PackagesController < ApplicationController
 
     package, space = PackageFetcher.new.fetch(hashed_params[:guid])
     package_not_found! unless package && permission_queryer.can_read_from_space?(space.id, space.organization_id)
+    unprocessable_non_docker_package_update! if package.type != PackageModel::DOCKER_TYPE && (message.username || message.password)
     unauthorized! unless permission_queryer.can_write_to_active_space?(space.id)
     suspended! unless permission_queryer.is_space_active?(space.id)
 
@@ -210,6 +211,10 @@ class PackagesController < ApplicationController
 
   def unprocessable_non_docker_package!
     unprocessable!('Cannot create bits package for a Docker app.')
+  end
+
+  def unprocessable_non_docker_package_update!
+    unprocessable!('Cannot update Docker credentials for a buildpack app.')
   end
 
   def unprocessable_app!
--- a/app/messages/package_update_message.rb
+++ b/app/messages/package_update_message.rb
@@ -2,7 +2,7 @@ require 'messages/metadata_base_message'
 
 module VCAP::CloudController
   class PackageUpdateMessage < MetadataBaseMessage
-    register_allowed_keys []
+    register_allowed_keys %i[username password]
 
     validates_with NoAdditionalKeysValidator
   end
--- a/spec/unit/actions/package_update_spec.rb
+++ b/spec/unit/actions/package_update_spec.rb
@@ -30,5 +30,63 @@ module VCAP::CloudController
         expect(package).to have_annotations({ key_name: 'tokyo', value: 'grapes' })
       end
     end
+
+    describe 'update docker credentials' do
+      let(:body) do
+        {
+          username: 'Udo',
+          password: 'It is a secret'
+        }
+      end
+      let(:package) { PackageModel.make(type: 'docker', docker_image: 'image-magick.com') }
+      let(:message) { PackageUpdateMessage.new(body) }
+
+      it "updates the package's docker credentials"  do
+        expect(message).to be_valid
+        package_update.update(package, message)
+
+        package.reload
+        expect(package.docker_username).to eq('Udo')
+        expect(package.docker_password).to eq('It is a secret')
+      end
+    end
+
+    describe 'update docker username' do
+      let(:body) do
+        {
+          username: 'Walz'
+        }
+      end
+      let(:package) { PackageModel.make(type: 'docker', docker_image: 'image-magick.com') }
+      let(:message) { PackageUpdateMessage.new(body) }
+
+      it "updates the package's docker username" do
+        expect(message).to be_valid
+        package_update.update(package, message)
+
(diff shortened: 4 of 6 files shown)


moleske (commented):
Can you update the [docs](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/docs/v3/source/includes/resources/packages/_update.md.erb) to show this new support for username and password?

Otherwise seems good.  You could consider adding a test at the controller integration level if you wanted to test the flow the whole way through (this [test file](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/request/packages_spec.rb#L1096).)  This would allow you to exercise the whole flow with real components, but some might consider it too much for this change.  I can probably be convinced either way

kathap:
> Can you update the [docs](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/docs/v3/source/includes/resources/packages/_update.md.erb) to show this new support for username and password?
> 
> Otherwise seems good. You could consider adding a test at the controller integration level if you wanted to test the flow the whole way through (this [test file](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/request/packages_spec.rb#L1096).) This would allow you to exercise the whole flow with real components, but some might consider it too much for this change. I can probably be convinced either way

Sure, documentation will be updated. Maybe I set it to draft, still working on tests and cases where only password or username is updated :) 
Maybe also this [documentation](https://docs.cloudfoundry.org/devguide/deploy-apps/push-docker.html#registry:~:text=If%20container%20registry%20credentials%20change%2C%20you%20must%20push%20the%20app%20with%20the%20new%20credentials.%20Apps%20require%20access%20to%20the%20container%20registry%20when%20starting.%20If%20you%20do%20not%20push%20the%20app%20with%20the%20new%20credentials%2C%20Cloud%20Foundry%20fails%20to%20start%20the%20app.) needs then an update..

app/actions/package_update.rb
```diff
@@ -9,12 +9,21 @@ def initialize
...
       end
       @logger.info("Finished updating metadata on package #{package.guid}")
       package
     rescue Sequel::ValidationFailed => e
       raise InvalidPackage.new(e.message)
     end
+
+    def unprocessable!(message)
+      raise CloudController::Errors::ApiError.new_from_details('UnprocessableEntity', message)
+    end
```
johha, line 27: Not needed anymore

kathap: Ah yes, true

kathap:
Tested the new PATCH, works with a restage after the PATCH, you can update username and password or just one of them.

app/actions/package_update.rb
```diff
@@ -9,6 +9,11 @@ def initialize
...
     def update(package, message)
       package.db.transaction do
+        if package.type == PackageModel::DOCKER_TYPE
+          package.docker_username = message.username unless message.username.nil?
+          package.docker_password = message.password unless message.password.nil?
+          package.save
+        end
         MetadataUpdate.update(package, message)
       end
       @logger.info("Finished updating metadata on package #{package.guid}")
```
philippthun, line 19: It's not only metadata anymore.

```suggestion
      @logger.info("Finished updating package #{package.guid}")
```

app/controllers/v3/packages_controller.rb
```diff
@@ -125,6 +125,7 @@ def update
 
     package, space = PackageFetcher.new.fetch(hashed_params[:guid])
     package_not_found! unless package && permission_queryer.can_read_from_space?(space.id, space.organization_id)
+    unprocessable_non_docker_package_update! if package.type != PackageModel::DOCKER_TYPE && (message.username || message.password)
```
philippthun, line 128: `.present?`

app/messages/package_update_message.rb
```diff
@@ -2,7 +2,8 @@
 
 module VCAP::CloudController
   class PackageUpdateMessage < MetadataBaseMessage
-    register_allowed_keys []
+    register_allowed_keys %i[username password]
+    attr_reader :package
```
philippthun, line 6: Is this needed?

spec/unit/actions/package_update_spec.rb
```diff
@@ -30,5 +30,63 @@ module VCAP::CloudController
...
+      let(:body) do
+        {
+          username: 'Udo',
+          password: 'Won\'t tell you'
+        }
+      end
+      let(:package) { PackageModel.make(type: 'docker', docker_image: 'image-magick.com') }
+      let(:message) { PackageUpdateMessage.new(body) }
+
+      it 'updates the package\'s docker credentials' do
```
philippthun, line 44: When using a `'` inside the text, it's better (more readable) to use `"` for the whole text instead of escaping:
```
it "updates the package's docker credentials" do
```

spec/unit/controllers/v3/packages_controller_spec.rb
```diff
@@ -443,6 +443,49 @@
...
+          let(:message) do
+            {
+              username: 'Walz',
+              password: 'Absolutely secret'
+            }
+          end
+
+          it 'fails' do
+            patch :update, params: { guid: package.guid }.merge(message), as: :json
+            expect(response).to have_http_status(:unprocessable_entity)
```
philippthun, line 462: What about adding a check for the error message?
```
expect(response).to have_error_message(/... the expected text .../)
```

app/actions/package_update.rb
```diff
@@ -9,6 +9,11 @@ def initialize
 
     def update(package, message)
       package.db.transaction do
+        if package.type == PackageModel::DOCKER_TYPE
+          package.docker_username = message.username unless message.username.nil?
```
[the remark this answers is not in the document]
kathap: I decided for .nil? because you can push a docker app like this:
`CF_DOCKER_PASSWORD='' cf push my-test-app -o cloudfoundry/test-app --docker-username ''`
so I thought I be consistent and allow empty string. But if I change it to .present? and you do a package update with empty string as username or password, the package doesn't get updated. What would be the best solution?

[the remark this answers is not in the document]
kathap: Ok, if you update an app with an empty name, you get an error `{"errors":[{"detail":"name presence, name format","title":"CF-UnprocessableEntity","code":10008}]}`. And if you try to update with empty lifecycle or type then there are also errors thrown `{"errors":[{"detail":"Lifecycle data must be an object","title":"CF-UnprocessableEntity","code":10008}]}`,
`{"errors":[{"detail":"Lifecycle type is not included in the list: buildpack, docker","title":"CF-UnprocessableEntity","code":10008}]}`. Maybe this would also be an option if we decide for not allowing empty strings to improve the error message.

app/messages/package_update_message.rb
philippthun: I meant the second line (`attr_reader :package`). I don't see `package` being used in this file.

app/actions/package_update.rb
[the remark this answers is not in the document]
philippthun: Okay, if you can create a package with empty strings for username and password, that should also be supported for the update scenario. So `nil?` seems to be the correct check.

PR #3474, opened 2023-10-11
Refactoring UAA client

philippthun:
While working on the "Support UAA identity zones in CC" PoC [1], I did some refactoring that I would like to keep although the PoC is not going to get implemented (see [2]).

[1] https://github.com/cloudfoundry/cloud_controller_ng/pull/3341
[2] https://github.com/cloudfoundry/uaa/issues/2505

### DIFF
--- a/app/actions/role_create.rb
+++ b/app/actions/role_create.rb
@@ -14,8 +14,7 @@ module VCAP::CloudController
     def create_space_role(type:, user:, space:)
       error!("Users cannot be assigned roles in a space if they do not have a role in that space's organization.") unless space.in_organization?(user)
 
-      uaa_client = CloudController::DependencyLocator.instance.uaa_client
-      UsernamePopulator.new(uaa_client).transform(user)
+      UsernamePopulator.new(uaa_username_lookup_client).transform(user)
 
       case type
       when RoleTypes::SPACE_AUDITOR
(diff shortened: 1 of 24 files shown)


kathap:
I can see use of `uaa_client` in https://github.com/cloudfoundry/cloud_controller_ng/blob/b2464ff47735bd4f376cf4fe3969d38b3c13d1a3/spec/unit/collection_transformers/username_populator_spec.rb#L6 
and https://github.com/cloudfoundry/cloud_controller_ng/blob/b2464ff47735bd4f376cf4fe3969d38b3c13d1a3/spec/unit/collection_transformers/usernames_and_roles_populator_spec.rb#L6
maybe there it should also be `uaa_username_lookup_client`

philippthun:
> I can see use of `uaa_client` in
> 
> https://github.com/cloudfoundry/cloud_controller_ng/blob/b2464ff47735bd4f376cf4fe3969d38b3c13d1a3/spec/unit/collection_transformers/username_populator_spec.rb#L6
> 
> 
> and
> https://github.com/cloudfoundry/cloud_controller_ng/blob/b2464ff47735bd4f376cf4fe3969d38b3c13d1a3/spec/unit/collection_transformers/usernames_and_roles_populator_spec.rb#L6
> 
> 
> maybe there it should also be `uaa_username_lookup_client`

I've changed the name to `uaa_username_lookup_client` in the `DependencyLocator` to incorporate the use case this UAA client is used for (and the permissions it has). And then I've adapted all places where the client is fetched from the `DependencyLocator`. I did not change other places where it is just a variable holding an instance of class `UaaClient`. In the end this particular `uaa_username_lookup_client` would also be used inside `UsernamePopulator` and `UsernamesAndRolesPopulator`, but there the longer name would make reading the code more complicated without adding much benefit (i.e. there are not multiple UAA clients involved). I've looked again through the code and would rather keep `uaa_client` in all the tests where it is just a mock object.

moleske (approved):
looks good to me.  had same question that @kathap but your answer makes sense to me

PR #3475, opened 2023-10-11
Exclude the newly added Readme.md file when checking applied migrations

philippthun:
**kiki-back-compat-migrations** complains with this message [1]:
```
Expected latest migration 20230822173000_add_migration_views_for_annotations.rb to equal Readme.md
```

[1] https://ci.capi.land/teams/main/pipelines/capi/jobs/kiki-back-compat-migrations/builds/1279

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -129,7 +129,7 @@ namespace :db do
     db = VCAP::CloudController::DB.connect(RakeConfig.config.get(:db), db_logger)
 
     latest_migration_in_db = db[:schema_migrations].order(Sequel.desc(:filename)).first[:filename]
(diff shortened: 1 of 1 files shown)


moleske (approved):
Was wondering what was going on in kiki (I hadn't actually clicked the job yet)

moleske:
this fix is correct, but the [pipeline uses](https://github.com/cloudfoundry/capi-ci/blob/main/ci/migrations/apply_latest_migrations.sh#L103) what is already deployed on the bosh api vm, which is the latest cf deployment.  This means it does not use the local version of cloud_controller_ng when running the task `ensure_migrations_are_current` so won't contain this filter.  Not quite sure how to get out of this, the easy solution is to move the readme somewhere else, commit and let pipeline go, then move the readme back

PR #3483, opened 2023-10-19
Fix issue with rake spec creating 'cc_test_' database instead of 'cc_…

geofffranks:
…test'

### DIFF
--- a/spec/support/bootstrap/db_config.rb
+++ b/spec/support/bootstrap/db_config.rb
@@ -63,10 +63,10 @@ class DbConfig
   end
 
   def default_name
-    unless ENV['TEST_ENV_NUMBER'].nil? or ENV['TEST_ENV_NUMBER'].empty?
-      "cc_test_#{ENV['TEST_ENV_NUMBER']}"
-    else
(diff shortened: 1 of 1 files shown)


spec/support/bootstrap/db_config.rb
```diff
@@ -63,10 +63,10 @@ def default_connection_prefix(db_type)
   end
 
   def default_name
-    if ENV['TEST_ENV_NUMBER']
-      "cc_test_#{ENV['TEST_ENV_NUMBER']}"
-    else
+    if ENV['TEST_ENV_NUMBER'].nil? || ENV['TEST_ENV_NUMBER'].empty?
```
philippthun, line 66: I think the most concise way to write the check is `ENV['TEST_ENV_NUMBER'].presence` (from Rails' Active Support). If the key does not exist, this returns `nil`, if the key exists, but contains an empty string, also `nil` is returned, otherwise the string is returned.

geofffranks: Updated to use this instead

philippthun (approved):

Issue #2193, opened 2021-04-13
numbers with commas in space manifests get parsed as an integer, not a string


sethboyles:
## Issue

When pushing an application with the following manifest file: 

```
applications:
- name: dora

env:
INTEGER: 1234,567,89
```

When they run cf env the variable INTEGER is not handled correctly and is returned as follows: 

```
INTEGER: 123456789
```

The commas are removed and the field is parsed as an integer instead of a string.  This is different than CF CLI v6's behavior and contrary to the YAML spec definition of an integer (and float): https://yaml.org/spec/1.2/spec.html#id2803828

## Context

This is caused by an issue with Ruby's included YAML parsing library, Psych.  There is an [old issue from 2016](https://github.com/ruby/psych/issues/273) where they note it is not fixed due to backwards compatibility concerns.  

## Expected result

`INTEGER: 1234,567,89` should be parsed as a string (EDIT: actually maybe it should be an array?)

## Current result

`INTEGER: 1234,567,89` is parsed as an integer `1234,567,89`

## Possible Fix

We could make a PR into https://github.com/ruby/psych, or we could switch libraries (if there even are others to choose from).  

We could try pre-parsing the YAML before we pass it to Psych, however I would expect that to be difficult and fraught with other issues.

- @sethboyles & @MarcPaquette

thelangley:
hey it looks like a potential fix was merged into psych. https://github.com/ruby/psych/issues/273

do we need an official release of psych to be cut before testing this new strict_integer functionality?

PR #2807, opened 2022-05-17
Use newer Psych rather than ruby's built in YAML

xandroc:
- Use 4.0.4 psych gem

Co-authored-by: Alex Rocha <alexr1@vmware.com>
Co-authored-by: Michael Oleske <moleske@pivotal.io>
Co-authored-by: David Alvarado <alvaradoda@vmware.com>

Thanks for contributing to cloud_controller_ng. To speed up the process of reviewing your pull request please provide us with:

* A short explanation of the proposed change: 
  Use latest psych gem

* An explanation of the use cases your change solves
  Resolves [this](https://github.com/cloudfoundry/cloud_controller_ng/issues/2193) github issue which

* Links to any other associated PRs

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng#cf-acceptance-tests-cats)

PR #3486, opened 2023-10-20
Remove unneeded psych gem

moleske:
- in https://github.com/cloudfoundry/cloud_controller_ng/pull/2807 we manually set psych so we could resolve https://github.com/cloudfoundry/cloud_controller_ng/issues/2193
- the newer psych is available as a default gem in [ruby 3.2](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/)

We could keep psych in the gemfile if we feel value in staying on the latest version psych instead of what version of psych ruby provides.  I don't have strong feelings either way (well strong enough to make a pr, not strong enough to advocate deeply)

The key thing we needed to resolve the issue was `strict_integer: true` in all the parse commands, nothing else from psych we cared all that much about


philippthun (commented):
I also favor removing this, as long as we don't need a specific (newer) version.

philippthun (approved):

PR #3490, opened 2023-10-24
Implement additional audit events

johha:
Some API endpoints which return credentials do not write audit events. This change adds the following audit events:
- `audit.service_instance.show` & `audit.user_provided_service_instance.show` 
   Endpoints:
   - GET /v2/service_instances/:guid
   - GET /v3/service_instances/:guid/credentials

- `audit.service_binding.show`
  Endpoints:
  - GET /v3/service_credential_bindings/:guid/details
  - GET /v2/service_bindings/:guid

- `audit.service_key.show`
  Endpoints:
  - GET /v3/service_credential_bindings/:guid/details
  - GET /v2/service_keys/:guid

[Docs](https://docs.cloudfoundry.org/running/managing-cf/audit-events.html#types) will be updated in separate PR once this is merged

### DIFF
--- a/app/controllers/services/service_bindings_controller.rb
+++ b/app/controllers/services/service_bindings_controller.rb
@@ -22,11 +22,16 @@ module VCAP::CloudController
     get path_guid, :read
 
     def read(guid)
-      obj = find_guid(guid)
-      raise CloudController::Errors::ApiError.new_from_details('ServiceBindingNotFound', guid) if obj.v2_app.blank?
+      service_binding = find_guid(guid)
+      raise CloudController::Errors::ApiError.new_from_details('ServiceBindingNotFound', guid) if service_binding.v2_app.blank?
 
-      validate_access(:read, obj)
-      object_renderer.render_json(self.class, obj, @opts)
+      validate_access(:read, service_binding)
+
+      Repositories::ServiceGenericBindingEventRepository.new(
+        Repositories::ServiceGenericBindingEventRepository::SERVICE_APP_CREDENTIAL_BINDING
+      ).record_show(service_binding, UserAuditInfo.from_context(SecurityContext))
+
+      object_renderer.render_json(self.class, service_binding, @opts)
     end
 
     post path, :create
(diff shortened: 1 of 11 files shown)


app/controllers/services/service_bindings_controller.rb
```diff
@@ -22,11 +22,16 @@ class ServiceBindingsController < RestController::ModelController
     get path_guid, :read
 
     def read(guid)
-      obj = find_guid(guid)
-      raise CloudController::Errors::ApiError.new_from_details('ServiceBindingNotFound', guid) if obj.v2_app.blank?
+      service_binding = find_guid(guid)
```
moleske, line 25: Thanks for renaming variables to be more meaningful!

app/controllers/services/service_bindings_controller.rb
```diff
@@ -22,11 +22,16 @@ class ServiceBindingsController < RestController::ModelController
...
+      raise CloudController::Errors::ApiError.new_from_details('ServiceBindingNotFound', guid) if service_binding.v2_app.blank?
```
moleske, line 26: Little surprised that changes were made to v2 api since we've generally avoided making changes to it.  That said changes seem fine

app/controllers/v3/service_instances_controller.rb
```diff
@@ -199,6 +199,8 @@ def credentials
     service_instance_not_found! unless service_instance && can_read_service_instance?(service_instance)
     unauthorized! unless permission_queryer.can_read_secrets_in_space?(service_instance.space_id, service_instance.space.organization_id)
 
+    service_event_repository.record_user_provided_service_instance_event(:show, service_instance)
```
moleske, line 202: It took us a moment to figure out why there was only a `audit.user_provided_service_instance.show` event here and not also a `audit.service_instance.show`.  That is due to the way we have set up the endpoint such that there is only user provided service instance, so this seems good as well

moleske (approved):
@MerricdeLauney and I took a look and this looks good to us!

app/controllers/services/service_bindings_controller.rb
johha: We still have some users using v2 and thus it makes sense to do this in v2 as well

PR #3500, opened 2023-11-01
Enable CF API to present `routable` field for app processes

geofffranks:
This enables CF API clients to provide feedback to developers on whether apps are currently passing readiness checks and are routable.

https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0020-readiness-healthchecks.md#cli-changes

### DIFF
--- a/app/presenters/v3/process_stats_presenter.rb
+++ b/app/presenters/v3/process_stats_presenter.rb
@@ -35,6 +35,7 @@ module VCAP::CloudController
             type: @type,
             index: index,
             state: stats[:state],
+            routable: stats[:routable],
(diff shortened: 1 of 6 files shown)


geofffranks:
As is, this PR may cause confusion for clients if this code is used in conjunction with a diego release that doesn't support readiness checks. If diego doesn't provide `routable`, right now the endpoint will return `false`. So valid + routable processes would get flagged with `routable: false` on the CF API, simply because diego didn't know about app readiness checks.

geofffranks:
Ok, pushed up a commit that's more compatible with environments where CAPI supports readiness but diego doesn't.

PR #3509, opened 2023-11-14
Add missing indexes for foreign keys on postgres

johha:
Mysql automatically creates indexes for foreign keys. This makes the execution of joins, eager loading etc. faster.

Conversely, in Postgres indexes for for foreign keys aren't automatically created. This change adds the missing indexes for Postgres.

The migration has been created with the help of the following psql function which shows the foreign keys with missing indexes:

```
--
-- function:    missing_fk_indexes2
-- purpose:     List all foreing keys in the database without and index in the referencing table.
--              The listing contains create index sentences
-- author:      Based on the work of Laurenz Albe
-- see:         https://www.cybertec-postgresql.com/en/index-your-foreign-key/
--
create or replace function missing_fk_indexes2 ()
returns setof varchar
language sql as $$
  select
    -- create index sentence
    'create index on ' ||
    tc.conrelid::regclass ||
    '(' ||
    string_agg(ta.attname, ', ' order by tx.n) ||
    ')' as create_index

  from pg_catalog.pg_constraint tc

  -- enumerated key column numbers per foreign key
  cross join lateral unnest(tc.conkey) with ordinality as tx(attnum, n)

  -- name for each key column
  join pg_catalog.pg_attribute ta on ta.attnum = tx.attnum and ta.attrelid = tc.conrelid

  where not exists (
    -- is there ta matching index for the constraint?
    select 1 from pg_catalog.pg_index i
    where
      i.indrelid = tc.conrelid and
      -- the first index columns must be the same as the key columns, but order doesn't matter
      (i.indkey::smallint[])[0:cardinality(tc.conkey)-1] @> tc.conkey) and
      tc.contype = 'f'
    group by
      tc.conrelid,
      tc.conname,
      tc.confrelid
    order by
      pg_catalog.pg_relation_size(tc.conrelid) desc
$$;

```

Which can be executed like this:

```
select * from missing_fk_indexes2();
```

### DIFF
--- a/db/migrations/20231114130216_add_missing_fk_indexes.rb
+++ b/db/migrations/20231114130216_add_missing_fk_indexes.rb
@@ -1,0 +1,51 @@
+def migration_name(table, columns)
+  "#{table}_#{columns.join('_')}_index"
+end
+
+Sequel.migration do
+  foreign_key_indexes = [
+    { table: :organizations, columns: %i[quota_definition_id] },
+    { table: :organizations, columns: %i[guid default_isolation_segment_guid] },
+    { table: :domains, columns: %i[owning_organization_id] },
+    { table: :spaces, columns: %i[isolation_segment_guid] },
+    { table: :routes, columns: %i[domain_id] },
+    { table: :routes, columns: %i[space_id] },
+    { table: :users, columns: %i[default_space_id] },
+    { table: :organizations_users, columns: %i[user_id] },
+    { table: :organizations_managers, columns: %i[user_id] },
+    { table: :organizations_billing_managers, columns: %i[user_id] },
+    { table: :organizations_auditors, columns: %i[user_id] },
+    { table: :spaces_developers, columns: %i[user_id] },
+    { table: :spaces_managers, columns: %i[user_id] },
+    { table: :spaces_auditors, columns: %i[user_id] },
+    { table: :organizations_private_domains, columns: %i[private_domain_id] },
+    { table: :route_bindings, columns: %i[route_id] },
+    { table: :route_bindings, columns: %i[service_instance_id] },
+    { table: :organizations_isolation_segments, columns: %i[isolation_segment_guid] },
+    { table: :staging_security_groups_spaces, columns: %i[staging_space_id] },
+    { table: :service_instance_shares, columns: %i[target_space_guid] },
+    { table: :job_warnings, columns: %i[fk_jobs_id] },
+    { table: :service_broker_update_requests, columns: %i[fk_service_brokers_id] },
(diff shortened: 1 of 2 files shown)


db/migrations/20231114130216_add_missing_fk_indexes.rb
```diff
@@ -0,0 +1,61 @@
+Sequel.migration do
+  up do
+    if database_type == :postgres
+      add_index :organizations, :quota_definition_id, name: :organizations_quota_definition_id_index, options: [:concurrently]
```
FloThinksPi, line 4: The current way this creates indices does create one large transaction. With concurrently this might not be a problem since nothing is locked and on mysql there are implcit commits after every DDL statement. So i would be fine leaving it in a transaction.

One thing i would change for resilience is to just make the add_index call when no index exists yet thus making the up migration idempotent. In case of unknown errors while creating the index, the migration is not stuck in a state where neither the down nor the up part can be executed anymore and a operator needs to manipulate the db.

Either way one can look at https://github.com/cloudfoundry/cloud_controller_ng/blob/main/db/migrations/20230725110800_restructure_index_for_improved_pagination_order.rb#L553-L561 where there is one transaction per index and also the up and down parts are idempotent

db/migrations/20231114130216_add_missing_fk_indexes.rb
```diff
@@ -0,0 +1,61 @@
...
+      add_index :service_broker_update_requests, :fk_service_brokers_id, name: :service_broker_update_requests_fk_service_brokers_id_index, options: [:concurrently]
+      add_index :kpack_lifecycle_data, :app_guid, name: :kpack_lifecycle_data_app_guid_index, options: [:concurrently]
+      add_index :spaces_supporters, :user_id, name: :spaces_supporters_user_id_index, options: [:concurrently]
+      add_index :route_shares, :target_space_guid, name: :route_shares_target_space_guid_index, options: [:concurrently]
+    end
+  end
+
+  down do
+    if database_type == :postgres
+      drop_index :organizations, nil, name: :organizations_quota_definition_id_index, options: [:concurrently]
```
FloThinksPi, line 34: Also add `if_exists: true` here to make the down part idempotent

FloThinksPi (changes requested):
I think the current way this creates indices will do them in one transaction in PSQL

FloThinksPi (approved):

johha:
This change indeed broke the unit tests on postgres 10 due to a random order in the result sets. While it's unclear why this is only happening on postgres 10, a valid fix is to explicitly order the results -> https://github.com/cloudfoundry/cloud_controller_ng/pull/3518

PR #3520, opened 2023-11-21
Fix/Improve PumaRunner Setup and Configuration

philippthun:
- Use `Events:on_booted` instead of `Configuration:after_worker_fork`; the `periodic_updater` shall run in the main process only.
- Use `Configuration:on_worker_shutdown` instead of `Events:on_stopped` to log incomplete requests; this needs to run inside the worker processes.
- Don't try to write a log message from inside `Events:on_stopped`; this yields a **can't be called from trap context** error.
- Configure thread/worker shutdown timeouts.
- Default `workers` and `max_threads` to 1.
- Refactor tests.

### DIFF
--- a/lib/cloud_controller/runners/puma_runner.rb
+++ b/lib/cloud_controller/runners/puma_runner.rb
@@ -6,19 +6,7 @@ module VCAP::CloudController
   class PumaRunner
     def initialize(config, app, logger, periodic_updater, request_logs)
       @logger = logger
-      @periodic_updater = periodic_updater
-      @request_logs = request_logs
       puma_config = Puma::Configuration.new do |conf|
-        # this is actually called everytime a worker is started
-        # https://github.com/puma/puma/blob/5f3f489ee867317c47724d0fc5b1d906f1b23de6/lib/puma/dsl.rb#L607
-        # we probably want to come up with a different way to do this.  Perhaps a singleton?
-        conf.after_worker_fork do
-          Thread.new do
-            EM.run do
-              @periodic_updater.setup_updates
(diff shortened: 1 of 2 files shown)


lib/cloud_controller/runners/puma_runner.rb
```diff
@@ -6,35 +6,45 @@ module VCAP::CloudController
...
+        conf.threads(0, config.get(:puma, :max_threads) || 1)
+
+        # In theory there shouldn't be any open connections when shutting down Puma as they have either been gracefully
+        # drained or forcefully terminated (after cc.nginx_drain_timeout) by Nginx. Puma has some built-in (i.e. not
+        # changeable) timeouts as well as some configurable timeouts.
+
+        # Reduce the worker shutdown timeout to 15 seconds (default is 30).
+        conf.worker_shutdown_timeout(15)
+        # Reduce the thread shutdown timeout to 10 seconds (4 [force_shutdown_after] + 5 [SHUTDOWN_GRACE_TIME] + 1)
+        conf.force_shutdown_after(4)
```
johha, line 27: where are the remaining 6 secs?

philippthun: They are hard-coded in Puma. 5 seconds "shutdown grace time" + 1 additional second...

The [Puma docs](https://www.rubydoc.info/gems/puma/Puma%2FDSL:force_shutdown_after) are rather unspecific:

> Puma always waits a few seconds after killing a thread for it to try to finish up it’s work [...]

johha (approved):

PR #3521, opened 2023-11-21
Adapt Periodic Puma Metrics

philippthun:
- The current `thread_info` is tightly coupled to Thin/EventMachine; don't emit those metrics when running the Puma webserver; maybe we will come up with other thread-related metrics for Puma at a later point in time.
- Return the summed up memory bytes and cpu for the Puma process and its worker subprocesses.
- When using Puma, the `periodic_updater` shall run in the main process only, thus direct invocations of `update!` from a controller (executed within a worker process) must be removed.
- When running Puma, the `DirectFileStore` data store is used for prometheus metrics.
- Aggregation methods for (gauge) metrics are specified and applied when using the `DirectFileStore`.
  - For `cc_requests_outstanding_total` the values per worker are summed up.
  - For metrics that are collected by the `periodic_updater` running in the main process, the aggregation method MOST_RECENT has been specified to eliminate the unnecessary `pid` label.
- Replace `VCAP::CloudController::Metrics::StatsdUpdater.new` with `CloudController::DependencyLocator.instance.statsd_updater`.
- Replace `Statsd.new` with `CloudController::DependencyLocator.instance.statsd_client`.
- Add `statsd_host` + `statsd_port` to `ClockSchema`; don't rely on default values (see cloudfoundry/capi-release#359)

### DIFF
--- a/lib/cloud_controller/metrics/prometheus_updater.rb
+++ b/lib/cloud_controller/metrics/prometheus_updater.rb
@@ -5,23 +5,23 @@ module VCAP::CloudController::Metrics
     DURATION_BUCKETS = [5, 10, 30, 60, 300, 600, 890].freeze
 
     METRICS = [
-      { type: :gauge, name: :cc_job_queues_length_total, docstring: 'Job queues length of worker processes', labels: [:queue] },
-      { type: :gauge, name: :cc_failed_jobs_total, docstring: 'Number of failed jobs of worker processes', labels: [:queue] },
+      { type: :gauge, name: :cc_job_queues_length_total, docstring: 'Job queues length of worker processes', labels: [:queue], aggregation: :most_recent },
+      { type: :gauge, name: :cc_failed_jobs_total, docstring: 'Number of failed jobs of worker processes', labels: [:queue], aggregation: :most_recent },
       { type: :counter, name: :cc_staging_requests_total, docstring: 'Number of staging requests' },
       { type: :histogram, name: :cc_staging_succeeded_duration_seconds, docstring: 'Durations of successful staging events', buckets: DURATION_BUCKETS },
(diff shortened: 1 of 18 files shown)


lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -5,51 +5,40 @@ class PrometheusUpdater
...
-      { type: :gauge, name: :cc_job_queues_length_total, docstring: 'Job queues length of worker processes', labels: [:queue] },
-      { type: :gauge, name: :cc_failed_jobs_total, docstring: 'Number of failed jobs of worker processes', labels: [:queue] },
+      { type: :gauge, name: :cc_job_queues_length_total, docstring: 'Job queues length of worker processes', labels: [:queue], aggregation: :most_recent },
+      { type: :gauge, name: :cc_failed_jobs_total, docstring: 'Number of failed jobs of worker processes', labels: [:queue], aggregation: :most_recent },
       { type: :counter, name: :cc_staging_requests_total, docstring: 'Number of staging requests' },
       { type: :histogram, name: :cc_staging_succeeded_duration_seconds, docstring: 'Durations of successful staging events', buckets: DURATION_BUCKETS },
       { type: :histogram, name: :cc_staging_failed_duration_seconds, docstring: 'Durations of failed staging events', buckets: DURATION_BUCKETS },
-      { type: :gauge, name: :cc_requests_outstanding_total, docstring: 'Requests outstanding' },
+      { type: :gauge, name: :cc_requests_outstanding_total, docstring: 'Requests outstanding', aggregation: :sum },
```
svkrieger, line 13: Would be interesting to see how the metrics look like if we don't aggregate them. Would the different values emitted by each worker be stored in the same time series, using a label "worker", or would they simply overwrite each other and lead to wrong metrics?

svkrieger (commented):
Looks good! Just a few conceptual questions.

svkrieger (approved):

PR #3527, opened 2023-11-29
Deprecated Syslog Binding URL API removal

chombium:
Code clean up after the deprecation of the `/v4` API for the Syslog Binding URL.


spec/unit/controllers/internal/syslog_drain_urls_controller_spec.rb
```diff
@@ -12,280 +12,6 @@ module VCAP::CloudController
     let!(:binding_with_drain1) { ServiceBinding.make(syslog_drain_url: 'fish,finger', app: app_obj, service_instance: instance1) }
     let!(:binding_with_drain2) { ServiceBinding.make(syslog_drain_url: 'foobar', app: app_obj, service_instance: instance2) }
 
-    describe 'GET /internal/v4/syslog_drain_urls' do
```
mkocher, line 15: It seems like this PR removed the test coverage for the shared hostname_from_app_name method. It would be good to add the test cases to the v5 specs.

mkocher (commented):
Nice to get some old code deleted. We looked into this and think this will be a safe change to make in our environments.

mkocher:
We took another look and found one internal consumer of this internal API. We're going to work on moving it to v5 to not hold up this cleanup.

chombium:
Hi @mkocher, I've updated all the tests which were included in the v4 API to work with v5 API, so that we have all of the tests which we had before.

When I only run the whole `RSpec.describe SyslogDrainUrlsInternalController` all the tests are successful, but when I run `bundle exec rake` some of the tests fail and the PR validation fails. I will have to check what's going on :-/

chombium:
I've added few more examples for the test by mistake. Now I've cleaned up the spec and everything should be fine now.

ctlong:
@chombium we'd like to hold off on merging this PR for a little while (~1-2 months). It turns out that we have an internal agent that uses the old v4 endpoint that we need to migrate to the v5 endpoint.

Thanks for your diligence in cleaning up this cruft!

chombium:
@ctlong thanks for the clarification. I thought that as it counterpart in Loggregator agent release was merged, we can merge this as well. We're not in a hurry, take your time to do the needful.

fhambrec:
Hi @ctlong, I assume @philippthun added the `blocked` label as you were still working on patching the internal agent to the /v5 endpoint. Is this done in the meantime/can we merge the PR? @chombium I guess we should revisit this PR to resolve the conflicts that have come up in the past months.

ctlong:
Yes, I believe we're unblocked now. Please feel free to proceed with this PR, and thanks for waiting!

PR #3539, opened 2023-12-06
Fix: Polling stopped after failed service instance update last operation fetch

svkrieger:

The Cloud Controller stopped polling the last operation state of a service instance update request when the Service Broker was unavailable and a proxy returned a 502 for example. According to the OSBAPI "Responses with any other status code SHOULD be interpreted as an error or invalid response. The Platform SHOULD continue polling until the Service Broker returns a valid response or the maximum polling duration is reached."

Additionally the error handling for the polling for service instance deletion was adjusted to make it consistent with the behaviour of all other service related polling mechanisms. Now the last operation state will keep in progress when HTTPResponseErrors and the like are being raised and polling continues like before. For StandardErrors though, the last operation is set to failed and the polling stops. 

Added tests to all polling operations to make sure the behaviour is consistent.

### DIFF
--- a/app/actions/v3/service_instance_delete.rb
+++ b/app/actions/v3/service_instance_delete.rb
@@ -55,10 +55,18 @@ module VCAP::CloudController
       end
 
       def poll
-        result = client.fetch_service_instance_last_operation(
-          service_instance,
-          user_guid: service_event_repository.user_audit_info.user_guid
-        )
+        logger = Steno.logger('cc.action.service_instance_delete')
+
+        begin
+          result = client.fetch_service_instance_last_operation(
+            service_instance,
+            user_guid: service_event_repository.user_audit_info.user_guid
+          )
+        rescue HttpRequestError, HttpResponseError, Sequel::Error => e
+          logger.error("Error fetching last operation from broker for service instance #{service_instance.guid}", error: e)
+          return ContinuePolling.call(nil)
+        end
+
         case result[:last_operation][:state]
         when 'in progress'
           update_last_operation_with_description(result[:last_operation][:description])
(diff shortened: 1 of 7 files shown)


spec/unit/actions/v3/service_instance_delete_spec.rb
```diff
@@ -1,4 +1,4 @@
-require 'db_spec_helper'
+require 'spec_helper'
```
svkrieger, line 1: I couldn't run this test file locally until I switched the import to `spec_helper`.

svkrieger:
https://github.com/cloudfoundry/cloud_controller_ng/blob/ee95901a3a14a6fcd6edc683c916a5d5190333f0/app/actions/v3/service_instance_delete.rb#L75-L77

If a StandardError is raised, all other poll methods will set the last operation of the resource to failed and stop polling. Only service instance delete sets the state to failed and continues polling, which does not seem to make sense for me. Does anyone know whether this is done intentionally?

app/actions/v3/service_instance_delete.rb
```diff
@@ -55,10 +55,18 @@ def delete
...
-          user_guid: service_event_repository.user_audit_info.user_guid
-        )
+        logger = Steno.logger('cc.action.service_instance_delete')
+
+        begin
+          result = client.fetch_service_instance_last_operation(
+            service_instance,
+            user_guid: service_event_repository.user_audit_info.user_guid
+          )
+        rescue HttpRequestError, HttpResponseError, Sequel::Error => e
```
svkrieger, line 65: Before this change a service instance delete operation continued polling when an exception was thrown, because below `StandardErrors` are caught and then the polling continued (even though, the operation was set to failed confusingly).
After this change, when the poll request fails, the polling will continue and the operation will remain in state `in progress`.

johha (approved):

kathap:
LGTM

kathap (approved):

PR #3546, opened 2023-12-08
Remove PSQL 10 from unittests, add PSQL 16 and MYSQL 8.2

FloThinksPi:
PSQL 10 is out of support, 16 is the newest release so we should add this.
Similar to the new psql version we should also at least test against the
newest mysql release.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -41,7 +41,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image: ["postgres:11", "postgres:16"]
+        image: ["postgres:10", "postgres:11", "postgres:16"]
     services:
       postgres:
         image: ${{ matrix.image }}
@@ -72,7 +72,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image: ["mysql:5.7", "mysql:8.0"]
+        image: ["mysql:5.7", "mysql:8.0", "mysql:8.2"]
     services:
       mysql:
         image: ${{ matrix.image }}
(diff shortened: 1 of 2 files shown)


johha:
I'd first change this in the concourse pipeline. otherwise we might get different results in the PR validation and Concourse pipeline. This would require an update of the following docker image: https://github.com/cloudfoundry/capi-dockerfiles/blob/main/capi-ruby-units/Dockerfile#L1C1-L1C1

.github/workflows/unit_tests.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image: ["postgres:10", "postgres:14"]
+        image: ["postgres:11", "postgres:16"]
```
moleske, line 44: To address @johha's concern we could append 11 and 16 to the list instead of replacing it right away.  The precedent for this is that the current concourse based pipeline still uses mysql 5.7 and not 8.

Same thing could be done with mysql 8.2 below

moleske (approved):
I'm fine merging as is cause we can always just revert if we find something not working

Do have a suggestion mentioned if we are concerned about not doing the concourse pipeline at the same time

.github/workflows/unit_tests.yml
FloThinksPi: Ok i readded psql 10 and added mysql 8.2 as new job.
This leads to 6 actions beeing executed per PR now so ~30% increas in CI cost is likely but i think we`re fine with this temporarily until we adopt the release pipelines :)

Issue #3544, opened 2023-12-07
FIPS validation: OpenSSL::Digest::MD5 cannot be used


jochenehret:
## Issue

On a FIPS enabled stemcell, the `OpenSSL::Digest::MD5` algorithm cannot be used.

## Context

We are validating CF on a FIPS enabled stemcell, see https://github.com/cloudfoundry/cf-deployment/issues/1140 for more details and a list of other issues. We are using a development capi-release which already includes the openssl v3.2.0 bump: https://github.com/cloudfoundry/cloud_controller_ng/commit/ee95901a3a14a6fcd6edc683c916a5d5190333f0

## Steps to Reproduce

If you have access to a FIPS enabled stemcell, you can `bosh ssh` into an "api" node and run this simple test:
```
require 'openssl'
puts OpenSSL::Digest::MD5.hexdigest 'abc'
```
You can also push a test app and it will fail in the staging step.

## Expected result

Staging should succeed.

## Current result

Staging fails, in the `cloud_controller_ng.log` you see:
```
 "test_mode_info"=>{"detail"=>"Stager error: Digest initialization failed: initialization error"
 "title"=>"CF-StagerError"
 "backtrace"=>["/cloud_controller_ng/lib/cloud_controller/diego/stager.rb:36:in `rescue in send_stage_package_request'"
 "/cloud_controller_ng/lib/cloud_controller/diego/stager.rb:31:in `send_stage_package_request'"
 "/cloud_controller_ng/lib/cloud_controller/diego/stager.rb:9:in `stage'"
 "/cloud_controller_ng/app/actions/build_create.rb:87:in `create_and_stage'"
 "/cloud_controller_ng/app/controllers/v3/builds_controller.rb:51:in `create'"
```

## Possible Fix

Replace `OpenSSL::Digest::MD5` with `OpenSSL::Digest::SHA256`. I've found these locations:
https://github.com/cloudfoundry/cloud_controller_ng/blob/ee95901a3a14a6fcd6edc683c916a5d5190333f0/lib/cloud_controller/diego/buildpack/staging_action_builder.rb#L78
https://github.com/cloudfoundry/cloud_controller_ng/blob/ee95901a3a14a6fcd6edc683c916a5d5190333f0/lib/cloud_controller/diego/buildpack/staging_action_builder.rb#L110
[...]

PR #3555, opened 2023-12-14
Bump default key length to 2048 and change fingerprint to SHA1

rkoster:
FIPS 140-2 mode requires keys at least 2048 bits length for RSA The use of MD5 hashes is not FIPS-compliant

SHA1 fingerprints are already supported by diego-ssh: https://github.com/cloudfoundry/diego-ssh/blob/main/proxy/proxy.go#L367-L369

Addresses part of the issues identified in: https://github.com/cloudfoundry/cloud_controller_ng/issues/3544


lib/cloud_controller/diego/ssh_key.rb
```diff
@@ -21,7 +22,9 @@ def authorized_key
           end
         end
 
-        delegate :fingerprint, to: :key
+        def fingerprint
+          ::SSHKey.new(key.to_der).sha1_fingerprint
```
a-b, line 26: Do we know if this is backward compatible with earlier clients?

rkoster: Who is the client in this case? It was my understanding that these fingerprints get validated by the ssh-proxy but do not reach the cf cli.

rkoster: SHA1 support in the ssh-proxy was added 8 years ago: https://github.com/cloudfoundry/diego-ssh/commit/0f6f64f1d6db9e3c1888c98196aefab7da7a9888#diff-1d74870266949f1927a500e5f0002617239d3e3cd9bcceee6be014738e5e0f2bR230

a-b (approved):
LGTM

lib/cloud_controller/diego/ssh_key.rb
a-b: Thanks for providing context.

philippthun, line 26: Small remark: I would rather cache the fingerprint in case someone decides to call this method twice (which doesn't seem to be the case today, but the `authorized_key` is also cached), i.e.
```
@fingerprint ||= ...
```

philippthun (approved):

PR #3557, opened 2023-12-15
Use xxHash in Rate Limiter

philippthun:
- There is no need for a "cryptographic" hash algorithm.
- `XXhash.xxh64(guid)` is 10 times faster than `OpenSSL::Digest:: SHA1.hexdigest(guid).hex`.

Addresses one of the issues identified in: #3544


moleske (approved):
I like that the todo is addressed.  This makes sense to me as this offset doesn't need a cryptographic hash.

I am a little concerned about xxhash cause looking at the [repo](https://github.com/nashby/xxhash) it doesn't seem too actively maintained.  But looking at it, it is probably ok to include

Issue #3561, opened 2023-12-17
[FIPS] CF-StagerError: 170011: Stager error: Digest initialization failed: initialization error


rkoster:
With the follwing PRs applied:
- https://github.com/cloudfoundry/cloud_controller_ng/pull/3557
- https://github.com/cloudfoundry/cloud_controller_ng/pull/3558

I'm getting:
```bash
cf push cf-env --stack cflinuxfs4 --buildpack ruby_buildpack
Pushing app cf-env to org system / space dev as admin...
Applying manifest file /Users/rubenk/workspace/cf-env/manifest.yml...
Manifest applied
Packaging files to upload...
Uploading files...
 186.20 KiB / 186.20 KiB [=============================================================] 100.00% 2s

Waiting for API to complete processing files...

Staging app and tracing logs...
Unexpected Response
Response Code: 500
Request ID:    58eaa0f9-60a0-4e8f-4576-552bca391abe::e81b4d24-449b-4667-8c61-0aa9d20ef8c0
Code: 0, Title: , Detail: {"errors":[{"detail":"Stager error: Digest initialization failed: initialization error","title":"CF-StagerError","code":170011}]}
```

WIth CC stack trace:
```json
{
  "errors": [
    {
      "detail": "Stager error: Digest initialization failed: initialization error",
      "title": "CF-StagerError",
      "code": 170011,
      "test_mode_info": {
        "detail": "Stager error: Digest initialization failed: initialization error",
        "title": "CF-StagerError",
        "backtrace": [
          "/var/vcap/data/packages/cloud_controller_ng/2de9ff7be298a52784f0ab031db749f2a7d01ee5/cloud_controller_ng/lib/cloud_controller/diego/stager.rb:36:in `rescue in send_stage_package_request'",
          "/var/vcap/data/packages/cloud_controller_ng/2de9ff7be298a52784f0ab031db749f2a7d01ee5/cloud_controller_ng/lib/cloud_controller/diego/stager.rb:31:in `send_stage_package_request'",
          "/var/vcap/data/packages/cloud_controller_ng/2de9ff7be298a52784f0ab031db749f2a7d01ee5/cloud_controller_ng/lib/cloud_controller/diego/stager.rb:9:in `stage'",
          "/var/vcap/data/packages/cloud_controller_ng/2de9ff7be298a52784f0ab031db749f2a7d01ee5/cloud_controller_ng/app/actions/build_create.rb:87:in `create_and_stage'",
[...]

PR #3562, opened 2023-12-18
Use SHA256 instead of MD5 when calcualting buildpack mount points

rkoster:
Needs similiar changes on the buildpacklifecycle side

Addresses: https://github.com/cloudfoundry/cloud_controller_ng/issues/3561


philippthun (commented):
I think we need a different approach to prevent staging errors during the update (for existing foundations). Somehow we need to support both paths in parallel for some time...

rkoster:
It was my understanding that the [lifecycle_bundles get co-located with the cloud-controller](https://github.com/cloudfoundry/cf-deployment/blob/main/cf-deployment.yml#L882) so we just need to make sure both cloud-controller and the diego-release will be bumped at the same time, into cf-deployment (or TAS).

moleske:
like @philippthun I am leaning more toward keeping things working in existing foundations.  Mostly cause I'm skeptical of keeping things co located. Though I can be convinced if we think most users are using cf-deployment and we want to put the effort on cf deployment getting a release with both this fix and lifecycle bundle fix at the same time

rkoster:
@jochenehret do you think it would be feasible to coordinate the bump of Cloud Controller and Diego so that both would switch to sha256 sums simultaneously?

cc @ameowlia

jochenehret:
I'm not into the details, but I would doubt that a simultaneous update to sha256 will work without downtimes. For a seamless update, in a first step the receiver of the message should be able to accept the old md5 plus the new sha256 messages. In the next step, the sender could be switched from md5 to sha256. We could release both steps as major cf-deployment versions so that users are required to install both.

rkoster:
caller and receiver are colocated on the same instance, so they would be updated at the same time if the releases are bumped at the same time (cross-working group coordination effort, between CAPI, Diego & CF-deployment).

jochenehret:
Which jobs on which instance_groups are involved here, precisely? Jobs are updated sequentially, so there could still be a small downtime during an update. @philippthun has started to think about solutions and could provide more details.

rkoster:
at this point its just unclear where the lifecycle binary is executed, is it being invoked by the cloud controller directly or is it executed in the diego container during staging and if so where does diego get it from. Are these lifecycle bits uploaded to a blobstore at cloud controller startup and referenced later? or are the bits taken from disk once a staging request comes in.

philippthun:
To be honest, I did not fully understand the workflow yet...

My understanding is that CC creates a message that includes these paths with MD5 or SHA256 checksums and sends this to Diego. Diego spins up a staging container and loads the buildpack_app_lifecycle.tgz from the [file-server](https://github.com/cloudfoundry/cf-deployment/blob/4ec18551b83a948f1b53bfa73ba9d0276270781b/cf-deployment.yml#L193) which seems to run on the api instance-group. So during an update there could be all possible combinations, i.e. new message + old lifecycle, old message + new lifecycle...

rkoster:
When is `buildpack_app_lifecycle.tgz` uploaded to the file-server? Is this done during cloud_controller start? or is this done every time an app gets staged? Also, are these bundles versioned somehow?

rkoster:
This does not look that hopeful:
```
==> file_server.stdout.log <==
{"timestamp":"2023-12-19T09:10:55.592678374Z","level":"info","source":"file-server","message":"file-server.ready","data":{}}
{"timestamp":"2023-12-19T09:31:13.834461087Z","level":"info","source":"file-server","message":"file-server.static-file.response","data":{"method":"GET","session":"1","size":58026358,"status":200,"uri":"/v1/static/buildpack_app_lifecycle/buildpack_app_lifecycle.tgz"}}
{"timestamp":"2023-12-19T09:32:18.436372800Z","level":"info","source":"file-server","message":"file-server.static-file.response","data":{"method":"GET","session":"2","size":0,"status":304,"uri":"/v1/static/buildpack_app_lifecycle/buildpack_app_lifecycle.tgz"}}
{"timestamp":"2023-12-19T10:09:46.756621977Z","level":"info","source":"file-server","message":"file-server.static-file.response","data":{"method":"GET","session":"3","size":0,"status":304,"uri":"/v1/static/buildpack_app_lifecycle/buildpack_app_lifecycle.tgz"}}
```

rkoster:
```
/var/vcap/sys/log/file_server# jq -r .static_directory /var/vcap/jobs/file_server/config/file_server.json
/var/vcap/jobs/file_server/packages/
```

philippthun:
The `buildpack_app_lifecycle.tgz` is "uploaded" during the BOSH deployment, the `buildpack_app_lifecycle` package is a dependency of the `file_server` job.

The file server adds an [Etag header](https://github.com/cloudfoundry/fileserver/blob/ddeb9bd6dfb9e2b23e6dc04fed787c7cf5f97236/handlers/static/fileserver.go#L61). Thus the HTTP status code 304 in your log, @rkoster.

rkoster:
Yeah, it seems like it's just serving the whole packages directory. So depending on if the file_server instance has been updated or not by the bosh director it will server to old or the new version of `buildpack_app_lifecycle.tgz`.

philippthun:
So we need to do this in three steps:
1. Update `buildpack_app_lifecycle` to support both (MD5 and SHA256) path variants (see [comment](https://github.com/cloudfoundry/buildpackapplifecycle/pull/65#issuecomment-1862768039)).
2. Switch CC from using MD5 to SHA256 paths.
3. Update `buildpack_app_lifecycle` to only support SHA256 paths.

- Each step should be done in a major cf-deployment release.
- Open question: Do we really want to use a cryptographic hashing algorithm although this is not required?

rkoster:
If step 1 would be implemented with a fallback (only calculate md5 if sha256 based path does not exist). We could get away with 2 deploys. 

We might be able to do it with 1 deploy if the file-server URL used by the cloud-controller [would point to the instance which has created the staging request](https://bosh.io/docs/dns/#placeholder-alias). The only downside here is that these request would fail if the instance would get updated.

rkoster:
With regards cryptographic hashing, I'm open to using something like xxhash. Inline with https://github.com/cloudfoundry/cloud_controller_ng/pull/3557

rkoster:
Ideally the use_xxhash_for_buildpack_staging_paths code path could be enable via a cloud controller feature flag.

This way it could option be switched on via a post deploy lifecycle script. Which means all changes could be rolled out in a single deploy.

philippthun:
> Ideally the use_xxhash_for_buildpack_staging_paths code path could be enable via a cloud controller feature flag.

Yes, that would be a good shortcut. But we should have this flag only temporarily and remove it together with the obsolete MD5 coding in a subsequent (major cf-deployment) release.

rkoster:
After giving it some more thought, controlling the hashing algorithm via a bosh property `cc.staging_buildpack_path_algorithm` which would default to `xxh64` but would also support `md5` for the time being. From the VMware side, we probably would want to keep this option around for at least a year since we can only change the default in our major releases which we do twice a year.

rkoster:
Before bumping this into capi release, the `legacy_md5_buildpack_paths_enabled` property should be exposed via the release.
@jochenehret When bumping capi release the `legacy_md5_buildpack_paths_enabled` property should be set to true, until the next major release.

winkingturtle-vmw:
Can we use a more standard algorithm like SHA256 instead? I don't know if speed is of an issue here, but I think using a more known algorithm like SHA256 would be a better pattern.

winkingturtle-vmw:
@rkoster My understanding is that we are doing this because MD5 is failing on FIPS compliant stemcells, not necessarily because MD5/SHA256 is safe or not. I think using some algorithm that is more known would be safer option here. I don't know who maintains `xxh64` whereas SHA256 implementation is maintained by the Go team.

rkoster:
Had a call with @winkingturtle-vmw to do a bit more research on xxhash to address the well-known algorithm concern.
It turns out xxhash is being used in the [linux kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=59e1a2f4bf83744e748636415fde7d1e9f557e05), as well as [promethus](https://github.com/prometheus/prometheus/blob/6cd24d87b0d4e3acd4503b061d32c5aa6d8883a9/model/labels/labels.go#L22).

spec/unit/lib/cloud_controller/diego/buildpack/staging_action_builder_spec.rb
```diff
@@ -45,13 +45,14 @@ module Buildpack
...
         end
+        let(:legacy_md5_buildpack_paths_enabled) { false }
         let(:buildpacks) { [] }
         let(:generated_environment) { [::Diego::Bbs::Models::EnvironmentVariable.new(name: 'generated-environment', value: 'generated-value')] }
 
         before do
           allow(LifecycleBundleUriGenerator).to receive(:uri).with('the-buildpack-bundle').and_return('generated-uri')
           allow(BbsEnvironmentBuilder).to receive(:build).with(env).and_return(generated_environment)
-          TestConfig.override(credhub_api: nil)
+          TestConfig.override(staging: { legacy_md5_buildpack_paths_enabled: }, credhub_api: nil)
```
philippthun, line 55: Other config options are defined above (`let(:config)`) and passed to the constructor. `legacy_md5_buildpack_paths_enabled` should also be added there.

spec/unit/lib/cloud_controller/diego/buildpack/staging_action_builder_spec.rb
```diff
@@ -20,6 +21,7 @@ module Buildpack
                          enable_declarative_asset_downloads: enable_declarative_asset_downloads
                        },
                        staging: {
+                         legacy_md5_buildpack_paths_enabled:,
```
philippthun, line 24:
```suggestion
                         legacy_md5_buildpack_paths_enabled: legacy_md5_buildpack_paths_enabled,
```

PR #3570, opened 2023-12-21
Fix action unit_tests_backwards_compatibility

FloThinksPi:


FloThinksPi:
Backwards Compatibility Unit Tests only work when the rake task `spec:without_migration` is available in the main branch (as it tries to run new schema with old code). So these failures can be ignored for now.

kathap (approved):

PR #3585, opened 2024-01-05
Introduce config option to disable statsd metrics

kathap:
This config option allows to send metrics only via prometheus on api vms. The default is true (i.e. enabled). In the default case, metrics are sent via statsd and prometheus.

#3526, #3528, #3521

### DIFF
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -349,14 +349,11 @@ module CloudController
     def statsd_client
       if @dependencies[:statsd_client]
         @dependencies[:statsd_client]
+      elsif config.get(:enable_statsd_metrics) == true || config.get(:enable_statsd_metrics).nil?
+        Statsd.logger = Steno.logger('statsd.client')
+        register(:statsd_client, Statsd.new(config.get(:statsd_host), config.get(:statsd_port)))
       else
-        config = CloudController::DependencyLocator.instance.config
-        if config.get(:enable_statsd_metrics) == true
-          Statsd.logger = Steno.logger('statsd.client')
-          register(:statsd_client, Statsd.new(config.get(:statsd_host), config.get(:statsd_port)))
-        else
-          register(:statsd_client, NullStatsdClient.new)
-        end
+        register(:statsd_client, NullStatsdClient.new)
       end
     end
 
--- a/lib/cloud_controller/dependency_locator.rb
+++ b/lib/cloud_controller/dependency_locator.rb
@@ -351,7 +351,7 @@ module CloudController
         @dependencies[:statsd_client]
       else
         config = CloudController::DependencyLocator.instance.config
-        if config.get(:disable_statsd_metrics) == false
+        if config.get(:enable_statsd_metrics) == true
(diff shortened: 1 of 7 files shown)


Gerg:
What do you think about naming the config value `enable_statsd_metrics` instead? That way we don't need to worry about double negatives.

lib/cloud_controller/dependency_locator.rb
```diff
@@ -350,8 +350,13 @@ def statsd_client
       if @dependencies[:statsd_client]
         @dependencies[:statsd_client]
       else
-        Statsd.logger = Steno.logger('statsd.client')
-        register(:statsd_client, Statsd.new(config.get(:statsd_host), config.get(:statsd_port)))
+        config = CloudController::DependencyLocator.instance.config
+        if config.get(:enable_statsd_metrics) == true
```
philippthun, line 354: This now evaluates to false for clock and deployment updater, but it shouldn't.

spec/unit/lib/cloud_controller/dependency_locator_spec.rb
```diff
@@ -484,6 +484,41 @@
...
+        and_return(expected_client)
+
+      expect(locator.statsd_client).to eq(expected_client)
+    end
+
+    it 'returns the statsd client for other than api vms' do
+      host = 'test-host'
+      port = 1234
+      TestConfig.context = :deployment_updater
+      TestConfig.override(
```
philippthun, line 510: This takes the default values from the `cloud_controller.yml`, i.e. `enable_statsd_metrics: true`. I suggest that you overwrite it with `nil` in this test case.

lib/cloud_controller/dependency_locator.rb
```diff
@@ -350,8 +350,13 @@ def statsd_client
       if @dependencies[:statsd_client]
         @dependencies[:statsd_client]
       else
-        Statsd.logger = Steno.logger('statsd.client')
-        register(:statsd_client, Statsd.new(config.get(:statsd_host), config.get(:statsd_port)))
+        config = CloudController::DependencyLocator.instance.config
```
philippthun, line 353: This line is not needed. `config` can be accessed directly.

spec/unit/lib/cloud_controller/dependency_locator_spec.rb
```diff
@@ -484,6 +484,41 @@
 
       expect(locator.statsd_client).to eq(expected_client)
     end
+
+    it 'returns the null statsd client' do
```
philippthun, line 488: Somehow the test context is missing, i.e. "when does it return the null statsd client"?

```suggestion
    it 'returns the null statsd client if enable_statsd_metrics is set to false' do
```

spec/unit/lib/cloud_controller/dependency_locator_spec.rb
```diff
@@ -484,6 +484,41 @@
...
+      host = 'test-host'
+      port = 1234
+
+      TestConfig.override(
+        statsd_host: host,
```
philippthun, line 493: For this test case host and port are not needed. Removing them makes the test shorter and easier to read.

philippthun (approved):

PR #3612, opened 2024-01-24
Task labels were not getting deleted causing issues with foreign key violations.

MarcPaquette:

When applying labels to tasks, it generate a foreign key constraint between tasks and task labels. When deleting the task, the underlying label does not cascade it's delete. This fix deletes the label of the tasks that are being marked for deletion.

### DIFF
--- a/app/jobs/runtime/prune_completed_tasks.rb
+++ b/app/jobs/runtime/prune_completed_tasks.rb
@@ -13,10 +13,12 @@ module VCAP::CloudController
         def perform
           logger = Steno.logger('cc.background')
           logger.info('Cleaning up old TaskModel rows')
-
           tasks_to_delete = TaskModel.where(state: prunable_states).where(Sequel.lit('updated_at < ?', cutoff_age))
-          deleted_count   = Database::BatchDelete.new(tasks_to_delete).delete
+          task_labels_to_delete = TaskLabelModel.where(task: tasks_to_delete)
+          deleted_label_count = Database::BatchDelete.new(task_labels_to_delete).delete
+          deleted_count = Database::BatchDelete.new(tasks_to_delete).delete
 
+          logger.info("Cleaned up #{deleted_label_count} TaskLabelModel rows")
           logger.info("Cleaned up #{deleted_count} TaskModel rows")
         end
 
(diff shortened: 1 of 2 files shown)


app/jobs/runtime/prune_completed_tasks.rb
```diff
@@ -13,10 +13,12 @@ def initialize(cutoff_age_in_days)
         def perform
           logger = Steno.logger('cc.background')
           logger.info('Cleaning up old TaskModel rows')
-
           tasks_to_delete = TaskModel.where(state: prunable_states).where(Sequel.lit('updated_at < ?', cutoff_age))
-          deleted_count   = Database::BatchDelete.new(tasks_to_delete).delete
+          task_labels_to_delete = TaskLabelModel.where(task: tasks_to_delete)
+          deleted_label_count = Database::BatchDelete.new(task_labels_to_delete).delete
```
sethboyles, line 18: Can you also do this for [Task Annotations](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/models/runtime/task_annotation_model.rb), which are the [second component of the CCNG metadata concept](https://v3-apidocs.cloudfoundry.org/version/3.155.0/index.html#annotations)?  I think they could cause a similar error.

app/jobs/runtime/prune_completed_tasks.rb
```diff
@@ -13,10 +13,12 @@ def initialize(cutoff_age_in_days)
...
+          deleted_count = Database::BatchDelete.new(tasks_to_delete).delete
```
sethboyles, line 19: It might be good to wrap the deletions in a `transaction` block in case the final task deletion fails--then the label deletion could be rolled back.  

Here's an example of a transaction block with the Sequel ORM:

https://github.com/cloudfoundry/cloud_controller_ng/blob/ee25971edd577137ecc1e7104ad343e55043de9f/app/actions/buildpack_create.rb#L11-L23

(it doesn't matter which Model or resource you use to get to the `db` object)

app/jobs/runtime/prune_completed_tasks.rb
MarcPaquette: Added annotation removal [in this commit](https://github.com/cloudfoundry/cloud_controller_ng/pull/3612/commits/c817d92ab222193d5a917dd2341f93166a6ebead)

app/jobs/runtime/prune_completed_tasks.rb
MarcPaquette: Added database transaction [in this commit.](https://github.com/cloudfoundry/cloud_controller_ng/pull/3612/commits/c817d92ab222193d5a917dd2341f93166a6ebead)

PR #3636, opened 2024-02-09
Improve performance of decorators

philippthun:
- Use direct attributes (e.g. `space.organization_id`) instead of accessor methods that require another object to be loaded from the database (e.g. `space.organization_guid` which is a helper method for `space.organization.guid`).
- When another object is needed to create the list of ids/guids that should be included (e.g. `service_plan.service.service_broker.space_id` requires `service` and `service_broker` objects), a test case has been added that ensures these objects are eager loaded when the decorator is used in a list endpoint.
- Sort by `created_at` as part of the database query, don't sort the result set afterwards.
- Handle `nil` values correctly, i.e. compact arrays and don't issue database queries if nothing would get selected. Add tests that ensure this behavior.

### DIFF
--- a/app/decorators/include_role_organization_decorator.rb
+++ b/app/decorators/include_role_organization_decorator.rb
@@ -9,11 +9,13 @@ module VCAP::CloudController
 
       def decorate(hash, roles)
         hash[:included] ||= {}
-        organization_guids = roles.map { |role| role.organization_guid unless role.for_space? }.uniq
-        organizations = Organization.where(guid: organization_guids).
-                        order(:created_at).eager(Presenters::V3::OrganizationPresenter.associated_resources).all
+        organization_ids = roles.reject(&:for_space?).map(&:organization_id).uniq
+        unless organization_ids.empty?
+          organizations = Organization.where(id: organization_ids).order(:created_at).
+                          eager(Presenters::V3::OrganizationPresenter.associated_resources).all
+        end
 
-        hash[:included][:organizations] = organizations.map { |organization| Presenters::V3::OrganizationPresenter.new(organization).to_hash }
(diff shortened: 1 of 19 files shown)


app/decorators/include_role_space_decorator.rb
```diff
@@ -7,11 +7,13 @@ def match?(include_params)
 
       def decorate(hash, roles)
         hash[:included] ||= {}
-        space_guids = roles.map(&:space_guid).uniq
-        spaces = Space.where(guid: space_guids).order(:created_at).
-                 eager(Presenters::V3::SpacePresenter.associated_resources).all
+        space_ids = roles.map { |role| role.space_id if role.for_space? }.compact.uniq
```
sethboyles, line 10: might be a bit cleaner to do something like

```
roles.select(&:for_space?).map(&:space_id).uniq
```

app/decorators/include_role_organization_decorator.rb
```diff
@@ -9,11 +9,13 @@ def match?(include_params)
 
       def decorate(hash, roles)
         hash[:included] ||= {}
-        organization_guids = roles.map { |role| role.organization_guid unless role.for_space? }.uniq
-        organizations = Organization.where(guid: organization_guids).
-                        order(:created_at).eager(Presenters::V3::OrganizationPresenter.associated_resources).all
+        organization_ids = roles.map { |role| role.organization_id unless role.for_space? }.compact.uniq
```
sethboyles, line 12: similarly could refactor this to

```
organization_ids = roles.select(&:for_space?).map(&:organization_id).uniq
```

sethboyles (approved):

app/decorators/include_role_space_decorator.rb
philippthun: Yes, that's better. Thanks!

PR #3644, opened 2024-02-17
Add support for comma-delimited destinations in ASGs

jrussett:

Allow operators to configure Cloud Controller to support ASGs with comma-delimited destinations.
For example, the following ASG definition would be valid:

```
[
 {
   "protocol": "tcp",
   "destination": "1.2.3.4,10.0.0.0/24,25.0.0.0-26.0.0.0",
   "ports": "65432",
   "description": "Valid comma delimited list of destinations"
 }
]
```

Currently, if a developer wants to open connections from their cf app to two non-consecutive IP addresses (e.g. 8.8.8.8 and 10.10.10.10), they'd have to create two separate ASGs. This creates duplication and performance implications on some large environments with many ASGs that are also running networking fabrics _other_ than silk.

One way to reduce said duplication is to enable comma-delimited destinations, so that the firewall rules are consolidated.

### Links

- https://github.com/cloudfoundry/capi-release/pull/386
- https://github.com/cloudfoundry/cloud_controller_ng/pull/3644
- https://github.com/cloudfoundry/silk-release/pull/107

- [#186770494](https://www.pivotaltracker.com/story/show/186770494)

##### Checklist


sethboyles:
We probably want to clarify in the docs that destinations can be a comma delimited list in our docs: https://v3-apidocs.cloudfoundry.org/version/3.159.0/index.html#the-security-group-object (if enabled)

app/messages/validators/security_group_rule_validator.rb
```diff
@@ -80,48 +80,31 @@ def validate_icmp_protocol(rule, record, index)
...
-
-      return true
-    end
-
-    false
-  end
-
-  def port_in_range(port)
-    port > 0 && port < 65_536
+    CloudController::TransportRuleValidator.validate_port(ports)
```
sethboyles, line 89: 👍

app/messages/validators/security_group_rule_validator.rb
```diff
@@ -131,34 +114,42 @@ def valid_destination_type(destination, record, index)
...
+      unless comma_delimited_destinations_enabled
+        add_rule_error(error_message, record, index)
+        return
+      end
+
+      destinations = destination.partition(',')
+      destination = destinations.first
+      remainder = destinations.last
+
+      validate_destination(remainder, record, index)
```
sethboyles, line 132: I personally think using recursion here is less clear than doing something like `destination.split(',').each { |foo| validate_destination(foo)}` up on line 30

jrussett: I tried implementing this with iteration instead of recursion in this commit:
- https://github.com/cloudfoundry/cloud_controller_ng/pull/3644/commits/c5c437db1546823259b4153aa970d6895002ddaa

Please let me know if you think that is more legible.

jrussett:
Hi @sethboyles
Thank you for looking over the PR. I have addressed your comments thus far. Please let me know if there is anything else I can answer or fix. 🙇

app/messages/validators/security_group_rule_validator.rb
sethboyles: That looks good to me.  Thanks for changing it!

a-b:
I'm curious if you consider having destinations as an [ ] array as an alternative to a comma-separated string.

sethboyles:
@a-b it's worth noting that CCNG already accepts ports in ASG definitions as a comma-separated list: https://v3-apidocs.cloudfoundry.org/version/3.159.0/index.html#the-security-group-object, so to do so with IP addresses is at least consistent with that

sethboyles (approved):

sethboyles:
@jrussett unfortunately this has gotten conflicts now so I can't merge directly. Sorry that this got delayed so long that this happened. I can attempt to resolve it myself when I have time, but I think the commits should also be squashed/combined into one. Would you like to resolve this or should I give it a shot?

sethboyles:
Oops the conflicts were only on the 'rebase' option, seems like I can squash & merge no problem

PR #3651, opened 2024-02-23
Use DISTINCT only when required (service plan|offering list fetcher)

philippthun:
As part of PR #3620 it has been validated that there is no overlap between the different sources (`UNION`ed datasets). But as these single datasets are created by `JOIN`ing several tables, there might be duplicates depending on the cardinality of those relations. Thus the overall `DISTINCT` is replaced by `DISTINCT`s added when `JOIN`ing a table with a one-to-many relation. Calling `.distinct` multiple times does not have any negative effect as this simply sets a flag on the Sequel dataset.

This change should improve performance of requests to the following endpoints whenever those `JOIN`s are not needed (e.g. depending on request parameters):
- /v3/service_offerings
- /v3/service_plans

Tests now ensure that there are:
- multiple plans per offering
- multiple service instances per plan
- multiple visibilities per plan
- multiple spaces per organization

### DIFF
--- a/app/fetchers/base_service_list_fetcher.rb
+++ b/app/fetchers/base_service_list_fetcher.rb
@@ -63,7 +63,7 @@ module VCAP::CloudController
                     end.from_self(alias: klass.table_name)
                   end
 
-        dataset.distinct.eager(eager_loaded_associations)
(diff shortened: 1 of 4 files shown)


johha (approved):
LGTM

PR #3656, opened 2024-02-27
Enable new rubocops

philippthun:

### DIFF
--- a/.rubocop_cc.yml
+++ b/.rubocop_cc.yml
@@ -9,7 +9,6 @@ require:
   - ./spec/linters/migration/require_primary_key.rb
   - ./spec/linters/match_requires_with_includes.rb
 
-
 AllCops:
   TargetRubyVersion: 3.2
   Exclude:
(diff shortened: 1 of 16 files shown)


moleske (approved):
Woo!  I definitely was failing at this.  My mistake was apparently never adding the `to be(true)` off the ends of the `match?`

philippthun:
> Woo! I definitely was failing at this. My mistake was apparently never adding the `to be(true)` off the ends of the `match?`

Sometimes rubocop is very persistent in forcing you into the wrong direction...

Initially the test was written as follows - which is very readable from my point of view:
```
expect(decorator.match?(%w[potato app turnip])).to be_truthy
```

Then rubocop wanted it this way (already less readable):
```
expect(decorator).to be_match(%w[potato app turnip])
```

And now the suggestion was:
```
expect(decorator).to match(%w[potato app turnip])
```

As this changed the semantics and thus failed the test, I reverted back to the original version, but with `be(true)` instead of `be_truthy`. This silenced rubocop!

PR #3659, opened 2024-02-27
update lock information logging to happen on info and to happen only once per state switch

Benjamintf1:

> update lock information logging to happen on info and to happen only once per state switch

> makes lock information easier to find in logs, but less verbose so to be able to bring it to the info level

### DIFF
--- a/lib/locket/lock_runner.rb
+++ b/lib/locket/lock_runner.rb
@@ -25,13 +25,20 @@ module Locket
       raise Error.new('Cannot start more than once') if @thread
 
       @thread = Thread.new do
+        failed = false
         loop do
           begin
             service.lock(build_lock_request)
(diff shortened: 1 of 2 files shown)


Benjamintf1:
Would we want to keep these additional logs at a debug level, or remove all debug logs.

Benjamintf1:
I just figured if you get one error, and a different error, it might be nice to see both if it's set to debug. Especially because I think saving the error off is going to lead to error spamming(I believe the errors have embeded timestamps). But I'm fine with just 1 log on both case.

sethboyles (approved):

sethboyles:
I agree that seeing more logs would be nice if you set to debug, but not sure how many operators would set to debug.  I saw that is your call--I don't think its necessary, however.

PR #3664, opened 2024-02-29
Remove migration views

kathap:
After merging #3394 and #3393 we can remove the views we introduced for annotations and labels tables.

This is a breaking change, migrating from a CF-Deployment version before PR 3394 to a version that removes the view will cause a downtime on the old CloudControllers since they still use the views.

In this database migration the annotations and labels migrations views will be dropped. They were introduced to in #3394 to help implementing the uniqueness of labels and annotations, they have no other use in the code. Now that #3394 and #3393 are through, the views are not needed anymore and can be deleted.

#3394 and #3393

### DIFF
--- a/db/migrations/20240229113008_drop_annotations_and_labels_migration_views.rb
+++ b/db/migrations/20240229113008_drop_annotations_and_labels_migration_views.rb
@@ -1,0 +1,48 @@
+Sequel.migration do
+  table_base_names = %w[
+    app
+    build
+    buildpack
+    deployment
+    domain
+    droplet
+    isolation_segment
+    organization
+    package
+    process
+    revision
+    route_binding
+    route
+    service_binding
+    service_broker
+    service_broker_update_request
+    service_instance
+    service_key
+    service_offering
+    service_plan
+    space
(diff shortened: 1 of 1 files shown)


db/migrations/20240229113008_drop_annotations_and_labels_migration_views.rb
```diff
@@ -0,0 +1,48 @@
...
+      transaction do
+        drop_view(:"#{table}_migration_view", if_exists: true)
+      end
+    end
+  end
+
+  down do
+    (annotation_tables + label_tables).each do |table|
+      transaction do
+        create_view(:"#{table}_migration_view", self[table.to_sym].select { [id, guid, created_at, updated_at, resource_guid, key_prefix, key_name, value] }, if_exists: true)
```
philippthun, line 44: Is there an `if_exists` option for `create_view`?

kathap: No, only for drop_view, good point

philippthun (approved):

PR #3672, opened 2024-03-04
Stream CCDB migration logs to ELK

WeiQuan0605:
DB migration logs are not available in ELK. It is difficult to analyse DB migration issues. 

In capi-release there is a config [cc.run_prestart_migrations](https://github.com/cloudfoundry/capi-release/blob/develop/jobs/cloud_controller_ng/spec#L399). If this config is false, it means that the PostgreSQL job is deployed to the same VM as Cloud Controller, and migration will not run in pre-start of Cloud Controller. With this config, logging can be passed to syslog's blackbox, then it can be seen in ELK. 

This Commit tests if CC, nginx_cc and ccng_monit_http_healthcheck work when there is a long migration. 


johha (approved):
LGTM

lib/cloud_controller/db.rb
```diff
@@ -139,6 +139,12 @@ def empty_from_sql
   end
 end
 
+class Sequel::Migration
+  def logger
```
philippthun, line 143: This is not used in the test migration.

Ideally one could use `log_info` from the `Sequel::Database` class in a migration ([here](https://github.com/jeremyevans/sequel/blob/70bcc08e47b060957e37584ee85c304d5698ef7c/lib/sequel/database/logging.rb#L31)).

johha (approved):
LGTM

lib/tasks/db.rake
```diff
@@ -265,6 +265,12 @@ namespace :db do
     DBMigrator.from_config(RakeConfig.config, db_logger).rollback(number_to_rollback)
   end
 
+  def logging_output
```
philippthun, line 268: The task `ensure_migrations_are_current` also calls `Steno.init`. This can be replaced with `logging_output` as well.

spec/migrations/migration_logs_to_elk_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
```
philippthun, line 1: `_to_elk` is not tested here, thus I would rename the file to `migration_logs_spec.rb` as it tests logs written during a migration.

spec/migrations/migration_logs_to_elk_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+RSpec.describe 'migration ccdb logs to elk', isolation: :truncation, type: :migration do
```
philippthun, line 3:
```suggestion
RSpec.describe 'migration logs', isolation: :truncation, type: :migration do
```

spec/migrations/migration_logs_to_elk_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  let(:db) { Sequel::Model.db }
+  let(:tmp_migrations_dir) { Dir.mktmpdir }
+
+  before do
+    long_time_migration = <<-MIGRATION
+      Sequel.migration do
+        up do
+          10.times do
+            VCAP::Migration.logging('Still Migrating')
```
philippthun, line 12: When changing the method in `VCAP::Migration` as described above, this should be:
```
VCAP::Migration.logger.info(...)
```

spec/migrations/migration_logs_to_elk_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+      end
+    MIGRATION
+
+    migration_file = "#{tmp_migrations_dir}/001_test_for_long_time_migration.rb"
+    File.write(migration_file, long_time_migration)
+  end
+
+  after do
+    Sequel::Migrator.run(db, tmp_migrations_dir, allow_missing_migration_files: true)
```
philippthun, line 26: I think the `after` block should revert the migration, i.e. add param `target: 0`.

spec/migrations/migration_logs_to_elk_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    FileUtils.rm_rf(tmp_migrations_dir)
+  end
+
+  it 'runs the up migration without errors' do
+    allow(VCAP::Migration).to receive(:logging)
+    expect(VCAP::Migration).to receive(:logging).with('Still Migrating').exactly(10).times
+    expect(VCAP::Migration).to receive(:logging).with('migration finished').once
+    expect { Sequel::Migrator.run(db, tmp_migrations_dir, target: 0, allow_missing_migration_files: true) }.not_to raise_error
```
philippthun, line 34: Here the `target: 0` param is wrong as we want to run the migration, not revert it.

spec/migrations/migration_logs_to_elk_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    File.write(migration_file, long_time_migration)
+  end
+
+  after do
+    Sequel::Migrator.run(db, tmp_migrations_dir, allow_missing_migration_files: true)
+    FileUtils.rm_rf(tmp_migrations_dir)
+  end
+
+  it 'runs the up migration without errors' do
+    allow(VCAP::Migration).to receive(:logging)
```
philippthun, line 31: To test the changed method in `VCAP::Migration` I suggest the following:

1. Define a `logger` test variable: `let(:logger) { instance_double(Steno::Logger) }`
2. Allow `Steno.logger` to be called only with the correct name (and return the mock from above): `allow(Steno).to receive(:logger).with('cc.db.migrations').and_return(logger)`
3. Expect that the correct messages are sent to `logger.info`: `expect(logger).to receive(:info).with(...)`

philippthun (changes requested):

philippthun (approved):

PR #3675, opened 2024-03-07
error log deserialization errors

Benjamintf1:

This would allow users to retrieve applications without environment variables causing problems. Hypothetically this error should *never* be thrown, but if it is, then it's likely not something we can recover the environment variables that were there before. It's maybe better to treat them like they just got wiped out. This does not preclude work to try and fix how we got it corrupted, but we couldn't reproduce. 

We're trying to look into a interesting problem. What seems to have happened, although we can't identify the cause, is that the environment variables of an application got corrupted. Specifically, it went from json in the db to a null terminated string that isn't json. We never got the specific data.

This causes extreme platform instability from a certain perspective. Specifically, it causes inability to read all the applications from the platform. Anything scanning /v2/apps will be unable to obtain all the applications without randomly getting an error partway through if *any* application's environment variables are corrupted. 

### DIFF
--- a/lib/cloud_controller/serializer.rb
+++ b/lib/cloud_controller/serializer.rb
@@ -10,7 +10,7 @@ module VCAP::CloudController::Serializer
         begin
           MultiJson.load string
         rescue MultiJson::ParseError
-          logger ||= Steno.logger('cc.serializer')
+          logger = Steno.logger('cc.serializer')
           logger.error("Failed to deserialize #{guid}")
           { 'corrupted-env': string }
         end
--- a/lib/cloud_controller/serializer.rb
+++ b/lib/cloud_controller/serializer.rb
@@ -12,7 +12,7 @@ module VCAP::CloudController::Serializer
         rescue MultiJson::ParseError
           logger ||= Steno.logger('cc.serializer')
           logger.error("Failed to deserialize #{guid}")
-          nil
+          { 'corrupted-env': string }
         end
       end
       alias_method "#{accessor_method_name}_without_serialization", accessor_method_name
--- a/lib/cloud_controller/serializer.rb
+++ b/lib/cloud_controller/serializer.rb
@@ -10,7 +10,8 @@ module VCAP::CloudController::Serializer
         begin
           MultiJson.load string
         rescue MultiJson::ParseError
-          raise CloudController::Errors::ApiError.new_from_details('DeserializationError', "Failed to deserialize #{guid} for object type #{self.class}. Trying to deserialize #{string}. You may have to delete and recreate the object")
+          error = "Failed to deserialize #{guid} for object type #{self.class}. Trying to deserialize #{string}. You may have to delete and recreate the object"
+          raise CloudController::Errors::ApiError.new_from_details('DeserializationError', error)
         end
       end
       alias_method "#{accessor_method_name}_without_serialization", accessor_method_name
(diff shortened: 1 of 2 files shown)


Benjamintf1:
I'm going to update this in just a second.

Benjamintf1:
image
tried to make this change and for some reason it's causing my tests to hang?

lib/cloud_controller/serializer.rb
```diff
@@ -7,7 +7,13 @@ def serializes_via_json(accessor_method_name)
         string = send("#{accessor_method_name}_without_serialization")
         return if string.blank?
 
-        MultiJson.load string
+        begin
+          MultiJson.load string
+        rescue MultiJson::ParseError
+          logger = Steno.logger('cc.serializer')
+          logger.error("Failed to deserialize #{guid}")
+          { 'corrupted-env': string }
```
philippthun, line 15: As these methods are also defined for other fields, returning `corrupted-env` seems wrong to me.

lib/cloud_controller/serializer.rb
```diff
@@ -7,7 +7,13 @@ def serializes_via_json(accessor_method_name)
         string = send("#{accessor_method_name}_without_serialization")
         return if string.blank?
 
-        MultiJson.load string
+        begin
+          MultiJson.load string
+        rescue MultiJson::ParseError
+          logger = Steno.logger('cc.serializer')
+          logger.error("Failed to deserialize #{guid}")
```
philippthun, line 14: I guess it's safe to assume that there is always a `guid` - but as these methods are defined for different models, it might be useful to also log the model type along with the guid.

lib/cloud_controller/serializer.rb
```diff
@@ -7,7 +7,13 @@ def serializes_via_json(accessor_method_name)
         string = send("#{accessor_method_name}_without_serialization")
         return if string.blank?
 
-        MultiJson.load string
+        begin
+          MultiJson.load string
```
philippthun, line 11: There seem to be other places where JSON data is loaded in a similar way, e.g. https://github.com/cloudfoundry/cloud_controller_ng/blob/e719d017ea4625397a97c2c14352ebdee66febe9/app/models/runtime/environment_variable_group.rb#L35

I think we should have a similar exception handling there as well.

lib/cloud_controller/serializer.rb
Benjamintf1: Good point.

Benjamintf1: Fixed this, easy fix. As for the others, I'm going to chat with some of my teammates either late this week or early next week about this.

Benjamintf1:
Let me update this pr branch more. Also https://github.com/cloudfoundry/cloud_controller_ng/issues/3700 made a issue.

lib/cloud_controller/serializer.rb
Benjamintf1: Got it.

lib/cloud_controller/serializer.rb
Benjamintf1: Strangely enough, I was unable to reproduce this error through the api, like how I was for applications. Perhaps this is only used on the backend?

Benjamintf1:
image
New error example

Benjamintf1:
image

version with changes. Still wasn't able to get the environemnt_variables_group to trigger.

sethboyles (approved):

lib/cloud_controller/serializer.rb
Benjamintf1: @philippthun lmk if you have any further questions, or concerns

philippthun (approved):

philippthun:
Please squash commits before merging or use "squash and merge".

PR #3694, opened 2024-03-22
Add new metric job_queue_load for cc-worker

WeiQuan0605:
Relying only on the length of the cc-worker job queue to decide when to scale out resources isn't always accurate. This is because the queue might contain jobs that aren't ready to run and aren't using any resources.

We propose to add a new metric to measure the system load. The new metric will show the number of jobs that are ready to start, not just all the jobs in the queue. 

Related PR for cf doc: [cloudfoundry/docs-running-cf/120](https://github.com/cloudfoundry/docs-running-cf/pull/120)

### DIFF
--- a/spec/unit/lib/cloud_controller/metrics/periodic_updater_spec.rb
+++ b/spec/unit/lib/cloud_controller/metrics/periodic_updater_spec.rb
@@ -72,6 +72,7 @@ module VCAP::CloudController::Metrics
       before do
         allow(statsd_updater).to receive(:update_user_count)
         allow(statsd_updater).to receive(:update_job_queue_length)
+        allow(statsd_updater).to receive(:update_job_queue_load)
         allow(statsd_updater).to receive(:update_thread_info_thin)
         allow(statsd_updater).to receive(:update_failed_job_count)
         allow(statsd_updater).to receive(:update_vitals)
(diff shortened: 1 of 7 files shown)


spec/unit/lib/cloud_controller/metrics/periodic_updater_spec.rb
```diff
@@ -353,6 +368,124 @@ module VCAP::CloudController::Metrics
...
+          cc_generic: 1,
+          'cc-api-0': 0
+        }
+        expected_total = 2
+
+        expect(statsd_updater).to have_received(:update_job_queue_load).with(expected_pending_job_queue_load, expected_total)
+        expect(prometheus_updater).to have_received(:update_job_queue_load).with(expected_pending_job_queue_load)
+      end
+
+      it 'ignores jobs that have already been attempted' do
```
svkrieger, line 445: Did you check whether there are any jobs which attempt's are increased and they get rerun?

svkrieger (commented):
Looks good! I would just change a small thing in the tests. And I'm not sure whether we have any jobs which attempt's increase and get rerun. In that case we should not exclude jobs which have attempts > 0.

spec/unit/lib/cloud_controller/metrics/periodic_updater_spec.rb
WeiQuan0605: I've checked, and what you're talking about is possible, but we have yet to realize it

WeiQuan0605: Conclusion: instead of using `attempts`, using `failed_at` to filter all jobs that are not failed before

svkrieger (approved):

PR #3697, opened 2024-03-25
Fix retry for UniqueConstraintViolation in RouteCreate (v3)

philippthun:
The creation of internal routes depends on finding the next `vip_offset`. When creating multiple routes in parallel, they find the same `vip_offset` and only the very first write operation succeeds. All the others fail with a `UniqueConstraintViolation` error. For this error a retry is triggered in the `VCAP::CloudController::RouteCreate` action (v3).

This retry did not work due to a bug (`undefined local variable or method 'user_audit_info'`). Besides fixing this bug a test has been added to ensure that the parallel creation of internal routes is working.

When running 5 parallel threads and each thread is creating an internal route, the following logs are produced:
```
Duplicate entry '1' for key 'routes.routes_vip_offset_index', retrying
Duplicate entry '1' for key 'routes.routes_vip_offset_index', retrying
Duplicate entry '1' for key 'routes.routes_vip_offset_index', retrying
Duplicate entry '1' for key 'routes.routes_vip_offset_index', retrying
Duplicate entry '2' for key 'routes.routes_vip_offset_index', retrying
Duplicate entry '2' for key 'routes.routes_vip_offset_index', retrying
Duplicate entry '2' for key 'routes.routes_vip_offset_index', retrying
Duplicate entry '3' for key 'routes.routes_vip_offset_index', retrying
Duplicate entry '3' for key 'routes.routes_vip_offset_index', retrying
Duplicate entry '4' for key 'routes.routes_vip_offset_index', retrying
```
This indicates that all the write operations are being executed at the same point in time and only one of them succeeds whereas the others are failing, i.e. 4 out of 5. For each retry there is one conflicting route less, so the number of errors goes down from 4 to 3 to 2 to 1 until the last route could be saved (after 4 retries).

### DIFF
--- a/app/actions/route_create.rb
+++ b/app/actions/route_create.rb
@@ -35,8 +35,8 @@ module VCAP::CloudController
     rescue Sequel::ValidationFailed => e
       validation_error!(e, route.host, route.path, route.port, space, domain)
     rescue Sequel::UniqueConstraintViolation => e
-      logger.warn("error creating route #{e}, retrying once")
-      RouteCreate.new(user_audit_info).create(message:, space:, domain:, manifest_triggered:)
+      logger.warn("error creating route #{e}, retrying")
+      RouteCreate.new(@user_audit_info).create(message:, space:, domain:, manifest_triggered:)
     end
 
     private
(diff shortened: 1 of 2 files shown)


tcdowney (approved):
This looks good to me.

My question is, why are we even assigning VIPs to internal routes anymore? I thought that was something for Istio Copilot. 
🤔 https://www.pivotaltracker.com/story/show/161849240 

**Update:** Created https://github.com/cloudfoundry/cloud_controller_ng/issues/3712 to track that.

PR #3717, opened 2024-04-02
Validate manifest contains valid utf-8 data

kathap:
This PR introduces an additional check when the manifest gets parsed, that the parsed data (Strings) contain only valid utf-8 data.

Until now you could write non-valid utf-8 data into the manifest, e.g. as binary encoded url for the buildpack 
```
---
applications:
- name: test
  buildpacks: 
  - !!binary |-
      aHR0cHM6Ly9naXRodWIuY29tL2Nsb3VkZm91bmRyeS9uZ2lueC1idWlsZHBhY2suZ2l0ww==
```
(url here = https://github.com/cloudfoundry/nginx-buildpack.git�)
and there was no error thrown which lead to an PG::Error while the worker tried to save the error message 
```
"errors":[{"title":"UnknownError", "detail":"An unknown error occurred.", "code":10001, "test_mode_info":{"detail":"PG::Error: incomplete multibyte character\n", "title":"CF-DatabaseError"
```
With the utf-8 check for the manifest an appropriate error will be raised telling the user that he has invalid utf-8 data in his manifest.

### DIFF
--- a/app/controllers/v3/application_controller.rb
+++ b/app/controllers/v3/application_controller.rb
@@ -34,6 +34,22 @@ module V3ErrorsHelper
     raise CloudController::Errors::ApiError.new_from_details('BadRequest', message)
   end
 
+  def check_utf8_encoding!(yaml_data)
+    if yaml_data.is_a?(Hash)
+      yaml_data.each_value do |value|
+        check_utf8_encoding!(value)
+      end
+    elsif yaml_data.is_a?(Array)
+      yaml_data.each do |value|
+        check_utf8_encoding!(value)
+      end
+    else
+      return unless yaml_data.is_a?(String) && !yaml_data.force_encoding('UTF-8').valid_encoding?
+
+      message_parse_error!('Invalid UTF-8 encoding in YAML data')
+    end
+  end
+
   def message_parse_error!(message)
     raise CloudController::Errors::ApiError.new_from_details('MessageParseError', message)
   end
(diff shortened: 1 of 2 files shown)


app/controllers/v3/application_controller.rb
```diff
@@ -34,6 +34,31 @@ def bad_request!(message)
...
   end
 
+  def is_a_valid_utf_8_string?(item)
+    return false unless item.is_a?(String)
+    return false if item.force_encoding('UTF-8').valid_encoding?
+
+    message_parse_error!('Invalid UTF-8 encoding in YAML data')
+  end
+
+  def utf8_error!(yaml_data)
```
svkrieger, line 44: As recursive functions like this can run into performance issues, it would be good to test how long this takes if an app manifest with the maximum allowed size has been provided. If it takes too long it could be a potential DoS attack surface.

app/controllers/v3/application_controller.rb
```diff
@@ -34,6 +34,31 @@ def bad_request!(message)
...
+    return unless yaml_data.is_a?(Hash) || item.is_a?(Array)
```
svkrieger, line 45: `item` is never set at this point. I think you meant `yaml_data.is_a?(Array)`?

svkrieger (changes requested):
Looks good! Just a small thing which needs to be fixed and testing the performance of this with a huge manifest would be good.

app/controllers/v3/application_controller.rb
kathap: The yaml parsing (Psych.safe_load) takes ~ 0,5 seconds, the utf-8 check takes ~0,02 s for a manifest file with 1MB (~ 39200 lines)

philippthun (approved):

PR #3719, opened 2024-04-03
Mysql throws error when app name contains emoji

kathap:
In postgres there is no issue when you name your app e.g. 🦄🦄🦄, but the tests for mysql fail with:
```
Sequel::DatabaseError:
Mysql2::error: Incorrect string value: '\xF0\x9F\xA6\x84\xF0\x9F...' for column 'name' at row 1
```
Question: consistency -> app names should have the same options in mysql and postgres?

### DIFF
--- a/spec/request/space_manifests_spec.rb
+++ b/spec/request/space_manifests_spec.rb
@@ -218,6 +218,36 @@ RSpec.describe 'Space Manifests' do
       )
     end
 
+
+    context 'when the manifest contains valid binary-encoded URL(s) for the buildpack(s)' do
+      let(:app1_model) { VCAP::CloudController::AppModel.make(name: '🦄🦄🦄', space: space) }
+      let(:yml_manifest_with_binary_valid_buildpacks) do
+        "---
+            applications:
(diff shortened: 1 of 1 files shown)


FloThinksPi:
If emojies/non ASCII characters behave differently on MYSQL and PSQL one may consider a more generic approach, e.g. on sequel level, throw an exception that gets handled accordingly and the API returns a proper error. 
The reason why: non ASCII characters can be found in all entities(strings) customers can provide over the api. For example orgs names, space names, labels, annotations and so on. 
Also ive seen such resources like org names with emojies in production - also when introducing a exception one must be sure to not break workflows when e.g. an operations is conducted on an app(e.g. scale) that has a non compliant name already on psql. :)

philippthun:
Heads-up: #5121 merged, replacing machinist + Sham with FactoryBot. This PR will need a rebase and a one-line fixup:

- VCAP::CloudController::AppModel.make(name: '🦄🦄🦄', space: space) → create(:app_model, name: '🦄🦄🦄', space: space)

PR #3758, opened 2024-04-17
Fix backwards compatibility github actions

svkrieger:
###  A short explanation of the proposed change:
`bundle exec rake db:parallel:migrate` (which is also used in the backwards compatibility tests) failed since introducing new migration logging (6d2f0fa). The error was `DB_CONNECTION_STRING not set`. This is because when using parallel test execution `POSTGRES_CONNECTION_PREFIX` is used. `DB_CONNECTION_STRING` is set after `DbConfig.new` was called. So the new logging setup has been moved after this command.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -242,26 +242,32 @@ namespace :db do
   end
 
   def migrate
-    logging_output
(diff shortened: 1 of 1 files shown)


svkrieger:
Failing tests for postgres11 are unrelated.

PR #3778, opened 2024-05-02
[RFC0028] Implement Cloud Native Buildpacks lifecycle

pbusko:

Implementation of the [RFC-0028](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0028-cnb-lifecycle.md)

This PR adds third lifecycle option: `cnb`. This lifecycle type utilises Cloud Native Buildpacks to produce droplets.

https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0028-cnb-lifecycle.md


lib/cloud_controller/diego/lifecycles/app_cnb_lifecycle.rb
```diff
@@ -0,0 +1,62 @@
+module VCAP::CloudController
+  class AppCNBLifecycle
```
johha, line 2: Very similar to `AppBuildpackLifecycle`. Reuse possible?

lib/cloud_controller/diego/lifecycles/cnb_lifecycle.rb
```diff
@@ -0,0 +1,56 @@
+require 'cloud_controller/diego/lifecycles/buildpack_info'
+require 'cloud_controller/diego/lifecycles/buildpack_lifecycle_data_validator'
+require 'fetchers/buildpack_lifecycle_fetcher'
+
+module VCAP::CloudController
+  class CNBLifecycle
```
johha, line 6: Is it possible to reuse `BuildpackLifecycle`?

johha (changes requested):
Still need to go through the unit tests. So far my main concerns are:
- Do we need the `buildpack_url` in the cnb lifecycle model? Might be needed for revisions 
- Code duplication in diego part. I'd prefer to use inheritance as much as possible

I'll look into why the backward compatibility unit tests are failing for `rotate_database_key_spec.rb`. I assume because the table is generated after the test runs.

spec/unit/lib/cloud_controller/errands/rotate_database_key_spec.rb
```diff
@@ -36,6 +36,7 @@ module VCAP::CloudController
             'VCAP::CloudController::PackageModel' => PackageModel.make(:docker, package_hash: Sham.guid, error: 'a-error', docker_image: 'image', docker_username: 'user'),
             'VCAP::CloudController::DropletModel' => DropletModel.make(:all_fields),
             'VCAP::CloudController::BuildpackLifecycleDataModel' => BuildpackLifecycleDataModel.make(:all_fields),
+            'VCAP::CloudController::CNBLifecycleDataModel' => CNBLifecycleDataModel.make(:all_fields),
```
johha, line 39: Can be removed if `CNBLifecycleDataModel` does not need any encrypted fields (see other comment). This would also fix the backward compatibility tests.

johha (changes requested):

johha (changes requested):

johha:
Related to my previous comment about the code duplication in the diego part it might be worth it to check if the tests could be also simplified, e.g. by using [shared examples](https://rspec.info/features/3-12/rspec-core/example-groups/shared-examples/)

lib/utils/uri_utils.rb
```diff
@@ -34,4 +49,43 @@ def self.uri_escape(uri)
...
+      host = ''
+      path = docker_uri
+    end
+
+    path = 'library/' + path if (official_docker_registry(name_parts[0]) || missing_registry(name_parts)) && path.exclude?('/')
+
+    path, tag = parse_docker_repository_tag(path)
+
+    raise InvalidDockerURI.new "Invalid image name [#{path}]" unless %r{\A[a-z0-9_\-\.\/]{2,255}\Z} =~ path
+    raise InvalidDockerURI.new "Invalid image tag [#{tag}]" if tag && !(/\A[a-zA-Z0-9_\-\.]{1,128}(@sha256:[a-z0-9]{64})?\Z/ =~ tag)
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: We add tests. The idea is to validate that the docker url is a valid one.

app/actions/droplet_create.rb
```diff
@@ -54,7 +54,8 @@ def create_buildpack_droplet(build)
 
       DropletModel.db.transaction do
         droplet.save
-        droplet.buildpack_lifecycle_data = build.buildpack_lifecycle_data
+        droplet.buildpack_lifecycle_data = build.buildpack_lifecycle_data if build.buildpack_lifecycle?
+        droplet.cnb_lifecycle_data = build.cnb_lifecycle_data if build.cnb_lifecycle?
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: We rearranged it, so that if ` build.cnb_lifecycle?` is `false`, the behavior should be the same as before, Do you think that is good enough?

app/models/runtime/app_model.rb
```diff
@@ -2,6 +2,7 @@
 require 'cloud_controller/serializer'
 require 'models/helpers/process_types'
 require 'hashdiff'
+require 'models/runtime/cnb_lifecycle_data_model'
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: You are right, it works without.

lib/utils/uri_utils.rb
```diff
@@ -34,4 +49,43 @@ def self.uri_escape(uri)
...
+
+    [host, path, tag]
+  end
+
+  private_class_method def self.official_docker_registry(host)
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: The reason was `rubocop` :) `private` will only work within a class.

c0d1ngm0nk3y:
> Still need to go through the unit tests. So far my main concerns are:
> 
> * Do we need the `buildpack_url` in the cnb lifecycle model? Might be needed for revisions

As discussed in person. We don't need the `buildpack_url` to be encrypted, but we reuse the `buildpack_lifecycle_buildpacks` relation.

> * Code duplication in diego part. I'd prefer to use inheritance as much as possible

We removed the code duplication where we saw beneficial. Could you have another look.

> I'll look into why the backward compatibility unit tests are failing for `rotate_database_key_spec.rb`. I assume because the table is generated after the test runs.

The tests were green in the last commit, but after fixing a rubocop offense, the tests fail again. Any idea?

lib/cloud_controller/diego/cnb/staging_action_builder.rb
```diff
@@ -0,0 +1,208 @@
+require 'credhub/config_helpers'
+require 'diego/action_builder'
+require 'digest/xxhash'
+
+module VCAP::CloudController
+  module Diego
+    module CNB
+      class StagingActionBuilder
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: We decided against inheritance since there is actually very little code in common.

lib/cloud_controller/diego/cnb/staging_completion_handler.rb
```diff
@@ -0,0 +1,53 @@
+require 'cloud_controller/diego/staging_completion_handler'
+require 'utils/uri_utils'
+
+module VCAP::CloudController
+  module Diego
+    module CNB
+      class StagingCompletionHandler < VCAP::CloudController::Diego::StagingCompletionHandler
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: We decided against inheritance since there is actually very little code in common.

lib/cloud_controller/diego/lifecycles/cnb_lifecycle.rb
c0d1ngm0nk3y: done

c0d1ngm0nk3y:
@johha We think we have addressed all review comments. Please let us know if we missed something.

db/migrations/20240514113500_create_cnb_lifecycle_data.rb
```diff
@@ -0,0 +1,45 @@
...
+      drop_foreign_key [:cnb_lifecycle_data_guid]
+      # TODO: fails with undefined 'database_type' - does not work in alter_table?
+      # VCAP::Migration.with_concurrent_timeout(self) do
+      drop_index :cnb_lifecycle_data_guid, name: :bl_cnb_bldata_guid_index, concurrently: @database_type == :postgres
+      # end
+      drop_column :cnb_lifecycle_data_guid
+    end
+    drop_table :cnb_lifecycle_data
+  end
+end
```
johha, line 45: I modified the migration by adding transactions and `with_concurrent_timeout`. This ensures that the table is not in a inconsistent state in case of errors and prevents that the default statement timeout cancels the index creation: 
```suggestion
Sequel.migration do
  # adding an index concurrently cannot be done within a transaction
  no_transaction

  up do
    transaction do
      create_table(:cnb_lifecycle_data) do
        VCAP::Migration.common(self, :cnb_lifecycle_data)

        String :build_guid, size: 255
        foreign_key [:build_guid], :builds, key: :guid, name: :fk_cnb_lifecycle_build_guid, on_delete: :cascade
        index [:build_guid], name: :fk_cnb_lifecycle_build_guid_index

        String :app_guid, size: 255
        index [:app_guid], name: :fk_cnb_lifecycle_app_guid_index

        String :droplet_guid, size: 255
        index [:droplet_guid], name: :fk_cnb_lifecycle_droplet_guid_index

        String :stack, size: 255
      end

      add_column :buildpack_lifecycle_buildpacks, :cnb_lifecycle_data_guid, String, size: 255
      alter_table(:buildpack_lifecycle_buildpacks) do
        add_foreign_key [:cnb_lifecycle_data_guid], :cnb_lifecycle_data, key: :guid, name: :fk_blcnb_bldata_guid, on_delete: :cascade
      end
    end

    VCAP::Migration.with_concurrent_timeout(self) do
      add_index :buildpack_lifecycle_buildpacks, :cnb_lifecycle_data_guid, name: :bl_cnb_bldata_guid_index, concurrently: true if database_type == :postgres
    end
  end

  down do
    VCAP::Migration.with_concurrent_timeout(self) do
      drop_index :buildpack_lifecycle_buildpacks, :cnb_lifecycle_data_guid, name: :bl_cnb_bldata_guid_index, concurrently: true if database_type == :postgres
    end

    transaction do
      alter_table(:buildpack_lifecycle_buildpacks) do
        drop_foreign_key [:cnb_lifecycle_data_guid]
      end
      drop_column :buildpack_lifecycle_buildpacks, :cnb_lifecycle_data_guid
      drop_table :cnb_lifecycle_data
    end
  end
end

```

Let me know what you think of it!

johha (commented):
I modified the db migration for better robustness

db/migrations/20240514113500_create_cnb_lifecycle_data.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    transaction do
+      alter_table(:buildpack_lifecycle_buildpacks) do
+        drop_foreign_key [:cnb_lifecycle_data_guid]
+      end
+      drop_column :buildpack_lifecycle_buildpacks, :cnb_lifecycle_data_guid
+      drop_table :cnb_lifecycle_data
+    end
+  end
+end
```
johha, line 47: To safeguard the migration further we should add `if_exists` and `if_not_exists`. This will help in case the migration needs to be re-triggered e.g. due to a db failure. 

```suggestion
Sequel.migration do
  # adding an index concurrently cannot be done within a transaction
  no_transaction

  up do
    transaction do
      create_table(:cnb_lifecycle_data) do
        VCAP::Migration.common(self, :cnb_lifecycle_data)

        String :build_guid, size: 255
        foreign_key [:build_guid], :builds, key: :guid, name: :fk_cnb_lifecycle_build_guid, on_delete: :cascade
        add_index [:build_guid], name: :fk_cnb_lifecycle_build_guid_index, if_not_exists: true

        String :app_guid, size: 255
        add_index [:app_guid], name: :fk_cnb_lifecycle_app_guid_index, if_not_exists: true

        String :droplet_guid, size: 255
        add_index [:droplet_guid], name: :fk_cnb_lifecycle_droplet_guid_index, if_not_exists: true

        String :stack, size: 255
      end

      add_column :buildpack_lifecycle_buildpacks, :cnb_lifecycle_data_guid, String, size: 255, if_not_exists: true
      alter_table(:buildpack_lifecycle_buildpacks) do
        add_foreign_key [:cnb_lifecycle_data_guid], :cnb_lifecycle_data, key: :guid, name: :fk_blcnb_bldata_guid, on_delete: :cascade, if_not_exists: true
      end
    end

    VCAP::Migration.with_concurrent_timeout(self) do
      add_index :buildpack_lifecycle_buildpacks, :cnb_lifecycle_data_guid, name: :bl_cnb_bldata_guid_index, concurrently: true, if_not_exists: true if database_type == :postgres
    end
  end

  down do
    VCAP::Migration.with_concurrent_timeout(self) do
      drop_index :buildpack_lifecycle_buildpacks, :cnb_lifecycle_data_guid, name: :bl_cnb_bldata_guid_index, concurrently: true, if_exists: true if database_type == :postgres
    end

    transaction do
      alter_table(:buildpack_lifecycle_buildpacks) do
        drop_foreign_key [:cnb_lifecycle_data_guid], if_exists: true
      end
      drop_column :buildpack_lifecycle_buildpacks, :cnb_lifecycle_data_guid, if_exists: true
      drop_table :cnb_lifecycle_data
    end
  end
end

```

spec/unit/lib/cloud_controller/errands/rotate_database_key_spec.rb
johha: This should be still removed as as `CNBLifecycleDataModel` doesn't have any encrypted field.

johha (changes requested):
small migration optimization

c0d1ngm0nk3y:
@johha I had to do small changes to your suggestion of the migration file. Could you have a look here at [da18b18ee4e46888](https://github.com/cloudfoundry/cloud_controller_ng/pull/3778/commits/da18b18ee4e46888ac23007f267bd28e39118855)? I think it still fits what you intended to change, right?

c0d1ngm0nk3y:
```
PG::UndefinedTable: ERROR:  relation \"schema_info\" does not exist
```

@johha Any idea why the tests fail again?

c0d1ngm0nk3y:
@johha I did more testing of the `revisions` feature and the app behaves the same with both lifecycles. There were some glitches, but those could be seen for both lifecycles. So the expectation that it should be the same was not wrong after all. It just another way in producing the `droplet`. Hence I have enabled revisions again with https://github.com/cloudfoundry/cloud_controller_ng/pull/3778/commits/dc01106106f75424143d13d02cb9d28aa39f6c7b.

db/migrations/20240514113500_create_cnb_lifecycle_data.rb
```diff
@@ -38,10 +37,10 @@
 
     transaction do
       alter_table(:buildpack_lifecycle_buildpacks) do
-        drop_foreign_key [:cnb_lifecycle_data_guid]
+        drop_foreign_key [:cnb_lifecycle_data_guid] if @db.foreign_key_list(:buildpack_lifecycle_buildpacks).any? { |fk| fk[:name] == :fk_blcnb_bldata_guid }
```
c0d1ngm0nk3y, line 40: @johha But isn't that the same as for `add_foreign_key` (line 22)? Just in a different order? Should/could we do it in the same way (if -> alter_table).

johha: rubocop wanted it that way...

c0d1ngm0nk3y: Good argument :)

johha: I'll change it, rubocop actually complained about something else

c0d1ngm0nk3y:
> All checks have passed

👍

johha (approved):
lgtm 👍 🤞

c0d1ngm0nk3y:
`acceptance` were successful with this version deployed 👍

Issue #3687, opened 2024-03-18
RouteDestinationUpdate does not trigger a DesiredLRPUpdate call


geofffranks:
Thanks for submitting an issue to `cloud_controller_ng`. We are always trying to improve! To help us, please fill out the following template.

## Issue

The `cf update-destination` command requires apps to restart for the new destination information to take effect, unlike other route related changes for an app/route/backend.

## Context

[RouteDestinationUpdate.update()](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/actions/route_destination_update.rb) does not trigger DesiredLRPUpdate calls. It looks like this is handled for [UpdateRouteDestinations.update()](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/actions/update_route_destinations.rb#L94) via an update_process method, telling the bbs client to emit a change event.

## Steps to Reproduce

1. Push [cf-acceptance-test's http2 app](https://github.com/cloudfoundry/cf-acceptance-tests/tree/main/assets/http2) via `cf push potato2`
2. `curl https://potato2.<app-domain>`
3. See that the output complains that HTTP2 must be used.
4. `cf update-destination potato2 <app-domain> -n potato2 --app-protocol http2`
5. Wait 60 seconds
6. `curl https://potato2.<app-domain>`
7. See that the output still complains that HTTP2 must be used.
8. `cf restart potato2`
9. `curl https://potato2.<app-domain>`
10. See that the output finally says 'Hello' and 'Protocol: HTTP/2.0'

## Expected result

Step 7 should see that the output says 'Hello' and 'Protocol: HTTP/2.0'

## Current result

Step 7 still shows that HTTP2 is not being used

## Possible Fix

Leverage the [update_process logic ](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/actions/update_route_destinations.rb#L94) from the UpdateRouteDestinations action to issue a DesiredLRPUpdate event to BBS from inside the RouteDestinationUpdate action.

PR #3787, opened 2024-05-08
Updating a route destination protocol informs backend

tcdowney:

Updates the RouteDestinationUpdate action to behave like other route update endpoints and notifiy Diego immediately

Prior to this change users would need to manually restart their app to get a route destination protocol change to take effect.

Fixes https://github.com/cloudfoundry/cloud_controller_ng/issues/3687

### DIFF
--- a/app/actions/route_destination_update.rb
+++ b/app/actions/route_destination_update.rb
@@ -15,9 +15,8 @@ module VCAP::CloudController
           destination.save
         end
 
-        runners = CloudController::DependencyLocator.instance.runners
         destination.processes.each do |process|
-          notify_backend_of_route_update(process, runners)
+          ProcessRouteHandler.new(process).notify_backend_of_route_update
         end
 
         destination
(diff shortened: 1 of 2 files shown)


app/actions/route_destination_update.rb
```diff
@@ -14,17 +14,34 @@ def update(destination, message)
...
+        destination.processes.each do |process|
+          notify_backend_of_route_update(process, runners)
+        end
+
+        destination
       end
 
       private
 
+      def notify_backend_of_route_update(process, runners)
```
philippthun, line 28: The implementation looks similar to `ProcessRouteHandler`. What about reusing this handler?

tcdowney: Are you suggesting reusing the `ProcessRouteHandler` entirely? Or just its `notify_backend_of_route_update` method? I don't think `ProcessRouteHandler` itself should be invoked since there is actually no work to do on the processes themselves, but I could look at modifying the `notify_backend_of_route_update` function to support both. It would just need refactored because it relies a lot on instance variable state.

philippthun: I was actually thinking about simply reusing the `ProcessRouteHandler` as done [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/6d325497d5c1e1f6604456f970ce7a7bb9f8e392/app/models/runtime/route.rb#L257) and [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/6d325497d5c1e1f6604456f970ce7a7bb9f8e392/app/models/services/route_binding.rb#L28).

tcdowney: Makes sense, I don't know why I thought `notify_backend_of_route_update` was a private method in there.

geofffranks (approved):

app/actions/route_destination_update.rb
tcdowney: Much cleaner, thanks for the suggestion @philippthun !
https://github.com/cloudfoundry/cloud_controller_ng/pull/3787/commits/13f67275814d5d187fcaac118e8c815cc5e6e377

xandroc (approved):

sethboyles (approved):

Issue #2587, opened 2021-12-09
 Applications deployed with `strategy --rolling` do not produce crash logs.


dzaslavskiy:
## Issue

Applications deployed with `strategy --rolling` do not produce crash logs.

## Context

Normally, when an application crashes it produces a log with a message detailing the crash:
```
App instance exited with guid f94d3889-3229-4045-befd-af282865dc2a payload: {"instance"=>"1016193b-39d2-42dc-69c8-27d9", "index"=>0, "cell_id"=>"b701ee4f-885a-4c07-80c1-036e3d494e0a", "reason"=>"CRASHED", "exit_description"=>"APP/PROC/WEB: Exited with status 1", "crash_count"=>1, "crash_timestamp"=>1638762864524964378, "version"=>"0d53cd6f-6270-49e9-8ba8-1acd5d9efb58"}
```
However, if the application is deployed with `strategy --rolling` and then redeployed. It will not produce these log messages.

There is another log message that is associated with crashes. This message is produced for both types of deployment strategies but it does not contain all of the information of the above message.
```
Process has crashed with type: "web"
```

## Steps to Reproduce

Deploy application with `strategy --rolling` and then redeploy it again. Cause app to crash. It will not produce above log message

## Expected result

The application produces above log message regardless of deployment strategy.

## Current result

The application does not produce the above log.

## Possible Fix

This may be related to fact that with  `strategy --rolling`  the process GUID changes in differant way then with normal deployments.

tcdowney:
Interesting, good find.

> This may be related to fact that with strategy --rolling the process GUID changes in differant way then with normal deployments.

I think your hunch is correct here. To make the v2->v3 migration easier the `web` process starts out with a GUID that is identical to its parent app's GUID. This allows it to be used almost interchangeably. Rolling deployments create new `web` processes so that causes the GUIDs to drift.

I think this also means crash events are broken for multi-process apps. We are sending the process GUID all the way through from [tps-watcher](https://github.com/cloudfoundry/tps/blob/fde3272c264a3e6538e8b289b7a93af450ff77a8/watcher/watcher.go#L136) to Cloud Controller, but at the point the event is created it is considering it to be an ["app" and not a process](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/controllers/internal/app_crashed_controller.rb#L21). I wonder if we could fix things by changing this call to actually pass through the app instead of a process.

PR #3790, opened 2024-05-09
Fix app crash events for apps using rolling deployments

tcdowney:
This change updates the endpoint to make app crash events for the app guid instead of the process guid which should resolve the rolling deployment issue and support crash events for non-web process types.

The internal app crash event endpoint assumed that the web process guid == app guid (legacy behavior from the v2->v3 data model migrations). Rolling deployments create new web processes that have different guids which made it so this endpoint no longer created events associated with the parent app.

Issue: https://github.com/cloudfoundry/cloud_controller_ng/issues/2587

### DIFF
--- a/spec/unit/controllers/internal/app_crashed_controller_spec.rb
+++ b/spec/unit/controllers/internal/app_crashed_controller_spec.rb
@@ -5,7 +5,8 @@ require 'spec_helper'
 module VCAP::CloudController
   RSpec.describe AppCrashedController do
     describe 'POST /internal/v4/apps/:process_guid/crashed' do
-      let(:diego_process) { ProcessModelFactory.make(state: 'STARTED', diego: true) }
+      let(:diego_process) { ProcessModelFactory.make(state: 'STARTED', diego: true, type: 'rolled-web') }
+      let(:app_model) { diego_process.app }
       let(:process_guid) { Diego::ProcessGuid.from(diego_process.guid, 'some-version-guid') }
       let(:url) { "/internal/v4/apps/#{process_guid}/crashed" }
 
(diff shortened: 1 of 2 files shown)


spec/unit/controllers/internal/app_crashed_controller_spec.rb
```diff
@@ -35,13 +35,14 @@ module VCAP::CloudController
         expect(last_response.status).to eq(200)
         expect(last_response.body).to eq '{}'
 
-        app_event = Event.find(actee: diego_process.guid, actor_type: 'app')
+        app_guid = diego_process.app.guid
+        app_event = Event.find(actee: app_guid, actor_type: 'app')
 
         expect(app_event).to be
         expect(app_event.space).to eq(diego_process.space)
```
philippthun, line 42: It's only a very minor thing; as this is an **app** event, it would be more consistent to expect that the `app_event.space` equals `diego_process.app.space`. I know, it's the same, but somewhat easier to understand from my point of view.

Maybe you define a `let(:app) { diego_process.app }` and then use `app.guid` and `app.space` in the tests.

tcdowney: Sounds good. Named the `let` slightly differently since `app` shadows something the controller specs use to implement `post`, `get`, etc.

https://github.com/cloudfoundry/cloud_controller_ng/commit/a90ad04913fc3fc224b83fcd6d6d23c51746078e#diff-124ca0f3421fb8907f4ce42b34e471d2ac64ede838ce5d1b16f93bfaebd31323R9

philippthun (approved):

PR #3808, opened 2024-05-21
Use DISTINCT only when UNIONing multiple datasets

philippthun:
When `UNION`ing multiple datasets, invoke _abstract_ method `distinct_union` which is implemented differently by the `ServiceOfferingListFetcher` and the `ServicePlanListFetcher`. For fetched service plans we already know that there is no overlap between the subqueries; for fetched services there can be duplicates and thus a `DISTINCT` has to be added to the query.


johha (approved):
lgtm 👍

PR #3829, opened 2024-06-04
Do not allow ASGs with leading zeros

jrussett:

This change adds validation logic to prevent the creation of ASGs with leading zeros in addresses. e.g.
- `10.0.0.0/8` and `192.168.1.156` are valid 
- `010.000.000.0/0` `10.0.000.10` are no longer valid.

The motivation for this change in behavior is to prevent malformed IPs from making their way into the CC database to begin with. We only want to accept valid addresses in ASG destinations going forward.

Cloud controller blindly stores the ASG destinations in the database and then passes the value off to downstream clients. Not all downstream clients have the same IP parsing functionality that ruby does, therefore jobs like `vxlan-policy-agent` in silk can stop working when a destination with leading zeros is bound via `cf bind-security-group`; when these problems happen, they tend to also take down the diego cell too by preventing routing to all running containers.

### Links to any other associated PRs, etc...

- [#187494337](https://www.pivotaltracker.com/story/show/187494337)
- https://github.com/cloudfoundry/silk-release/pull/130

### Checkboxes

### DIFF
--- a/lib/cloud_controller/rule_validator.rb
+++ b/lib/cloud_controller/rule_validator.rb
@@ -66,7 +66,9 @@ module CloudController
 
     def self.parse_ip(val)
       if val.is_a?(Array)
-        val.map { |ip| NetAddr::IPv4.parse(ip) }
+        val.map do |ip|
+          NetAddr::IPv4.parse(ip)
+        end
       else
         NetAddr::IPv4Net.parse(val)
       end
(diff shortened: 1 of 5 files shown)


ameowlia:
Adding acceptance steps here

1. Save this asg to a file called asg.json
```
[
  {
    "protocol": "tcp",
    "destination": "010.0.0.53",
    "ports": "443"
  },
  {
    "protocol": "tcp",
    "destination": "10.000.0.53",
    "ports": "443"
  },
  {
    "protocol": "tcp",
    "destination": "10.0.000.53",
    "ports": "443"
  },
  {
    "protocol": "tcp",
    "destination": "10.0.0.053",
    "ports": "443"
  },
  {
    "protocol": "tcp",
    "destination": "010.000.000.053",
    "ports": "443"
  }
]
```
2. Run `cf create-security-group asg asg.json`
3. See the following failure.
```
Rules[0]: destination octets cannot contain leading zeros, Rules[1]: destination octets cannot contain leading zeros, Rules[2]: destination octets cannot contain leading zeros, Rules[3]: destination octets cannot contain leading zeros, Rules[4]: destination octets cannot contain leading zeros
FAILED
```

Samze (approved):
LGTM and ran through acceptance

lib/cloud_controller/rule_validator.rb
```diff
@@ -78,5 +80,29 @@ def self.comma_delimited_destinations_enabled?
       config = VCAP::CloudController::Config.config
       config.get(:security_groups, :enable_comma_delimited_destinations)
     end
+
+    def self.no_leading_zeros(destination)
+      return no_leading_zeros_in_address(destination) unless destination.is_a?(Array)
+
+      no_zeros = true
+      destination.each do |address|
+        no_zeros &&= no_leading_zeros_in_address(address)
```
sethboyles, line 89: 🤯 

In the future you might prefer something a bit more rubyist approved:

```
destination.all? { |address|  no_leading_zeros_in_address(address) }
```

PR #3839, opened 2024-06-12
JSON encode values in SCIM filter string

philippthun:
When building the SCIM filter string, all values (e.g. user ids, names) are now JSON encoded (*), i.e. `#{Oj.dump(str)}` instead of `"#{str}"`.

Furthermore when receiving a `TargetError`, the 'error_info' (i.e. JSON response from UAA containing error details) is now logged.

(*) Excerpt from RFC7644:
```
SCIM filters MUST conform to the following ABNF [...] rules as specified below:

[...]

compValue = false / null / true / number / string; rules from JSON (RFC 7159)

[...]

In the above ABNF rules, the "compValue" (comparison value) rule is built on JSON Data Interchange format ABNF rules as specified in [RFC7159] [...]
```


johha (approved):
lgtm 👍

Issue #1312, opened 2019-03-19
Killing `cloud_controller_ng` can cause counter metrics like `requests.outstanding` to get stuck


tcdowney:
## Issue
Logging this a Github issue for discoverability.

The  `cc.requests.outstanding` counter metric can get stuck if Cloud Controller is stopped/killed in the middle of handling requests.

## Context

Cloud Controller emits the `cc.requests.outstanding` to give operators insight into the number of active requests being handled by the `cloud_controller_ng` Ruby process.

> Number of requests that are currently being processed. Emitted for each Cloud Controller request.
More information on emitted metrics: https://docs.cloudfoundry.org/running/all_metrics.html

Cloud Controller manages this metric in https://github.com/cloudfoundry/cloud_controller_ng/blob/90bbddabf983a353297a11aa8a2bcc1f98b81503/middleware/request_metrics.rb#L14.

So if the Cloud Controller is stopped/killed before it has a chance to call `complete_request` [here](https://github.com/cloudfoundry/cloud_controller_ng/blob/90bbddabf983a353297a11aa8a2bcc1f98b81503/lib/cloud_controller/metrics/request_metrics.rb#L14), none of the active requests can decrement the counter in statsd. This is similar to #1294 .

## Steps to Reproduce

1. Install the [CF Loggregator Firehose Plugin](https://docs.cloudfoundry.org/loggregator/cli-plugin.html)

2. Connect to the firehose and grep out the metric:
```
cf nozzle -f ValueMetric | grep --line-buffered cc.*outstanding
```

3. Hammer the Cloud Controllers with requests:
```
hey -n 1000000 -c 50 -H "Authorization: $(cf oauth-token)" http://api.arya.capi.land/v2/apps
```

4. Observe something similar to the following from `cf nozzle`
```
origin:"cc" eventType:ValueMetric timestamp:1553034399010006107 deployment:"cf" job:"api" index:"c032ab53-66f7-45fc-a229-d25d509967dc" ip:"10.0.1.11" tags:<key:"source_id" value:"cc" > valueMetric:<name:"requests.outstanding" value:19 unit:"counter" >
origin:"cc" eventType:ValueMetric timestamp:1553034399011308663 deployment:"cf" job:"api" index:"c032ab53-66f7-45fc-a229-d25d509967dc" ip:"10.0.1.11" tags:<key:"source_id" value:"cc" > valueMetric:<name:"requests.outstanding" value:20 unit:"counter" >
[...]

PR #3841, opened 2024-06-12
Fix cc.requests.outstanding.gauge when using puma web server

Samze:
https://github.com/cloudfoundry/cloud_controller_ng/issues/1312 introduced `cc.requests.outstanding.gauge` which holds the counter in memory. With the introduction of puma there may be multiple processes, so each would emit its own value for this metric. This would cause the gauge to flop between values. This metric is listed as an important kpi for capi scaling https://docs.cloudfoundry.org/running/managing-cf/scaling-cloud-controller.html#cloud_controller_ng. 

This fix for puma will instead uses Redis for the gauge.

* Refactor requests_metric to use statsd_updater
* Statsd_updater now contains statsd requests logic
* Add Redis/Inmemory store to statsd_updager
* Add a missing mutex to counter for thread safety for the in memory implementation.
* Chose to prefix the entry in redis with `metrics:` but open to feedback here.

Inspired by https://github.com/cloudfoundry/cloud_controller_ng/commit/4539e596ab6ae64556b170e4387633e9ebd55292

An alternative considered, was to read the prometheus metric and re-emit that to StatsD, however we observed performance degradation. Presumably because of the number of reads from disk for the [DirectFileStorage](https://github.com/prometheus/client_ruby?tab=readme-ov-file#directfilestore-caveats-and-things-to-keep-in-mind) to aggregate the metric across processes and so Redis seemed the best approach.

cc @sethboyles / @pivotalgeorge

---

### DIFF
--- a/lib/cloud_controller/metrics/request_metrics.rb
+++ b/lib/cloud_controller/metrics/request_metrics.rb
@@ -3,33 +3,21 @@ require 'statsd'
 module VCAP::CloudController
   module Metrics
     class RequestMetrics
-      def initialize(statsd=CloudController::DependencyLocator.instance.statsd_client, prometheus_updater=CloudController::DependencyLocator.instance.prometheus_updater)
-        @counter = 0
-        @statsd = statsd
+      def initialize(statsd_updater=CloudController::DependencyLocator.instance.statsd_updater, prometheus_updater=CloudController::DependencyLocator.instance.prometheus_updater)
+        @statsd_updater = statsd_updater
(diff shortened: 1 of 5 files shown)


sethboyles:
Code looks good, I'll do acceptance tomorrow

joaopapereira (approved):
LGTM, my only concern here would be that this store can only save one metric, but looking at the statsd_updater it looks like we only have 1 type of metric so it should be fine.

PR #3848, opened 2024-06-17
Add delete buildpack cache per app endpoint

georgessap:


app/controllers/v3/apps_controller.rb
```diff
@@ -225,6 +225,19 @@ def restart
...
 
+  def clear_buildpack_cache
+    app, space = AppFetcher.new.fetch(hashed_params[:guid])
+    app_not_found! unless app && permission_queryer.can_read_from_space?(space.id, space.organization_id)
+    unauthorized! unless permission_queryer.can_manage_apps_in_active_space?(space.id)
+    suspended! unless permission_queryer.is_space_active?(space.id)
+
+    delete_job = Jobs::V3::BuildpackCacheDelete.new(app.guid)
+    Jobs::Enqueuer.new(delete_job, queue: Jobs::Queues.generic).enqueue
+    VCAP::AppLogEmitter.emit(app.guid, "Enqueued job to delete app with guid #{app.guid}")
```
svkrieger, line 236: The message should state that a job has been enqueued, which will delete the app's buildpack cache, not the app itself.

app/controllers/v3/apps_controller.rb
```diff
@@ -225,6 +225,19 @@ def restart
...
+
+    head :ok, 'Location' => url_builder.build_url(path: "/v3/jobs/#{app.guid}")
```
svkrieger, line 238: 202 should be returned, as we create an async job, which will delete the cache.

spec/unit/controllers/v3/apps_controller_spec.rb
```diff
@@ -1489,6 +1489,81 @@
...
+
+    before do
+      set_current_user(user)
+      allow_user_read_access_for(user, spaces: [space])
+      allow_user_write_access(user, space:)
+      VCAP::CloudController::BuildpackLifecycleDataModel.make(app: app_model, buildpacks: nil, stack: VCAP::CloudController::Stack.default.name)
+    end
+
+    it 'returns a 200 and the app' do
+      put :clear_buildpack_cache, params: { guid: app_model.guid }, as: :json
```
svkrieger, line 1507: I wonder why `put` is used here and also in other tests e.g. for :start, :restart and so on. Shouldn't we use `post`?

svkrieger (changes requested):
Looks good to me in general. Just found a few minor things, which need to be changed.

docs/v3/source/includes/resources/apps/_clear_buildpack_cache.md.erb
```diff
@@ -0,0 +1,36 @@
...
+```
+Example Response
+```
+
+```http
+HTTP/1.1 202 Accepted
+Location: https://api.example.org/v3/jobs/[guid]
+```
+
+This endpoint will delete all of the existing buildpack caches for a specified app.
```
stephanme, line 22: This endpoint will delete the buildpack cache for a specified app.
(there only one cache per app)

docs/v3/source/includes/resources/apps/_clear_buildpack_cache.md.erb
```diff
@@ -0,0 +1,36 @@
...
+The buildpack cache is used during staging by buildpacks as a way to
+cache certain resources, e.g. downloaded Ruby gems. A user who wants to
+decrease the size of their blobstore could use this endpoint to delete
+unnecessary blobs.
```
stephanme, line 26: The last sentence makes sense only for the global cache deletion.
The main reason for deleting a buildpack cache is that an app doesn't stage anymore due to out-of-disk caused by a large buildpack cache content. Often the real RC is that the buildpack didn't clean up e.g. old runtime versions that are not used/needed anymore.

spec/unit/controllers/v3/apps_controller_spec.rb
```diff
@@ -1489,6 +1489,81 @@
...
+
+        it 'returns a 404 ResourceNotFound error' do
+          post :clear_buildpack_cache, params: { guid: app_model.guid }, as: :json
+
+          expect(response).to have_http_status :not_found
+          expect(response.body).to include 'ResourceNotFound'
+        end
+      end
+
+      context 'when the user can read but cannot write to the app' do
```
stephanme, line 1551: Isn't there a way to test the CF roles directly?
SpaceDeveloper+SpaceSupport can clear the cache. Other roles (or no roles) can't.
Would be a perfect documentation as well for the intended authorisation.

Disclaimer: I'm not an expert in the capi unit tests. Maybe this is the standard way how to test it and then it is good.

stephanme (changes requested):

spec/unit/controllers/v3/apps_controller_spec.rb
georgessap: I have based these tests off the `start`, `stop` and `restart` method tests. So tried to keep them consistent across these controller actions.

georgessap: Something like this you @stephanme ?

https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/unit/controllers/v3/apps_controller_spec.rb#L1307

app/controllers/v3/apps_controller.rb
```diff
@@ -225,6 +225,19 @@ def restart
...
+    app, space = AppFetcher.new.fetch(hashed_params[:guid])
+    app_not_found! unless app && permission_queryer.can_read_from_space?(space.id, space.organization_id)
+    unauthorized! unless permission_queryer.can_delete_buildpack_cache?(space.id)
+    suspended! unless permission_queryer.is_space_active?(space.id)
+
+    delete_job = Jobs::V3::BuildpackCacheDelete.new(app.guid)
+    Jobs::Enqueuer.new(delete_job, queue: Jobs::Queues.generic).enqueue
+    VCAP::AppLogEmitter.emit(app.guid, "Enqueued job to delete app buildpack cache with guid #{app.guid}")
+
+    head HTTP::ACCEPTED, 'Location' => url_builder.build_url(path: "/v3/jobs/#{app.guid}")
```
svkrieger, line 238: Here the jobs guid needs to be returned, not the app guid

app/controllers/v3/apps_controller.rb
```diff
@@ -225,6 +225,19 @@ def restart
...
 
+  def clear_buildpack_cache
+    app, space = AppFetcher.new.fetch(hashed_params[:guid])
+    app_not_found! unless app && permission_queryer.can_read_from_space?(space.id, space.organization_id)
+    unauthorized! unless permission_queryer.can_delete_buildpack_cache?(space.id)
+    suspended! unless permission_queryer.is_space_active?(space.id)
+
+    delete_job = Jobs::V3::BuildpackCacheDelete.new(app.guid)
+    Jobs::Enqueuer.new(delete_job, queue: Jobs::Queues.generic).enqueue
+    VCAP::AppLogEmitter.emit(app.guid, "Enqueued job to delete app buildpack cache with guid #{app.guid}")
```
svkrieger, line 236: I would rephrase this to make clear that the guid is the app guid and not the guid of the buildpack cache.
Maybe `Enqueued job to delete buildpack cache of app with app guid #{app.guid}`.

spec/unit/controllers/v3/apps_controller_spec.rb
```diff
@@ -1489,6 +1489,109 @@
...
+    let(:user) { VCAP::CloudController::User.make }
+
+    before do
+      set_current_user(user)
+      allow_user_read_access_for(user, spaces: [space])
+      allow_user_write_access(user, space:)
+      VCAP::CloudController::BuildpackLifecycleDataModel.make(app: app_model, buildpacks: nil, stack: VCAP::CloudController::Stack.default.name)
+    end
+
+    it 'returns a 202 and the app' do
```
svkrieger, line 1506: the app is not returned here, but the location of the async deletion job. Expected result needs to be adjusted as well.

svkrieger (changes requested):

lib/cloud_controller/permissions.rb
```diff
@@ -107,6 +107,10 @@ def readable_org_guids_query
     membership.authorized_org_guids_subquery(ROLES_FOR_ORG_READING)
   end
 
+  def can_delete_buildpack_cache?(space_id)
+    roles.admin? || membership.role_applies?(ROLES_FOR_SPACE_SERVICES_READING, space_id)
```
tcdowney, line 111: Clearing the cache isn't a "READ" action. I think if we want `SPACE_DEVELOPER` and `SPACE_SUPPORTER` to be able to do this then the `ROLES_FOR_APP_MANAGING` constant is a bit more correct.

app/jobs/v3/buildpack_cache_delete.rb
```diff
@@ -2,10 +2,12 @@ module VCAP::CloudController
   module Jobs
     module V3
       class BuildpackCacheDelete < VCAP::CloudController::Jobs::CCJob
-        attr_accessor :app_guid
+        attr_accessor :app_guid, :resource_guid, :resource_type, :model_class
 
-        def initialize(app_guid)
+        def initialize(model_class, app_guid, resource_type=nil)
+          @model_class = model_class
```
tcdowney, line 8: What are `model_class` and `resource_type` used for here?

tcdowney (changes requested):

app/jobs/v3/buildpack_cache_delete.rb
kathap: Indeed, model_class is not necessary. resource_guid is expected by pollable_job_wrapper. But then we don't need app_guid. Will change app_guid to resource_guid.

lib/cloud_controller/permissions.rb
kathap: That is indeed true

tcdowney (approved):
LGTM

stephanme (approved):

PR #3854, opened 2024-06-19
fix: CPU Entitlement is zero when stats are missing

acrmp:
- In #3641 we exposed CPU Entitlement in process stats and returned null when stats are not yet available in Log Cache.
- Change missing process stats to instead have CPU Entitlement of zero for consistency with other existing metrics. Recent versions of the cf CLI will show zero rather than an empty column for CPU Entitlement.
- CPU Entitlement may still be null when the metric is not available on the deployment.


acrmp:
We think the GitHub action failures are unrelated to our changes, let us know if that's not the case.

philippthun (approved):

PR #3874, opened 2024-07-05
Move V2 API docs to Github Pages

pivotalgeorge:

Serve the V2 API docs as static GitHub Pages assets, deployed via GitHub Actions

As part of the VMware->Broadcom migration, the infrastructure hosting the current v2 apidocs is being replaced. This provides an opportunity to sunset unneeded apps, lowering the project's maintenance & support burden.

  - this PR makes no changes to CAPI behavior


moleske (commented):
Will [capi-ci](https://github.com/cloudfoundry/capi-ci/blob/c50ebee1b6faf66c1e87cbce5927c419151c9b63/ci/pipeline.yml#L847) need to change as well?  Also it seems like maybe this should trigger on capi releases?  It looks like it currently runs on pushes to `main` and `gh-pages`, along with manual workflows, but not on new tags/releases

moleske (commented):
Ah just saw that v3 has github pages now, so mimicking that.  This seems reasonable to move over as well.

Or, ya know, just end v2 support 😛

pivotalgeorge:
Yep, v3 is already deploying to GitHub Pages, and it seems like a switch to GitHub Actions is [in-progress](https://github.com/cloudfoundry/cloud_controller_ng/actions/workflows/deploy_v3_docs.yml). The corresponding change to capi-ci would be to remove doc-publishing behavior.

As far as triggering on tags/releases, that sounds good, but this repo (CCNG) doesn't use tags. I'm assuming you mean triggering on releases in [capi-release](https://github.com/cloudfoundry/capi-release), right?

sethboyles:
> As far as triggering on tags/releases, that sounds good, but this repo (CCNG) doesn't use tags. I'm assuming you mean triggering on releases in [capi-release](https://github.com/cloudfoundry/capi-release), right?

I think that's what we'd have to do.  Is that even possible?

evanfarrar:
> > As far as triggering on tags/releases, that sounds good, but this repo (CCNG) doesn't use tags. I'm assuming you mean triggering on releases in [capi-release](https://github.com/cloudfoundry/capi-release), right?
> 
> I think that's what we'd have to do. Is that even possible?

maybe the github actions should run in capi-release? we tag /release in capi release

moleske:
Could tag this repo with the version on the commit that is included in the capi release version.  Then wouldn't have to move the docs.  Also fine with moving the docs and action to capi release so only one place has to be tagged.

Might also be a way to trigger actions across repos but that seems unnecessarily complicated.

I was mostly asking cause I noticed trigger was just on main/gh-pages.  Which may be just fine as long as new versions are created properly.  I'm out of practice on how the old docs did all this cause "it just worked."  So triggering off of tags may not be needed if when a new version is created the pipeline just makes the new version, e.g. on commits like [this](https://github.com/cloudfoundry/cloud_controller_ng/commit/8446149ee6f549bc2f5e68718f4a9e4a4a6025e7), or [this](https://github.com/cloudfoundry/cloud_controller_ng/commit/bea81821077db45deec0d9890641e8fcb9323ad2), or [this](https://github.com/cloudfoundry/cloud_controller_ng/commit/fb16f6f575031154fc781dc7a24a4d8adf3fc67e).  And doc changes to main that aren't new versions just update the release candidate version.

I don't have strong feelings was just curious what the plan is as it is clear vmware/broadcom from George's message is uninterested in hosting the existing docs site (which I'm assuming includes hosting v3 in addition to the v2 George mentions in this pr).  Also don't think any of the above needs to block merging this pr since it has manual triggers and all that for folks to test out first before letting the bots run it

philippthun:
> Yep, v3 is already deploying to GitHub Pages, and it seems like a switch to GitHub Actions is [in-progress](https://github.com/cloudfoundry/cloud_controller_ng/actions/workflows/deploy_v3_docs.yml). The corresponding change to capi-ci would be to remove doc-publishing behavior.

We did not change the v3 docs build process, we only moved the script over to [capi-ci](https://github.com/cloudfoundry/capi-ci/blob/main/ci/docs/build_docs_v3.sh) - and did some improvements like getting rid of git ssh. So both, [release candidate](https://concourse.app-runtime-interfaces.ci.cloudfoundry.org/teams/capi-team/pipelines/capi/jobs/rc-docs-v3) and [final docs](https://concourse.app-runtime-interfaces.ci.cloudfoundry.org/teams/capi-team/pipelines/capi/jobs/publish-docs-v3), are still created by jobs in the Concourse pipeline and the GitHub action [deploy_v3_docs](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/.github/workflows/deploy_v3_docs.yml) is only used to upload the results (the previously used mechanism was discontinued by GitHub beginning of July).

pivotalgeorge:
@philippthun, thanks for the clarification. This PR is just about v2 docs. The Concourse vs. Github Actions decision will be relevant for v3, but as v3 docs are in GH-pages, they will not be impacted by the current impending shutdown.

@moleske
> Could tag this repo with the version on the commit that is included in the capi release version

is [picking-up](https://github.com/cloudfoundry/cloud_controller_ng/pull/3874/files#diff-e34cd92a2ae1a54775a2e16b2d1312915a8ce73162ab81b601b30f7086753cd0R9) version info from [config/version_v2](https://github.com/cloudfoundry/cloud_controller_ng/commits/main/config/version_v2) insufficient?

philippthun:
> This PR is just about v2 docs.

Okay, I see. Would you mind removing the file `.github/workflows/gh_pages_docs_v3.yml` from this PR?

pivotalgeorge:
@sethboyles & @evanfarrar — to followup on the capi-release triggering conversation:

I created a draft PR on the capi-release side: https://github.com/cloudfoundry/capi-release/pull/438
an example triggered run is [#12](https://github.com/pivotalgeorge/cloud_controller_ng/actions/runs/9897433090) - shows up as triggered by me since the secret was set to an access token tied to my account, but In Production™ this would be a bot account.

@moleske would this address your original concern?

pivotalgeorge:
Since today is the last day for the PCF env hosting v2 docs, I say we merge this in to avoid downtime

pivotalgeorge:
After discussing with @sethboyles and @Gerg: Moving the v2 workflow to capi-release. see https://github.com/cloudfoundry/capi-release/pull/440

sethboyles (approved):

Issue #3837, opened 2024-06-11
Canary Deployments Proposal


sethboyles:
# CF Canary Deployments 
Authors: @sethboyles @pivotalgeorge @joaopapereira 
Reviewers: @Gerg @tcdowney @Samze
CF Canary Deployments 

Draft Proposal

Authors: [Seth Boyles](mailto:seth.boyles@broadcom.com) [George Gelashvili](mailto:george.gelashvili@broadcom.com) [Joao De Almeida Pereira](mailto:joao.pereira@broadcom.com)

Reviewers: [Greg Cobb](mailto:greg.cobb@broadcom.com)

# Feature goals

Canary Deployments will allow App Developers to create a new Application Instance with a new version of their application. By monitoring the Canary Instance, App Developers will be able to ensure the reliability of this Canary Instance before promoting the new version to the rest of their application instances.

# How to Use It

## User Workflow

​​

![](https://lh7-us.googleusercontent.com/docsz/AD_4nXdJ6WdQsmmRVBDVpvd8lg9NHe3rs0RwJC9Sbm68zfw5B5VEYLv2uFv4WDiMne68aVTSq7VvaDFMD_fhQTtwnAfCOtkPP1ZmspxHs1A-69YBeFcic7PKedmVnHk7rzrfBampw4CA52nIwJHdVdRB9YVNyJ9j?key=T7KJJihw_qtugaxI6ahl9w)

### Starting a Canary Deployment

App Developers will be able to push an application with Canary Deployments by using the `--strategy` flag on the CLI, similar to Rolling Deployments:

    $ cf push myapp --strategy=canary

(this flag will also be available for other actions like `restart` and `restage`)

Once the Canary Deployment has brought up the canary instance, the CLI will exit.

### Observing the Canary Instance

App Developers will be able to monitor the canary instance’s status by:

- monitoring app logs by a tag identifying which logs originate from the Canary Instance

- routing requests to the canary instance directly

### Promoting the New Version

If the App Developer determines that the Canary Instance is reliable and wants to promote the new app version to all instances, they would execute the `cf continue-deployment` command:

    cf continue-deployment myapp

Cloud Controller will then promote the rest of the Application Instances to the new app version.

### Canceling the Canary Deployment

If the App Developer  determines that the Canary Instance is NOT reliable, they would execute the  `cf cancel-deployment` command:

    cf cancel-deployment myapp

Cloud Controller will then tear down the Canary Instance.

***

[...]

PR #3881, opened 2024-07-11
Add last_status_change to deployments object

joaopapereira:

Add a timestamp for the last time that a status changed on a deployment

This will help developers identify how long a Canary Deployment has been in the PAUSED state.

* #3837 

### DIFF
--- a/db/migrations/20240709234116_add_status_updated_at_to_deployments.rb
+++ b/db/migrations/20240709234116_add_status_updated_at_to_deployments.rb
@@ -1,0 +1,14 @@
+Sequel.migration do
+  up do
+    alter_table(:deployments) do
+      add_column :status_updated_at, DateTime, default: Sequel::CURRENT_TIMESTAMP, null: false
+    end
+    run 'update deployments set status_updated_at = updated_at where updated_at is not null'
+  end
+
+  down do
+    alter_table(:deployments) do
+      drop_column :status_updated_at
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


db/migrations/20240709234116_add_status_updated_at_to_deployments.rb
```diff
@@ -0,0 +1,7 @@
+Sequel.migration do
+  change do
+    alter_table(:deployments) do
+      add_column :status_updated_at, DateTime, default: Sequel::CURRENT_TIMESTAMP, null: false
```
Samze, line 4: **question/non-blocking** I suppose we can't set an accurate value here for existing deployments since we can't determine when the latest time status was changed right? So this field will be inaccurate until another deployment status change?

sethboyles: good q... maybe we should allow null and set this in Ruby?

Samze: Are you suggesting we set a default in Ruby or set it to null until there is a status update for existing deployments? Another option here that may be slightly closer to being correct is to set it to the value of `updated_at`.

sethboyles: > set it to null until there is a status update for existing deployments?

This one.  Which means old Deployments will have a null value for last_status_change.

Using `updated_at` might be a nice compromise if we want to avoid a null column...

Samze (approved):
LGTM

PR #3887, opened 2024-07-16
Add thread based delayed workers

johha:
Currently operators can configure the number of delayed job workers for the generic queue with two parameters:
-  number of cc-worker VMs in the cf bosh deployment
-  number of worker processes on each VM (capi parameter `number_of_workers`)

Most delayed jobs however are limited by I/O (e.g. database queries) and don't do a lot of processing in ruby.
To increase the throughput on a single worker VM while reducing resource consumption (threads vs processes) this change introduces the option to run multiple delayed worker threads in a single process. 
The number of workers calculates like this: `num_cc_workers X number_processes X num_threads` 

Tests comparing 8 worker processes versus 8 worker threads show that the throughput is the same while the memory consumption is  ~60% lower. Allowing operators to run more workers on a single VM.

All threads have a graceful shutdown period of 30 seconds ([like monit](https://mmonit.com/monit/documentation/monit.html#LIMITS)) so that long running jobs have a chance to finish.
In case a thread crashes with an error a graceful shutdown is triggered and the whole process will exit with an error. Delayed::Job already catches all errors during job execution (https://github.com/collectiveidea/delayed_job/blob/master/lib/delayed/worker.rb#L241-L244).

  - https://github.com/cloudfoundry/capi-release/pull/447

* [x] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats) -> Tested on a bbl env

### DIFF
--- a/lib/delayed_job/threaded_worker.rb
+++ b/lib/delayed_job/threaded_worker.rb
@@ -1,9 +1,7 @@
 require 'socket'
 
 class ThreadedWorker < Delayed::Worker
-  DEFAULT_THREAD_COUNT = 4
-
-  def initialize(options={}, thread_count=DEFAULT_THREAD_COUNT)
+  def initialize(thread_count, options={})
     super(options)
     @thread_count = thread_count
     @threads = []
--- a/lib/delayed_job/threaded_worker.rb
+++ b/lib/delayed_job/threaded_worker.rb
@@ -1,104 +1,102 @@
-require 'socket'
-
 class ThreadedWorker < Delayed::Worker
-  def initialize(thread_count, options={})
+  def initialize(thread_count, options={}, grace_period_seconds=30)
     super(options)
     @thread_count = thread_count
     @threads = []
-    @stop_signal = false
-    @name_prefix = options[:worker_name] || 'worker'
+    @unexpected_error = false
+    @mutex = Mutex.new
+    @grace_period_seconds = grace_period_seconds
   end
 
   def start
-    trap_signals
+    # add quit trap as in QuitTrap monkey patch
+    trap('QUIT') do
+      Thread.new { say 'Exiting...' }
+      stop
+    end
 
-    say "Starting multi-threaded job worker with #{@thread_count} threads"
+    trap('TERM') do
+      Thread.new { say 'Exiting...' }
+      stop
+      raise SignalException.new('TERM') if self.class.raise_signal_exceptions
+    end
 
-    @thread_count.times do |i|
-      thread_name = generate_thread_name(i + 1)
-      @threads << Thread.new do
-        thread_name(thread_name)
-        threaded_work_off
+    trap('INT') do
+      Thread.new { say 'Exiting...' }
+      stop
+      raise SignalException.new('INT') if self.class.raise_signal_exceptions && self.class.raise_signal_exceptions != :term
+    end
+
+    say "Starting threaded delayed worker with #{@thread_count} threads"
+
+    @thread_count.times do |thread_index|
+      thread = Thread.new do
+        Thread.current[:thread_name] = "thread:#{thread_index + 1}"
+        threaded_start
+      rescue Exception => e # rubocop:disable Lint/RescueException
+        @mutex.synchronize do
+          say "Unexpected error: #{e.message}\n#{e.backtrace.join("\n")}", 'error'
+          @unexpected_error = true
+        end
+        stop
+      end
+      @mutex.synchronize do
+        @threads << thread
       end
     end
 
-    @threads.each(&:join) # Wait for threads to finish
+    @threads.each(&:join)
+  ensure
+    raise 'Unexpected error occurred in one of the worker threads' if @unexpected_error
   end
 
   def name
-    Thread.current[:name] || @name
+    base_name = super
+    thread_name = Thread.current[:thread_name]
+    thread_name ? "#{base_name} #{thread_name}" : base_name
   end
 
-  def name=(name)
-    # Set the instance variable for compatibility && ensure thread-local storage is also updated
-    @name = name
-    Thread.current[:name] = name
-  end
+  def stop
+    Thread.new do
+      say 'Shutting down worker threads gracefully...'
+      super
 
-  private
-
-  def generate_thread_name(thread_index)
-    if @name
-      "#{@name}-thread:#{thread_index}"
-    else
-      "#{@name_prefix}-host:#{Socket.gethostname} pid:#{Process.pid} thread:#{thread_index}"
+      @threads.each do |t|
+        Thread.new do
+          t.join(@grace_period_seconds)
+          if t.alive?
+            say "Killing thread '#{t[:thread_name]}'"
+            t.kill
+          end
+        end
+      end.each(&:join) # Ensure all join threads complete
     end
   end
 
-  def thread_name(name)
-    # Store the name in thread-local storage && Set the name in the instance variable for compatibility
-    Thread.current[:name] = name
-    self.name = name
-  end
-
-  def stop_signal?
-    @stop_signal
-  end
-
-  def threaded_work_off
-    retry_attempts = 0
-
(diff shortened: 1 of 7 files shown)


lib/delayed_job/threaded_worker.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def generate_thread_name(thread_index)
+    if @name
+      "#{@name}-thread:#{thread_index}"
+    else
+      "#{@name_prefix}-host:#{Socket.gethostname} pid:#{Process.pid} thread:#{thread_index}"
+    end
+  end
+
+  def thread_name(name)
```
philippthun, line 48: Isn't this the same as `def name=(name)` above?

lib/delayed_job/threaded_worker.rb
```diff
@@ -0,0 +1,104 @@
...
+
+        count = @result[0] + @result[1]
+
+        if count.zero?
+          sleep(self.class.sleep_delay) unless stop_signal?
+        else
+          say sprintf("#{count} jobs processed at %.4f j/s, %d failed", count / runtime, @result.last)
+        end
+      rescue StandardError => e
+        say "Thread #{name} encountered an error: #{e.message}. Restarting thread..."
```
Samze, line 75: **Suggestion** This isn't actually restarting the thread right? Perhaps change the phrasing here to `Thread #{name} encountered an error: #{e.message}. Retrying...`

lib/delayed_job/threaded_worker.rb
```diff
@@ -0,0 +1,104 @@
...
+        retry_attempts += 1
+        if retry_attempts >= 5
+          say "Thread #{name} has failed #{retry_attempts} times. Exiting to prevent infinite loop."
```
Samze, line 78: **question** Just to understand here, with our current DelayedWorker, if there is an exception in the thread it will exit the process and monit will start it back up again.

With this implementation if one of our threads fails 5 times it will return, but the thread nor process will be restarted unless *all* threads error and the `join` above completes and we exit the process?

I'm not sure in what scenarios multiple exceptions are raised, but its possible over time our actual running thread count will reduce from `thread_count` downwards until finally all threads error and monit restarts the process. Is that correct?

lib/delayed_job/threaded_worker.rb
johha: The first one is the getter for `name` and the 2nd one is the setter

lib/delayed_job/threaded_worker.rb
johha: This could indeed become a problem. I think it would be better to use a thread pool (with `Concurrent::ThreadPoolExecutor`). I have a first working implementation but the tests still need some work. I'll convert this PR to a draft for now.

johha: We decided to simplify the implementation by removing the thread pool and retry mechanisms.
In case a thread errors we will log the error message trigger a graceful shutdown of 30 seconds and raise an error. Then monit can restart the whole process.
This is probably a edge case as delayed job already catches all errors when executing jobs: https://github.com/collectiveidea/delayed_job/blob/master/lib/delayed/worker.rb#L241-L244

Samze (approved):
LGTM

lib/delayed_job/threaded_worker.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    say "Starting threaded delayed worker with #{@num_threads} threads"
+
+    @num_threads.times do |thread_index|
+      thread = Thread.new do
+        Thread.current[:thread_name] = "thread:#{thread_index + 1}"
+        threaded_start
+      rescue Exception => e # rubocop:disable Lint/RescueException
+        @mutex.synchronize do
+          say "Unexpected error: #{e.message}\n#{e.backtrace.join("\n")}", 'error'
```
philippthun, line 38: The mutex is only required for setting the `unexpected_error` attribute, or? Can the message be moved out?

lib/delayed_job/threaded_worker.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    @threads.each(&:join)
+  ensure
+    raise 'Unexpected error occurred in one of the worker threads' if @unexpected_error
+  end
+
+  def name
+    base_name = super
+    thread_name = Thread.current[:thread_name]
+    thread_name ? "#{base_name} #{thread_name}" : base_name
```
philippthun, line 56: In other places we use the pattern `sth.nil? ? other : sth`

spec/unit/lib/delayed_job/threaded_worker_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+
+    it 'sets the exit flag in the parent worker' do
+      worker.stop
+      sleep 0.1 until worker.instance_variable_get(:@exit)
+      expect(worker.instance_variable_get(:@exit)).to be true
+    end
+
+    it 'allows threads to finish their work without being killed prematurely' do
+      allow(worker).to receive(:threaded_start) do
+        5.times do
```
philippthun, line 99: The implementation for these threads could be:
```
sleep grace_period_seconds / 2 until worker.instance_variable_get(:@exit) == true
```

spec/unit/lib/delayed_job/threaded_worker_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+          break if worker.instance_variable_get(:@exit)
+
+          sleep 0.5
+        end
+      end
+
+      worker_thread = Thread.new { worker.start }
+      sleep(0.5)
+      expect(worker.instance_variable_get(:@threads).all?(&:alive?)).to be true
```
philippthun, line 108: I like the `sleep .. until ...` pattern. What about:
```
sleep 0.1 until worker.instance_variable_get(:@threads).length == num_threads && worker.instance_variable_get(:@threads).all?(&:alive?)
```

spec/unit/lib/delayed_job/threaded_worker_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+      worker.instance_variable_get(:@threads).each { |t| allow(t).to receive(:kill).and_call_original }
+
+      Thread.new { worker.stop }.join
+      worker.instance_variable_get(:@threads).each(&:join)
```
philippthun, line 112: This `join` is done internally in the ThreadedWorker and should not be repeated in the test.

spec/unit/lib/delayed_job/threaded_worker_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+      expect(worker.instance_variable_get(:@threads).all?(&:alive?)).to be false
```
philippthun, line 113: From my point of view the other check (i.e. kill not called) is enough. I would remove this expect.

spec/unit/lib/delayed_job/threaded_worker_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+      worker_thread.join
+      worker.instance_variable_get(:@threads).each { |t| expect(t).not_to have_received(:kill) }
+    end
+
+    it 'kills threads that exceed the grace period during shutdown' do
```
philippthun, line 118: Should be adapted in a similar way as the test above...

spec/unit/lib/delayed_job/threaded_worker_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+      worker = ThreadedWorker.new(num_threads, {}, 3)
+      allow(worker).to receive(:threaded_start) do
+        10.times do
```
philippthun, line 121: Here the implementation could be:
```
sleep grace_period_seconds * 2 until worker.instance_variable_get(:@exit) == true
```

philippthun (approved):

PR #3890, opened 2024-07-18
Update CAPI v2 docs links to new subdomain

pivotalgeorge:
This PR updates CAPI v2 docs links from `apidocs.cloudfoundry.org` to `v2-apidocs.cloudfoundry.org`.
The goal is to signal that v3 is the primary API, also a necessary step in eventually sunsetting v2.

~~The infrastructure hosting the app serving docs at apidocs.cloudfoundry.org is expected to go down imminently, so this change is needed to avoid dead links.~~
`apidocs.cloudfoundry.org` has been converted to a redirect to `v3-apidocs.cloudfoundry.org`, so this change is needed to avoid incorrect links

See capi-release PRs [#440](https://github.com/cloudfoundry/capi-release/pull/440) and [#441](https://github.com/cloudfoundry/capi-release/pull/441)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 ## Helpful Resources
 
 * [V3 API Docs](http://v3-apidocs.cloudfoundry.org)
-* [V2 API Docs](http://apidocs.cloudfoundry.org)
+* [V2 API Docs](http://v2-apidocs.cloudfoundry.org)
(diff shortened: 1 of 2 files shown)


moleske (approved):
yep those are pretty new links, thanks for the updating!

PR #3891, opened 2024-07-18
Add max_in_flight to /v3/deployments

evanfarrar:
__* A short explanation of the proposed change:__
** This exposes a new option for the deployment resource that 
__* An explanation of the use cases your change solves__
** Currently, rolling deploys deploy one instance at a time, but if an app has many instances this can take quite a while
** there is a CLI feature in progress, will link when it is a PR

### DIFF
--- a/app/actions/deployment_create.rb
+++ b/app/actions/deployment_create.rb
@@ -49,7 +49,8 @@ module VCAP::CloudController
             original_web_process_instance_count: desired_instances(app.oldest_web_process, previous_deployment),
             revision_guid: revision&.guid,
             revision_version: revision&.version,
-            strategy: DeploymentModel::ROLLING_STRATEGY
+            strategy: DeploymentModel::ROLLING_STRATEGY,
+            max_in_flight: message.options ? message.options[:max_in_flight] : 1
           )
           MetadataUpdate.update(deployment, message)
 
@@ -168,7 +169,8 @@ module VCAP::CloudController
           original_web_process_instance_count: desired_instances(app.oldest_web_process, previous_deployment),
           revision_guid: revision&.guid,
           revision_version: revision&.version,
-          strategy: DeploymentModel::ROLLING_STRATEGY
+          strategy: DeploymentModel::ROLLING_STRATEGY,
+          max_in_flight: message.options ? message.options[:max_in_flight] : 1
         )
 
         MetadataUpdate.update(deployment, message)
--- a/app/messages/deployment_create_message.rb
+++ b/app/messages/deployment_create_message.rb
@@ -7,6 +7,7 @@ module VCAP::CloudController
       droplet
       revision
       strategy
+      options
     ]
 
     validates_with NoAdditionalKeysValidator
--- a/lib/cloud_controller/deployment_updater/updater.rb
+++ b/lib/cloud_controller/deployment_updater/updater.rb
@@ -84,7 +84,7 @@ module VCAP::CloudController
           end
 
           scale_down_oldest_web_process_with_instances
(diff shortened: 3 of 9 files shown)


lib/cloud_controller/deployment_updater/updater.rb
```diff
@@ -84,7 +84,7 @@ def scale_deployment
           end
 
           scale_down_oldest_web_process_with_instances
-          deploying_web_process.update(instances: deploying_web_process.instances + 1)
+          deploying_web_process.update(instances: deploying_web_process.instances + deployment.max_in_flight)
```
sethboyles, line 87: in cases where total number of instances is not divisible by `max_in_flight`, will this result in extra instances ?  Should there be a check like this?

```
[deploying_web_process.instances + deployment.max_in_flight, deployment.original_web_process_instance_count].min
```

app/actions/deployment_create.rb
```diff
@@ -49,7 +49,8 @@ def create(app:, user_audit_info:, message:)
             original_web_process_instance_count: desired_instances(app.oldest_web_process, previous_deployment),
             revision_guid: revision&.guid,
             revision_version: revision&.version,
-            strategy: DeploymentModel::ROLLING_STRATEGY
+            strategy: DeploymentModel::ROLLING_STRATEGY,
+            max_in_flight: message.options ? message.options[:max_in_flight] : 1
```
sethboyles, line 53: Would this be cleaner if it was encapsulated in the message?

spec/unit/lib/cloud_controller/deployment_updater/updater_spec.rb
```diff
@@ -466,6 +513,26 @@ module VCAP::CloudController
...
+              max_in_flight: 10
+            )
+          end
+
+          it 'scales up the coerced web process by 10' do
+            expect do
+              subject.scale
+            end.to change {
+              deploying_web_process.reload.instances
+            }.by(10)
```
sethboyles, line 533: similar to my above comment, if `original_web_process_instance_count` is `6` (not sure but from my reading of the test file it seems to be), then I wouldn't expect the actual instances to exceed `6`

sethboyles (commented):
Mostly looks good, once the 'max instances' case is handled.  I do like how simple the change is if we are OK with the batch updates.  It is "max" in flight, after all...

app/actions/deployment_create.rb
```diff
@@ -168,7 +169,8 @@ def deployment_for_stopped_app(app, message, previous_deployment, previous_dropl
           original_web_process_instance_count: desired_instances(app.oldest_web_process, previous_deployment),
           revision_guid: revision&.guid,
           revision_version: revision&.version,
-          strategy: DeploymentModel::ROLLING_STRATEGY
+          strategy: DeploymentModel::ROLLING_STRATEGY,
+          max_in_flight: message.options ? message.options[:max_in_flight] : 1
```
Samze, line 173: **todo** We need to handle when options is set to an object, but the `max_in_flight` key is not set. I believe this would currently evaluate to null

app/messages/deployment_create_message.rb
```diff
@@ -7,6 +7,7 @@ class DeploymentCreateMessage < MetadataBaseMessage
       droplet
       revision
       strategy
+      options
```
Samze, line 10: *thought/todo* I believe we need validation to ensure max_in_flight must be >= 1

evanfarrar: I noticed that we don't have validations that instances are >= 1...

PR #3894, opened 2024-07-22
Add option to check for deprecated stacks in pre-start rake task

xandroc:
* Add optional check to ensure operators can upgrade without hanging dependencies to old or unwanted stacks

### DIFF
--- a/lib/tasks/stack_check.rake
+++ b/lib/tasks/stack_check.rake
@@ -1,23 +1,10 @@
 namespace :stacks do
   desc 'Check Installed Stacks'
   task stack_check: :environment do
-    VCAP::CloudController::Encryptor.db_encryption_key = RakeConfig.config.get(:db_encryption_key)
-    require 'models/runtime/buildpack_lifecycle_data_model'
-    require 'models/runtime/stack'
+    require 'cloud_controller/check_stacks'
 
-    deprecated_stack = 'cflinuxfs3'
-    stack_config = VCAP::CloudController::Stack::ConfigFile.new(RakeConfig.config.get(:stacks_file))
-    configured_stacks = stack_config.stacks
-    deprecated_stack_in_config = (configured_stacks.find { |stack| stack['name'] == deprecated_stack }).nil?
-
-    exit(0) if deprecated_stack_in_config
-
-    logger = Steno.logger('cc.stack')
-    VCAP::CloudController::DB.connect(RakeConfig.config.get(:db), logger)
-    deprecated_stack_in_db = VCAP::CloudController::Stack.first(name: deprecated_stack).present?
-    if deprecated_stack_in_db
-      logger.error('rake task \'stack_check\' failed, ' + deprecated_stack + ' is not supported')
-      exit(1)
+    BackgroundJobEnvironment.new(RakeConfig.config).setup_environment do
+      VCAP::CloudController::CheckStacks.new(RakeConfig.config).validate_stacks
     end
   end
 end
--- a/lib/tasks/stack_check.rake
+++ b/lib/tasks/stack_check.rake
@@ -1,4 +1,4 @@
-namespace :buildpacks do
+namespace :stacks do
   desc 'Check Installed Stacks'
   task stack_check: :environment do
     VCAP::CloudController::Encryptor.db_encryption_key = RakeConfig.config.get(:db_encryption_key)
@@ -7,12 +7,8 @@ namespace :buildpacks do
 
     deprecated_stack = 'cflinuxfs3'
     stack_config = VCAP::CloudController::Stack::ConfigFile.new(RakeConfig.config.get(:stacks_file))
-    p stack_config.stacks
     configured_stacks = stack_config.stacks
     deprecated_stack_in_config = (configured_stacks.find { |stack| stack['name'] == deprecated_stack }).nil?
-    p(configured_stacks.find { |stack| stack['name'] == deprecated_stack })
-    p deprecated_stack_in_config
-    p deprecated_stack_in_config.nil?
 
     exit(0) if deprecated_stack_in_config
 
--- a/lib/tasks/stack_check.rake
+++ b/lib/tasks/stack_check.rake
@@ -1,0 +1,27 @@
+namespace :buildpacks do
+  desc 'Check Installed Stacks'
+  task stack_check: :environment do
+    VCAP::CloudController::Encryptor.db_encryption_key = RakeConfig.config.get(:db_encryption_key)
+    require 'models/runtime/buildpack_lifecycle_data_model'
+    require 'models/runtime/stack'
+
+    deprecated_stack = 'cflinuxfs3'
+    stack_config = VCAP::CloudController::Stack::ConfigFile.new(RakeConfig.config.get(:stacks_file))
+    p stack_config.stacks
+    configured_stacks = stack_config.stacks
+    deprecated_stack_in_config = (configured_stacks.find { |stack| stack['name'] == deprecated_stack }).nil?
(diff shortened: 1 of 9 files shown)


spec/unit/lib/cloud_controller/check_stacks_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+        ],
+        'deprecated_stacks' => 'cflinuxfs3'
+      }
+    end
+
+    let(:cflinuxfs3) { { 'name' => 'cflinuxfs3', 'description' => 'fs3' } }
+    let(:cflinuxfs4) { { 'name' => 'cflinuxfs4', 'description' => 'fs4' } }
+
+    before do
+      # binding.pry
```
joaopapereira, line 22: Do we still need this?

spec/unit/lib/cloud_controller/check_stacks_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+      file.write(stack_file_contents.to_yaml)
+      file.close
+      Stack.configure(file)
+      Stack.populate
+      TestConfig.override(stacks_file: file.path)
+    end
+
+    let(:stack_checker) { CheckStacks.new(TestConfig.config_instance) }
+
+    describe 'there deprecated stacks is nil' do
```
joaopapereira, line 34:
```suggestion
    describe 'the deprecated stacks is nil' do
```

joaopapereira (approved):
LGTM, some minor nit's

spec/unit/lib/cloud_controller/check_stacks_spec.rb
xandroc: ah, good catch! I'll remove it

spec/unit/lib/cloud_controller/check_stacks_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+          'default' => 'cflinuxfs4',
+          'stacks' => [
+            cflinuxfs3,
+            cflinuxfs4
+          ],
+          'deprecated_stacks' => ['cflinuxfs3']
+        }
+      end
+
+      describe 'the deprecated stack is in the db' do
```
Samze, line 79: *question* Do we need these describes for the db behaviour? Looking at the code, if the deprecated stack is in the stacks list we don't even check the DB. We return early here https://github.com/cloudfoundry/cloud_controller_ng/pull/3894/files#diff-b51e853460880f38ea614d45c335bfdaa98ebf81f6166ce4c13ffab95bfbf871R23

spec/unit/lib/cloud_controller/check_stacks_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+        it 'does not raise an error' do
+          expect { stack_checker.validate_stacks }.not_to raise_error
+        end
+      end
+
+      describe 'the deprecated stack is not in the db' do
```
Samze, line 85: Same as https://github.com/cloudfoundry/cloud_controller_ng/pull/3894/files#r1693280842

lib/tasks/stack_check.rake
```diff
@@ -0,0 +1,14 @@
+namespace :stacks do
+  desc 'Check Installed Stacks'
+  task stack_check: :environment do
+    logger = Steno.logger('cc.stack')
+    VCAP::CloudController::Encryptor.db_encryption_key = RakeConfig.config.get(:db_encryption_key)
```
Samze, line 5: **issue/question** Not entirely sure why we do this, but I *believe* `VCAP::CloudController::Encryptor` has two methods of loading an encryption key. There is the historic way which is setting `db_encryption_key` but there is also an array of encryption keys that allow rotation. See [capi-release](https://github.com/cloudfoundry/capi-release/blob/develop/jobs/cloud_controller_ng/spec#L766-L781). You can see this logic here https://github.com/cloudfoundry/cloud_controller_ng/blob/3d09a707b253acbdf54dff659a329f4267c840f3/lib/cloud_controller/encryptor.rb#L49-L51. Here we seem to only set the old style encryption key, so this does not factor in the encryption array/ledger.

So I'm not sure what the consequence is of loading this Encrypter this way. Do we need to create it with more configuration, somewhat like https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/cloud_controller/config.rb#L110-L116 ?

xandroc: Not entirely sure how we use `config.rb`, we load the encryption key the same way as db migrations,
https://github.com/cloudfoundry/cloud_controller_ng/blob/3d09a707b253acbdf54dff659a329f4267c840f3/lib/tasks/db.rake#L127

@Gerg @tcdowney any input here? is there a correct or preferred way to load the db encryption key?

tcdowney: I think @Samze is correct and I think maybe we've found a bug in `db.rake` -- unless something somewhere is calling `@config.configure_components` and it gets set up properly as a side effect.

I think both here and `db.rake` need to do the full thing and load the encryption key all the ways in these rake tasks (maybe this could be put into a function on the config object).

```
Encryptor.db_encryption_key = get(:db_encryption_key)

      if get(:database_encryption)
        Encryptor.database_encryption_keys = get(:database_encryption)[:keys]
        Encryptor.current_encryption_key_label = get(:database_encryption)[:current_key_label]
        Encryptor.pbkdf2_hmac_iterations = get(:database_encryption)[:pbkdf2_hmac_iterations]
      end
```

The legacy way of using `db_encryption_key` works when that's set, but it's a deprecated field and the array of properties is the preferred way of doing this.

https://github.com/cloudfoundry/cf-deployment/blob/2893dec3cb0afaab310f8f23b235969f21c9cd28/cf-deployment.yml#L946-L950

I think this might be able to cause some subtle bugs with the way it currently is... 🤔

tcdowney: It does the same thing again here:
https://github.com/cloudfoundry/cloud_controller_ng/blob/3d09a707b253acbdf54dff659a329f4267c840f3/lib/cloud_controller/db_migrator.rb#L8

🤔

tcdowney: Yeah I put a `binding.pry` in the DBMigrator and confirmed the proper one is empty:
```
[4] pry(#<DBMigrator>)> VCAP::CloudController::Encryptor.send(:database_encryption_keys)
=> {}
```

I think we might want to make a separate bug and fix that @Samze @xandroc ^ Good catch

xandroc: I'll write a story for this, thanks for verifying @tcdowney

xandroc: @Samze updated

Samze: @xandroc I don't know much about whether just calling `@config.configure_components` here has any additional consequences or if it is safe, perhaps @tcdowney knows? Just want to make sure we're not introducing additional problems here.

tcdowney: I don't know for sure. The `ProcessObserver` setup and `run_initializers` might not be something we want to do, but I'm not positive.

https://github.com/cloudfoundry/cloud_controller_ng/blob/3d09a707b253acbdf54dff659a329f4267c840f3/lib/cloud_controller/config.rb#L121-L123

To be safe you might want to extract the db encryptor steps out into its own method and call that in configure_components and these rake tasks.

xandroc: Broke db setup out into to its own method and calling it separately

tcdowney (approved):
LGTM

Samze (approved):

PR #3899, opened 2024-07-24
Prevent 500 on parallel service instance creation

kathap:
We observed 500 responses on ```POST /v3/service_instances``` when requests run in parallel and the db is small.
This occurs if the one POST bypasses the other one and inserts the record into the db after the first POST has passed the validate function but before the actual insert. This leads to a ```PG::UniqueViolation: ERROR: duplicate key value violates unique constraint``` which is not caught and thus results in a 500 response. To prevent this, we add a Sequel::UniqueConstraintViolation rescue to catch the ```PG::UniqueViolation``` error and to return an adequate 422 error "The service instance name is taken.."

Do not return 500 on parallel service instance creation on small db but a 422.

We observed 500 responses on ```POST /v3/service_instances``` when requests run in parallel and the db is small. Now an appropriate error is returned.

### DIFF
--- a/app/models/services/service_instance.rb
+++ b/app/models/services/service_instance.rb
@@ -110,6 +110,15 @@ module VCAP::CloudController
       end
     end
 
+    def around_save
+      yield
+    rescue Sequel::UniqueConstraintViolation => e
+      raise e unless e.message.include?('si_space_id_name_index')
+
+      errors.add(:name, :unique)
+      raise validation_failed_error
+    end
+
     def validate
       validates_presence :name
       validates_presence :space
--- a/spec/unit/actions/v3/service_instance_create_managed_spec.rb
+++ b/spec/unit/actions/v3/service_instance_create_managed_spec.rb
@@ -214,6 +214,23 @@ module VCAP::CloudController
           end
         end
 
+        context 'parallel creation of managed service instances' do
+          it 'ensures one creation is successful and the other fails due to name conflict' do
+            # First request, should succeed
+            expect do
+              action.precursor(message:, service_plan:)
+            end.not_to raise_error
+
+            # Mock the validation for the second request to simulate the race condition and trigger a unique constraint violation
+            allow_any_instance_of(ManagedServiceInstance).to receive(:validate).and_return(true)
+
+            # Second request, should fail with correct error message
+            expect do
+              action.precursor(message:, service_plan:)
+            end.to raise_error(ServiceInstanceCreateManaged::InvalidManagedServiceInstance, 'The service instance name is taken: si-test-name.')
+          end
+        end
+
         context 'quotas' do
           context 'when service instance limit has been reached for the space' do
             before do


joaopapereira (approved):
LGTM

tcdowney (dismissed):

kathap:
Need to enhance the parallel test still

spec/unit/actions/v3/service_instance_create_managed_spec.rb
```diff
@@ -214,6 +214,37 @@ module V3
...
+
+        context 'parallel creation of managed service instances' do
+          it 'ensures one creation is successful and the other fails due to name conflict' do
+            # First request, should succeed
+            expect do
+              action.precursor(message:, service_plan:)
+            end.not_to raise_error
+
+            # Mock the validation for the second request to simulate the race condition and trigger a unique constraint violation
+            allow_any_instance_of(VCAP::CloudController::ManagedServiceInstance).to receive(:validate).and_return(true)
```
philippthun, line 236: The `VCAP::CloudController` namespace is not needed here.

spec/unit/actions/v3/service_instance_create_managed_spec.rb
```diff
@@ -214,6 +214,37 @@ module V3
...
+
+            # Second request, should fail with correct error message
+            expect do
+              action.precursor(message:, service_plan:)
+            end.to raise_error(ServiceInstanceCreateManaged::InvalidManagedServiceInstance, 'The service instance name is taken: si-test-name.')
+
+            instances = VCAP::CloudController::ServiceInstance.where(name: name, space_id: space.id)
+            expect(instances.count).to eq(1)
```
philippthun, line 244: I think this check is not necessary as it's impossible that two instances with the same name and space_id exist...

spec/unit/actions/v3/service_instance_create_managed_spec.rb
kathap: You're right

spec/unit/actions/v3/service_instance_create_managed_spec.rb
kathap: Indeed

app/models/services/service_instance.rb
```diff
@@ -110,6 +111,14 @@ def name_clashes
       end
     end
 
+    def around_save
+      yield
+    rescue Sequel::UniqueConstraintViolation => e
+      raise UniqueConstraintViolationError.new(e) if e.message.try(:include?, 'si_space_id_name_index')
```
philippthun, line 117: Maybe we just raise a `Sequel::ValidationFailed` error with the correct information:

```
if e.message.try(:include?, 'si_space_id_name_index')
  errors.add(:name, :unique)
  raise validation_failed_error
end
```

This would mean that the `validation_error!` method in the `ServiceInstanceCreateMixin` does not need to be touched at all. And this could also be a pattern for all the other models having the same problem, i.e. simply translate a `UniqueConstraintViolation` into the corresponding `ValidationFailed` error.

kathap: sounds good

philippthun (approved):

PR #3918, opened 2024-08-02
Prevent 500 on parallel object creation

kathap:
While fixing #3899, we found further possible occurrences of 500 errors in severel _create actions when doing parallel requests. In this PR we want to address them by adding an around_save to the corresponding model in which we catch Sequel::UniqueConstraintViolation errors for the case when  ne POST bypasses the other one and inserts the record into the db after the first POST has passed the validate function but before the actual insert.

Adding an around_save to the corresponding model in which we catch Sequel::UniqueConstraintViolation errors for the case when  ne POST bypasses the other one and inserts the record into the db after the first POST has passed the validate function but before the actual insert
No more 500 errors on parallel request execution

#3899
#3924

### DIFF
--- a/spec/unit/actions/service_credential_binding_app_create_spec.rb
+++ b/spec/unit/actions/service_credential_binding_app_create_spec.rb
@@ -145,6 +145,57 @@ module VCAP::CloudController
             end
           end
 
+          context 'when creating bindings with the same name concurrently' do
+            it 'raises an error when the binding name already exists' do
+              # First request, should succeed
+              expect do
+                action.precursor(service_instance, app:, message:)
+              end.not_to raise_error
+
+              # Mock the validation for the second request to simulate the race condition and trigger a
+              # unique constraint violation on app_guid + name
+              allow_any_instance_of(ServiceCredentialBindingAppCreate).to receive(:validate_binding!).and_return(true)
+              allow_any_instance_of(ServiceBinding).to receive(:validate).and_return(true)
+              binding = instance_double(ServiceBinding, destroy: false)
+              allow(ServiceBinding).to receive(:first).with(service_instance:, app:).and_return(binding)
+
+              allow_any_instance_of(ServiceBinding).to receive(:save_with_attributes_and_new_operation) do |instance|
+                instance.errors.add(%i[app_guid name], :unique)
+                raise Sequel::ValidationFailed.new('app_guid and name unique')
+              end
+
+              # Second request, should fail with correct error
+              expect do
+                action.precursor(service_instance, app:, message:)
+              end.to raise_error(ServiceBindingCreate::UnprocessableCreate,
+                                 "The binding name is invalid. App binding names must be unique. The app already has a binding with name '#{name}'.")
+            end
+
+            it 'raises an error when the app is already bound to the service instance' do
+              # First request, should succeed
+              expect do
+                action.precursor(service_instance, app:, message:)
+              end.not_to raise_error
+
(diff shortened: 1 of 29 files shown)


spec/unit/actions/service_credential_binding_app_create_spec.rb
```diff
@@ -145,6 +145,57 @@ module V3
...
+              end.not_to raise_error
+
+              # Mock the validation for the second request to simulate the race condition and trigger a
+              # unique constraint violation on service_instance_guid + app_guid
+              allow_any_instance_of(ServiceCredentialBindingAppCreate).to receive(:validate_binding!).and_return(true)
+              allow_any_instance_of(ServiceBinding).to receive(:validate).and_return(true)
+              binding = instance_double(ServiceBinding, destroy: false)
+              allow(ServiceBinding).to receive(:first).with(service_instance:, app:).and_return(binding)
+
+              allow_any_instance_of(ServiceBinding).to receive(:save_with_attributes_and_new_operation) do |instance|
```
svkrieger, line 187: Same question here

spec/unit/isolation_segment_create_spec.rb
```diff
@@ -44,6 +44,24 @@ module VCAP::CloudController
...
+            IsolationSegmentCreate.create(message)
+          end.not_to raise_error
+
+          # Mock the validation for the second request to simulate the race condition and trigger a unique constraint violation
+          allow_any_instance_of(IsolationSegmentModel).to receive(:validate).and_return(true)
+
+          # Second request, should fail with correct error
+          expect do
+            IsolationSegmentCreate.create(message)
+          end.to raise_error(IsolationSegmentCreate::Error)
```
svkrieger, line 62: Would be good to test for the error string here, like done in all other tests.

svkrieger (commented):
Looks great! Just a question about mocking in two tests and a minor finding.

EDIT: Also do we plan to get rid of the unique checks in the coding? If we catch the unique constraint errors thrown by the DB reliably now, it should not be necessary to check for uniqueness beforehand.

spec/unit/actions/service_credential_binding_app_create_spec.rb
```diff
@@ -145,6 +145,57 @@ module V3
...
+              end.not_to raise_error
+
+              # Mock the validation for the second request to simulate the race condition and trigger a
+              # unique constraint violation on app_guid + name
+              allow_any_instance_of(ServiceCredentialBindingAppCreate).to receive(:validate_binding!).and_return(true)
+              allow_any_instance_of(ServiceBinding).to receive(:validate).and_return(true)
+              binding = instance_double(ServiceBinding, destroy: false)
+              allow(ServiceBinding).to receive(:first).with(service_instance:, app:).and_return(binding)
+
+              allow_any_instance_of(ServiceBinding).to receive(:save_with_attributes_and_new_operation) do |instance|
```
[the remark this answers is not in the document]
kathap: Good point!

spec/unit/actions/service_credential_binding_key_create_spec.rb
```diff
@@ -250,6 +250,34 @@ module V3
             end
           end
         end
+
+        context 'when creating keys with the same name concurrently' do
+          let(:service_instance2) { ManagedServiceInstance.make(space:) }
```
svkrieger, line 255: We don't need a second service instance, name and message here. The constraint is on the name of the keys per service instance.

svkrieger (changes requested):

svkrieger (approved):

PR #3926, opened 2024-08-09
Clear the app's current droplet guid when droplet is deleted

philippthun:


Samze (approved):

Gerg:
This would put the app into a state where it could no longer be restarted, right? Would it be better to block deletion of droplets if they are the current droplet for an app?

stephanme:
I like the idea of blocking droplet deletion as long as it used as current droplet of an app. This should be backed by an FK constraint.

philippthun:
Closed in favor of #3960.

PR #3954, opened 2024-09-06
Optimize service instances SQL query efficiency

a-b:
Service instances cannot have duplicate values due to the primary key constraint on GUID, so using distinct is not only redundant but also affects query performance.

Co-authored-by: @shahnavaz

### DIFF
--- a/app/fetchers/service_instance_list_fetcher.rb
+++ b/app/fetchers/service_instance_list_fetcher.rb
@@ -22,7 +22,7 @@ module VCAP::CloudController
         end
 
         filter(dataset, message).
-          select_all(:service_instances).
+          select_all(:service_instances)
       end
 
       private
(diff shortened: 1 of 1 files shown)


sethboyles (changes requested):
It's unfortunate we didn't have a test in place for why this is needed, but I'm adding one with this PR:

https://github.com/cloudfoundry/cloud_controller_ng/pull/3955

We need this DISTINCT clause due to the left join on `service_instance_shares`. If a Service Instance has been shared with two or more spaces the user belongs to, that will result in duplicate SIs in the query results

a-b:
@sethboyles, thanks for your feedback. At least we improved test coverage. We will keep iterating on this PR to find alternative ways to improve query performance. Our current goals are:

1. [_must_] Find possible SQL optimizations for service instances query.
1. [_attractive_] Keep it SQL dialect agnostic.

Samze:
@a-b Perhaps you could close this and if you have suggestions for optimizations create a new PR for them?

PR #3961, opened 2024-09-10
Add db connection metrics

svkrieger:

Add metrics about DB connection usage.

To better understand the DB connection usage, especially when using Puma, we introduce metrics which track how many connections have been opened, are currently used, available and the hold and wait duration.

None

### DIFF
--- a/lib/cloud_controller/db.rb
+++ b/lib/cloud_controller/db.rb
@@ -42,6 +42,7 @@ module VCAP::CloudController
       add_connection_expiration_extension(db, opts)
       add_connection_validator_extension(db, opts)
       db.extension(:requires_unique_column_names_in_subquery)
+      add_connection_metrics_extension(db)
       db
     end
 
@@ -66,6 +67,15 @@ module VCAP::CloudController
       db.pool.connection_expiration_timeout = opts[:connection_expiration_timeout]
       db.pool.connection_expiration_random_delay = opts[:connection_expiration_random_delay] if opts[:connection_expiration_random_delay]
       # So that there are no existing connections without an expiration timestamp
+      db.disconnect
+    end
+
+    def self.add_connection_metrics_extension(db)
+      # only add the metrics for api processes. Otherwise e.g. rake db:migrate would also initialize metric updaters, which need additional config
+      return if Object.const_defined?(:RakeConfig)
+
+      db.extension(:connection_metrics)
+      # so that we gather connection metrics from the beginning
       db.disconnect
     end
 
--- a/lib/cloud_controller/metrics/prometheus_updater.rb
+++ b/lib/cloud_controller/metrics/prometheus_updater.rb
@@ -16,6 +16,7 @@ module VCAP::CloudController::Metrics
     end
 
     DURATION_BUCKETS = [5, 10, 30, 60, 300, 600, 890].freeze
+    CONNECTION_DURATION_BUCKETS = [0.001, 0.005, 0.01, 0.05, 0.1, 0.5, 1, 5, 10].freeze
 
     METRICS = [
       { type: :gauge, name: :cc_job_queues_length_total, docstring: 'Job queues length of worker processes', labels: [:queue], aggregation: :most_recent },
@@ -35,7 +36,15 @@ module VCAP::CloudController::Metrics
       { type: :gauge, name: :cc_running_tasks_total, docstring: 'Total running tasks', aggregation: :most_recent },
       { type: :gauge, name: :cc_running_tasks_memory_bytes, docstring: 'Total memory consumed by running tasks', aggregation: :most_recent },
       { type: :gauge, name: :cc_users_total, docstring: 'Number of users', aggregation: :most_recent },
-      { type: :gauge, name: :cc_deployments_in_progress_total, docstring: 'Number of in progress deployments', aggregation: :most_recent }
+      { type: :gauge, name: :cc_deployments_in_progress_total, docstring: 'Number of in progress deployments', aggregation: :most_recent },
+      { type: :gauge, name: :cc_acquired_db_connections_total, labels: %i[process_type], docstring: 'Number of acquired DB connections' },
+      { type: :histogram, name: :cc_db_connection_hold_duration_seconds, docstring: 'The time threads were holding DB connections', buckets: CONNECTION_DURATION_BUCKETS },
+      # cc_connection_pool_timeouts_total must be a gauge metric, because otherwise we cannot match them with processes
+      { type: :gauge, name: :cc_db_connection_pool_timeouts_total, labels: %i[process_type],
+        docstring: 'Number of threads which failed to acquire a free DB connection from the pool within the timeout' },
+      { type: :gauge, name: :cc_open_db_connections_total, labels: %i[process_type], docstring: 'Number of open DB connections (acquired + available)' },
+      { type: :histogram, name: :cc_db_connection_wait_duration_seconds, docstring: 'The time threads were waiting for an available DB connection',
+        buckets: CONNECTION_DURATION_BUCKETS }
     ].freeze
 
     THIN_METRICS = [
@@ -61,20 +70,20 @@ module VCAP::CloudController::Metrics
 
       # Register all metrics, to initialize them for discoverability
       METRICS.each { |metric| register(metric) }
-      THIN_METRICS.each { |metric| register(metric) } if VCAP::CloudController::Config.config.get(:webserver) == 'thin'
-      PUMA_METRICS.each { |metric| register(metric) } if VCAP::CloudController::Config.config.get(:webserver) == 'puma'
+      THIN_METRICS.each { |metric| register(metric) } if VCAP::CloudController::Config.config&.get(:webserver) == 'thin'
+      PUMA_METRICS.each { |metric| register(metric) } if VCAP::CloudController::Config.config&.get(:webserver) == 'puma'
     end
 
     def update_gauge_metric(metric, value, labels: {})
       @registry.get(metric).set(value, labels:)
     end
 
-    def increment_gauge_metric(metric)
-      @registry.get(metric).increment
+    def increment_gauge_metric(metric, labels: {})
+      @registry.get(metric).increment(labels:)
     end
 
-    def decrement_gauge_metric(metric)
-      @registry.get(metric).decrement
+    def decrement_gauge_metric(metric, labels: {})
+      @registry.get(metric).decrement(labels:)
     end
 
     def increment_counter_metric(metric)
--- a/lib/cloud_controller/runner.rb
+++ b/lib/cloud_controller/runner.rb
@@ -29,6 +29,11 @@ module VCAP::CloudController
       secrets_hash = parse_secrets
       parse_config(secrets_hash)
 
+      # DB connection metrics have a label to determine whether the process accessing the connection is the
+      # main or a worker process. We need to set this env variable before `setup_db` otherwise the main process
+      # will show up twice in the metrics as main and worker. Thin metrics will be labeled with main as well.
+      ENV['PROCESS_TYPE'] = 'main'
+
       setup_cloud_controller
 
       request_logs = VCAP::CloudController::Logs::RequestLogs.new(Steno.logger('cc.api'))
--- a/lib/cloud_controller/runners/puma_runner.rb
+++ b/lib/cloud_controller/runners/puma_runner.rb
@@ -36,6 +36,10 @@ module VCAP::CloudController
         conf.before_fork do
           Sequel::Model.db.disconnect
         end
+        conf.on_worker_boot do
+          ENV['PROCESS_TYPE'] = 'puma_worker'
+          prometheus_updater.update_gauge_metric(:cc_db_connection_pool_timeouts_total, 0, labels: { process_type: 'puma_worker' })
+        end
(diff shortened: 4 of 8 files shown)


lib/cloud_controller/runner.rb
```diff
@@ -29,6 +29,11 @@ def initialize(argv)
       secrets_hash = parse_secrets
       parse_config(secrets_hash)
 
+      # DB connection metrics have a label to determine whether the process accessing the connection is the
+      # main or a worker process. We need to set this env variable before `setup_db` otherwise the main process
+      # will show up twice in the metrics as main and worker. Thin metrics will be labeled with main as well.
+      ENV['IS_MAIN'] = 'true'
```
Samze, line 35: **nit** I wonder if something like `ENV['PROCESS_TYPE'] = 'main'` and then set it to `worker` in the puma runner would be more expressive?

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -35,7 +36,15 @@ def self.allow_pid_label
       { type: :gauge, name: :cc_running_tasks_total, docstring: 'Total running tasks', aggregation: :most_recent },
       { type: :gauge, name: :cc_running_tasks_memory_bytes, docstring: 'Total memory consumed by running tasks', aggregation: :most_recent },
       { type: :gauge, name: :cc_users_total, docstring: 'Number of users', aggregation: :most_recent },
-      { type: :gauge, name: :cc_deployments_in_progress_total, docstring: 'Number of in progress deployments', aggregation: :most_recent }
+      { type: :gauge, name: :cc_deployments_in_progress_total, docstring: 'Number of in progress deployments', aggregation: :most_recent },
+      { type: :gauge, name: :cc_acquired_db_connections_total, labels: %i[process_type], docstring: 'Number of acquired DB connections' },
+      { type: :histogram, name: :cc_db_connection_hold_duration_seconds, docstring: 'Durations of connections being hold by threads', buckets: CONNECTION_HOLD_DURATION_BUCKETS },
```
Samze, line 41: *nit/suggestion* Should this be past tesnse? e.g. `cc_db_connection_held_duration_seconds` as we only emit this metric on release

lib/cloud_controller/db.rb
```diff
@@ -69,6 +70,15 @@ def self.add_connection_expiration_extension(db, opts)
       db.disconnect
     end
 
+    def self.connection_metrics_extension(db)
+      # only add the metrics for api processes. Otherwise e.g. rake db:migrate would also initialize metric updaters, which need additional config
```
Samze, line 74: **question** Does this also add connection metrics to cc_workers/deployment_updater/clock too right?

lib/sequel/extensions/connection_metrics.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    private
+
+    def acquire(thread)
+      begin
+        if (conn = super)
+          acquired_at = Time.now
+          @prometheus_updater.increment_gauge_metric(:cc_acquired_db_connections_total, labels: { process_type: })
+          @connection_info[thread] ||= {}
```
Samze, line 42: **question** I don't know ruby concurrency well enough, do we not need any mutex here because the key being read/written to in the hash is unique per thread?

lib/cloud_controller/runners/puma_runner.rb
```diff
@@ -36,6 +36,10 @@ def initialize(config, app, logger, periodic_updater, request_logs)
         conf.before_fork do
           Sequel::Model.db.disconnect
         end
+        conf.on_worker_boot do
+          ENV['IS_MAIN'] = 'false'
+          prometheus_updater.update_gauge_metric(:cc_db_connection_pool_timeouts_total, 0, labels: { process_type: 'worker' })
```
Samze, line 41: **question/thought** If I'm understanding correctly, this will aggregate all worker metrics into one metric with the single `worker` label. Given the pools are per process, do we care about db metrics per worker?

lib/sequel/extensions/connection_metrics.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    def disconnect_connection(conn)
+      super
+      @prometheus_updater.update_gauge_metric(:cc_open_db_connections_total, size, labels: { process_type: })
+    end
+
+    def release(thread)
+      super
+
+      # acquired_at should be always set, but as a safeguard we check that it is present before accessi
```
Samze, line 85:
```suggestion
      # acquired_at should be always set, but as a safeguard we check that it is present before accessing
```

lib/cloud_controller/db.rb
```diff
@@ -42,6 +42,7 @@ def self.connect(opts, logger)
       add_connection_expiration_extension(db, opts)
       add_connection_validator_extension(db, opts)
       db.extension(:requires_unique_column_names_in_subquery)
+      connection_metrics_extension(db)
```
Samze, line 45: **praise** Thanks for doing this! Will be great to have db metrics

Samze, line 45: **thought** As part of this we should make sure we update the scaling docs  https://docs.cloudfoundry.org/running/managing-cf/scaling-cloud-controller.html and the metrics doc https://docs.cloudfoundry.org/running/all_metrics.html#cc

lib/sequel/extensions/connection_metrics.rb
```diff
@@ -0,0 +1,100 @@
...
+      if @connection_info[thread] && @connection_info[thread].key?(:acquired_at)
+        @prometheus_updater.update_histogram_metric :cc_db_connection_hold_duration_seconds, (Time.now - @connection_info[thread][:acquired_at]).seconds
+      end
+
+      @prometheus_updater.decrement_gauge_metric(:cc_acquired_db_connections_total, labels: { process_type: })
+      @connection_info.delete(thread)
+    end
+
+    def process_type
+      ENV['IS_MAIN'] == 'true' ? 'main' : 'worker'
```
Samze, line 95: **thought** Could `worker` here be confused with CC worker as opposed to puma worker?

a-b: @Samze, would something like 'web_worker' help to avoid potential confusion?

lib/sequel/extensions/connection_metrics.rb
svkrieger: As you said, the key is unique for each thread and each thread is only accessing its own key. So I'd argue we do not need a mutex here.

lib/cloud_controller/runners/puma_runner.rb
svkrieger: The prometheus registry stores metrics per process. The metrics will have the `pid` label by default, and won't be aggregated. So in this case we are able to see how many connection pool timeouts were thrown by each process. I only added the `process_type` labels for convenience, so that it is easier in the monitoring system to see which process is a puma worker process and which is a puma main process. Otherwise you'd only see the pid and have to check other metrics like worker_backlog for the pid's of the worker processes.

lib/cloud_controller/runners/puma_runner.rb
```diff
@@ -36,6 +36,10 @@ def initialize(config, app, logger, periodic_updater, request_logs)
         conf.before_fork do
           Sequel::Model.db.disconnect
         end
+        conf.on_worker_boot do
+          ENV['IS_MAIN'] = 'false'
```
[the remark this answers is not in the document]
philippthun: The `on_thread_start` hook does not run inside the newly created thread, but just before is is getting created; so this does not work.

lib/cloud_controller/db.rb
svkrieger: No it does not. All of the other jobs are rake tasks and are excluded by this line. Adding DB metrics to those other jobs would come with quite some effort. On the worker VMs there is no web server running, which means we cannot emit prometheus metrics there. Currently we are only interested in the DB connection usage of Puma processes. I guess we could add the metrics on other jobs later if needed.

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -61,24 +70,24 @@ def initialize(registry=Prometheus::Client.registry)
...
 
-    def decrement_gauge_metric(metric)
-      @registry.get(metric).decrement
+    def decrement_gauge_metric(metric, labels: {})
+      @registry.get(metric).decrement labels:
     end
 
-    def increment_counter_metric(metric)
-      @registry.get(metric).increment
+    def increment_counter_metric(metric, labels: {})
```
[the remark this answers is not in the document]
svkrieger: Let's remove it. I tried to use a counter metric before, but it did not work.

lib/sequel/extensions/connection_metrics.rb
```diff
@@ -0,0 +1,100 @@
...
+# connection pool types it might need adjustments.
+#
+# Related module: Sequel::ConnectionMetrics
+
+module Sequel
+  module ConnectionMetrics
+    # Initialize the data structures used by this extension.
+    def self.extended(pool)
+      case pool.pool_type
+      when :single, :sharded_single, :timed_queue, :sharded_threaded, :sharded_timed_queue
```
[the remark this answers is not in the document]
svkrieger: Indeed :D

lib/sequel/extensions/connection_metrics.rb
```diff
@@ -0,0 +1,100 @@
...
+      # if assign_connection does not return a connection, the pool is exhausted and the thread has to wait
+      unless conn
+        waiting_since = Time.now
+        @connection_info[thread] = { waiting_since: } unless @connection_info[thread] && @connection_info[thread].key?(:waiting_since)
+      end
+      conn
+    end
+
+    def make_new(server)
+      conn = super
```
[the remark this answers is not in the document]
svkrieger: In this case we need it, because we would have to put super at the end of the method, to return the connection. But `size` will be updated by super, so we have to emit the metric after super has returned. Additionally if super fails, we'd need error handling. So I think this way it's easier.

lib/cloud_controller/runners/puma_runner.rb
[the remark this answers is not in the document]
svkrieger: Would have been nice!

lib/sequel/extensions/connection_metrics.rb
svkrieger: I renamed it to `puma_worker`. Currently the CC workers do not emit those metrics, but if they did in the future it might be confusing indeed.

lib/cloud_controller/db.rb
svkrieger: Good point! PR is open for the docs.

lib/cloud_controller/metrics/prometheus_updater.rb
svkrieger: Hmm somehow `hold_duration` sounds more intuitive to me as the action of holding connections is being measured. Don't have a strong opinion here though.

Samze: 👍  no strong opinions either - will leave it up to you.

spec/unit/lib/sequel/extensions/connection_metrics_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+
+      it 'increments the connection pool timeouts metric and raises' do
+        expect(prometheus_updater).to receive(:increment_gauge_metric).with(:cc_db_connection_pool_timeouts_total, labels: { process_type: 'puma_worker' })
+
+        expect { db.pool.send(:acquire, thread) }.to raise_error(Sequel::PoolTimeout)
+      end
+
+      it 'emits the time the thread waited for the connection' do
+        db.pool.instance_variable_set(:@connection_info, db.pool.instance_variable_get(:@connection_info).merge!({ thread => { waiting_since: Time.now - 60 } }))
+        expect(prometheus_updater).to receive(:update_histogram_metric).with(:cc_db_connection_wait_duration_seconds, any_args)
```
philippthun, line 61: Maybe you can test that it's a duration (i.e. `an_instance_of(ActiveSupport::Duration)`).

spec/unit/lib/sequel/extensions/connection_metrics_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+  describe 'assign_connection' do
+    context 'when assign_connection returns a connection' do
+      it 'does not set the waiting_since timestamp' do
+        db.pool.send(:assign_connection, thread)
+        expect(db.pool.instance_variable_get(:@connection_info)[thread]).to be_nil
+      end
+    end
+
+    context 'when assign_connection does not return a connection' do
+      let(:original_max_size) { db.pool.instance_variable_get(:@max_size) }
```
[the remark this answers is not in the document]
svkrieger: Oh good catch! I just realised that's a leftover from a previous version where I did not give each test its own connection pool. I'm gonna remove it, as it is not needed anymore.

spec/unit/lib/sequel/extensions/connection_metrics_spec.rb
svkrieger: Good idea! I replaced all `any_args` and `anything` with more precise evaluations.

philippthun (approved):

PR #3973, opened 2024-09-13
Orphan mitigation for instances/bindings when reaching the "end timeout" for an asynchronous operation

kathap:
When a creation of a service instance runs into the ```broker_client_max_async_poll_duration_minutes ``` timeout, the service broker sets the state to create failed with ```Service Broker failed to provision within the required time```. In such a case the service broker may have kept some metadata, which should be deleted. Therefore adding an orphan mitigation when service instance creation job runs in this timeout and create has state failed.

Adding an orphan mitigation when service instance creation job runs in a timeout and create has state failed
When a creation of a service instance times out and state is create failed, service broker may have kept some metadata, which should be deleted. 

### DIFF
--- a/app/jobs/v3/create_binding_async_job.rb
+++ b/app/jobs/v3/create_binding_async_job.rb
@@ -86,6 +86,7 @@ module VCAP::CloudController
             description: "Service Broker failed to #{operation} within the required time."
           }
         )
+        mitigate_orphaned_binding
       end
 
       private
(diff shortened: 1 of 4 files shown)


svkrieger (approved):
Changes look good! Just a few thoughts:
- Should we add a CATS test for those scenarios?
- I saw that if the broker returns a 202 on the deletion request that the CC stops orphan mitigation. Might be fine, but the deletion could still fail. Though this might be something the service broker should deal with then.

kathap:
We will add cats test for orphan mitigation for state create failed, but not for the timeout scenario (hard to implement in the cats test).

PR #3984, opened 2024-09-20
Make STOPPING state optional

sethboyles:
* new STOPPING state could be viewed as a breaking change

related: https://github.com/cloudfoundry/capi-release/pull/473

### DIFF
--- a/config/cloud_controller.yml
+++ b/config/cloud_controller.yml
@@ -393,3 +393,4 @@ max_manifest_service_binding_poll_duration_in_seconds: 60
 update_metric_tags_on_rename: true
 
 app_log_revision: true
+app_instance_stopping_state: true
--- a/lib/cloud_controller/config_schemas/base/api_schema.rb
+++ b/lib/cloud_controller/config_schemas/base/api_schema.rb
@@ -381,7 +381,8 @@ module VCAP::CloudController
               dataset: String
             },
 
-            update_metric_tags_on_rename: bool
+            update_metric_tags_on_rename: bool,
+            app_instance_stopping_state: bool
           }
         end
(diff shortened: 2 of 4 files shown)


config/cloud_controller.yml
```diff
@@ -393,3 +393,4 @@ max_manifest_service_binding_poll_duration_in_seconds: 60
 update_metric_tags_on_rename: true
 
 app_log_revision: true
+app_stopping_state: true
```
Samze, line 396: **thought** Should this be process_stopping_state? Since the app `state` field is the desired state.

config/cloud_controller.yml
```diff
@@ -393,3 +393,4 @@ max_manifest_service_binding_poll_duration_in_seconds: 60
 update_metric_tags_on_rename: true
```
Samze, line 393: **thought** In terms of docs, is there anything we think we need to add so that its not unexpected that `STOPPING` state won't be available if this flag is off. Is there any precedence here?

sethboyles: I don't think we usually document these types of configurations elsewhere.  We also don't have much documentation about what to expect from the process stats endpoint in general: https://v3-apidocs.cloudfoundry.org/version/3.176.0/index.html#get-a-process

Samze (approved):

config/cloud_controller.yml
Samze: I suppose the docs do say `The state of the instance; valid values are RUNNING, CRASHED, STARTING, STOPPING, DOWN` which is not a lie, just in some TAS deployments stopping won't appear.

PR #3987, opened 2024-09-24
Raise error if a previous packer job attempt is detected

a-b:
This could be helpful if a previous attempt crashed.

### DIFF
--- a/lib/cloud_controller/packager/local_bits_packer.rb
+++ b/lib/cloud_controller/packager/local_bits_packer.rb
@@ -14,11 +14,11 @@ module CloudController
         tmp_dir = VCAP::CloudController::Config.config.get(:directories, :tmpdir)
         local_bits_packer_path = File.join(tmp_dir, "local_bits_packer-#{blobstore_key}")
 
-        begin
-          Dir.mkdir(local_bits_packer_path)
-        rescue StandardError => e
-          raise ConflictError.new("Found a leftover directory that might be from the previous worker’s unfinished job: #{e.message}")
+        if Dir.exist?(local_bits_packer_path)
+          raise ConflictError.new("Found a leftover directory that might be from the previous worker's unfinished job: #{local_bits_packer_path}")
         end
+
+        Dir.mkdir(local_bits_packer_path)
 
         begin
           complete_package_path = match_resources_and_validate_package(local_bits_packer_path, uploaded_package_zip, cached_files_fingerprints)
--- a/lib/cloud_controller/packager/local_bits_packer.rb
+++ b/lib/cloud_controller/packager/local_bits_packer.rb
@@ -4,13 +4,24 @@ require 'cloud_controller/packager/shared_bits_packer'
 
 module CloudController
   module Packager
+    class ConflictError < StandardError
+    end
+
     class LocalBitsPacker
       include Packager::SharedBitsPacker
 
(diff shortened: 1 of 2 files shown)


lib/cloud_controller/packager/local_bits_packer.rb
```diff
@@ -4,20 +4,33 @@
...
       def send_package_to_blobstore(blobstore_key, uploaded_package_zip, cached_files_fingerprints)
         tmp_dir = VCAP::CloudController::Config.config.get(:directories, :tmpdir)
-        Dir.mktmpdir('local_bits_packer', tmp_dir) do |root_path|
-          complete_package_path = match_resources_and_validate_package(root_path, uploaded_package_zip, cached_files_fingerprints)
+        local_bits_packer_path = File.join(tmp_dir, "local_bits_packer-#{blobstore_key}")
+
+        begin
+          Dir.mkdir(local_bits_packer_path)
+        rescue StandardError => e
+          raise ConflictError.new("Found a leftover directory that might be from the previous worker’s unfinished job: #{e.message}")
```
Samze, line 20: **thought** We don't know for sure this error was generated due to a leftover directory since we're rescuing a generic StandardError right? Would it be more precise to explicitly check if the directory exists before throwing a Conflict error?

lib/cloud_controller/packager/local_bits_packer.rb
```diff
@@ -4,20 +4,33 @@
...
+          complete_package_path = match_resources_and_validate_package(local_bits_packer_path, uploaded_package_zip, cached_files_fingerprints)
 
           package_blobstore.cp_to_blobstore(complete_package_path, blobstore_key)
 
           {
             sha1: Digester.new.digest_path(complete_package_path),
             sha256: Digester.new(algorithm: OpenSSL::Digest::SHA256).digest_path(complete_package_path)
           }
+        ensure
+          FileUtils.remove_dir(local_bits_packer_path)
```
Samze, line 33: **suggestion** Curious, what was happening to directories before this change? Also should we have a test for this?

a-b: I added a test that reproduces the scenario where a folder for a specific GUID already exists on the filesystem.

lib/cloud_controller/packager/local_bits_packer.rb
a-b: I appreciate your suggestion and will make the necessary updates.

a-b: Implemented in 042970dc6f70b8c4915cc25bdd91ea5607abfcd2

lib/cloud_controller/packager/local_bits_packer.rb
a-b: This code mimics the cleanup part of  [Dir.mktmpdir](https://rubyapi.org/o/Dir.mktmpdir#method-c-mktmpdir). I guess we can add something to test this behavior, too.

sethboyles (changes requested):
✅ package uploads failed successfully:
1. when a foundation is stuck in a unzip loop, deploying this change will cause the job to finally transition to 'failed'
2. new pushes of malformed zips will also transition to 'failed'

I think we can merge after removing the redundant `remove_dir`

sethboyles (approved):

PR #3988, opened 2024-09-25
Orphan mitigation for instances/bindings in state "create failed"

kathap:
It might be that a service broker that responds with create failed on a service instance/binding creation keeps
some meta data about the resource. The OSBAPI spec recommends that the platform performs orphan mitigation in such a case. Adding orphan mitigation to instance and binding creations, if they are in state create failed.

Adding orphan mitigation to instance and binding creations, if they are in state create failed.
It might be that a service broker that responds with create failed on a service instance/binding creation keeps
some meta data about the resource. The OSBAPI spec recommends that the platform performs orphan mitigation in such a case. 
[3973](https://github.com/cloudfoundry/cloud_controller_ng/pull/3973)

### DIFF
--- a/app/actions/service_credential_binding_app_create.rb
+++ b/app/actions/service_credential_binding_app_create.rb
@@ -32,9 +32,8 @@ module VCAP::CloudController
 
         ServiceBinding.new.tap do |b|
           ServiceBinding.db.transaction do
-            existing_binding = binding
-            binding.destroy if existing_binding
-            VCAP::Services::ServiceBrokers::V2::OrphanMitigator.new.cleanup_failed_bind(binding) if existing_binding
+            binding&.destroy
+            VCAP::Services::ServiceBrokers::V2::OrphanMitigator.new.cleanup_failed_bind(binding) if binding
             b.save_with_attributes_and_new_operation(
               binding_details,
               CREATE_INITIAL_OPERATION
--- a/app/actions/service_credential_binding_app_create.rb
+++ b/app/actions/service_credential_binding_app_create.rb
@@ -32,7 +32,9 @@ module VCAP::CloudController
 
         ServiceBinding.new.tap do |b|
           ServiceBinding.db.transaction do
-            binding.destroy if binding
+            existing_binding = binding
+            binding.destroy if existing_binding
+            VCAP::Services::ServiceBrokers::V2::OrphanMitigator.new.cleanup_failed_bind(binding) if existing_binding
             b.save_with_attributes_and_new_operation(
               binding_details,
               CREATE_INITIAL_OPERATION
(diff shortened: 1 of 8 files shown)


app/actions/v3/service_instance_create_managed.rb
```diff
@@ -39,7 +40,10 @@ def precursor(message:, service_plan:)
 
         ManagedServiceInstance.new.tap do |i|
           ManagedServiceInstance.db.transaction do
-            instance.destroy if instance&.create_failed?
+            if instance&.create_failed?
+              instance.destroy
+              VCAP::Services::ServiceBrokers::V2::OrphanMitigator.new.cleanup_failed_provision(instance)
```
Samze, line 45: **thought** I wonder if this may cause issues for certain Service Brokers since doing this orphan mitigation here means I believe this will queue up two background jobs, one to deprovision and one to provision right? Since these requests are both for the same guid I'm not sure how brokers would response.

kathap: @philippthun pointed out that the guid from ```instance``` (resource that gets deleted) and the guid from ```i``` (instance that gets created) are different, the newly created instance (```i```) gets a new guid. We could add this in a test.

Samze: Makes sense 👍

johha (approved):
LGTM


</details>

app/actions/service_credential_binding_app_create.rb
```diff
@@ -32,7 +32,8 @@ def precursor(service_instance, message:, app: nil, volume_mount_services_enable
 
         ServiceBinding.new.tap do |b|
           ServiceBinding.db.transaction do
-            binding.destroy if binding
+            binding&.destroy
```
svkrieger, line 35: Probably we could use a normal if block here instead of using `binding&` and `if binding`?

svkrieger (commented):
Verified the following behaviour on a foundation:

Operation | Broker   Response | Orphan   mitigation triggered
-- | -- | --
Create   service instance | 500   (sync) | Immediately   and again when create-service is being retried
Poll   create service instance | 200 -   "state failed" | Triggered   when user retries instance creation
  |   |  
Create   app binding | 500   (sync) | Immediately   and again when bind-service is being retried
Poll   create app binding | 200 -   "state failed" | Triggered   when user retries binding creation
  |   |  
Create   service key | 500   (sync) | Immediately   and again when create-service-key is being retried
Poll   create service key | 200 -   "state failed" | Triggered   when user retries key creation
  |   |  
Create   route binding | 500   (sync) | Immediately   and again when bind-route-service is being retried
Poll   create route binding | 200 -   "state failed" | Triggered   when user retries route binding creation


For the synchronous cases I'm not sure if we want to have the orphan mitigation triggered twice. I compared the behaviour in this branch with main and in the main branch orphan mitigation is already triggered automatically when a synchronous provisioning/binding request fails.

kathap:
We previously discussed the potential scenario of double orphan mitigation, which could occur under certain conditions. After review, we agreed that performing orphan mitigation twice is preferable to having no orphan mitigation at all, and it does not introduce any harmful side effects (as for our knowledge so far).

svkrieger (approved):

PR #4018, opened 2024-10-11
Remove mysql 5.7 from unit tests

### LINKED ISSUE #3949: Remove Mysql5.7 support / remove Mysql5.7 from PR validation
## Mysql5.7 is not supported any more, remove support for Mysql5.7
[MySQL@5.7 is not supported any more upstream](https://endoflife.date/mysql) and also not in [cf-deployment](https://github.com/cloudfoundry/cf-deployment/blob/main/texts/deployment-guide.md#:~:text=The%20following%20databases%20are%20not,MySQL%20%3C8.0). PR validation could be sped up by removing the checks for MySQL5.7. Extra lines in migrations, that are needed to be compliant with Mysql5.7, would not be necessary (e.g. removing unique constraint in MySQL 5.7 does not include removing the corresponding index, in higher versions it does).

## Steps to Reproduce

Write a migration to drop a unique constraint, like
```
Sequel.migration do
  up do
    if self.class.name.match?(/mysql/i)
      alter_table :quota_definitions do
        drop_constraint :name, if_exists: true
      end
    elsif self.class.name.match?(/postgres/i)
      alter_table :quota_definitions do
        drop_constraint :quota_definitions_name_key, if_exists: true
      end
    end
  end

  down do
    if self.class.name.match?(/mysql/i)
      alter_table :quota_definitions do
        add_unique_constraint :name, name: :name, if_not_exists: true
      end
    elsif self.class.name.match?(/postgres/i)
      alter_table :quota_definitions do
        add_unique_constraint :name, name: :quota_definitions_name_key, if_not_exists: true
      end
    end
  end
end
```
and open a PR. It will fail with ```Mysql2::Error: Duplicate key name 'name'``` for MySQL5.7.
## Expected result

PR validation runs without check for MySQL5.7, same for Backwards Compatibility Unit Tests.

## Current result
Migrations and migration specs need to be compliant for MySQL5.7

## Possible Fix

Remove support for MySQL5.7 and the MySQL5.7 PR validation checks.

stephanme:

Take out mysql 5.7 from the unit test matrix as it is explicitly not supported for [cf-deployment](https://github.com/cloudfoundry/cf-deployment/blob/main/texts/deployment-guide.md#databases).

Allows to use mysql features and rely on bug fixes that are not available in mysql 5.7.
Example: Dropping of duplicated unique constraints in DB migrations.

Fixes #3949 and unblocks #3952

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -76,7 +76,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image: ["mysql:5.7", "mysql:8.0", "mysql:8.2"]
+        image: ["mysql:8.0", "mysql:8.2"]
     services:
       mysql:
         image: ${{ matrix.image }}
--- a/.github/workflows/unit_tests_backwards_compatibility.yml
+++ b/.github/workflows/unit_tests_backwards_compatibility.yml
@@ -85,7 +85,7 @@ jobs:
       fail-fast: false
       max-parallel: 3
       matrix:
-        image: ["mysql:5.7", "mysql:8.0", "mysql:8.2"]
+        image: ["mysql:8.0", "mysql:8.2"]
     services:
       mysql:
         image: ${{ matrix.image }}


sethboyles:
Our preference would be to hold off on this until May, as we support foundations using MySQL 5.7 until then.

stephanme:
OK, we need then to adapt #3949 to work with mysql 5.7.
Labeled it as blocked until May 2025.

johha:
@sethboyles @Samze Can we proceed on this? `2025-07-31` is over 😉

sethboyles:
Yes, please go ahead.

sethboyles:
(and thank you for your patience)

PR #4021, opened 2024-10-15
Fix flaky test in security_group_unapply_spec

svkrieger:
Just a small fix for a flaky test.

### DIFF
--- a/spec/unit/actions/security_group_apply_spec.rb
+++ b/spec/unit/actions/security_group_apply_spec.rb
@@ -30,8 +30,9 @@ module VCAP::CloudController
         end
 
         it 'updates the latest security group update table' do
+          last_update = AsgLatestUpdate.last_update
           subject.apply_running(security_group, message, visible_space_guids:, all_spaces_visible:)
-          expect(AsgLatestUpdate.last_update).to be > 1.second.ago
+          expect(AsgLatestUpdate.last_update).to be > last_update
         end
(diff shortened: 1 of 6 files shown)


svkrieger:
@moleske as you are the initial author of these tests, could you verify that my change here does not mess with the semantics of the tests? I'm not sure why `> 1.second.ago` was used and whether it is important.

moleske:
Looking at the original PR https://github.com/cloudfoundry/cloud_controller_ng/pull/2832 I think we just wanted to make sure that some time had passed.  This was made for the diego feature of dynamic asg and we wanted to make sure the time in the table was later than the original time.  I don't think there's anything special about 1 second, just a small amount of time we felt was reasonable.  Tagging @xandroc as I see his name on the PR and he may have thoughts since no else on that pr is particularly active in the project anymore

If this solution is working out here, I'd suggest going through the linked pr and fixing other tests that also do `> 1.second.ago` as they are likely flakey as well

xandroc:
👍🏽 I think this is a good change, agree with @moleske it's likely worth updating the remaining tests

svkrieger:
@moleske @xandroc thank you for the feedback. I adjusted the other tests as well.

xandroc:
lgtm!

xandroc (approved):

moleske (approved):

Issue #3559, opened 2023-12-15
Ruby 3.3 support


moleske:
Thanks for submitting an issue to `cloud_controller_ng`. We are always trying to improve! To help us, please fill out the following template.

## Issue

Ruby 3.3 requires some changes based on quick tests with ruby-3.3.0-rc

## Context

Two issues (so far) related to `Regexp.new` expecting 2 arguments but getting three

- [rfc822](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/Gemfile#L37) - (used in cloud controller in these [two](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/vcap/rest_api/message.rb#L61) [places](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/sequel_plugins/vcap_validations.rb#L16) only) https://github.com/mspanc/rfc822/blob/master/lib/rfc822.rb#L16
- [cookiejar](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/Gemfile.lock#L141) - (which is a dependency of em-http-request) https://github.com/dwaite/cookiejar/pull/52

There may be more but this is what I hit so far

## Steps to Reproduce

run `bundle exec rake` while using ruby 3.3

## Expected result

it works

## Current result

it's sad

## Possible Fix

see context

## `name of issue` screenshot

no

moleske:
I think those are the only issues for `bundle exec rake`.  Since there's no bosh ruby package yet I didn't bother trying to hack something for testing with CATS/SITS/BARAS

The terrible hacks I did to test that we definitely shouldn't do are
- manually edit the cookiejar gem at `gems/cookiejar-0.3.3/lib/cookiejar/cookie_validation.rb`
- copy and paste the chunk of code from the rfc822 into our code

see this aptly named branch for what I did (and let's not do what I did on the branch, or at least do it not so quick and dirty)
https://github.com/cloudfoundry/cloud_controller_ng/tree/terrible-things-to-get-to-ruby-3.3

sethboyles:
We're blocked on this bug introduced in Bundler 2.5.17, since the bosh ruby package provides bundler 2.5.22 and BOSH compilation VMs do not have git present:

https://github.com/rubygems/rubygems/issues/8217

philippthun:
Ruby 3.3 support has shipped - main currently runs Ruby 3.3.11. Both blockers noted above are resolved:
[...]

PR #4069, opened 2024-11-04
Remove unused em-http-request

sethboyles:
* this gem is blocking us from upgrading to Ruby 3.3, but it seems we don't actually use it

https://github.com/cloudfoundry/cloud_controller_ng/issues/3559

### DIFF
--- a/lib/vcap/services/api/async_requests.rb
+++ b/lib/vcap/services/api/async_requests.rb
@@ -1,6 +1,4 @@
 # Copyright (c) 2009-2011 VMware, Inc.
-require 'eventmachine'
-require 'em-http-request'
 require 'httpclient'
 
 require 'vcap/services/api/const'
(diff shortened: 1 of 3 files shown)


lib/vcap/services/api/async_requests.rb
```diff
@@ -13,38 +11,6 @@ module Api
...
-        raise UnexpectedResponse.new("Error sending request #{msg.extract.to_json} to gateway #{@url}: #{http.error}") unless http.error.empty?
-
-        code = http.response_header.status.to_i
-        body = http.response
-        [code, body]
-      end
-    end
-  end
-
   module SynchronousHttpRequest
```
Samze, line 14: **question** I see `SynchronousHttpRequest` is used https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/models/services/managed_service_instance.rb#L99 but is `requester` method called elsewhere? Quick grep would suggest maybe not. 

```
rg requester
src/code.cloudfoundry.org/cc-uploader/handlers/upload_build_artifacts/upload_build_artifacts_test.go
131:		Context("when the requester (client) goes away", func() {

src/code.cloudfoundry.org/cc-uploader/handlers/upload_droplet/upload_droplet_test.go
178:		Context("when the requester (client) goes away", func() {

src/cloud_controller_ng/app/models/services/managed_service_instance.rb
98:    def requester

src/github.com/cloudfoundry/blobstore_url_signer/server/handler_test.go
49:			It("writes the signed URL back to requester", func() {
74:			It("writes the signed URL back to requester", func() {
```

Seems like all of the `lib/vcap/services/api` stuff is from the very old v1 service broker contract, which I dont *think* is used right?

Samze (approved):

lib/vcap/services/api/async_requests.rb
sethboyles: I didn't think about it too hard at the time because with this PR I was trying to unblock the Ruby 3.3 upgrade, but I think you're right.  I made another PR that removes the other old services code to keep the priority/purpose of this PR distinct: #4085

PR #4070, opened 2024-11-04
Ruby 3.3

sethboyles:
https://github.com/cloudfoundry/cloud_controller_ng/issues/3559


Samze:
I wonder if we should pre-emptively run https://github.com/cloudfoundry/cf-performance-tests against this change to see if theres any noticeable positive/negative effect.

sethboyles:
Wouldn't hurt, I suppose.  If we are starting to care about pre-release data, maybe we should have cf-performance-tests that run on prereleases, not just after capi is bumped in cf-deployment?

That said, I *think* most of our issues are due to DB queries, so I would be surprised if this makes a significant impact.  If Ruby code turns out to have a big impact maybe we should push the BOSH team get Rust in the bosh ruby package so we can turn on YJIT

PR #4072, opened 2024-11-04
Remove unused code

sethboyles:
I was trying out this gem [debride](https://github.com/seattlerb/debride), which attempts to identify unused code in Ruby.

Due to the dynamic and meta-programmatic nature of Ruby/CCNG, there were a lot of false positives, and these were a few that I believe can safely be removed.


lib/cloud_controller/runner.rb
```diff
@@ -17,7 +17,7 @@
 
 module VCAP::CloudController
   class Runner
-    attr_reader :config, :config_file, :insert_seed_data, :secrets_file
```
a-b, line 20: Is it possible that this is used by a rake task or pipeline that wasn't loaded into the debride analysis context?

sethboyles: There's no reference to it in cloud_controller_ng or capi-release.  [capi-release has references to CCNG's Drain code that debride didn't catch](https://github.com/search?q=repo%3Acloudfoundry%2Fcapi-release%20CloudController%3A%3ADrain&type=code), but as far as I can tell that is the only place CCNG's code is referenced externally (and arguably should be replaced with a rake task)

nookala (approved):

moleske (approved):
My favorite kind of code change, deletion

PR #4081, opened 2024-11-09
Validate checksums when fetching cached files for app packaging

sethboyles:
Although resource matching stores files by their sha1 checksum, when later fetching the resource to include in app packaging, the bits packer does not validate the checksum still matches.  The packer will still include the corrupted file, and users (most likely) won't receive an error until the staging process.  This state is only recoverable by clearing the resource cache, which relies on operator intervention.

With this change, when a bad checksum is found, the cached resource(s) are deleted, and an error is raised, causing faster feedback. Since the cached resources have been deleted, a second push/package upload should succeed.

### DIFF
--- a/lib/cloud_controller/packager/shared_bits_packer.rb
+++ b/lib/cloud_controller/packager/shared_bits_packer.rb
@@ -42,9 +42,19 @@ module CloudController
         cached_resources_dir = File.join(root_path, 'cached_resources_dir')
 
         FileUtils.mkdir(cached_resources_dir)
+        checksum_mismatch = false
         matched_resources.each do |local_destination, file_sha, mode|
-          global_app_bits_cache.download_from_blobstore(file_sha, File.join(cached_resources_dir, local_destination), mode:)
+          file_path = File.join(cached_resources_dir, local_destination)
+          global_app_bits_cache.download_from_blobstore(file_sha, file_path, mode:)
+
+          if Digester.new.digest_path(file_path) != file_sha
+            checksum_mismatch = true
+            global_app_bits_cache.delete(file_sha)
+          end
         end
+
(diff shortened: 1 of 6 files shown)


nookala (approved):

lib/cloud_controller/packager/shared_bits_packer.rb
```diff
@@ -42,9 +42,19 @@ def append_matched_resources(app_packager, cached_files_fingerprints, root_path)
         cached_resources_dir = File.join(root_path, 'cached_resources_dir')
 
         FileUtils.mkdir(cached_resources_dir)
+        checksum_mismatch = false
         matched_resources.each do |local_destination, file_sha, mode|
+          file_path = File.join(cached_resources_dir, local_destination)
           global_app_bits_cache.download_from_blobstore(file_sha, File.join(cached_resources_dir, local_destination), mode:)
```
philippthun, line 48: You can use `file_path` here.

spec/unit/controllers/runtime/app_bits_upload_controller_spec.rb
```diff
@@ -132,7 +132,7 @@ def make_request
           end
 
           context 'with at least one resource and an application' do
-            let(:req_body) { { resources: Oj.dump([{ 'fn' => 'lol', 'sha1' => 'abc', 'size' => 2048 }]), application: valid_zip } }
+            let(:req_body) { { resources: Oj.dump([{ 'fn' => 'lol', 'sha1' => 'da39a3ee5e6b4b0d3255bfef95601890afd80709', 'size' => 2048 }]), application: valid_zip } }
```
a-b, line 135: **question (non-blocking)**: why do we want to have valid sha in units? Aren't we mocking actual library?

a-b (approved):

Samze (approved):

PR #4101, opened 2024-11-20
Deployment updater refactor

sethboyles:

* [X] I have run capi-bara-tests


lib/cloud_controller/deployment_updater/actions/finalize.rb
```diff
@@ -0,0 +1,55 @@
...
+            state: DeploymentModel::DEPLOYED_STATE,
+            status_value: DeploymentModel::FINALIZED_STATUS_VALUE,
+            status_reason: DeploymentModel::DEPLOYED_STATUS_REASON
+          )
+        end
+
+        private
+
+        def promote_deploying_web_process
+          deploying_web_process.update(type: ProcessTypes::WEB)
```
sethboyles, line 31: Not sure if there is a point to this.  Tests do not fail if I comment it out.

lib/cloud_controller/deployment_updater/actions/scale_down_superseded.rb
```diff
@@ -0,0 +1,30 @@
+module VCAP::CloudController
+  module DeploymentUpdater
+    module Actions
+      class ScaleDownSuperseded
```
sethboyles, line 4: This might have to be Superseded+Canceled, as apparently Superseded+Deployed, is a separate case

lib/cloud_controller/deployment_updater/actions/cancel.rb
```diff
@@ -0,0 +1,41 @@
...
+          deployment.db.transaction do
+            app.lock!
+            return unless deployment.lock!.state == DeploymentModel::CANCELING_STATE
+
+            deploying_web_process.lock!
+
+            prior_web_process = Calculators::FindInterimWebProcess.new(deployment).call || app.oldest_web_process
+            prior_web_process.lock!
+
+            prior_web_process.update(instances: deployment.original_web_process_instance_count, type: ProcessTypes::WEB)
```
sethboyles, line 27: again, why update to WEB process here?  Isn't it already of type `web`?

sethboyles:
I think the 'interim' process/ deployment train logic can be revisited.  It doesn't necessarily work as intended.  When having multiple deployments we can get into cases where significantly more instances than intended are being used.  After chaining several deployments together, I was able to get it into a state where there were up to 15 instances at the same time, when my app was scaled to 10 instances and max-in-flight of 1:

Screenshot 2024-11-20 at 4 25 05 PM


(this screenshot is from a pre-refactor version of CC)

sethboyles:
Right now the deployment updater assumes previous deployment train interactions went as intended, rather than taking stock of all current running instances for all web processes. I wonder if instead we can have some logic at the beginning of the deployment updater that takes into account a few things:

* current number of instances for all old (non-deploying) revisions (right now it just looks at the oldest)
* current deployment number of running instances
* current deployment number of starting instances

This was we can rely less on a smooth continual process and more on checking the state of the world each loop, and updating for the next step. So even for cases like the above, where there are more instances than desired, the deployment updater can just correct for the desired number of instances.

We know that if we have # of desired instances at 10, and max-in-flight of 3, we always want 13 running instances (until the final deployment has reached 10 instances)

So the deployment updater can simply count the number of running instances for the final deployment. If it gets 6, we know that the number of old instances (across all old deployments) should be 10 desired instances - 6 new instances + 3 max-in-flight = 7 old instances.

The Updater can then take the known number of old instances (in this case, 15) and then iterate through each deployment with the intention of tearing down 8 instances, going through the oldest deployments first.

lib/cloud_controller/deployment_updater/actions/finalize.rb
sethboyles: I think this is an artifact of 'webish' processes: https://github.com/cloudfoundry/cloud_controller_ng/commit/617fadec7210213afa087806196fa245287e97a0#diff-23c8e0c518bc3fdeaa2541c3275a7c737543c5efc35962d98494aa265c509446R74

I'm going to delete

lib/cloud_controller/deployment_updater/actions/scale_down_superseded.rb
Samze: Do you know why we support superseded+deployed ?https://github.com/cloudfoundry/cloud_controller_ng/blob/f3834d2504f4b47de6cc23366d9483d484f1a8f7/app/actions/deployment_create.rb#L201 Seems like Superseded+Canceled is a more accurate state.

It seems like in the Superseded+Deployed case we don't scale down the deployment until we finalize the deployment https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/cloud_controller/deployment_updater/updater.rb#L172C11-L172C30 . Is that because if our old deployment was half way through, we don't want to scale those down as we'll reduce the overall app instance number before we've rolled out our new deployment? 🤔

lib/cloud_controller/deployment_updater/calculators/find_interim_web_process.rb
```diff
@@ -0,0 +1,37 @@
...
+
+        def call
+          # Find newest interim web process that (a) belongs to a SUPERSEDED (DEPLOYED) deployment and (b) has at least
+          # one running instance.
+          app.web_processes_dataset.
+            qualify.
+            join(:deployments, deploying_web_process_guid: :guid).
+            where(deployments__state: DeploymentModel::DEPLOYED_STATE).
+            where(deployments__status_reason: DeploymentModel::SUPERSEDED_STATUS_REASON).
+            order(Sequel.desc(:created_at), Sequel.desc(:id)).
```
Samze, line 20: **question** It was like this previously, but do you know why the order is important if we're just cleaning these processes up?

lib/cloud_controller/deployment_updater/actions/scale_down_superseded.rb
sethboyles: >  Is that because if our old deployment was half way through, we don't want to scale those down as we'll reduce the overall app instance number before we've rolled out our new deployment? 

That's right, so instead those are finalized and their AI's are kept running.

I haven't quite wrapped my head around why/when we want to supersede + cancel (though looking at the code it seems like we just do that when we deploy over any other state, including prepaused/paused)

lib/cloud_controller/deployment_updater/calculators/find_interim_web_process.rb
sethboyles: This is used in the Cancel action, so the goal here is to find a previous web process to restore state to. So once an interim process is found, then that process is scaled UP to the desired number of instances immediately.

lib/cloud_controller/deployment_updater/actions/cancel.rb
Samze, line 27: **question** Again existing behaviour, but does this mean if I do a deployment, then supersede it with another deployment, and then cancel my new deployment, it will revert to the superseded deployment with the desired number of instances?

lib/cloud_controller/deployment_updater/actions/finalize.rb
Samze: I see the name and behaviour were taken from the old implementation, interesting that the method name here really doesn't match the code.

The only thing I can think of is that would updating this trigger some Model hooks that would do something with diego thats important?

lib/cloud_controller/deployment_updater/actions/finalize.rb
```diff
@@ -0,0 +1,55 @@
...
+        end
+
+        def update_non_web_processes
+          return if deploying_web_process.revision.nil?
```
Samze, line 35: **question** Do we know what happens here if revisions are disable for the app?

lib/cloud_controller/deployment_updater/actions/scale_down_old_process.rb
```diff
@@ -0,0 +1,35 @@
...
+
+        def initialize(deployment, process, instances_to_scale_down)
+          @deployment = deployment
+          @app = deployment.app
+          @process = process
+          @instances_to_scale_down = instances_to_scale_down
+        end
+
+        def call
+          if process.instances <= deployment.max_in_flight && is_interim_process?(process)
```
Samze, line 15: **thought** Does this need to be here? Will this not happen as part of the eventual finalize deployment?

lib/cloud_controller/deployment_updater/calculators/instances_to_scale_down.rb
```diff
@@ -0,0 +1,18 @@
+module VCAP::CloudController
+  module DeploymentUpdater
+    module Calculators
+      class InstancesToScaleDown
```
Samze, line 4: **thought** I wonder if the calculators should just be dealing with numbers rather than deployments/processes, might make them easier to test, especially if the logic gets more knarly in the future e.g.

```Ruby
def initialize(max_in_flight, instances)
```

lib/cloud_controller/deployment_updater/calculators/instances_to_scale_up.rb
```diff
@@ -0,0 +1,17 @@
+module VCAP::CloudController
+  module DeploymentUpdater
+    module Calculators
+      class InstancesToScaleUp
```
Samze, line 4: **thought** Do we think theres value in these being separate classes? Vs a Scale calculator with 2 methods (scale up/ scale down)

lib/cloud_controller/deployment_updater/calculators/all_instances_routable.rb
```diff
@@ -0,0 +1,26 @@
...
+    module Calculators
+      class AllInstancesRoutable
+        attr_reader :deployment, :logger
+
+        def initialize(deployment, logger)
+          @deployment = deployment
+          @logger = logger
+        end
+
+        def call
```
Samze, line 12: **suggestion** I'm not sure about having `call` be the general purpose function call for all the classes here, especially when they return different thing (bool/ints/processes). Are you envisioning some future refactor benefit?

lib/cloud_controller/deployment_updater/calculators/find_interim_web_process.rb
sethboyles: That said if it does revert to the most recent Deployment, I don't see anywhere that updates the worker/non-web processes to have the same droplet 🤔

lib/cloud_controller/deployment_updater/actions/cancel.rb
sethboyles: I believe so.  I'll have to try it to be sure.  I'm beginning to think we need a doc to track the actual behavior of different Deployment scenarios

lib/cloud_controller/deployment_updater/actions/finalize.rb
sethboyles: I think so.

lib/cloud_controller/deployment_updater/actions/scale_down_old_process.rb
sethboyles: It would probably eliminate the need for this action entirely

lib/cloud_controller/deployment_updater/calculators/all_instances_routable.rb
```diff
@@ -0,0 +1,26 @@
+module VCAP::CloudController
+  module DeploymentUpdater
+    module Calculators
+      class AllInstancesRoutable
```
Samze, line 4: **nit** I wonder if this could be less specific, something like `AllInstancesReady` or `AllInstancesRunning` ?

lib/cloud_controller/deployment_updater/actions/scale.rb
```diff
@@ -0,0 +1,63 @@
...
+            instances_to_scale_down = Calculators::InstancesToScaleDown.new(deployment, oldest_web_process_with_instances).call
+
+            deployment.update(
+              last_healthy_at: Time.now,
+              state: DeploymentModel::DEPLOYING_STATE,
+              status_value: DeploymentModel::ACTIVE_STATUS_VALUE,
+              status_reason: DeploymentModel::DEPLOYING_STATUS_REASON
+            )
+
+            if deploying_web_process.instances >= deployment.original_web_process_instance_count
```
Samze, line 39: **thought** Is it the case that when `instances_to_scale_up == 0` and `instances_to_scale_down == 0` we're done? I wonder if we can use the Calculator here to push some more logic in there

lib/cloud_controller/deployment_updater/actions/cleanup_web_processes.rb
```diff
@@ -0,0 +1,21 @@
+module VCAP::CloudController
+  module DeploymentUpdater
+    module Actions
+      class CleanupWebProcesses
+        attr_reader :deployment, :app, :protected_process
+
+        def initialize(deployment, process)
```
Samze, line 7: *nit* process -> exclude_process/protected_process

lib/cloud_controller/deployment_updater/calculators/all_instances_routable.rb
sethboyles: I think the main benefit here is to prevent mental overhead of trying to remember which Calculator has which specific method, when they are all designed to do one thing and one thing only. It's simpler (and less redundant feeling) than calling something like `AllInstancesRoutable.new(...).all_instances_routable?`

That said, I'm open to a different name/pattern.

Historically, `call` is usually chosen for these kinds of single-purpose Ruby objects as it allows them to be used interchangeably with lambdas, as they are executed with `call`.

lib/cloud_controller/deployment_updater/calculators/instances_to_scale_up.rb
sethboyles: My starting guidelines with this refactor were to create dumb single-functions classes that only responded to `call` so I could focus on reorganizing the logic.  I think it's fine to combine these.

lib/cloud_controller/deployment_updater/actions/cleanup_web_processes.rb
```diff
@@ -0,0 +1,21 @@
+module VCAP::CloudController
+  module DeploymentUpdater
+    module Actions
+      class CleanupWebProcesses
```
Samze, line 4: **thought** I wonder if it would be cleaner if `CleanupWebProcesses`/`ScaleDownOldProcess`/`ScaleDownSuperceded` could be one class whose responsibly is to handle the scaling/destruction of processes. `ProcessScaler` or something

lib/cloud_controller/deployment_updater/actions/cancel.rb
Samze: I wonder if thats expected, I think you could argue it should instead restore to the last known complete deployment/ original web processes.

lib/cloud_controller/deployment_updater/calculators/all_instances_routable.rb
Samze: Makes sense, then I if its`call` is basically determined by whether the refactor is single funciton classes or not.

lib/cloud_controller/deployment_updater/actions/finalize.rb
sethboyles: Yep, that's how I understand it.

lib/cloud_controller/deployment_updater/actions/scale_down_superseded.rb
sethboyles: Well, if you superseded over and over, the DeploymentUpdater will start to take down instances from those superseded deployments one by one as it brings up the latest/current deployment.

However, there is definitely a bug in the implementation.  You can definitely get an excessive number of running instances by having multiple superseding deployments.  See this comment for an example: https://github.com/cloudfoundry/cloud_controller_ng/pull/4101#issuecomment-2489827283

Samze:
Once this refactor is complete, to gain a bit more confidence we may want to explicitly run pre-merge.
* https://github.com/cloudfoundry/capi-bara-tests/blob/main/baras/deployments.go
* https://github.com/cloudfoundry/cf-acceptance-tests/blob/main/v3/deployment.go

lib/cloud_controller/deployment_updater/actions/finalize.rb
Samze: non-blocking, but did we ever understand this?

lib/cloud_controller/deployment_updater/actions/scale.rb
```diff
@@ -0,0 +1,77 @@
...
+              Finalize.new(deployment).call
+              return
+            end
+
+            ScaleDownCanceled.new(deployment).call
```
Samze, line 38: **meganit** I initially read this as "the scaling down is to be cancelled" rather than "scale down the canceled" process. Is there maybe a more expressive name? Also do we care at this level that they are cancelled? Could this be `ScaleDownIntermProcesses` or something?

lib/cloud_controller/deployment_updater/actions/finalize.rb
sethboyles: So this method doesn't have anything to do with the droplet, just gets the command from the revision. If revisions are disabled, then the processes get their commands from the Procfile or app manifest, probably in the apply manifest job (not sure about when the Procfile gets read).

sethboyles:
Both the capi-bara-tests and CATs tests (in the deployments tests, I didn't run the entire suites)

lib/cloud_controller/deployment_updater/actions/scale.rb
sethboyles: Well, it's not scaling down the Interim processes, only the Superseded and Canceled Processes.

I don't think this is a real case, btw--I can't get it to occur in my bosh-lite. Not completely sure, though

sethboyles: Nevermind.  It's if you quickly deploy over a canceling deployment: https://github.com/cloudfoundry/cloud_controller_ng/blob/deployment-updater-refactor/spec/unit/actions/deployment_create_spec.rb#L451-L460

PR #4102, opened 2024-11-21
Expire packages in "Docker" lifecycle

jochenehret:
Mark Docker packages as "EXPIRED" so that they can be properly cleaned up.

There is no place where Docker packages are marked as "EXPIRED". This can lead to a high number of packages in "READY" state. So call the `expire_packages!` method when staging completed.

I haven't found a better place for this method call. There is no "PackageUploadHandler" class for the Docker lifecycle.


jochenehret:
Alternative solution: https://github.com/cloudfoundry/cloud_controller_ng/pull/4105

PR #4106, opened 2024-11-25
Use raw sql for stack check

gururajsh:
We currently run stack_checker on the first cc startup node to verify we don't load deprecated stacks. If a deprecated stack is found, an error is raised and startup of the initial cc instance fails.

To prevent potential inconsistencies between models and the db, we'd like to avoid loading models at all before we run migrations. This means we need to use raw sql to both verify the stacks table exists (we're not in a clean install) and to check whether any deprecated stacks exists. 


lib/tasks/stack_check.rake
```diff
@@ -7,9 +7,8 @@ namespace :stacks do
     next unless db.table_exists?(:buildpack_lifecycle_data)
 
     RakeConfig.config.load_db_encryption_key
```
Samze, line 9: **suggestion/question** Since we're not loading the models, I don't think we need need to load the encryption keys and nothing in the stacks table is encrypted either.

lib/tasks/stack_check.rake
```diff
@@ -7,9 +7,8 @@ namespace :stacks do
     next unless db.table_exists?(:buildpack_lifecycle_data)
```
Samze, line 7: **suggestion** I don't think we need to check for the `buildpack_lifecycle_data` table anymore since we no longer need to load it because we don't load the stacks model.

lib/tasks/stack_check.rake
Samze: @xandroc / @sethboyles can you confirm if my understanding is right?

spec/tasks/stack_check_spec.rb
```diff
@@ -52,6 +52,7 @@
       before do
         allow(db_double).to receive(:table_exists?).with(:stacks).and_return true
         allow(VCAP::CloudController::DB).to receive(:connect).and_return(db_double)
+        allow(db_double).to receive(:fetch).with('SELECT * FROM stacks WHERE name LIKE ? ', 'cflinuxfs3').and_return('1')
```
Samze, line 55: *nit* Would this actually return 1? Wouldn't it return an entire row? Maybe it doesn't matter.

lib/tasks/stack_check.rake
sethboyles: Yep, we shouldn't need this (or `buildpack_lifecycle_data`) anymore

lib/cloud_controller/check_stacks.rb
```diff
@@ -19,11 +20,12 @@ def validate_stacks
     def validate_stack(deprecated_stack)
       configured_stacks = @stack_config.stacks
       deprecated_stack_in_config = (configured_stacks.find { |stack| stack['name'] == deprecated_stack }).present?
-
       return if deprecated_stack_in_config
 
-      deprecated_stack_in_db = VCAP::CloudController::Stack.first(name: deprecated_stack).present?
-
+      deprecated_stack_in_db = false
+      @db.fetch('SELECT * FROM stacks WHERE name LIKE ? ', deprecated_stack) do |_row|
```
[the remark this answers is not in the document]
Samze: Nit comment so non-blocking, but `empty?` method works on any dataset and `fetch` returns a dataset.

```
db.fetch('SELECT * from stacks where name like ?', "cflinuxfs4").empty?
=> false

.db.fetch('SELECT * from stacks where name like ?', "cflinuxfs3").empty?
=> true
```

PR #4124, opened 2024-12-03
Granular max-in-flight updates

sethboyles:
* Scale action now compares number of starting/running non-routable instances to max-in-flight, instead of waiting for all instances to become routable
* Scale action does not continue scaling up when any instances are 'unhealthy' (e.g. 'crashed', 'down', etc) as it's difficult to determine if unhealthy instances belong to the 'max in flight' group
* Number of desired nondeploying instances now recalculated each iteration instead of decrementing by 'max-in-flight' without checking if it's in a correct state. This mitigates bugs where deployment trains (continually creating new deployments before the previous had completed) could result in number app instances exceeding the max-in-flight limit.

See this thread for some of the issues this PR aims to solve: https://github.com/cloudfoundry/cloud_controller_ng/pull/4101#issuecomment-2489827283

* [x]  I have run BARAS

### DIFF
--- a/lib/cloud_controller/deployment_updater/actions/scale.rb
+++ b/lib/cloud_controller/deployment_updater/actions/scale.rb
@@ -6,6 +6,7 @@ module VCAP::CloudController
   module DeploymentUpdater
     module Actions
       class Scale
+        HEALTHY_STATES = [VCAP::CloudController::Diego::LRP_RUNNING, VCAP::CloudController::Diego::LRP_STARTING].freeze
         attr_reader :deployment, :logger, :app
 
         def initialize(deployment, logger)
@@ -17,14 +18,15 @@ module VCAP::CloudController
         def call
           deployment.db.transaction do
             return unless deployment.lock!.state == DeploymentModel::DEPLOYING_STATE
-            return unless all_instances_routable?
+
+            return unless can_scale? || can_downscale?
 
             app.lock!
+
             oldest_web_process_with_instances.lock!
             deploying_web_process.lock!
 
             deployment.update(
-              last_healthy_at: Time.now,
               state: DeploymentModel::DEPLOYING_STATE,
               status_value: DeploymentModel::ACTIVE_STATUS_VALUE,
               status_reason: DeploymentModel::DEPLOYING_STATUS_REASON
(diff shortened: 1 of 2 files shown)


lib/cloud_controller/deployment_updater/actions/scale.rb
```diff
@@ -36,36 +38,86 @@ def call
...
         end
 
-        def all_instances_routable?
-          instances = instance_reporters.all_instances_for_app(deployment.deploying_web_process)
-          instances.all? { |_, val| val[:state] == VCAP::CloudController::Diego::LRP_RUNNING && val[:routable] }
-        rescue CloudController::Errors::ApiError # the instances_reporter re-raises InstancesUnavailable as ApiError
-          logger.info("skipping-deployment-update-for-#{deployment.guid}")
-          false
+        def starting_instances
+          healthy_instances.reject { |_, val| val[:state] == VCAP::CloudController::Diego::LRP_RUNNING && val[:routable] }
```
Samze, line 104: **nit** Would it be more explicit to filter by `LRP_STARTING` rather then reject `LRP_RUNNING`?

lib/cloud_controller/deployment_updater/actions/scale.rb
```diff
@@ -36,36 +38,86 @@ def call
...
+        def healthy_instances
+          reported_instances.select { |_, val| HEALTHY_STATES.include?(val[:state]) }
+        end
+
+        def unhealthy_instances
+          reported_instances.reject { |_, val| HEALTHY_STATES.include?(val[:state]) }
+        end
+
+        def reported_instances
+          @reported_instances = instance_reporters.all_instances_for_app(deployment.deploying_web_process)
```
Samze, line 120: **nit** This could use `deploying_web_process` method over `deployment.deploying_web_process`

lib/cloud_controller/deployment_updater/actions/scale.rb
```diff
@@ -36,36 +38,86 @@ def call
...
+          false
+        end
+
+        def desired_non_deploying_instances
+          [deployment.original_web_process_instance_count - routable_instances.count, 0].max
         end
 
         def desired_new_instances
-          [deploying_web_process.instances + deployment.max_in_flight, deployment.original_web_process_instance_count].min
+          [deploying_web_process.instances + deployment.max_in_flight - starting_instances.count, deployment.original_web_process_instance_count].min
```
Samze, line 88: **question/non-blocking** How come we need to minus the starting instance count here to determine the desired new instances? Shouldn't those instances not matter for the desired state?

lib/cloud_controller/deployment_updater/actions/scale.rb
```diff
@@ -36,36 +38,86 @@ def call
...
-          instances = instance_reporters.all_instances_for_app(deployment.deploying_web_process)
-          instances.all? { |_, val| val[:state] == VCAP::CloudController::Diego::LRP_RUNNING && val[:routable] }
-        rescue CloudController::Errors::ApiError # the instances_reporter re-raises InstancesUnavailable as ApiError
-          logger.info("skipping-deployment-update-for-#{deployment.guid}")
-          false
+        def starting_instances
+          healthy_instances.reject { |_, val| val[:state] == VCAP::CloudController::Diego::LRP_RUNNING && val[:routable] }
+        end
+
+        def routable_instances
```
Samze, line 107: **thought/non-blocking** `routable_instances/healthy_instances/unhealthy_instances/reported_instances` are all methods for the newly deploying process right? I wonder if theres any way we can be more explicit about that. As technically those functions could apply to the older web process. We could prefix with `new` or `deploying` or something....

lib/cloud_controller/deployment_updater/actions/scale.rb
sethboyles: I think that would be the case if `deploying_web_process.instances` were the actual state, but it's the desired state. So we subtract starting instances to determine how many of the new max-in-flight instances are finished, and only increase by that amount (since they are no longer 'in flight')

lib/cloud_controller/deployment_updater/actions/scale.rb
sethboyles: That's a good point.  I do worry the prefixes might hurt readability if every method is called `deploying_abc_instances` or something.  

Maybe an alternative is to separate the scale down and scale up into separate actions entirely?  I think their logic is fairly disentangled at this point.

lib/cloud_controller/deployment_updater/actions/scale.rb
sethboyles: I changed this line to only look at routable_instances, but added a safeguard that prevents the scaling action from acting if there are fewer routable instances than there should be

Samze (approved):

PR #4113, opened 2024-11-28
Implement shadow user creation by org manager

svkrieger:
Thanks for contributing to cloud_controller_ng. To speed up the process of reviewing your pull request please provide us with:

* A short explanation of the proposed change:

* An explanation of the use cases your change solves

* Links to any other associated PRs

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)

PR #4171, opened 2025-01-14
Handle rate limited UAA

svkrieger:
UAA might return 429 responses. This is especially relevant for the newly added UAA shadow user creation https://github.com/cloudfoundry/cloud_controller_ng/pull/4113
This changes the error code the CC returns to the user if UAA responded with 429. Before that the user got back a 503 "CF-UaaUnavailable". Now CC returns 429 "CF-UaaRateLimited" including a random Retry-After header. (UAA does not return a Retry-After header yet).

### DIFF
--- a/lib/cloud_controller/uaa/uaa_client.rb
+++ b/lib/cloud_controller/uaa/uaa_client.rb
@@ -102,6 +102,14 @@ module VCAP::CloudController
       raise e unless e.info['error'] == 'scim_resource_already_exists'
 
       { 'id' => e.info['user_id'] }
+    rescue CF::UAA::BadResponse => e
+      unless e.message == 'invalid status response: 429'
+        logger.error("UAA request for creating a user failed: #{e.inspect}")
+        raise UaaUnavailable
+      end
+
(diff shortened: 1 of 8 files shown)


lib/cloud_controller/uaa/uaa_client.rb
```diff
@@ -102,6 +102,14 @@ def create_shadow_user(username, origin)
       raise e unless e.info['error'] == 'scim_resource_already_exists'
 
       { 'id' => e.info['user_id'] }
+    rescue CF::UAA::BadResponse => e
+      unless e.message == 'invalid status response: 429'
+        logger.error("UAA request for creating a user failed: #{e.inspect}")
+        raise e
```
philippthun, line 108: raise `UaaUnavailable` instead

philippthun (approved):

lib/cloud_controller/uaa/uaa_client.rb
svkrieger: good catch!

PR #4174, opened 2025-01-16
Remove foreign key contraint in table job_warnings on 'fk_jobs_id'

svkrieger:

Deletion of foreign key constraint in `job_warnings` table on `fk_jobs_id`. This also deletes the column `fk_jobs_id`.
This will drastically increase the speed of the cleanup job of the pollable jobs table.

The job_warnings table was created in a wrong way. There are two columns 'job_id' and 'fk_jobs_id'. There is a foreign key constraint on 'fk_jobs_id', which references jobs(id). CC coding never inserts data into 'fk_jobs_id', but only into 'job_id'. So the foreign key constraint is not needed but causes way slower cleanup of pollable jobs, which can even run into DB statement timeouts and cause an ever growing table. CC already takes care of deleting job_warnings when jobs are deleted.

### DIFF
--- a/db/migrations/20250116144231_remove_unnecessary_fk_in_job_warnings.rb
+++ b/db/migrations/20250116144231_remove_unnecessary_fk_in_job_warnings.rb
@@ -1,0 +1,9 @@
+Sequel.migration do
+  up do
+    unless foreign_key_list(:job_warnings).empty?
+      alter_table :job_warnings do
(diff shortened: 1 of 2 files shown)


philippthun (approved):

philippthun:
> CC already takes care of deleting job_warnings when jobs are deleted.

Unfortunately this is only partially true. See #4175.

PR #4187, opened 2025-01-23
Reduce logcache timeout from 250 to 10s

WeiQuan0605:

A shorter timeout might prevent the CF PUSH timeout because CLI will continue to poll for process stats and a retry may succeed.

If a Cloud Controller request to Log Cache takes too long,  `Connection to Log Cache timed out ` error will be triggered earlier.

### DIFF
--- a/lib/logcache/client.rb
+++ b/lib/logcache/client.rb
@@ -44,13 +44,14 @@ module Logcache
 
     private
 
-    def with_request_error_handling(_source_guid)
+    def with_request_error_handling(source_guid)
       tries ||= 3
       start_time = Time.now
-
-      # convert to milliseconds to get more precise information
-      time_taken = (Time.now - start_time) * 1000
-      logger.info("Response time: #{time_taken.round(2)} ms")
+      time_taken = (Time.now - start_time) * 1000 # convert to milliseconds to get more precise information
+      logger.info("Response time: #{time_taken.round(2)} ms",
+                  { source_id: source_guid,
+                    event: 'log_cache_request',
+                    time_taken_in_ms: time_taken.round(2) })
       yield
     rescue StandardError => e
       raise CloudController::Errors::ApiError.new_from_details('ServiceUnavailable', 'Connection to Log Cache timed out') if e.is_a?(GRPC::DeadlineExceeded)
--- a/lib/logcache/client.rb
+++ b/lib/logcache/client.rb
@@ -16,13 +16,13 @@ module Logcache
           "#{host}:#{port}",
           GRPC::Core::ChannelCredentials.new(client_ca, client_key, client_cert),
           channel_args: { GRPC::Core::Channel::SSL_TARGET => tls_subject_name },
-          timeout: 250
+          timeout: 10
         )
       else
         @service = Logcache::V1::Egress::Stub.new(
           "#{host}:#{port}",
           :this_channel_is_insecure,
-          timeout: 250
+          timeout: 10
         )
       end
     end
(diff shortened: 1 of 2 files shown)


lib/logcache/client.rb
```diff
@@ -46,6 +46,11 @@ def container_metrics(source_guid:, start_time:, end_time:, envelope_limit: DEFA
 
     def with_request_error_handling(_source_guid)
       tries ||= 3
+      start_time = Time.now
+
+      # convert to milliseconds to get more precise information
+      time_taken = (Time.now - start_time) * 1000
+      logger.info("Response time: #{time_taken.round(2)} ms")
```
stephanme, line 53: Is there a way to add the response time additionally in a parsable formation (json data like we do for vcap-request-id)?

WeiQuan0605: I can take a look

lib/logcache/client.rb
```diff
@@ -44,8 +44,14 @@ def container_metrics(source_guid:, start_time:, end_time:, envelope_limit: DEFA
 
     private
 
-    def with_request_error_handling(_source_guid)
+    def with_request_error_handling(source_guid)
       tries ||= 3
+      start_time = Time.now
+      time_taken = (Time.now - start_time) * 1000 # convert to milliseconds to get more precise information
+      logger.info("Response time: #{time_taken.round(2)} ms",
```
svkrieger, line 51: Other client's logs look rather like `lrp.instances.request`, so I'd prefer the actual log to look like `logcache.response`. `event` could then be removed from the hash.

lib/logcache/client.rb
```diff
@@ -44,8 +44,14 @@ def container_metrics(source_guid:, start_time:, end_time:, envelope_limit: DEFA
...
+                  { source_id: source_guid,
+                    event: 'log_cache_request',
+                    time_taken_in_ms: time_taken.round(2) })
```
svkrieger, line 54:
```suggestion
                    response_time_ms: response_time_ms })
```
Just personal naming preference.

lib/logcache/client.rb
```diff
@@ -44,8 +44,14 @@ def container_metrics(source_guid:, start_time:, end_time:, envelope_limit: DEFA
 
     private
 
-    def with_request_error_handling(_source_guid)
+    def with_request_error_handling(source_guid)
       tries ||= 3
+      start_time = Time.now
+      time_taken = (Time.now - start_time) * 1000 # convert to milliseconds to get more precise information
```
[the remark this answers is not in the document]
WeiQuan0605: I agree with integer, and should the variable name be similar or consistent with other codes? For example, like: [time_taken in request_log](https://github.com/cloudfoundry/cloud_controller_ng/blob/b66529f815b7b3a6cf477d1fc229066c314be255/middleware/request_logs.rb#L16)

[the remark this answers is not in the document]
svkrieger: Ah I'm fine with `time_taken_in_ms` if we used it somewhere else already

lib/logcache/client.rb
svkrieger: No I meant replacing the log message `"Response time: #{time_taken.round(2)} ms"` with `"logcache.response"`.

WeiQuan0605: I've already updated the code based on your review

lib/logcache/client.rb
```diff
@@ -44,8 +44,13 @@ def container_metrics(source_guid:, start_time:, end_time:, envelope_limit: DEFA
 
     private
 
-    def with_request_error_handling(_source_guid)
+    def with_request_error_handling(source_guid)
       tries ||= 3
+      start_time = Time.now
+      time_taken_in_ms = ((Time.now - start_time) * 1000).to_i # convert to milliseconds to get more precise information
```
svkrieger, line 50: Calculation of `time_taken_in_ms` and `logger.info` must be placed after `yield`.

svkrieger (changes requested):

svkrieger (approved):

PR #4191, opened 2025-01-28
Add draft ADR for bigint id column migration

johha:
To avoid overflows of the `id` primary key column we should migrate it to `bigint`.
The purpose of this ADR is to find a good approach and to document all relevant decisions.

Feedback is highly appreciated 🙂

### DIFF
--- a/decisions/0013-migrating-int-to-bigint-for-primary-keys.md
+++ b/decisions/0013-migrating-int-to-bigint-for-primary-keys.md
@@ -55,13 +55,14 @@ The migration will be conducted in multiple steps to ensure minimal risk.
 - If foreign keys exist, drop the corresponding foreign key constraints.
 - Remove the sync triggers.
 - Switch the primary key by renaming `id_bigint` to `id`.
+- Add PK constraint on `id` column and configure `id` generator.
 - If foreign keys exist, rename `id_bigint` to `id` in referencing tables accordingly.
 - Everything is done in a single transaction to ensure consistency.
 
 ### Database Specifics
 
 #### PostgreSQL
-The default value of the `id` column could be either a sequence (for older PostgreSQL versions) or an identity column (for newer PostgreSQL versions).
+The default value of the `id` column could be either a sequence (for PostgreSQL versions < 10) or an identity column (for newer PostgreSQL versions).
 This depends on the version of PostgreSQL which was used when the table was initially created.
 The migration script needs to handle both cases.
 
@@ -70,6 +71,7 @@ Unknown at this point. Further investigation is needed.
 
 ### Rollback Mechanism
 The old `id` column is no longer retained, as the `CHECK` constraint ensures correctness during migration.
+Step 3b (switch over) will be executed in a single transaction and will be rolled back if any issues occur.
 If unexpected issues occur, the migration will fail explicitly, requiring intervention.
 If rollback is needed, either backups could be restored or the migration needs to be reverted manually.
 
@@ -89,7 +91,7 @@ For the `events` table there is a default cleanup interval of 31 days. Therefore
 ### Positive Consequences
 
 - Future-proofing the schema for tables with high record counts.
-- Minimal downtime due to the phased approach.
+- Minimal locking during step 3b (actual migration) could result in slower queries or minimal downtime.
 - A standardized process for similar migrations across the database.
 
 ### Negative Consequences
@@ -98,10 +100,6 @@ For the `events` table there is a default cleanup interval of 31 days. Therefore
 - Potentially long runtimes for backfilling data in case tables have millions of records.
 - Requires careful coordination across multiple CAPI/CF-Deployment versions.
 - If backfilling encounters edge cases (e.g., missing cleanup jobs), the migration may be delayed until operators intervene.
-
-## Rollback Plan
-Keep the original id column intact until the final cleanup phase.
-If issues arise during the switch, revert the primary key to id.
 
 ## Alternatives Considered
 
--- a/decisions/0013-migrating-int-to-bigint-for-primary-keys.md
+++ b/decisions/0013-migrating-int-to-bigint-for-primary-keys.md
@@ -21,7 +21,7 @@ This migration must:
 - Handle tables with millions of records efficiently. 
 - Provide a safe rollback mechanism in case of issues during the migration.
 - Be reusable for other tables in the future.
-- Ensure that migration only is executed when id_bigint is fully populated.
+- Ensure that migration only is executed when the new `id_bigint` column is fully populated.
 
 The largest tables in a long-running foundation are `events`, `delayed_jobs`, `jobs`, and `app_usage_events`.
 
@@ -42,31 +42,40 @@ The migration will be conducted in multiple steps to ensure minimal risk.
 #### Step 2 - Backfill
 - Use a batch-processing script (e.g. a delayed job) to populate `id_bigint` for existing rows in both the primary table and, if applicable, all foreign key references.
 - Table locks will be avoided by using a batch processing approach.
-- In case the table has configurable cleanup duration, the backfill job will only process records which are beyond the cleanup duration to reduce the number of records to be processed. 
+- In case the table has a configurable cleanup duration, the backfill job will only process records which are beyond the cleanup duration to reduce the number of records to be processed. 
 - Backfill will be executed outside the migration due to its potentially long runtime.
-- If necessary the backfill will run over multiple releases.
+- If necessary the backfill will run for multiple weeks to ensure all records are processed.
 
 #### Step 3a - Migration Pre Check
-- Double check that `id_bigint` is fully populated before proceeding.
+- Add a `CHECK` constraint to verify that id_bigint is fully populated (`id_bigint == id & id_bigint != NULL`).
 - In case the backfill is not yet complete or the `id_bigint` column is not fully populated the migration exits gracefully and is retried in the next deploy.
 #### Step 3b - Actual Migration
-- Retain the `id` column as a backup by renaming it to `id_old`.
-- If foreign keys exist, rename the corresponding id columns in referencing tables to `<ref>_id_old` as well.
+- Remove the `CHECK` constraint once verified.
+- Drop the primary key constraint on id.
+- If foreign keys exist, drop the corresponding foreign key constraints.
 - Remove the sync triggers.
 - Switch the primary key by renaming `id_bigint` to `id`.
 - If foreign keys exist, rename `id_bigint` to `id` in referencing tables accordingly.
-- Create new sync triggers for keeping `id` and `id_bigint` in sync for newly inserted records.
 - Everything is done in a single transaction to ensure consistency.
-#### Step 4 - Cleanup
-- Remove remaining sync triggers.
-- Drop the `id_old` column after verifying stability and ensuring the system fully relies on the migrated id column
-- If foreign keys exist, drop the `id_old` columns in referencing tables after verification.
+
+### Database Specifics
+
+#### PostgreSQL
+The default value of the `id` column could be either a sequence (for older PostgreSQL versions) or an identity column (for newer PostgreSQL versions).
+This depends on the version of PostgreSQL which was used when the table was initially created.
+The migration script needs to handle both cases.
+
+#### MySQL
+Unknown at this point. Further investigation is needed.
 
 ### Rollback Mechanism
-Retain the original `id` column until the cleanup phase, allowing for reversion to the previous state if needed.
+The old `id` column is no longer retained, as the `CHECK` constraint ensures correctness during migration.
+If unexpected issues occur, the migration will fail explicitly, requiring intervention.
+If rollback is needed, either backups could be restored or the migration needs to be reverted manually.
 
 ### Standardized Approach
-Write reusable scripts for adding `id_bigin`t, setting up triggers, backfilling data, and verifying migration readiness.
+Write reusable scripts for adding `id_bigint`, setting up triggers, backfilling data, and verifying migration readiness.
+These scripts can be reused for other tables in the future.
 
 ### Release Strategy
 
@@ -74,7 +83,7 @@ Steps 1-2 will be released as a cf-deployment major release to ensure that the d
 Steps 3-4 will be released as a subsequent cf-deployment major release to complete the migration.  
 Between these releases there should be a reasonable time to allow the backfill to complete.
 
-For the `events` table there is a default cleanup interval of 31 days. Therefore for the `events` table the gap between the releases should be around 60 days.
+For the `events` table there is a default cleanup interval of 31 days. Therefore, for the `events` table the gap between the releases should be around 60 days.
 
 ## Consequences
 ### Positive Consequences
@@ -86,7 +95,7 @@ For the `events` table there is a default cleanup interval of 31 days. Therefore
 ### Negative Consequences
 
 - Increased complexity in the migration process.
-- Potentially long runtimes for backfilling data in tables with millions of records.
+- Potentially long runtimes for backfilling data in case tables have millions of records.
 - Requires careful coordination across multiple CAPI/CF-Deployment versions.
 - If backfilling encounters edge cases (e.g., missing cleanup jobs), the migration may be delayed until operators intervene.
 
@@ -122,7 +131,7 @@ Cons: Requires downtime, locks the table for the duration of the migration, and 
 Reason Rejected: Downtimes are unacceptable for productive foundations.
 
 
-## Example Migration Scripts With PostgreSQL Syntax
+## Example Migration Scripts With PostgreSQL Syntax For `events` Table
 
 ### Step 1 - Preparation
 ```sql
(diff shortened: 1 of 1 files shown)


decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,243 @@
...
+```
+PG::SequenceGeneratorLimitExceeded: ERROR:  nextval: reached maximum value of sequence "events_id_seq"
+```
+The goal is to migrate these primary key `id` columns from `int` to `bigint` without causing downtime and to ensure compatibility across PostgreSQL and MySQL databases.
+This migration must:
+- Avoid downtime since the CF API is actively used in production.
+- Handle tables with millions of records efficiently. 
+- Provide a safe rollback mechanism in case of issues during the migration.
+- Be reusable for other tables in the future.
+- Ensure that migration only is executed when id_bigint is fully populated.
```
jochenehret, line 24: Didn't understand the last requirement, is "id_bigint" the new index or column? It's likely explained below...

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,243 @@
...
+- Add a new column `id_bigint` of type `bigint` to the target table. If the `id` column is referenced as a foreign key in other tables, also add an `<ref>_id_bigint` column in those referencing tables.
+- Create triggers to keep `id_bigint` in sync with `id` when new records are inserted.
+- Add constraints and indexes to `id_bigint` as required to match primary key and foreign key requirements.
+
+#### Step 2 - Backfill
+- Use a batch-processing script (e.g. a delayed job) to populate `id_bigint` for existing rows in both the primary table and, if applicable, all foreign key references.
+- Table locks will be avoided by using a batch processing approach.
+- In case the table has configurable cleanup duration, the backfill job will only process records which are beyond the cleanup duration to reduce the number of records to be processed. 
+- Backfill will be executed outside the migration due to its potentially long runtime.
+- If necessary the backfill will run over multiple releases.
```
jochenehret, line 47: What "release" is meant here?

jochenehret (commented):
Are there any considerations for the Ruby sequel library necessary?

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,243 @@
...
+#### Step 3a - Migration Pre Check
+- Double check that `id_bigint` is fully populated before proceeding.
+- In case the backfill is not yet complete or the `id_bigint` column is not fully populated the migration exits gracefully and is retried in the next deploy.
+#### Step 3b - Actual Migration
+- Retain the `id` column as a backup by renaming it to `id_old`.
+- If foreign keys exist, rename the corresponding id columns in referencing tables to `<ref>_id_old` as well.
+- Remove the sync triggers.
+- Switch the primary key by renaming `id_bigint` to `id`.
+- If foreign keys exist, rename `id_bigint` to `id` in referencing tables accordingly.
+- Create new sync triggers for keeping `id` and `id_bigint` in sync for newly inserted records.
```
stephanme, line 58: Why is this needed? From now on, the `id_old` is not used anymore. If the cleanup step is fast, I could imagine to merge step 3b and 4. Removing `id_old` has the advantage that we prove that it is not used anymore.
PK constraints on `id` (formerly `id_bigint`) should prove that the id migration is really complete.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,243 @@
...
+- Everything is done in a single transaction to ensure consistency.
+#### Step 4 - Cleanup
+- Remove remaining sync triggers.
+- Drop the `id_old` column after verifying stability and ensuring the system fully relies on the migrated id column
+- If foreign keys exist, drop the `id_old` columns in referencing tables after verification.
+
+### Rollback Mechanism
+Retain the original `id` column until the cleanup phase, allowing for reversion to the previous state if needed.
```
stephanme, line 66: I see, that's why keeping `id_old`. But this rollback relies on working triggers (i.e. to keep `id_old` in sync with the new bigint `id`). Not sure if this is worth it if we can "prove" by constraints that `id_bigint` is completely filled.

I can't imagine of realistic scenarios for which a rollback would help and we don't cheat ourselves by relying on migration coding/triggers for the rollback. Maybe performance problems of bigint compared to int.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,243 @@
...
+UPDATE events
+SET id_bigint = events.id
+FROM batch
+WHERE events.id = batch.id;
+```
+
+
+### Step 3a - Migration Pre Check
+```sql
+SELECT COUNT(*) FROM events WHERE id_bigint IS NULL;
```
stephanme, line 172: Maybe better try to add an `id==id_bigint` constraint (which is indirectly also a PK constraint for `id_bigint`). This ensures that the migration condition is not broken between pre-check and actual migration.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,243 @@
...
+The migration will be conducted in multiple steps to ensure minimal risk.
+#### Step 1 - Preparation
+- Add a new column `id_bigint` of type `bigint` to the target table. If the `id` column is referenced as a foreign key in other tables, also add an `<ref>_id_bigint` column in those referencing tables.
+- Create triggers to keep `id_bigint` in sync with `id` when new records are inserted.
+- Add constraints and indexes to `id_bigint` as required to match primary key and foreign key requirements.
+
+#### Step 2 - Backfill
+- Use a batch-processing script (e.g. a delayed job) to populate `id_bigint` for existing rows in both the primary table and, if applicable, all foreign key references.
+- Table locks will be avoided by using a batch processing approach.
+- In case the table has configurable cleanup duration, the backfill job will only process records which are beyond the cleanup duration to reduce the number of records to be processed.
```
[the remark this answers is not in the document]
johha: I've found records in our ccdbs from 2017. Not sure why those weren't cleaned up but for those case a backfill would be helpful.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,246 @@
...
+
+#### Step 3a - Migration Pre Check
+- Add a `CHECK` constraint to verify that id_bigint is fully populated (`id_bigint == id & id_bigint != NULL`).
+- In case the backfill is not yet complete or the `id_bigint` column is not fully populated the migration exits gracefully and is retried in the next deploy.
+#### Step 3b - Actual Migration
+- Remove the `CHECK` constraint once verified.
+- Drop the primary key constraint on id.
+- If foreign keys exist, drop the corresponding foreign key constraints.
+- Remove the sync triggers.
+- Switch the primary key by renaming `id_bigint` to `id`.
```
stephanme, line 57: Missing: Add PK constraint on `id` column and configure id generator.
(this is part of the SQL stmts below)

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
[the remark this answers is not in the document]
stephanme: If step 1 and step 2 shall run at the same time, the idea with limiting the backfill to really old records makes sense.

Alternative would be:
- run step 1
- wait cleanup duration
- run step 2 (w/o special handling, i.e. backfill all remaining records that don't yet have a `id_bigint` value)

The result should be the same. Its more a tradeoff between more complex backfill vs. more complex rollout procedure. Implementation detail.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,246 @@
...
+- If foreign keys exist, rename `id_bigint` to `id` in referencing tables accordingly.
+- Everything is done in a single transaction to ensure consistency.
+
+### Database Specifics
+
+#### PostgreSQL
+The default value of the `id` column could be either a sequence (for older PostgreSQL versions) or an identity column (for newer PostgreSQL versions).
```
stephanme, line 64: Which Postgres version introduced identity column? If CC still supports version w/o indentity column, we may have adapt step 3b (configure id generator).

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,246 @@
...
+This depends on the version of PostgreSQL which was used when the table was initially created.
+The migration script needs to handle both cases.
+
+#### MySQL
+Unknown at this point. Further investigation is needed.
+
+### Rollback Mechanism
+The old `id` column is no longer retained, as the `CHECK` constraint ensures correctness during migration.
+If unexpected issues occur, the migration will fail explicitly, requiring intervention.
```
stephanme, line 73: Could add: step 3b (the switch over) runs in one transaction and will rollback in case of error.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,246 @@
...
+Steps 3-4 will be released as a subsequent cf-deployment major release to complete the migration.  
+Between these releases there should be a reasonable time to allow the backfill to complete.
+
+For the `events` table there is a default cleanup interval of 31 days. Therefore, for the `events` table the gap between the releases should be around 60 days.
+
+## Consequences
+### Positive Consequences
+
+- Future-proofing the schema for tables with high record counts.
+- Minimal downtime due to the phased approach.
```
stephanme, line 92: No downtime instead of minimal downtime. Maybe: Minimal table lock time due to ...

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Negative Consequences
+
+- Increased complexity in the migration process.
+- Potentially long runtimes for backfilling data in case tables have millions of records.
+- Requires careful coordination across multiple CAPI/CF-Deployment versions.
+- If backfilling encounters edge cases (e.g., missing cleanup jobs), the migration may be delayed until operators intervene.
+
+## Rollback Plan
+Keep the original id column intact until the final cleanup phase.
```
stephanme, line 103: This not part anymore of the migration. plan.

stephanme (changes requested):

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
johha: Generators were introduces with postgres 10 which is already end of life

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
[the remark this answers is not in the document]
johha: I also see this as a implementation detail.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
johha: fixed

Samze:
Just as a heads up, even though MySQL 5.7 is EOL. Since its still available from certain cloud providers, we technically still support it until 2025-07-31.

So if we're thinking of doing this before that date, we'd need to test against that MySQL version.

johha:
> Just as a heads up, even though MySQL 5.7 is EOL. Since its still available from certain cloud providers, we technically still support it until 2025-07-31.
> 
> So if we're thinking of doing this before that date, we'd need to test against that MySQL version.

👍 According to the mysql docs there should no downtime when renaming columns ([source](https://dev.mysql.com/doc/refman/5.7/en/innodb-online-ddl-operations.html#online-ddl-column-operations)). Still someone who has access to a copy of a productive ccdb should try out the migration procedure.

stephanme (dismissed):

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,244 @@
...
+#### PostgreSQL
+The default value of the `id` column could be either a sequence (for PostgreSQL versions < 10) or an identity column (for newer PostgreSQL versions).
+This depends on the version of PostgreSQL which was used when the table was initially created.
+The migration script needs to handle both cases.
+
+#### MySQL
+Unknown at this point. Further investigation is needed.
+
+### Rollback Mechanism
+The old `id` column is no longer retained, as the `CHECK` constraint ensures correctness during migration.
```
nookala, line 73: Why not rename the old id column and drop later if required?

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,244 @@
...
+- Add constraints and indexes to `id_bigint` as required to match primary key and foreign key requirements.
+
+#### Step 2 - Backfill
+- Use a batch-processing script (e.g. a delayed job) to populate `id_bigint` for existing rows in both the primary table and, if applicable, all foreign key references.
+- Table locks will be avoided by using a batch processing approach.
+- In case the table has a configurable cleanup duration, the backfill job will only process records which are beyond the cleanup duration to reduce the number of records to be processed. 
+- Backfill will be executed outside the migration due to its potentially long runtime.
+- If necessary the backfill will run for multiple weeks to ensure all records are processed.
+
+#### Step 3a - Migration Pre Check
```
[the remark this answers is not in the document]
johha: Step 3a is just a check to ensure we can start the migration.
The plan is to ship 3a & 3b with the same CAPI/CF-D release. We will have one major release for steps 1 & 2 which prepares the migration and another major for steps 3a & 3b for the actual migration.
Between those two releases there should be reasonable time - maybe something like 60 days.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,244 @@
...
+- Add a `CHECK` constraint to verify that id_bigint is fully populated (`id_bigint == id & id_bigint != NULL`).
+- In case the backfill is not yet complete or the `id_bigint` column is not fully populated the migration exits gracefully and is retried in the next deploy.
+#### Step 3b - Actual Migration
+- Remove the `CHECK` constraint once verified.
```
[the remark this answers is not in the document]
johha: Adding the check constraint in 3a would fail if there are still missing entries.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
```diff
@@ -0,0 +1,244 @@
...
+#### Step 2 - Backfill
+- Use a batch-processing script (e.g. a delayed job) to populate `id_bigint` for existing rows in both the primary table and, if applicable, all foreign key references.
+- Table locks will be avoided by using a batch processing approach.
+- In case the table has a configurable cleanup duration, the backfill job will only process records which are beyond the cleanup duration to reduce the number of records to be processed. 
+- Backfill will be executed outside the migration due to its potentially long runtime.
+- If necessary the backfill will run for multiple weeks to ensure all records are processed.
+
+#### Step 3a - Migration Pre Check
+- Add a `CHECK` constraint to verify that id_bigint is fully populated (`id_bigint == id & id_bigint != NULL`).
+- In case the backfill is not yet complete or the `id_bigint` column is not fully populated the migration exits gracefully and is retried in the next deploy.
```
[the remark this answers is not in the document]
johha: Maybe it would be actually better if the migration fails and then the next deploy would retry it.
We'll check this again.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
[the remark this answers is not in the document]
Samze: Whats the behaviour if someone goes from a cfd version before this to a cfd version that contains all the changes? 

Will the migration still work just with a long deploy time?

sethboyles:
@johha what do you think about making this an opt-in process?  Either the whole process, or at least making Phase 3 able to be deferred at Operator discretion.

We're still in the process of validating this, but we don't think this is a concern for us in the short to medium term.  The fact that this is a multi phase process will make it hard to coordinate (as we don't use CFD).  We'd prefer to choose when we have our foundations run these migrations.  

I'm not sure how easy it would be to set up a second migrations directory or schema_migrations table, and what issues might arise from having it uncertain if `id` columns are int or bigint (I think that the ORM will take care of that for us, though).

The tables mentioned (`events`, `delayed_jobs`, `jobs`, and `app_usage_events`) are probably ones that could easily have their increment markers reset, so if it became an issue for us I think that's the first thing we would look toward.

johha:
@sethboyles I understand that smaller foundations will probably never run into this problem and thus would like have to option to skip the bigint migration.
The easiest way to implement this would be to add a config flag and if this flag is set all related migrations will result in a no-op (they would still appear in the `schema_versions` table). Operators choosing this option would be required to apply those migrations manually when the id gets close to a overflow.
Therefore I'd prefer a opt-out process so that operators need to take a conscious decision.
Would this be a valid option for you? 

We're also discussing currently if we should change the default type for id columns to bigint when creating new schemas. With this at least newly created foundations would never run into a overflow situation. Any thoughts on that?

sethboyles:
I think that's a potential solution.  There's still a chance that instead of _never_ running into this problem it won't be a problem for 5-10 years, and then we will need to think if we want some different solution.  We're still collecting numbers to validate this.

For new foundations to have bigint as the default instead of int, that seems reasonable to me.

decisions/0013-migrating-int-to-bigint-for-primary-keys.md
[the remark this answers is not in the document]
stephanme: +1 for letting the migration fail if backfill is not yet complete.

This will be the case when an update over multiple major cf-d versions gets applied so that step 1 and step 3 run directly one after each other (i.e. step 2 = backfill doesn't have a chance to run).
We need a procedure how to resolve such situations. E.g. running the backfill manually as a rake task.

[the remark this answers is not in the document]
johha: I adjusted migration step 3 so that it will fail.

PR #4205, opened 2025-02-03
Add prometheus metrics to cc workers

svkrieger:
Add a small webserver to the first worker process on cc-worker VM, which publishes cc-worker metrics in a prometheus endpoint. Metrics of all worker processes are stored in the same DirectFileStore, so that we only need one endpoint to publish all their metrics.

Currently we only publish metrics on the API VMs. We are also interested in metrics on worker VMs e.g. DB connection metrics. With this change metrics are being published in the same way and can be scraped using the `prom_scraper` job of the loggregator release.

Configuration in CAPI: https://github.com/cloudfoundry/capi-release/pull/507

### DIFF
--- a/lib/cloud_controller/config_schemas/base/worker_schema.rb
+++ b/lib/cloud_controller/config_schemas/base/worker_schema.rb
@@ -37,6 +37,10 @@ module VCAP::CloudController
 
             log_audit_events: bool,
 
+            directories: {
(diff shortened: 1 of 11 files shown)


jochenehret (commented):
Looks good, only question is if http is ok or if we should require mTLS.

jochenehret (approved):

PR #4212, opened 2025-02-08
Deployment scale refactor

sethboyles:

### DIFF
--- a/lib/cloud_controller/deployment_updater/actions/down_scaler.rb
+++ b/lib/cloud_controller/deployment_updater/actions/down_scaler.rb
@@ -1,0 +1,48 @@
+require 'cloud_controller/deployment_updater/actions/scale_down_old_process'
+module VCAP::CloudController
+  module DeploymentUpdater
+    module Actions
+      class DownScaler
+        attr_reader :deployment, :logger, :app
+
+        def initialize(deployment, logger, target_total_instance_count, routable_instance_count)
+          @deployment = deployment
+          @app = deployment.app
+          @logger = logger
+          @target_total_instance_count = target_total_instance_count
+          @routable_instance_count = routable_instance_count
+        end
+
+        def scale_down
+          instances_to_reduce = non_deploying_web_processes.map(&:instances).sum - desired_non_deploying_instances
+
+          return if instances_to_reduce <= 0
+
+          non_deploying_web_processes.each do |process|
+            if instances_to_reduce < process.instances
+              ScaleDownOldProcess.new(deployment, process, process.instances - instances_to_reduce).call
+              break
+            end
+
+            instances_to_reduce -= process.instances
+            ScaleDownOldProcess.new(deployment, process, 0).call
+          end
+        end
+
+        def can_downscale?
+          non_deploying_web_processes.map(&:instances).sum > desired_non_deploying_instances
+        end
+
+        def desired_non_deploying_instances
+          [@target_total_instance_count - @routable_instance_count, 0].max
+        end
+
+        private
+
+        def non_deploying_web_processes
+          app.web_processes.reject { |process| process.guid == deployment.deploying_web_process.guid }.sort_by { |p| [p.created_at, p.id] }
+        end
+      end
+    end
+  end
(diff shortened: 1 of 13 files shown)


spec/unit/lib/cloud_controller/deployment_updater/actions/up_scaler_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+        expect(up_scaler.finished_scaling?).to be true
+      end
+
+      it 'returns false if current instances have not reached the target number of instances' do
+        summary = double(starting_instances_count: 0, routable_instances_count: 5, healthy_instances_count: 5, unhealthy_instances_count: 0)
+        up_scaler = DeploymentUpdater::Actions::UpScaler.new(deployment, logger, 6, summary)
+        expect(up_scaler.finished_scaling?).to be false
+      end
+
+      # TODO: Currently, deployments transition to COMPLETED even if there are some instances that have not started and become routable
```
sethboyles, line 104: I was able to reproduce this by adding this to Dora's healthcheck endpoint:
```
 if ENV['CF_INSTANCE_INDEX'].to_i==3
     status 500
     return
 end

```

Given 6 instances and max of flight=2, the deployment will complete before the instance has started/become routable.  Will fix in a future PR

lib/cloud_controller/metrics/prometheus_updater.rb
```diff
@@ -1,4 +1,5 @@
 require 'prometheus/client'
+require 'prometheus/client/data_stores/direct_file_store'
```
sethboyles, line 2: this was so I could use the lightweight_spec_helper for some tests

lib/cloud_controller/deployment_updater/actions/down_scaler.rb
```diff
@@ -0,0 +1,48 @@
...
+            ScaleDownOldProcess.new(deployment, process, 0).call
+          end
+        end
+
+        def can_downscale?
+          non_deploying_web_processes.map(&:instances).sum > desired_non_deploying_instances
+        end
+
+        def desired_non_deploying_instances
+          [target_total_instance_count - @routable_instance_count, 0].max
```
Samze, line 37: **nit** Should we be consistent on direct access vs attribute reader here?

lib/cloud_controller/deployment_updater/actions/scale.rb
```diff
@@ -1,127 +1,70 @@
...
-            end
-          end
-        end
-
-        private
+            down_scaler.scale_down if down_scaler.can_downscale?
 
-        def scale_down_old_processes
-          instances_to_reduce = non_deploying_web_processes.map(&:instances).sum - desired_non_deploying_instances
+            return true if up_scaler.finished_scaling?
```
Samze, line 43: **question** Can we return true here if the downscaler is still doing its thing?

lib/cloud_controller/diego/reporters/instances_reporter.rb
```diff
@@ -105,6 +106,17 @@ def crashed_instances_for_app(process)
         raise CloudController::Errors::InstancesUnavailable.new(e)
       end
 
+      def instance_count_summary(process)
```
Samze, line 109: *praise* big fan of pushing this out of the updater and into the reporter

lib/cloud_controller/deployment_updater/actions/scale.rb
sethboyles: Yes, I think so.  In rolling deployments, the updater will call `Finalize` after this returns true, deleting the old web processes anyway.  In other words, once all instances are scaled up for the new web process, we can just wrap things up.

That's the logic I preserved in the refactor, anyway.  I noted a few bugs I found with it.  We'll also have to think if that still makes sense for Canary deployments when this will return true on each step.  I suppose that question is--does a Canary deployment transition to PAUSED when all Canary instances are brought up or do we also have to wait for old instances to be brought down?

Samze (approved):

PR #4213, opened 2025-02-11
Better error message and support logs for /v3/service_offerings and plans

kathap:
Users may be confused why they receive a 422 with error message "Invalid service plan. Ensure that the service plan exists, is available, and you have access to it." [unprocessable_service_plan!](https://github.com/cloudfoundry/cloud_controller_ng/blob/f10b4c29bd27bbd89faf3d86109abd4d544b2a97/app/controllers/v3/service_instances_controller.rb#L423) when trying to create a service instance.
The root cause of this can have different origins:
1) service plan does not exist
2) user can not see the service plan in any org
3) user can see service plan in one of his orgs but service plan got removed and is still in use by service instances (service plan is set to active=false)
4) user sees the plan in of his orgs but not in the one he is creating the service instance in 

Split the error into different errors depending on the root cause. Error message will stay the same if user can not see the service plan in any org. 
For case 1 and 2 error message will stay:
```Invalid service plan. Ensure that the service plan exists, is available, and you have access to it.```
For case 3 error message will now be:
```Invalid service plan. The service plan 'service_plan' with guid 'some-guid' has been removed from the service broker's catalog. It is not possible to create new service instances using this plan.```
For case 4 error message will now be:
```Invalid service plan. This could be due to a space-scoped broker which is offering the service plan 'service_plan.name' with guid 'service_plan.guid' in another space or that the plan is not enabled in this organization. Ensure that the service plan is visible in your current space 'space.name' with guid 'space.guid'.```

More fine granulated error messages, helping the user to solve the error better.

### DIFF
--- a/app/controllers/v3/service_instances_controller.rb
+++ b/app/controllers/v3/service_instances_controller.rb
@@ -255,9 +255,8 @@ class ServiceInstancesV3Controller < ApplicationController
   def create_managed(message, space:)
     service_plan = ServicePlan.first(guid: message.service_plan_guid)
     unprocessable_service_plan! unless service_plan_valid?(service_plan)
-    unavailable_service_plan!(service_plan.name, service_plan.guid) unless service_plan_active?(service_plan)
-    service_plan_not_visible_in_space!(service_plan.name, service_plan.guid, space.name, space.guid) \
-      unless service_plan_exists_in_space?(service_plan, space)
+    unavailable_service_plan!(service_plan) unless service_plan_active?(service_plan)
+    service_plan_not_visible_in_space!(service_plan, space) unless service_plan_exists_in_space?(service_plan, space)
 
     action = V3::ServiceInstanceCreateManaged.new(user_audit_info, message.audit_hash)
     VCAP::CloudController::ManagedServiceInstance.db.transaction do
@@ -415,9 +414,8 @@ class ServiceInstancesV3Controller < ApplicationController
 
     service_plan = ServicePlan.first(guid: message.service_plan_guid)
     unprocessable_service_plan! unless service_plan_valid?(service_plan)
-    unavailable_service_plan!(service_plan.name, service_plan.guid) unless service_plan_active?(service_plan)
-    service_plan_not_visible_in_space!(service_plan.name, service_plan.guid, space.name, space.guid) \
-      unless service_plan_exists_in_space?(service_plan, service_instance.space)
+    unavailable_service_plan!(service_plan) unless service_plan_active?(service_plan)
+    service_plan_not_visible_in_space!(service_plan, space) unless service_plan_exists_in_space?(service_plan, service_instance.space)
     invalid_service_plan_relation! unless service_plan.service == service_instance.service
   end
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by jochenehret — APPROVED** [review]

File: app/controllers/v3/service_instances_controller.rb [hunk 1]
```diff
@@ -424,6 +437,16 @@ def unprocessable_service_plan!
     unprocessable!('Invalid service plan. Ensure that the service plan exists, is available, and you have access to it.')
   end
 
+  def unavailable_service_plan!(service_plan, service_plan_guid)
+    unprocessable!("Invalid service plan. The service plan '#{service_plan}' with guid '#{service_plan_guid}' has been removed from the service broker's catalog." \
+                   'It is not possible to create new service instances using this plan.')
+  end
+
+  def service_plan_not_visible_in_space!(service_plan, service_plan_guid, space_name, space_guid)
+    unprocessable!("Invalid service plan. Ensure that the service plan '#{service_plan}' with guid '#{service_plan_guid}' \
```
svkrieger, line 446: Should we make this message more clear? I.e. mention that this could be due to a space-scoped broker offering this plan or that the plan is not enabled in this org?

app/controllers/v3/service_instances_controller.rb
```diff
@@ -424,6 +437,16 @@ def unprocessable_service_plan!
     unprocessable!('Invalid service plan. Ensure that the service plan exists, is available, and you have access to it.')
   end
 
+  def unavailable_service_plan!(service_plan, service_plan_guid)
+    unprocessable!("Invalid service plan. The service plan '#{service_plan}' with guid '#{service_plan_guid}' has been removed from the service broker's catalog." \
+                   'It is not possible to create new service instances using this plan.')
```
svkrieger, line 442: Space missing after `.`

app/controllers/v3/service_instances_controller.rb
```diff
@@ -254,7 +254,10 @@ def create_user_provided(message)
 
   def create_managed(message, space:)
     service_plan = ServicePlan.first(guid: message.service_plan_guid)
-    unprocessable_service_plan! unless service_plan_valid?(service_plan, space)
+    unprocessable_service_plan! unless service_plan_valid?(service_plan)
+    unavailable_service_plan!(service_plan.name, service_plan.guid) unless service_plan_active?(service_plan)
+    service_plan_not_visible_in_space!(service_plan.name, service_plan.guid, space.name, space.guid) \
```
svkrieger, line 259: This syntax is hard to read. I'd prefer passing `(service_plan, space)` to the method to make the line shorter and move the unless into the same line. Same for other method calls.

app/controllers/v3/service_instances_controller.rb
```diff
@@ -394,17 +396,26 @@ def admin?
...
   end
 
   def raise_if_invalid_service_plan!(service_instance, message)
     return unless message.service_plan_guid
 
     service_plan = ServicePlan.first(guid: message.service_plan_guid)
-    unprocessable_service_plan! unless service_plan_valid?(service_plan, service_instance.space)
+    unprocessable_service_plan! unless service_plan_valid?(service_plan)
+    unavailable_service_plan!(service_plan) unless service_plan_active?(service_plan)
+    service_plan_not_visible_in_space!(service_plan, space) unless service_plan_exists_in_space?(service_plan, service_instance.space)
```
svkrieger, line 418: `space` is not defined. This leads to a 500 error.
```suggestion
    service_plan_not_visible_in_space!(service_plan, service_instance.space) unless service_plan_exists_in_space?(service_plan, service_instance.space)
```

spec/request/service_instances_spec.rb
```diff
@@ -1260,7 +1260,10 @@ def check_filtered_instances(*instances)
             expect(last_response).to have_status_code(422)
             expect(parsed_response['errors']).to include(
               include({
-                        'detail' => 'Invalid service plan. Ensure that the service plan exists, is available, and you have access to it.',
+                        'detail' => 'Invalid service plan. This could be due to a space-scoped broker which is offering the service plan ' \
```
svkrieger, line 1263: Would be good to have the same tests for the update case (PATCH). If we had those, above bug which results in a 500 would have been detected.

app/controllers/v3/service_instances_controller.rb
```diff
@@ -424,6 +435,18 @@ def unprocessable_service_plan!
...
   end
 
+  def unavailable_service_plan!(service_plan)
+    unprocessable!("Invalid service plan. The service plan '#{service_plan.name}' with guid '#{service_plan.guid}' has been removed from the service broker's catalog. " \
+                   'It is not possible to create new service instances using this plan.')
+  end
+
+  def service_plan_not_visible_in_space!(service_plan, space)
+    unprocessable!('Invalid service plan. This could be due to a space-scoped broker which is offering the service plan ' \
+                   "'#{service_plan.name}' with guid '#{service_plan.guid} in another space or that the plan " \
```
svkrieger, line 445:
```suggestion
                   "'#{service_plan.name}' with guid '#{service_plan.guid}' in another space or that the plan " \
```

svkrieger (changes requested):

spec/request/service_instances_spec.rb
kathap: Good point, will add it

svkrieger (approved):

PR #4230, opened 2025-02-21
Log service plan visibility on create or update service instances

kathap:
Enhancement of https://github.com/cloudfoundry/cloud_controller_ng/pull/4230.

For case 1 and 2 a log info is additionally added to have it easier to find the root cause when users get 422 CF-UnprocessableEntity error. Due to security reasons, this is just added as log info and not as error message.

Make it easier to find the root cause when users get 422 CF-UnprocessableEntity error
#4213

### DIFF
--- a/app/controllers/v3/service_instances_controller.rb
+++ b/app/controllers/v3/service_instances_controller.rb
@@ -254,6 +254,7 @@ class ServiceInstancesV3Controller < ApplicationController
 
   def create_managed(message, space:)
     service_plan = ServicePlan.first(guid: message.service_plan_guid)
+    log_service_plan_visibility(service_plan)
     unprocessable_service_plan! unless service_plan_valid?(service_plan)
     unavailable_service_plan!(service_plan) unless service_plan_active?(service_plan)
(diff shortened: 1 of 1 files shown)


app/controllers/v3/service_instances_controller.rb
```diff
@@ -431,7 +428,15 @@ def unprocessable_space!
...
 
-  def unprocessable_service_plan!
+  def service_plan_does_not_exist!
+    logger.info('Service Plan does not exist.')
+    unprocessable!('Invalid service plan. Ensure that the service plan exists, is available, and you have access to it.')
+  end
+
+  def service_plan_not_visible_to_user!(service_plan)
+    user = VCAP::CloudController::SecurityContext.current_user
+    logger.info("Service Plan with guid '#{service_plan.guid}' and id '#{service_plan.id}' is not visible in current org to user with guid '#{user.guid}'.")
```
svkrieger, line 438: 'in current org' is not correct here as the user can see the plan if he has access to it in any org

svkrieger (approved):

PR #4231, opened 2025-02-21
delayed_jobs lock_with_read_ahead for postgres

stephanme:

- monkey patch for delayed_jobs to allow configurable lock method
- config via jobs.read_ahead (number of jobs to read)
  - postgres default = 0 (use lock_with_for_update)
  - mysql default = 5 (use lock_with_read_ahead)
- jobs.read_ahead = 0: lock_with_for_update = lock jobs using SELECT FOR UPDATE
- jobs.read_ahead > 0: lock_with_read_ahead = optimistic locking of jobs
- add rake task generate_load to create jobs

Using lock_with_for_update showed severe performance problems on postgres when there is a high number of jobs in the queue (>50k). Job processing starves due to row locks and temp IO gets high.
lock_with_read_ahead performs much better under such load situations.

Performance on low/normal job load is comparable.

### Checklist

### DIFF
--- a/lib/delayed_job/delayed_worker.rb
+++ b/lib/delayed_job/delayed_worker.rb
@@ -4,6 +4,9 @@ require 'puma'
 require 'prometheus/middleware/exporter'
 
 class CloudController::DelayedWorker
+  DEFAULT_READ_AHEAD_POSTGRES = 0
+  DEFAULT_READ_AHEAD_MYSQL = Delayed::Worker::DEFAULT_READ_AHEAD
+
   def initialize(options)
     @queue_options = {
       min_priority: ENV.fetch('MIN_PRIORITY', nil),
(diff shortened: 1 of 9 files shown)


lib/delayed_job/delayed_worker.rb
```diff
@@ -56,6 +58,16 @@ def get_initialized_delayed_worker(config, logger)
...
     Delayed::Worker.sleep_delay = config.get(:jobs, :global, :worker_sleep_delay_in_seconds)
     Delayed::Worker.logger = logger
+    if ::Sequel::Model.db.database_type == :mysql
+      read_ahead = config.get(:jobs, :read_ahead) || Delayed::Worker::DEFAULT_READ_AHEAD
+      # lock for update is not configurable for mysql
+      read_ahead = Delayed::Worker::DEFAULT_READ_AHEAD if read_ahead <= 0
+    else
+      # read_ahead 0 = lock for update (default for postgres)
+      read_ahead = config.get(:jobs, :read_ahead) || 0
+      read_ahead = 0 if read_ahead < 0
```
jochenehret, line 68: Would it make sense to extract the 0 constant to a variable?
E.g. `Delayed::Worker::DEFAULT_READ_AHEAD_POSTGRES`?

lib/delayed_job/delayed_worker.rb
```diff
@@ -25,6 +25,8 @@ def start_working
     setup_metrics(config) if @publish_metrics
     BackgroundJobEnvironment.new(config).setup_environment(readiness_port)
 
+    # load monkey patch for sequel backend to support configurable job lock method (postgres only)
+    require 'delayed_job/sequel_patch'
```
johha, line 32: `require` might fit better in `db.rb`:
https://github.com/cloudfoundry/cloud_controller_ng/blob/9cf8c4581aaab3d361e5dc694fb97b86a315dbfa/lib/cloud_controller/db.rb#L87-L87

lib/tasks/jobs.rake
```diff
@@ -75,4 +75,24 @@ namespace :jobs do
                                        thread_grace_period_seconds: args.thread_grace_period_seconds,
                                        publish_metrics: publish_metrics).start_working
   end
+
+  desc 'Generate dummy jobs for tests. Usage: rake jobs:generate_load\'[num,delay]\''
+  task :generate_load, %i[num delay] => :environment do |_t, args|
```
johha, line 80: I'm not sure if we should have test helpers in the rake tasks. 
There are already scripts like `scripts/seed_service_instances.rb` which are used for similar use cases.
Those can be executed by pasting  the content into a ccng console (`./bin/console`).

johha (changes requested):

lib/delayed_job/delayed_worker.rb
stephanme: Moved and works well.

lib/tasks/jobs.rake
stephanme: Converted rake task into generate_jobs script. Works well locally but not yet tested on a bosh VM.

johha (approved):
LGTM


</details>

jochenehret (approved):

PR #4236, opened 2025-02-26
Ruby 3.4

sethboyles:
Seems like we might have to skip upgrading to Ruby 3.3 because it doesn't look like the default version of Bundler will get the bugfix we need.  Trying to upgrade to Ruby 3.4 instead, but am running into more language change issues (particularly around keyword arguments). 

We may have to update more fog gems: https://github.com/fog/fog-local/pull/34

capi-release PR: https://github.com/cloudfoundry/capi-release/pull/517


db/migrations/20220601183937_create_asg_lastest_updates.rb
```diff
@@ -1,7 +1,7 @@
 Sequel.migration do
   change do
     create_table :asg_timestamps do
-      primary_key name: :id
+      primary_key :id, name: :id
```
sethboyles, line 4: I think enough time has passed that we don't need to worry about someone having run this migration, but not the followup that corrected it.  With Ruby 3.4, this migration would no longer behave consistently, anyway (`name: id` get serialized to `{name: :id}` instead of `{:name=>:id}`

sethboyles: I initially tried to make these migrations more explicit, but that caused issues with the `rename_column` on MySQL 5.7 tests.  Which we don't actually support, so a different potential workaround is dropping those tests.

Samze (approved):

xandroc (approved):

philippthun:
@sethboyles On which infrastructure providers did you test this already? You mention in the description possible issues with fog, so I guess we should test this on all the providers we support...

sethboyles:
@philippthun I only tested on singleton blobstore.  The fog-local change ended up being minor, thankfully.  But testing on other providers is a good idea... This might be a difficult now that we no longer have the blobstore-fanout, unfortunately.

db/migrations/20220601183937_create_asg_lastest_updates.rb
moleske: Is it worth just dropping [5.7 from the pipeline](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/.github/workflows/unit_tests.yml#L79) then?

PR #4247, opened 2025-03-07
Allow scaling with Deployments directly

sethboyles:

### DIFF
--- a/app/actions/deployment_create.rb
+++ b/app/actions/deployment_create.rb
@@ -43,6 +43,8 @@ module VCAP::CloudController
 
           desired_instances = desired_instances(app.oldest_web_process, previous_deployment)
 
+          validate_quota!(message, app)
+
           deployment = DeploymentModel.create(
             app: app,
             state: starting_state(message),
@@ -55,7 +57,8 @@ module VCAP::CloudController
             revision_version: revision&.version,
             strategy: message.strategy,
             max_in_flight: message.max_in_flight,
-            canary_steps: message.options&.dig(:canary, :steps)
+            canary_steps: message.options&.dig(:canary, :steps),
+            web_instances: message.web_instances || desired_instances
           )
 
           MetadataUpdate.update(deployment, message)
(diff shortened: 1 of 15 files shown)


app/actions/deployment_create.rb
```diff
@@ -163,7 +166,18 @@ def record_audit_event(deployment, droplet, user_audit_info, message)
 
       private
 
+      def validate_quota!(message, app)
+        return if message.web_instances.blank?
+
+        current_web_process = app.newest_web_process
+        current_web_process.instances = message.web_instances
```
Samze, line 173: **question** Does this modified state persist outside of the function?

app/messages/deployment_create_message.rb
```diff
@@ -74,6 +80,14 @@ def validate_max_in_flight
       errors.add(:max_in_flight, 'must be an integer greater than 0')
     end
 
+    def validate_web_instances
+      web_instances = options[:web_instances]
+
+      return unless !web_instances.is_a?(Integer) || web_instances < 1
```
Samze, line 86: **thought** I was curious if we support 0 today, seems like this:
```
cf push "static" -b staticfile_buildpack -i 0 --strategy=rolling
```

Actually results in a stuck deployment, it scales down processes to 0, starts a new process with 0 and I'm guessing theres some logic that expecting a process to be routable.

```
cf push "static" -b staticfile_buildpack -i 0 --strategy=canary
```

Does seem to work, but doesn't seem like a valid usecase.

Do we think theres any valid useful for the rolling case? Not sure it makes sense to deploy a new app revision and want it to scale down to 0 simulatiously.

app/models/runtime/deployment_model.rb
```diff
@@ -113,6 +113,14 @@ def continuable?
       state == DeploymentModel::PAUSED_STATE
     end
 
+    def desired_web_instances
```
Samze, line 116: **question** Could you elaborate on `this should handle deployments created on old API vms mid bosh deployment`? Not sure I follow.

sethboyles: `web_instances` will be NULL for deployments created prior to this change rolling out.  Since Deployments can potentially take a long time (even longer than a BOSH deployment), we need to make sure this code doesn't assume `web_instances` is present, even though it will be set at creation time moving forward.

app/messages/deployment_create_message.rb
Samze: Created #4250

app/actions/deployment_create.rb
sethboyles: good catch.  It is indeed cached via the ORM relations.  I've added a `reload` below and added a test for this.

app/messages/deployment_create_message.rb
sethboyles: Some app developers run 0 instance apps just for tasks.  I can't think of a strong use case for needing to support scaling with deployments to 0 (especially as it's not like rolling deployments scale down 1 by 1)

sethboyles: When I try it with Canary, I end up with 1 web instance when it should be 0. Again, weird use case.

Samze (approved):

PR #4269, opened 2025-03-18
Ensure asg-latest-update is renewed with microsecond precision even in mysql

geofffranks:

There was a bug with the ASG Latest Update modelwhere renewing would insert `CURRENT_TIMESTAMP` which in mysql defaults to second-level precision, even though the database column is set for microsecond precision. This resolves this by creating the timstamp in CAPI instead of relying on the SQL keyword. 

### DIFF
--- a/app/models/runtime/asg_latest_update.rb
+++ b/app/models/runtime/asg_latest_update.rb
@@ -8,9 +8,9 @@ module VCAP::CloudController
     def self.renew
       old_update = AsgTimestamp.first
       if old_update
-        old_update.update(last_update: Sequel::CURRENT_TIMESTAMP)
+        old_update.update(last_update: Time.now.utc)
       else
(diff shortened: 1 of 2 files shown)


app/models/runtime/asg_latest_update.rb
```diff
@@ -8,9 +8,9 @@ class AsgTimestamp < Sequel::Model
     def self.renew
       old_update = AsgTimestamp.first
       if old_update
-        old_update.update(last_update: Sequel::CURRENT_TIMESTAMP)
+        old_update.update(last_update: Time.now)
```
tcdowney, line 11:
```suggestion
        old_update.update(last_update: Time.now.utc)
```

Can we explicitly make this be in UTC? It probably is in practice

geofffranks: Fixed

tcdowney (approved):
LGTM

PR #4280, opened 2025-03-27
Keep /v2/info endpoint when v2 is disabled

kathap:
/v2/info was currently excluded from v2 rate limiting. Don't rate limit and keep /v2/info even when CF API v2 is switched off.  
/v2/info endpoint is still available when cc.temporary_enable_v2=false and indicates that v2 is disabled. 

### DIFF
--- a/app/controllers/runtime/info_controller.rb
+++ b/app/controllers/runtime/info_controller.rb
@@ -4,6 +4,26 @@ module VCAP::CloudController
 
     get '/v2/info', :read
     def read
+      unless @config.get(:temporary_enable_v2)
+        return Oj.dump({
+                         name: @config.get(:info, :name),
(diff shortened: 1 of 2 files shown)


app/controllers/runtime/info_controller.rb
```diff
@@ -4,6 +4,26 @@ class InfoController < RestController::BaseController
 
     get '/v2/info', :read
     def read
+      unless @config.get(:temporary_enable_v2)
+        return Oj.dump({
+                         name: @config.get(:info, :name),
```
jochenehret, line 9: Can't you use the "info" map from below and change the "support" and "api_version" values?

kathap: Good idea

jochenehret (approved):

Issue #4286, opened 2025-04-01
Updating a Route does not Propagate to Diego for Running Applications


maxmoehl:
When updating a route via [`PATCH /v3/routes/:guid`](https://v3-apidocs.cloudfoundry.org/version/3.190.0/index.html#update-a-route) changes to the options do not propagate to diego and therefore do not update the route which is emitted by route-emitter for any currently running app which is bound to that route.

From @philippthun via [Slack](https://cloudfoundry.slack.com/archives/C02HNDJB31R/p1743452982424319):

> Looking at the implementation of the [controller](https://github.com/cloudfoundry/cloud_controller_ng/blob/81e3e18a196d2886a66c96c5d3a29bf34f1fd7b3/app/controllers/v3/routes_controller.rb#L108) and the underlying [action](https://github.com/cloudfoundry/cloud_controller_ng/blob/81e3e18a196d2886a66c96c5d3a29bf34f1fd7b3/app/actions/route_update.rb#L5-L7), it seems as if only the database is updated.
> There is an `update_routes` [method](https://github.com/cloudfoundry/cloud_controller_ng/blob/81e3e18a196d2886a66c96c5d3a29bf34f1fd7b3/lib/cloud_controller/diego/runner.rb#L32) in `VCAP::CloudController::Diego::Runner`, but it does not seem to be used at all.
> In a different [action](https://github.com/cloudfoundry/cloud_controller_ng/blob/81e3e18a196d2886a66c96c5d3a29bf34f1fd7b3/app/actions/app_update.rb#L48-L54) the method update_metric_tags is called; maybe this could be used as a blueprint in case the bahviour should be changed.

Thanks to @ameowlia for reporting this.

maxmoehl:
This also needs a fix in gorouter: https://github.com/cloudfoundry/gorouter/pull/472

maxmoehl:
Turns out the fix in gorouter is more cosmetic. The CC PR is now ready for review.

PR #4298, opened 2025-04-10
feat: notify diego of route updates

maxmoehl:
After updating a route via the new update route endpoint, it was only stored in the database. Since routes can be changed without restarting the application this commit adds logic to also push the update to diego.

Resolves: https://github.com/cloudfoundry/cloud_controller_ng/issues/4286

### DIFF
--- a/app/actions/route_update.rb
+++ b/app/actions/route_update.rb
@@ -6,6 +6,10 @@ module VCAP::CloudController
 
         route.save
         MetadataUpdate.update(route, message)
+
+        route.apps.each do |process|
+          ProcessRouteHandler.new(process).notify_backend_of_route_update
+        end
       end
 
       route
--- a/app/actions/route_update.rb
+++ b/app/actions/route_update.rb
@@ -2,16 +2,15 @@ module VCAP::CloudController
   class RouteUpdate
     def update(route:, message:)
       Route.db.transaction do
-        route.options = route.options.symbolize_keys.merge(message.options).compact if message.requested?(:options)
-
+        if message.requested?(:options)
+          route.options = route.options.symbolize_keys.merge(message.options).compact
+          route.apps.each do |process|
+            ProcessRouteHandler.new(process).notify_backend_of_route_update
+          end
+        end
         route.save
         MetadataUpdate.update(route, message)
-
-        route.apps.each do |process|
-          ProcessRouteHandler.new(process).notify_backend_of_route_update
-        end
       end
-
       route
     end
   end
--- a/app/actions/route_update.rb
+++ b/app/actions/route_update.rb
@@ -2,14 +2,15 @@ module VCAP::CloudController
   class RouteUpdate
     def update(route:, message:)
       Route.db.transaction do
-        if message.requested?(:options)
-          route.options = route.options.symbolize_keys.merge(message.options).compact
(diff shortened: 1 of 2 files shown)


spec/unit/actions/route_update_spec.rb
```diff
@@ -41,7 +41,8 @@ module VCAP::CloudController
     end
 
     let(:message) { RouteUpdateMessage.new(body) }
-    let(:route) { Route.make }
```
philippthun, line 44: I would create the following variables:
```
let(:process) { ProcessModel.make }
let(:route_mapping) { RouteMappingModel.make(app: process.app) }
let(:route) { route_mapping.route }
```

spec/unit/actions/route_update_spec.rb
```diff
@@ -143,6 +144,13 @@ module VCAP::CloudController
 
           it 'adds no options' do
             expect(message).to be_valid
+
```
philippthun, line 147: Instead of adding the same code block at different places, I would suggest the following (right at the top of the `describe '#update options'` block:
```
describe '#update options' do
  let(:fake_route_handler) { instance_double(ProcessRouteHandler) }

  before do
    allow(ProcessRouteHandler).to receive(:new).with(process).and_return(fake_route_handler)
    allow(fake_route_handler).to receive(:notify_backend_of_route_update)
  end
```

And then **for each context below**, add one of the following expectations:
```
  it 'does not notify the backend' do
    expect(fake_route_handler).not_to receive(:notify_backend_of_route_update)

    subject.update(route:, message:)
  end

  it 'notifies the backend' do
    expect(fake_route_handler).to receive(:notify_backend_of_route_update)

    subject.update(route:, message:)
  end
```

spec/unit/actions/route_update_spec.rb
```diff
@@ -41,7 +41,8 @@ module VCAP::CloudController
     end
 
     let(:message) { RouteUpdateMessage.new(body) }
-    let(:route) { Route.make }
+    let(:space) { Space.make }
+    let(:route) { Route.make(space:) }
 
     subject { RouteUpdate.new }
     describe '#update metadata' do
```
philippthun, line 48: For the `describe '#update metadata'` block, you should add this:
```
before do
  expect(ProcessRouteHandler).not_to receive(:new)
end
```

This will also force you to change the implementation, as I would expect that Diego should only be notified if some route option changed.

philippthun (changes requested):

app/actions/route_update.rb
```diff
@@ -2,12 +2,15 @@ module VCAP::CloudController
   class RouteUpdate
     def update(route:, message:)
       Route.db.transaction do
-        route.options = route.options.symbolize_keys.merge(message.options).compact if message.requested?(:options)
-
+        if message.requested?(:options)
+          route.options = route.options.symbolize_keys.merge(message.options).compact
+          route.apps.each do |process|
```
philippthun, line 7: I think it would be better to move this communication with Diego outside of the database transaction, so in pseudo code:
```
transaction (
  if new/changed options (
    update options
    save
  )
  update metadata
)

if new/changed options (
  notify diego
)

return route
```

philippthun (commented):
One additional thing that came into my mind...

PR #4320, opened 2025-04-24
Add cnb file support

sethboyles:

### DIFF
--- a/app/fetchers/buildpack_list_fetcher.rb
+++ b/app/fetchers/buildpack_list_fetcher.rb
@@ -19,7 +19,8 @@ module VCAP::CloudController
         dataset = dataset.where(name: message.names) if message.requested?(:names)
 
         dataset = NullFilterQueryGenerator.add_filter(dataset, :stack, message.stacks) if message.requested?(:stacks)
-        dataset = dataset.where(lifecycle: message.requested?(:lifecycle) ? message.lifecycle : Config.config.get(:default_app_lifecycle))
+
+        dataset = dataset.where(lifecycle: message.lifecycle) if message.requested?(:lifecycle)
 
         if message.requested?(:label_selector)
           dataset = LabelSelectorQueryGenerator.add_selector_queries(
--- a/app/messages/buildpacks_list_message.rb
+++ b/app/messages/buildpacks_list_message.rb
@@ -21,7 +21,8 @@ module VCAP::CloudController
 
     def initialize(params={})
       super
-      pagination_options.default_order_by = :position
+      pagination_options.default_order_by = :lifecycle
+      pagination_options.secondary_default_order_by = :position
     end
 
     def self.from_params(params)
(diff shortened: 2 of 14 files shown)


app/models/runtime/buildpack.rb
```diff
@@ -1,23 +1,35 @@
...
     add_association_dependencies annotations: :destroy
 
-    def self.list_admin_buildpacks(stack_name=nil)
+    def self.user_visibility_filter(_user)
+      full_dataset_filter
+    end
+
+    def self.list_admin_buildpacks(stack_name=nil, lifecycle=Config.config.get(:default_app_lifecycle))
       scoped = exclude(key: nil).exclude(key: '')
+      scoped = scoped.filter(lifecycle:)
```
sethboyles, line 32: I think we need to remove this line (or the default arg).  This hides all buildpacks of the non-default type on the list endpoint

sethboyles: I guess this is intentional: https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0031-system-cnb.md#proposal

It's very confusing though, and breaks the CLI if you are trying to update a buildpack that is not of the default lifecycle

sethboyles: @Gerg pointed out to me that line in the RFC is about POST, not list.  @pbusko was it your intention to have a default filter on the list endpoint?

pbusko: this filter is for backward compatibility, since the list endpoint has additional `lifecycle` query parameter (just like other buildpack related endpoints).
If we remove it, current versions of CF CLI will receive a soup of buildpacks with mixed lifecycles. I believe there's no way around it, it has to be filtered

app/messages/buildpack_upload_message.rb
```diff
@@ -53,7 +54,10 @@ def is_archive
 
       return if mime_bits =~ /^#{VCAP::CloudController::GZIP_MIME}/ || mime_bits =~ /^#{VCAP::CloudController::ZIP_MIME}/
 
-      errors.add(:base, "#{bits_name} is not a zip or gzip archive")
+      mime_bits_at_offset = File.read(bits_path, 8, 257)
+      return if mime_bits_at_offset =~  /^#{VCAP::CloudController::CNB_MIME}/
+
+      errors.add(:base, "#{bits_name} is not a zip or gzip archive or cnb file")
```
pbusko: CNB file has to specific format, it's simly a tarball

pbusko:
Would you mind opening a PR with your additions to the original (https://github.com/sap-contributions/cloud_controller_ng/tree/cnb-system-buildpacks)

app/messages/buildpack_upload_message.rb
tomkennedy513: the cnb file is part of the buildpack spec (https://github.com/buildpacks/spec/blob/main/distribution.md#buildpackage). Even though its just a tar, we want to imply that we are just following the cnb spec and if that format changes in the future we would expect to adapt to that change.

PR #4328, opened 2025-04-30
Enhance SB polling intervals

svkrieger:
- Currently the retry-after header is used in the exponential backoff calculation. This PR makes it more intuitive by using the retry-after value as is, if it is larger than our own calculated polling interval. With that service owners can defer polling when they know that provisioning takes longer.

- Introducing a configuration parameter `broker_client_max_async_poll_interval_seconds` for capping the polling interval. This helps preventing that polling intervals get too large. Default is still at 24h, which was hardcoded before.

https://github.com/cloudfoundry/capi-release/pull/542

### DIFF
--- a/app/jobs/reoccurring_job.rb
+++ b/app/jobs/reoccurring_job.rb
@@ -30,7 +30,7 @@ module VCAP::CloudController
       end
 
       def polling_interval_seconds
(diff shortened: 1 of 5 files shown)


jochenehret:
> ... if it is smaller than our own calculated polling interval ...
Not "larger"?

PR #4341, opened 2025-05-05
Can't move routes to other orgs if the space has the same name as owner space

nookala:
Compare the space id instead of name for route transfer

Given org `org1` with space `space` org `org2` with space `space`:

 `cf push dora` with no manifest (so default-route is used)
`cf move-route thawing.app-runtime-interfaces.ci.cloudfoundry.org --hostname dora -o org2 -s space`
this will return `OK`, thinking the spaces are equal, though it wont actually be moved
The fix compares the space ids instead of the name so that the move-route actually works in this case.

### DIFF
--- a/spec/unit/actions/route_transfer_owner_spec.rb
+++ b/spec/unit/actions/route_transfer_owner_spec.rb
@@ -7,6 +7,7 @@ module VCAP::CloudController
     let(:route) { Route.make domain: SharedDomain.make, space: original_owning_space }
     let(:original_owning_space) { Space.make name: 'original_owning_space' }
     let(:target_space) { Space.make name: 'target_space' }
+    let(:target_space_dup_name) { Space.make name: 'original_owning_space' }
     let(:shared_space) { Space.make name: 'shared_space' }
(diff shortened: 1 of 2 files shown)


spec/unit/actions/route_transfer_owner_spec.rb
```diff
@@ -26,25 +27,32 @@ module VCAP::CloudController
...
-          expect(route.space.name).to eq original_owning_space.name
+          expect(route.shared_spaces.map(&:id)).not_to include original_owning_space.id
+          expect(route.space.id).to eq original_owning_space.id
+        end
+      end
+
+      context 'target space has the same name as the owning space' do
+        it 'makes the target space with the same name the new owner' do
+          RouteTransferOwner.transfer(route, target_space_dup_name, user_audit_info)
+          expect(route.space.id).to eq target_space.id
```
philippthun, line 55: I guess this must be `expect(route.space.id).to eq target_space_dup_name.id`.

sethboyles (approved):

philippthun (approved):

PR #4342, opened 2025-05-05
Rebased cnb buildpack updates

sethboyles:
rebased version of https://github.com/sap-contributions/cloud_controller_ng/pull/374/file against main so we can see if that fixes the unit/backwards compatibility tests


app/jobs/runtime/buildpack_installer_factory.rb
```diff
@@ -37,8 +37,7 @@ def plan(buildpack_name, manifest_fields)
 
           planned_jobs = []
 
-          found_buildpacks = Buildpack.where(name: buildpack_name).all
-
+          found_buildpacks = Buildpack.where(name: buildpack_name, lifecycle: manifest_fields[0][:options][:lifecycle]).all
```
Samze, line 40: **question** Why do we take the first entry? And is `options.lifecycle` always set?

sethboyles: I think there is some data prep going on before we get to this point (that is, we're not just passing in the values straight from the manifest) and then the buildpacks are grouped by lifecycle: 

https://github.com/cloudfoundry/cloud_controller_ng/pull/4342/files#diff-4b8b4efbcadd70b60a4ed679100b017aa85113fdc425ce0184ad64010dd5dd77R60

Samze: Gotcha, and yeah looks like a default is set https://github.com/cloudfoundry/cloud_controller_ng/pull/4342/files#diff-4b8b4efbcadd70b60a4ed679100b017aa85113fdc425ce0184ad64010dd5dd77R19

lib/cloud_controller/install_buildpacks.rb
```diff
@@ -12,12 +12,12 @@ def install(buildpacks)
       CloudController::DependencyLocator.instance.buildpack_blobstore.ensure_bucket_exists
       job_factory = VCAP::CloudController::Jobs::Runtime::BuildpackInstallerFactory.new
 
-      buildpack_install_jobs = []
-
       factory_options = []
       buildpacks.each do |bpack|
         buildpack_opts = bpack.deep_symbolize_keys
 
+        buildpack_opts[:lifecycle] = Lifecycles::BUILDPACK if buildpack_opts[:lifecycle].nil?
```
Samze, line 19: **question** We default here to lifecycle buildpack, should this be default_app_lifecycle?

Samze: We discussed this and decided this should remain buildpack and that we should document the default nature (at least in the capi release spec )

Samze (dismissed):

Samze (approved):

PR #4344, opened 2025-05-08
Can't push apps with app manifests if routes were shared

nookala:
Check if route shared with space and don't throw error if shared

If the route is shared with the space don't throw error
Given org `org1` with space `space` org `org2` with space `space2`:

 `cf push dora` with no manifest (so default-route is used)
`cf share-route thawing.app-runtime-interfaces.ci.cloudfoundry.org --hostname dora -o org2 -s space2`
this will return `OK`, thinking the spaces are equal, though it wont actually be moved
NOTE: The space names do not have to be the same
`cf target -o org2 -s space2`
`cf push dora` again
Observe the error: `Routes cannot be mapped to destinations in different spaces`


app/actions/manifest_route_update.rb
```diff
@@ -93,7 +93,9 @@ def find_or_create_valid_route(app, manifest_route, user_audit_info)
               manifest_triggered: true
             )
           elsif route.space.guid != app.space_guid
-            raise InvalidRoute.new('Routes cannot be mapped to destinations in different spaces')
+            # check if route is shared with space
+            spaces = route.shared_spaces
+            raise InvalidRoute.new('Routes cannot be mapped to destinations in different spaces') if spaces.blank? || spaces.none? { |space| space.values[:id] == app.space.id }
```
tcdowney, line 98: This logic is duplicated in several places now (e.g. https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/controllers/v3/routes_controller.rb#L365).

Is there a shared place we could implement the check? Maybe a function on the `Route` itself?

app/models/runtime/route.rb
```diff
@@ -194,6 +194,10 @@ def self.user_visibility_filter(user)
       }
     end
 
+    def app_spaces_no_match?(app)
```
Samze, line 197: **suggestion** I find this name a little confusing, what about:
```
def available_in_space(space)
```

And pass `app.space`. So usage looks like
```
!route.available_in_space?(app.space)
```

Samze (approved):

PR #4352, opened 2025-05-09
Improve error handling for already deleted resources

johha:
If a user triggers two DELETE requests for the same resource within a few milliseconds two deletion jobs are created. If both jobs are be executed in parallel it might happen that one of them fails with a `Sequel::NoExistingObject` error. This change prevents this by catching those errors but only if the error message originates from the current model/ table. Therefore it is ensured that similar error messages when e.g. deleting sub resources still result in job failures.

The same applies if a user unbind routes but also deletes the route itself within a very short time period. For this case the route model now also catches `Sequel::NoExistingObject` errors related to route bindings.

Additionally, the error handling in the DeleteActionJob was improved to catch errors when the actions itself don't handle them.

### DIFF
--- a/app/jobs/delete_action_job.rb
+++ b/app/jobs/delete_action_job.rb
@@ -30,10 +30,10 @@ module VCAP::CloudController
         end
 
         # Ignore errors if the target resource has already been deleted (e.g., by a parallel job)
-        quoted_table_name = @model_class.db.quote_identifier(@model_class.table_name)
+        quoted_table_name = model_class.db.quote_identifier(model_class.table_name)
         errors.reject! { |err| err.is_a?(Sequel::NoExistingObject) && err.message.include?("DELETE FROM #{quoted_table_name}") } unless errors.frozen?
 
-        raise errors.first unless errors&.empty?
+        raise errors.first unless errors.empty?
 
         warnings
       end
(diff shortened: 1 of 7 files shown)


nookala (approved):

app/jobs/delete_action_job.rb
```diff
@@ -17,12 +17,22 @@ def perform
...
+          end
+        rescue StandardError => e
+          errors << e
         end
 
+        # Ignore errors if the target resource has already been deleted (e.g., by a parallel job)
+        quoted_table_name = @model_class.db.quote_identifier(@model_class.table_name)
+        errors.reject! { |err| err.is_a?(Sequel::NoExistingObject) && err.message.include?("DELETE FROM #{quoted_table_name}") } unless errors.frozen?
+
         raise errors.first unless errors&.empty?
```
philippthun, line 36:
```suggestion
        raise errors.first unless errors.empty?
```

As `errors` is now explicitly initialized, this safe-guard is not needed anymore.

app/jobs/delete_action_job.rb
```diff
@@ -17,12 +17,22 @@ def perform
...
+            errors, warnings = delete_action.delete(dataset)
+          else
+            errors = delete_action.delete(dataset)
+          end
+        rescue StandardError => e
+          errors << e
         end
 
+        # Ignore errors if the target resource has already been deleted (e.g., by a parallel job)
+        quoted_table_name = @model_class.db.quote_identifier(@model_class.table_name)
```
philippthun, line 33:
```suggestion
        quoted_table_name = model_class.db.quote_identifier(model_class.table_name)
```

There is an `attr_reader` defined for `model_class`.

PR #4358, opened 2025-05-13
Add feature flags to v1 app manifests

philippthun:
Allow app features to be set via manifest. Features are specified as follows (per app):
```
features:
    ssh: true
    revisions: true
    service-binding-k8s: false
    file-based-vcap-services: false
```

Features are shown in the generated manifest for an app (`GET /v3/apps/:guid/manifest`) and taken into account when calculating the manifest diff (`POST /v3/spaces/:guid/manifest_diff`).

### DIFF
--- a/app/actions/app_feature_update.rb
+++ b/app/actions/app_feature_update.rb
@@ -1,15 +1,41 @@
 module VCAP::CloudController
   class AppFeatureUpdate
-    def self.update(feature_name, app, message)
-      case feature_name
-      when AppFeatures::SSH_FEATURE
-        app.update(enable_ssh: message.enabled)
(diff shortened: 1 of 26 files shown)


app/actions/app_feature_update.rb
```diff
@@ -1,15 +1,41 @@
...
+
+        return if flags.empty?
+
+        check_invalid_combination!(app, flags)
+        app.update(flags)
+      end
+
+      private
+
+      def database_column(feature_name)
```
svkrieger, line 25: The name of this method confused me a bit. Could we rename it to something like `feature_column_name` or `database_column_name`?

Issue #4182, opened 2025-01-21
Problems related to pruning old AppUsageEvent and ServiceUsageEvent records


joyvuu-dave:
### Problem

A problem we've been facing involves the pruning of old AppUsageEvent and ServiceUsageEvent records. Often, these records are removed before the corresponding Apps or Services have actually stopped, making it difficult to determine how long those resources have been running. If a consumer starts polling after the `start` record is pruned, it may never know the true start time of that App or Service.

### Challenges and Use of Purge/Seed

A specific pain point relates to the `destructively_purge_all_and_reseed` endpoints for App and Service Usage Events. These endpoints are sometimes used when event tables become inconsistent — often because `start` records were removed prematurely. While `destructively_purge_all_and_reseed` recreates running resources in the database, it assigns new `start` timestamps that do not reflect actual creation or launch times. As a result, usage metrics can become misleading.

### Core Problems
* Pruning Before Completion
    * The system prunes old records to manage database growth. However, if an App/Service remains running for a long period, its `start` record may be deleted before the `stop` record exists
    * A newly added or recovering consumer will not see accurate start times
* Extended Downtime Leading to Missed Events
    * Sometimes, a usage-event polling service may go offline for an extended period (e.g. an unnoticed crash). By the time it resumes polling, older events may have been pruned, leaving gaps in historical data
* Accurate State Visibility
    * It becomes challenging to piece together which Apps or Services are still running when critical events have already been removed, forcing reliance on [destructively_purge_all_and_reseed](https://v3-apidocs.cloudfoundry.org/version/3.185.0/index.html#purge-and-seed-app-usage-events) to reset the data (where we lose accurate historical start times)

### Potential Approaches

After running into this issue repeatedly, I’ve created [a set of code changes](https://github.com/joyvuu-dave/cloud_controller_ng/pull/1) designed to:
* Keep `start` Records for Active Apps/Services
    * Records remain in place until the corresponding `stop` event is encountered, preventing the loss of essential lifecycle information.
* Consumer Registration
    * By including `consumer_guid` and `after_guid` in usage-event requests, consumers can register themselves, allowing the Cloud Controller to avoid pruning events they have not yet processed
* Threshold-Based Pruning
    * A configurable limit (`threshold_for_keeping_unprocessed_records`) ensures the database does not grow indefinitely if a registered consumer stays offline. If the record count exceeds this threshold, older entries can still be pruned
* Endpoints for Managing Consumers
[...]

PR #4374, opened 2025-05-27
Keep Usage Event records of running & un-processed Apps and Services

joyvuu-dave:
Addresses #4182 

Commit
Allows users to turn on or off keep_unprocessed_records feature
Leave it off by default

Commit
Seed Usage Event records
Useful for integration testing. May only be needed while validating Usage Consumer feature.

Commit
Fix a typo

Commit
Provide index, show, and destroy endpoints for usage consumers
Allows for usage consumers to de-register themselves with the cloud controller. Index and show methods allow consumers to make informed decisions before initiating a purge of usage events.

Usage event endpoint re-ordered to align with convention of other resources.

Commit
Set threshold for keeping un-processed records
Only keep un-processed records if usage event table size is below threshold. This is a safeguard to ensure row count does not grow unbounded in the event of a zombie consumer.

Commit
Maintain un-processed usage event records
Allow consumer_guid to be passed into request for AppUsageEvent and ServiceUsageEvent records. This establishes a consumer of these events. Un-processed consumer records will be maintained by cloud controller. Registered consumers are automatically deleted if associated event is removed.

Commit
Keep usage event records of running apps and services
AppUsageEvent and ServiceUsageEvent records can be used to track the lifecycle of apps and services in a foundation. However, if the start event is pruned before end event, the foundation no longer maintains an accurate state of its running apps and services. By holding on to the start events, consumers of these records are able to get an accurate picture of the current state whenever they choose to begin consuming usage event records.

Commit
Make cutoff_age_in_days a named parameter
Better aligns with style used throughout codebase. Done in anticipation of adding more parameters to the `initialize` method.

* [ ] I have viewed, signed, and submitted the Contributor License Agreement <-- in progress

### DIFF
--- a/db/migrations/20250527000000_add_unique_constraint_to_usage_consumers.rb
+++ b/db/migrations/20250527000000_add_unique_constraint_to_usage_consumers.rb
@@ -1,0 +1,13 @@
+Sequel.migration do
+  change do
+    alter_table :app_usage_consumers do
+      add_index :consumer_guid, unique: true, name: :app_usage_consumers_consumer_guid_unique
+      add_index :last_processed_guid, name: :app_usage_consumers_last_processed_guid_index
+    end
+
+    alter_table :service_usage_consumers do
+      add_index :consumer_guid, unique: true, name: :service_usage_consumers_consumer_guid_unique
+      add_index :last_processed_guid, name: :service_usage_consumers_last_processed_guid_index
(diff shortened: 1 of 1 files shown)


joyvuu-dave:
I should have CLA approval now -- is there a way to re-trigger this bot check?

PR #4383, opened 2025-05-30
Fixes for CVEs

nookala:
https://github.com/advisories/GHSA-vmwr-mc7x-5vc3
https://github.com/advisories/GHSA-gjh7-p2fx-99vx
https://github.com/advisories/GHSA-2gw2-8q9w-cw8p


moleske (commented):
I'm confused. Two of these dependencies are being added to the gemfile and aren't used anywhere.  And rack is at 2.2.15 which is higher than the cve you're referencing

nookala:
> I'm confused. Two of these dependencies are being added to the gemfile and aren't used anywhere. And rack is at 2.2.15 which is higher than the cve you're referencing

ffi already exists in the Gemfile.lock as a dependency   ffi (1.16.3) https://github.com/cloudfoundry/cloud_controller_ng/blob/c4a94b77311380d7dd94bc04f6864c23bd8b8300/Gemfile.lock#L155
rexml is also an existing dependency, but is 3.3.9 so I could pin it to >= 3.3.9 https://github.com/cloudfoundry/cloud_controller_ng/blob/c4a94b77311380d7dd94bc04f6864c23bd8b8300/Gemfile.lock#L290

> I'm confused. Two of these dependencies are being added to the gemfile and aren't used anywhere. And rack is at 2.2.15 which is higher than the cve you're referencing
Agreed on rack, could explicitly pin to >= 2.2.15

nookala:
> > ffi already exists in the Gemfile.lock as a dependency ffi (1.16.3)
> 
> This is why I'm confused, it is already brought in transitively and is updated. Adding it to the gemfile I don't believe helps cause the transitive dependency will bring in whatever version it needs. Same for the other dependencies. I also see no issue in the github security checks.
> 
> I'll let someone who knows ruby dependency management better than I (hi @sethboyles!) chime in to correct me if I'm misunderstanding
As per my understanding the gem version is to be added in the Gemfile to override a transitive dependency https://docs.veracode.com/r/Fix_Example_Transitive_Vulnerability_for_Ruby
bundle show after the change returns
* rack (2.2.16)
* rexml (3.4.1)
* ffi (1.16.3)

sethboyles:
I think you can run `bundle update --conservative rexml ffi rack` to update just those gems without adding to the Gemfile. 

it won't update rexml because it's already at 3.4.1.  Seems like a false alarm from the CVE scanner

sethboyles:
I misread your comment @nookala since the formatting makes it look like a quote. It's unfortunate that Veracode (is that the scanner Bcom uses?) recommends adding the gems to the gemfile directly, because it makes it less clear what gems are actually being used directly by CCNG.

I'd suggest trying the [validate](https://docs.veracode.com/r/Validating_Fixed_Agent_Based_Scan_Results) with the `--conservative` option I suggested above.  If that doesn't work, and we actually need to add to the Gemfile, maybe we can put transitive security fixes in their own section lower in the file with a comment explaining they are only explicitly set to satisfy a CVE scanner.

moleske:
Though I don't know the report being referenced, I think it is all false positives

- [ffi is at 1.16.3](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/Gemfile.lock#L155) which is newer than 1.9.24
- [rexml is at 3.4.1](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/Gemfile.lock#L441) which is newer than 3.3.6
- [rack is at 2.2.15](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/Gemfile.lock#L387) which is newer than 2.2.14

Some of dependencies that bring in ffi, rexml, and rack say that lower versions of those dependencies are acceptable

here are the links to the dependabot security updates that fixed some of the issues referenced in PR.  You'll only be able to see these if you are an approver in the capi project though.  I couldn't find one for ffi, but the ffi cve referenced is from 2018 and has been fixed for awhile

- rack GHSA-gjh7-p2fx-99vx - https://github.com/cloudfoundry/cloud_controller_ng/security/dependabot/167 and https://github.com/cloudfoundry/cloud_controller_ng/security/dependabot/166
- rexml GHSA-vmwr-mc7x-5vc3 - https://github.com/cloudfoundry/cloud_controller_ng/security/dependabot/121 and https://github.com/cloudfoundry/cloud_controller_ng/security/dependabot/120

nookala:
> Though I don't know the report being referenced, I think it is all false positives
> 
> * [ffi is at 1.16.3](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/Gemfile.lock#L155) which is newer than 1.9.24
> * [rexml is at 3.4.1](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/Gemfile.lock#L441) which is newer than 3.3.6
> * [rack is at 2.2.15](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/Gemfile.lock#L387) which is newer than 2.2.14
> 
> Some of dependencies that bring in ffi, rexml, and rack say that lower versions of those dependencies are acceptable, but the lock file is not bringing in those lower versions
> 
> here are the links to the dependabot security updates that fixed some of the issues referenced in PR. You'll only be able to see these if you are an approver in the capi project though. I couldn't find one for ffi, but the ffi cve referenced is from 2018 and has been fixed for awhile
> 
> * rack [GHSA-gjh7-p2fx-99vx](https://github.com/advisories/GHSA-gjh7-p2fx-99vx) - https://github.com/cloudfoundry/cloud_controller_ng/security/dependabot/167 and https://github.com/cloudfoundry/cloud_controller_ng/security/dependabot/166
> * rexml [GHSA-vmwr-mc7x-5vc3](https://github.com/advisories/GHSA-vmwr-mc7x-5vc3) - https://github.com/cloudfoundry/cloud_controller_ng/security/dependabot/121 and https://github.com/cloudfoundry/cloud_controller_ng/security/dependabot/120
> 
> edit - [found the ffi bump to resolve this cve from 2018](https://github.com/cloudfoundry/cloud_controller_ng/commit/87548eace9f891f7d01bd2dd0e22d4d01e2beea3). It was done before the advisory was released, so no dependabot security note for it

Thanks @moleske and @sethboyles I've marked the CVEs as false positives.

PR #4386, opened 2025-06-02
Add logging of service metadata during instance create/update/delete

kathap:
Inform operators which service plan, service offering and broker is involved in service instance creation/update/deletion

Add names of service plan, service offering and broker as logging information for service instance creation/update/deletion.
This change updates fetch_writable_service_instance to eagerly load associated service_plan, service, and service_broker relationships for managed service instances. This avoids N+1 queries when accessing these associations later in the request lifecycle.

This change improves operator visibility by logging service plan names, offerings, and broker names during service instance creation, update, and deletion.
Previously, logs only included GUIDs or IDs, requiring manual lookups or customer input to identify the service involved. With this change, operators can directly see which service is affected, making it easier to diagnose issues and support users/customers more efficiently.

### DIFF
--- a/app/controllers/v3/service_instances_controller.rb
+++ b/app/controllers/v3/service_instances_controller.rb
@@ -312,22 +312,9 @@ class ServiceInstancesV3Controller < ApplicationController
     action.preflight!
     if action.update_broker_needed?
 
-      service_plan_relations = if message.service_plan_guid
-                                 { Sequel[:service_plans][:guid] => message.service_plan_guid }
-                                 ServicePlan.eager_graph(service: :service_broker).
-                                   where(Sequel[:service_plans][:guid] => message.service_plan_guid).
-                                   all
-                               else
-                                 { Sequel[:service_plans][:id] => service_instance.service_plan_id }
-                                 ServicePlan.eager_graph(service: :service_broker).
-                                   where(Sequel[:service_plans][:id] => service_instance.service_plan_id).
-                                   all
-                               end
-
-      service_plan = service_plan_relations[0]
-      plan_name = service_plan.name
-      service_name = service_plan.service.name
-      broker_name = service_plan.service.service_broker.name
+      plan_name = service_instance.service_plan.name
+      service_name = service_instance.service_plan.service.label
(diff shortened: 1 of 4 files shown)


app/controllers/v3/service_instances_controller.rb
```diff
@@ -297,6 +311,13 @@ def update_managed(service_instance)
     action = V3::ServiceInstanceUpdateManaged.new(service_instance, message, user_audit_info, message.audit_hash)
     action.preflight!
     if action.update_broker_needed?
+      logger.info(
+        "Updating managed service instance with name '#{service_instance.name}' " \
+        "using service plan '#{service_instance.service_plan.name}' " \
+        "from service offering '#{service_instance.service_plan.service.label}' " \
+        "provided by broker '#{service_instance.service_plan.service.service_broker.name}'."
```
philippthun: When I tested this, a database query to fetch broker information was executed.

It seems as if `service_instance.reload` ([here](https://github.com/cloudfoundry/cloud_controller_ng/blob/511e64b9d933de1fe9454e2ccc5108e4a0e9b6a7/app/actions/v3/service_instance_update_managed.rb#L272)) removes the associated objects (or at least some of them).

So for the `update_managed` part, eager loading does not have the expected benefits.

Additional question: can the service plan be updated? This would even make the eager loading completely obsolete in such a case.

kathap: The service plan can be updated, yes.

svkrieger: Good catch with the reload! I removed the reload now and use a copy of the service instance instead for verifying the updated object, to ensure that follow-up preflight checks run with the original service instance object.

To your question: Yes, the plan is updatable. But only for the same service offering -> so eager loading the service and broker still provides a benefit because those will never change for a service instance.

I compared the number of DB queries in main with the ones from this PR:

Scenario/codebase | main | PR
-- | -- | --
create | 73 | <green>71<green/>
update | 95 | 91
delete | 48 | 48

-> even though we are fetching more information for the logs, this will reduce the number of queries for create/update cases and stays the same for the delete case.

philippthun (approved):

PR #4393, opened 2025-06-04
Support providing Windows gMSA credentials via a container-level env var for use by Garden/Winc

tcdowney:
We want to be able to provide gMSA credentials to apps that are used as part of container creation on Windows (see https://learn.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/gmsa-run-container). Service Bindings felt like a natural fit here... e.g. user either creates an UPSI with gMSA credentials or uses a broker (like CredHub service broker or a new broker) and these are provided via a Service Credential Binding.

The problem is that this data needs to be made available to garden/winc at container creation time. Not when it tries to exec a process in the container (which is when it has access to process-level env vars). We were initially thinking of making VCAP_SERVICES available as a container-level env var (https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/cloud_controller/diego/app_recipe_builder.rb#L71) which garden/winc would have access to.

Unfortunately there are two issues with putting `VCAP_SERVICES` up there.

1. If `VCAP_SERVICES` is too large to be an environment variable this will cause container creation to fail in weird and difficult to debug ways.
2.  File-based service bindings means that `VCAP_SERVICES` may not exist.

So instead this change introduces a new container env var called `WINDOWS_GMSA_CREDENTIAL_REF` that will be populated with the value of the first service binding credential that has a `'credhub-windows-gmsa-credential-ref'` key. This is a relatively niche use-case and will not affect most apps, but will provide a lot of value for Windows apps that need to run with distinct GMSA credentials.

Related Slack thread where we discussed this solution: https://cloudfoundry.slack.com/archives/C07C04W4Q/p1747164384236189

### DIFF
--- a/app/models/runtime/app_model.rb
+++ b/app/models/runtime/app_model.rb
@@ -125,6 +125,12 @@ module VCAP::CloudController
       DatabaseUriGenerator.new(service_binding_uris).database_uri
     end
 
+    def windows_gmsa_credential_refs
+      service_bindings.map do |binding|
+        binding.credentials['credhub-windows-gmsa-credential-ref'] if binding.credentials.present?
+      end.compact
+    end
+
     def staging_in_progress?
       builds_dataset.where(state: BuildModel::STAGING_STATE).any?
     end
--- a/app/models/runtime/app_model.rb
+++ b/app/models/runtime/app_model.rb
@@ -126,7 +126,7 @@ module VCAP::CloudController
     end
 
     def windows_gmsa_credential_refs
-      service_bindings.map do |binding|
+      service_bindings.sort_by(&:id).map do |binding|
         binding.credentials['credhub-windows-gmsa-credential-ref'] if binding.credentials.present?
       end.compact
     end
--- a/lib/cloud_controller/diego/windows_environment_sage.rb
+++ b/lib/cloud_controller/diego/windows_environment_sage.rb
@@ -4,7 +4,7 @@ require 'cloud_controller/diego/bbs_environment_builder'
 
 module VCAP::CloudController
   module Diego
-    class InvalidWindowsGMSACredentials < CloudController::Errors::ApiError; end
+    class InvalidWindowsGMSACredentials < StandardError; end
     class WindowsEnvironmentSage
       def self.ponder(app)
         windows_gmsa_credential_env(app)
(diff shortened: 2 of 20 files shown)


tcdowney:
Once this is approved/merged I will PR https://github.com/cloudfoundry/docs-dev-guide/blob/master/deploy-apps/environment-variable.html.md.erb to explain this env var.

app/models/runtime/app_model.rb
```diff
@@ -125,6 +125,12 @@ def database_uri
       DatabaseUriGenerator.new(service_binding_uris).database_uri
     end
 
+    def windows_gmsa_credential_refs
+      service_bindings.sort_by(&:id).map do |binding|
+        binding.credentials['credhub-windows-gmsa-credential-ref'] if binding.credentials.present?
```
Samze, line 130: **question** Is non-credhub sources of this credential a valid usecase?

lib/cloud_controller/diego/windows_environment_sage.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'cloud_controller/diego/bbs_environment_builder'
+
+module VCAP::CloudController
+  module Diego
+    class WindowsEnvironmentSage
```
Samze, line 7: **question** Just curious, but this will set the env var if that credential is present and it will be injected into a linux garden container. Is that a valid usecase? Using linux containers but with windwos GMSA creds?

tcdowney: It is not a valid use case, but also does not cause any problems. I thought about putting a `if stack.starts_with?('windows')` type of check, but it felt like it could be very brittle since `stack` names are operator-defined and can be completely arbitrary. E.g. someone could name `windows` to `win25` and break my check.

Cloud Controller does not actually know about Linux vs Windows so I didn't want to make this too clever.

app/models/runtime/app_model.rb
tcdowney: Maybe. So far from the feedback we've gotten there is a preference that they be in a secret management system other than CCDB. I could make it support both, but so far I'm not aware of anyone asking for it.

winkingturtle-vmw: No, using Credhub would be a requirement for this usecase.

lib/cloud_controller/diego/windows_environment_sage.rb
tcdowney: @winkingturtle-vmw mentioned to me that there may be future usecases for .NET Core running on Linux that could make use of this. Out of scope, but another reason not to attempt to restrict it to Windows only. Even in those cases the env var name is still valid, though, because it is for Integrated Windows Auth.

app/models/runtime/app_model.rb
Samze: 👍  If theres a future usecase seems easy enough to add another key here that populates the env var.

lib/cloud_controller/diego/windows_environment_sage.rb
Samze: Sounds good

Samze (approved):

PR #4395, opened 2025-06-05
Force delete very old delayed jobs

svkrieger:
Force delete delayed_jobs, which are older (run_at) then the configured cutoff_age_in_days + 1 day.

We found some old jobs in the delayed_jobs table, which are beyond the cutoff_age_in_days interval. They had either no failed_at set or were still locked. With this change the FailedJobsCleanup job will force delete all jobs, which are older than the cutoff_age_in_days + 1 day.

### DIFF
--- a/spec/unit/jobs/runtime/failed_jobs_cleanup_spec.rb
+++ b/spec/unit/jobs/runtime/failed_jobs_cleanup_spec.rb
@@ -64,9 +64,40 @@ module VCAP::CloudController
           end
 
           context 'when older than specified cut-off' do
-            let(:run_at) { Time.now.utc - 3.days }
+            let(:run_at) { Time.now.utc - 50.hours }
 
             it 'removes the job' do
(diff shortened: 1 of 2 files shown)


spec/unit/jobs/runtime/failed_jobs_cleanup_spec.rb
```diff
@@ -73,6 +73,38 @@ def max_attempts
                 Delayed::Job.find(id: @delayed_job.id)
               }.from(@delayed_job).to(nil)
             end
+
+            context 'when job is orphaned' do
+              it 'does not remove the job if it is not older than cut-off + 1 day' do
+                Sequel::Model.db[:delayed_jobs].where(id: @delayed_job.id).update(failed_at: nil, locked_by: nil)
+                puts @delayed_job.inspect
```
philippthun: Looks like a leftover...

PR #4402, opened 2025-06-11
Improve error messaging for creating a Deployment that will violate quotas with high max-in-flight

nookala:
Fix error message to be more descriptive

When creating a Deployment, it can be unclear that a quota violation is due to a high max-in-flight. Enhance the error message with better description and suggestions for remediation.

Previously the error was "memory space_quota_exceeded". With the change it's "memory space_quota_exceeded for space space1. This space's quota may not be large enough to support rolling deployments or your configured max-in-flight."
For an org level quota error after change is as follows "memory quota_exceeded for organization org1. This organization's quota may not be large enough to support rolling deployments or your configured max-in-flight."


app/models/runtime/constraints/max_memory_policy.rb
```diff
@@ -13,7 +13,8 @@ def validate
 
     return if policy_target.has_remaining_memory(requested_memory)
 
-    resource.errors.add(field, error_name)
+    resource.errors.add(field,
+                        "exceeded for space #{policy_target.name}. This space's quota may not be large enough to support rolling deployments or your configured max-in-flight")
```
sethboyles: I think this will affect the error message for all places we encounter the max memory policy, when we just want to show this particular message during Rolling Deployments.  You might want to take a look at the DeploymentCreate action to see if we can translate the error there or find some way of using a different policy.

It looks like we pass in the error name to the MaxMemoryPolicy (like [here](github.com/cloudfoundry/cloud_controller_ng/blob/main/app/models/runtime/process_model.rb#L257) so maybe we could call these explicitly in DeploymentCreate before doing the standard process validations, or something like that.

app/models/runtime/constraints/max_memory_policy.rb
```diff
@@ -13,7 +13,8 @@ def validate
 
     return if policy_target.has_remaining_memory(requested_memory)
 
-    resource.errors.add(field, error_name)
+    resource.errors.add(field,
```
sethboyles: We'd also want to update the error messages for the other aspects of quotas: disk, log rate limit, etc.

app/actions/deployment_create.rb
```diff
@@ -75,7 +75,13 @@ def create(app:, user_audit_info:, message:)
 
         deployment
       rescue RevisionResolver::NoUpdateRollback, Sequel::ValidationFailed, AppStart::InvalidApp => e
-        error = DeploymentCreate::Error.new(e.message)
+        space_quota_errors = [:space_quota_exceeded.to_s, :space_app_instance_limit_exceeded.to_s]
```
sethboyles: I think we'd want to do something similar for org-level quotas: https://github.com/cloudfoundry/cloud_controller_ng/blob/12ebacb24ad9c33b5e69c5b7a0bbc79a80c25974/app/models/runtime/process_model.rb#L257

app/actions/deployment_create.rb
```diff
@@ -83,9 +83,24 @@ def create(app:, user_audit_info:, message:)
 
         deployment
       rescue RevisionResolver::NoUpdateRollback, Sequel::ValidationFailed, AppStart::InvalidApp => e
-        error = DeploymentCreate::Error.new(e.message)
+        raise handle_deployment_create_error(e, app)
+      end
+
+      def handle_deployment_create_error(e, app)
+        space_quota_errors = [:space_quota_exceeded.to_s, :space_app_instance_limit_exceeded.to_s]
```
sethboyles: *nit* instead of making symbols and turning them to Strings, you could make an array of literals with this syntax: 

```
%w(space_quota_exceeded space_app_instance_limit_exceeded)
```

app/actions/deployment_create.rb
```diff
@@ -83,9 +83,24 @@ def create(app:, user_audit_info:, message:)
...
+        org_quota_errors = [:quota_exceeded.to_s, :app_instance_limit_exceeded.to_s]
+        if space_quota_errors.any? { |substring| e.message.include?(substring) }
+          space_error_msg = " for space #{app.space.name}. This space's quota may not be large enough to support rolling deployments or your configured max-in-flight."
+          error = DeploymentCreate::Error.new(e.message + space_error_msg)
+        elsif org_quota_errors.any? { |substring| e.message.include?(substring) }
+          org_error_msg_1 = " for organization #{app.organization.name}. "
+          org_error_msg_2 = "This organization's quota may not be large enough to support rolling deployments or your configured max-in-flight."
+          error = DeploymentCreate::Error.new(e.message + org_error_msg_1 + org_error_msg_2)
```
sethboyles: I wonder if it's worth all this extra logic to make the space/org distinction.  What if you checked for any quota error and simply added `One or more quotas may not be large enough to support rolling deployments or your configured max-in-flight.` to then end?  Could simplify the code a bunch.  What do you think?

sethboyles:
Tested this out:

```
Creating deployment for app dora...

memory space_quota_exceeded, memory quota_exceeded, app_instance_limit space_app_instance_limit_exceeded, log_rate_limit exceeds space log rate quota for space space. This space's quota may not be large enough to support rolling deployments or your configured max-in-flight.
FAILED
```

Definitely is more helpful in the context of deployments.  👍

app/actions/deployment_create.rb
nookala: I think it would be useful to have the distinction to narrow down the error and update the quota for the appropriate space/org.

app/actions/deployment_create.rb
```diff
@@ -83,7 +83,25 @@ def create(app:, user_audit_info:, message:)
 
         deployment
       rescue RevisionResolver::NoUpdateRollback, Sequel::ValidationFailed, AppStart::InvalidApp => e
-        error = DeploymentCreate::Error.new(e.message)
+        raise handle_deployment_create_error(e, app)
```
sethboyles: isn't there a `raise` inside the method already? I think it's clearer what's going on if you leave it out, but then maybe remove the `raise` on L106 and rename the method to something like `enhanced_deployment_create_error`

sethboyles: It think it's also fine to leave it in the method, but probably not in both places

sethboyles (approved):

Issue #4372, opened 2025-05-23
Support the ability to run Processes with a user other than`vcap`


tcdowney:
## Problem

Apps with the CNB or Buildpack lifecycle¹ always run as the `vcap` user. This effectively hardcodes assumptions about the [underlying stack](https://github.com/cloudfoundry/cflinuxfs4/blob/d10cd4a40f5b2add77807e1b5d9f532213a6b3c2/Dockerfile#L26) supporting a user named `vcap` in CAPI and is not flexible for other runtimes/stacks. Some stacks like Windows have permissions and behaviors for its built in users -- `ContainerUser` and `ContainerAdmin` -- that the `vcap` user cannot inherit (see: https://github.com/microsoft/Windows-Containers/issues/390).

¹ Docker apps have more flexibility here and can [run as the user declared in the image metadata](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/cloud_controller/diego/docker/desired_lrp_builder.rb#L71).

## Proposal

We make `user` a first-class property on the `Process` that when set will override `vcap`. To prevent misuse while maintaining flexibility we will introduce a new config property in capi-release to serve as an allow list of valid usernames.

Example BOSH spec:

```
cc.allowed_process_users:
    description: "Permitted users that processes may use"
    default: ['vcap']
```

We will modify the [Update a Process](https://v3-apidocs.cloudfoundry.org/version/3.195.0/index.html#update-a-process) API and app manifest to accept the `user` field.

Example manifest:

```
---
version: 1
applications:
- name: dot-net-app
  stack: windows
  processes:
  - type: web
    user: 'containeruser'
```

Any Process that has a `user` set will then use this value in place of any defaults.

### What if the app does not support the specified user?

If the app does not actually support running as this user then it will fail to start. Attempting to create and set up permissions for arbitrary users is out of scope of this proposal.

## Alternatives Considered

We considered making a new app feature for Windows apps called `windows-run-as-containeruser`. However this felt too inflexible and specific to a particular use case to be a general app feature. Rather than do that, we feel that making a proper developer-settable `user` field will account for unforeseen use cases in the future and open up the ability to support stacks/buildpacks that don't use `vcap`.

tcdowney:
Another thing to think about is how the proposed allow list applies to Docker apps which already have more complicated logic here:

```
        def action_user
          execution_metadata = Oj.load(@execution_metadata)
          user = execution_metadata['user']
[...]

PR #4407, opened 2025-06-13
Allow developer to specify run action user on Processes

tcdowney:
This adds a new field to the Process called `user` which is used to override the default run action user from `vcap` to something else. The use case this specifically solves is allowing Windows apps to opt-in to running as the built-in Windows `ContainerUser` user. It makes configuring user a bit more straightforward for Docker applications as well, but is not useful for the cflinux stacks since those stacks only support the `vcap` user.

Currently the list of permitted users is controlled by a new capi-release BOSH property called `cc.allowed_process_users` (see https://github.com/cloudfoundry/capi-release/pull/555), but a possible future enhancement would be to make `run_user` and `build_user` first-class properties on `Stack` resources so that stacks could declare what users they support explicitly.

See https://github.com/cloudfoundry/cloud_controller_ng/issues/4372 for more details on the use cases this solves.

I plan on submitting follow-up PRs to support this configuration via Manifests and on Tasks.


app/models/runtime/constraints/process_user_policy.rb
```diff
@@ -0,0 +1,16 @@
+class ProcessUserPolicy
+  ERROR_MSG = 'invalid (requested user %<requested_user>s not allowed, permitted users are: %<allowed_users>s)'.freeze
```
sethboyles, line 2: **nit** I think some scare quotes would helpful here:
```
user invalid (requested user 'root' not allowed, permitted users are: 'vcap')
```
instead of

```
user invalid (requested user root not allowed, permitted users are: vcap)
```

sethboyles:
Interestingly I can get dora to work with user `root`, which is not surprising, but rolling deployments no longer work.  The first instance comes up as 'running'.  I think there might be some extra checks on whether or not an instance is 'routable' that is preventing it from progressing.  I'd have to revisit the DeploymentUpdater code, though.

I think the same thing is happening with docker apps if you set 'root' user.  You can do a plain old restart, though.

Do you think `user` should be tracked on Revisions?

tcdowney:
I think `user` is a reserved column on MySQL (only tested with Postgres locally), I'm going to rename the column to `action_user` (but keep APIs the same).

app/models/runtime/constraints/process_user_policy.rb
tcdowney: Good idea, done: https://github.com/cloudfoundry/cloud_controller_ng/pull/4407/commits/102a954360d272bd837c04c0a59c0bf839d8061a

db/migrations/20250610212414_add_user_to_processes.rb
```diff
@@ -0,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table :processes do
+      add_column :user, String, null: true, default: nil, size: 255 unless @db.schema(:processes).map(&:first).include?(:user)
```
Samze, line 4: **question/non-blocking** Is 255 just a length we set to limit the size or is it actually some linux/windows limitation on user?

app/models/runtime/process_model.rb
```diff
@@ -564,6 +572,18 @@ def open_ports
 
     private
 
+    def permitted_users
+      Set.new([DEFAULT_USER]) + Config.config.get(:allowed_process_users)
```
Samze: **question** Do we want to always include the default (vcap) here ? Shouldn't it just be set in `allowed_process_users`?

I see its set as a default here https://github.com/cloudfoundry/capi-release/pull/555/files#diff-2840526eabf042bb611403427f10ee9c514ffd6944636ffab0129d1d7a0553d6R426

tcdowney: Since `vcap` is the default user that's used if the field is unset I wanted to make sure it was clear that `vcap` is always a permitted user for the sake of this validation.

It might be a bit paranoid, but if feels like `vcap` always being permitted is more a property of this codebase than it is capi-release. Things like this exist which I don't plan on unwinding at this time:

* https://github.com/cloudfoundry/cloud_controller_ng/blob/12ebacb24ad9c33b5e69c5b7a0bbc79a80c25974/lib/cloud_controller/diego/staging_action_builder.rb#L153
* https://github.com/cloudfoundry/cloud_controller_ng/blob/12ebacb24ad9c33b5e69c5b7a0bbc79a80c25974/lib/cloud_controller/diego/constants.rb#L10

Samze: Yeah I think always permitting vcap makes sense. https://github.com/cloudfoundry/capi-release/pull/555/files#diff-2840526eabf042bb611403427f10ee9c514ffd6944636ffab0129d1d7a0553d6R426 is a little confusing, as if vcap appears in the default list it suggests if I override the list and don't include vcap it will be removed. Perhaps that field description should be updated to explain that vcap is always allowed?

tcdowney: yeah I can do that 👍

tcdowney: Updated on that PR https://github.com/cloudfoundry/capi-release/pull/555/files#diff-df2c1f9acc90ddf6fd241b1a51c92f45a0940c1d54a68e65a22691883ea57c89R849

db/migrations/20250610212414_add_user_to_processes.rb
tcdowney: I set it to that because varchar(255) is the most common string type we use on MySQL at least.
https://github.com/search?q=repo%3Acloudfoundry%2Fcloud_controller_ng%20size%3A%20255&type=code

We only request something bigger if there is a good reason to. On Linux I think the maximum username length is 32 characters and on Windows it is generally around 20. But I think there are exceptions to that where it can be larger.

This is effectively gated by what the stack supports and what the operator has configured in the allow-list I decided not to think about it too much further. Even the characterset itself is difficult to lock down because we're talking about multiple OSes here and not even Unix is super prescriptive (e.g. https://datatracker.ietf.org/doc/html/rfc8265).

config/cloud_controller.yml
```diff
@@ -67,6 +67,7 @@ maximum_app_disk_in_mb: 2048
 max_retained_deployments_per_app: 100
 max_retained_builds_per_app: 100
 max_retained_revisions_per_app: 100
+allowed_process_users: ['vcap', 'ContainerUser']
```
Samze:
```suggestion
allowed_process_users: ['ContainerUser']
```

jochenehret:
This change causes the "Docker Application Lifecycle" tests of the CATs suite to fail:
https://concourse.app-runtime-interfaces.ci.cloudfoundry.org/teams/capi-team/pipelines/capi/jobs/gyro-exp-tests/builds/38
https://concourse.app-runtime-interfaces.ci.cloudfoundry.org/teams/capi-team/pipelines/capi/jobs/olaf-mysql-tests/builds/39

```
  [FAIL] [docker] Docker Application Lifecycle running a docker app without a start command [BeforeEach] handles docker-defined metadata and environment variables correctly
  /tmp/build/1f5f819e/cf-acceptance-tests/docker/docker_lifecycle.go:80
  [FAIL] [docker] Docker Application Lifecycle running a docker app with a start command [BeforeEach] retains its start command through starts and stops
  /tmp/build/1f5f819e/cf-acceptance-tests/docker/docker_lifecycle.go:53
  [FAIL] [docker] Docker Application Lifecycle running a docker app without a start command [BeforeEach] when env vars are set with 'cf set-env' prefers the env vars from cf set-env over those in the Dockerfile
  /tmp/build/1f5f819e/cf-acceptance-tests/docker/docker_lifecycle.go:80
  [FAIL] [docker] Docker App Lifecycle CredHub Integration when CredHub is configured service bindings [JustBeforeEach] in assisted mode the app should see the service creds
  /tmp/build/1f5f819e/cf-acceptance-tests/docker/credhub_enabled.go:105
```
I found the following stack trace in the cloud controller logs:
```
{"timestamp":"2025-06-23T11:22:28.000598027Z","message":"Request failed: 500: {\"errors\"=\u003e[{\"title\"=\u003e\"UnknownError\", \"detail\"=\u003e\"An unknown error occurred.\", \"code\"=\u003e10001, \"test_mode_info\"=\u003e{\"detail\"=\u003e\"unexpected character (after ) at line 1, column 1 [parse.c:706] in '\", \"title\"=\u003e\"CF-EncodingError\", \"backtrace\"=\u003e[\"/var/vcap/data/packages/cloud_controller_ng/15dbf508b71c470d5ed4a0e29f62d3d68928ae12/cloud_controller_ng/app/models/runtime/process_model.rb:582:in `load'\", \"/var/vcap/data/packages/cloud_controller_ng/15dbf508b71c470d5ed4a0e29f62d3d68928ae12/cloud_controller_ng/app/models/runtime/process_model.rb:582:in `docker_run_action_user'\", \"/var/vcap/data/packages/cloud_controller_ng/15dbf508b71c470d5ed4a0e29f62d3d68928ae12/cloud_controller_ng/app/models/runtime/process_model.rb:399:in `run_action_user'\", \"/var/vcap/data/packages/cloud_controller_ng/15dbf508b71c470d5ed4a0e29f62d3d68928ae12/cloud_controller_ng/app/presenters/v3/process_presenter.rb:31:in `to_hash'\", 
```
Can you please check and provide a fix?

PR #4442, opened 2025-07-10
Log DB query request metrics

philippthun:
- implement Sequel extension for counting DB queries and accumulating DB query times
- store DB request metrics in thread-local variables
- reset metrics in the `request_metrics` middleware (the first one)
- log metrics on request completion in the `request_logs` middleware

### DIFF
--- a/lib/cloud_controller/logs/request_logs.rb
+++ b/lib/cloud_controller/logs/request_logs.rb
@@ -26,17 +26,21 @@ module VCAP::CloudController
         @incomplete_requests.store(request_id, env)
       end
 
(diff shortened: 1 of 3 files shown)


jochenehret (changes requested):
Works fine if `log_db_queries` is `true`. If it is `false`, the keys `total_db_query_time_in_ms` and `db_query_count` are logged with constant values 0. `time_taken_in_ms` is correctly logged. The `db_query` keys should only be logged f `log_db_queries` is `true`.

jochenehret (approved):

PR #4452, opened 2025-07-14
Optionally disallow the use of 'root' user for Processes and Tasks of docker lifecycle Apps

acosta11:
### Proposed Change
Add a configuration option to allow or disallow the use of the 'root' and '0' user for processes and tasks of docker lifecycle apps.

~~Use 'vcap' as the default user in the case that the 'root' user is disallowed.~~ Edit: opting to leave the user as is given docker images need to add the desired user anyway, which makes a non-root default less clear.

### Use Case
Operators can secure their runtime infrastructure by preventing the possibility of cascading user privilege escalation to the root user of the underlying os namespace construct realizing the desired Process.

### Dev Notes
~~Given the scope of docker lifecycle apps, it seems like the validation logic for this feature has to live in the policy layer. If implemented in the message layer, I think it may have the unintended effect of allowing the root user in the net new context of buildpack lifecycle apps when configured to match the current default of allowed for docker apps (assuming the flag exists at this level of coarseness).~~ Edit: the user allow list functionality in the policy layer works to this end, but we still need to ensure docker apps that set the user in the Dockerfile are appropriately handled.

Note this implementation specifically only checks for the 'root' or '0' user, so apps that appropriately set non-root users in the Dockerfile do not have to go through the same additional allowed users list validation as the Process and Task model. This behavior is technically inconsistent with other lifecycles and could be confusing for some users, but it does offer the advantage of more compatibility with existing docker apps.  And technically the behavior is consistent with the spirit of preventing root user privilege escalation.

Confirmed in manual end to end testing that disallowing 'root' user on the cloud_controller_clock will cause an error in the nsync process for apps with the root user that were previously valid. Triggered a new sync by manually deleting the lrp on the diego cell. The error does not get surfaced to the api/cli. Instead, a user would have to go into the bosh logs and see the attempted sync on the scheduler job. Note that allowing the 'root' user on the clock job avoids this error, but still prevents new process/task updates from using the 'root' user, so it seems like a reasonable option for backwards compatibility.

Example error:
```
{
  "timestamp": "2025-07-29T20:36:27.040856189Z",
  "message": "error-updating-lrp-state",
  "log_level": "error",
  "source": "cc.diego.sync.processes",
  "data": {
    "error": "RunnerError",
    "error_message": "Runner error: 'root' user not permitted.",
    "error_backtrace": "/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/cloud_controller/diego/docker/lifecycle_protocol.rb:34:in `desired_lrp_builder'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/cloud_controller/diego/app_recipe_builder.rb:54:in `app_lrp_arguments'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/cloud_controller/diego/app_recipe_builder.rb:31:in `build_app_lrp'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/cloud_controller/diego/bbs_apps_client.rb:13:in `block in desire_app'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/cloud_controller/diego/bbs_apps_client.rb:107:in `handle_diego_errors'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/cloud_controller/diego/bbs_apps_client.rb:12:in `desire_app'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/cloud_controller/diego/processes_sync.rb:104:in `block (3 levels) in desire_lrps'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/utils/workpool.rb:41:in `block (3 levels) in create_workpool_thread'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/utils/workpool.rb:39:in `loop'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/utils/workpool.rb:39:in `block (2 levels) in create_workpool_thread'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/utils/workpool.rb:38:in `catch'\n/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/utils/workpool.rb:38:in `block in create_workpool_thread'\n..."
  },
  "thread_id": 59180,
  "fiber_id": 59200,
  "process_id": 7,
  "file": "/var/vcap/data/packages/cloud_controller_ng/5478c77a2ce8a293944ab99954deb8c0e045a5fd/cloud_controller_ng/lib/cloud_controller/diego/processes_sync.rb",
  "lineno": 80,
  "method": "block in process_workpool_exceptions"
}
```

### Checklist

### DIFF
--- a/lib/cloud_controller/diego/desire_app_handler.rb
+++ b/lib/cloud_controller/diego/desire_app_handler.rb
@@ -12,6 +12,8 @@ module VCAP::CloudController
               if e.name == 'RunnerError' && e.message['the requested resource already exists']
                 existing_lrp = client.get_app(process)
                 client.update_app(process, existing_lrp)
+              elsif e.name == 'UnprocessableEntity'
+                raise
               end
             end
           end
--- a/lib/cloud_controller/diego/docker/lifecycle_protocol.rb
+++ b/lib/cloud_controller/diego/docker/lifecycle_protocol.rb
@@ -8,6 +8,8 @@ module VCAP
     module Diego
       module Docker
         class LifecycleProtocol
+          ROOT_USERS = %w[root 0].freeze
+
           def lifecycle_data(staging_details)
             lifecycle_data              = Diego::Docker::LifecycleData.new
             lifecycle_data.docker_image = staging_details.package.image
@@ -22,10 +24,18 @@ module VCAP
           end
 
           def task_action_builder(config, task)
+            if task.docker? && !docker_run_action_user_permitted?(task.run_action_user)
+              raise ::CloudController::Errors::ApiError.new_from_details('UnprocessableEntity', 'Attempting to run task as root user, which is not permitted.')
+            end
+
             TaskActionBuilder.new(config, task, { droplet_path: task.droplet.docker_receipt_image })
           end
 
           def desired_lrp_builder(config, process)
+            if process.docker? && !docker_run_action_user_permitted?(process.run_action_user)
+              raise ::CloudController::Errors::ApiError.new_from_details('UnprocessableEntity', 'Attempting to run process as root user, which is not permitted.')
+            end
+
             DesiredLrpBuilder.new(config, builder_opts(process))
           end
 
@@ -45,6 +55,10 @@ module VCAP
           def container_env_vars_for_process(process)
             additional_env = []
             additional_env + WindowsEnvironmentSage.ponder(process.app)
+          end
+
+          def docker_run_action_user_permitted?(run_action_user)
+            Config.config.get(:allow_docker_root_user) || ROOT_USERS.exclude?(run_action_user)
           end
         end
       end
--- a/spec/unit/lib/cloud_controller/diego/desire_app_handler_spec.rb
+++ b/spec/unit/lib/cloud_controller/diego/desire_app_handler_spec.rb
@@ -51,6 +51,18 @@ module VCAP::CloudController
             expect(client).to have_received(:get_app).exactly(2).times
           end
         end
+
+        context 'when root user is not permitted' do
+          it 'raises a CloudController::Errors::ApiError' do
+            allow(client).to receive(:desire_app).and_raise(CloudController::Errors::ApiError.new_from_details('UnprocessableEntity',
+                                                                                                               'Attempting to run process as root user, which is not permitted.'))
+            expect { DesireAppHandler.create_or_update_app(process, client) }.to(raise_error do |error|
+              expect(error).to be_a(CloudController::Errors::ApiError)
+              expect(error.name).to eq('UnprocessableEntity')
+              expect(error.message).to include('Attempting to run process as root user, which is not permitted')
+            end)
+          end
+        end
       end
     end
   end
--- a/spec/unit/lib/cloud_controller/diego/docker/lifecycle_protocol_spec.rb
+++ b/spec/unit/lib/cloud_controller/diego/docker/lifecycle_protocol_spec.rb
@@ -118,6 +118,162 @@ module VCAP
                 end
               end
             end
+
+            context 'when root user is allowed' do
+              let(:app) { AppModel.make(:docker, { droplet: }) }
+
+              before do
+                TestConfig.override(allow_docker_root_user: true, additional_allowed_process_users: %w[root 0])
+              end
+
+              context 'and the process sets the root user' do
+                let(:process) { ProcessModel.make(:docker, { app: app, user: 'root' }) }
+
+                it 'creates a diego DesiredLrpBuilder' do
+                  expect do
+                    lifecycle_protocol.desired_lrp_builder(config, process)
+                  end.not_to raise_error
+                end
+              end
+
+              context 'and the process does not set a user' do
+                let(:process) { ProcessModel.make(:docker, { app: }) }
+
+                context 'and the droplet docker execution metadata sets the root user' do
+                  let(:droplet_execution_metadata) { '{"entrypoint":["/image-entrypoint.sh"],"user":"root"}' }
+                  let(:droplet) do
+                    DropletModel.make(:docker, {
+                                        state: DropletModel::STAGED_STATE,
+                                        docker_receipt_image: 'the-image',
+                                        execution_metadata: droplet_execution_metadata
+                                      })
+                  end
+
+                  it 'creates a diego DesiredLRPBuilder' do
+                    expect do
+                      lifecycle_protocol.desired_lrp_builder(config, process)
+                    end.not_to raise_error
+                  end
+                end
+
+                context 'and the droplet docker execution metadata sets the 0 user' do
+                  let(:droplet_execution_metadata) { '{"entrypoint":["/image-entrypoint.sh"],"user":"0"}' }
+                  let(:droplet) do
+                    DropletModel.make(:docker, {
+                                        state: DropletModel::STAGED_STATE,
+                                        docker_receipt_image: 'the-image',
+                                        execution_metadata: droplet_execution_metadata
+                                      })
+                  end
+
+                  it 'creates a diego TaskActionBuilder' do
+                    expect do
+                      lifecycle_protocol.desired_lrp_builder(config, process)
+                    end.not_to raise_error
+                  end
+                end
+
+                context 'and the droplet docker execution metadata does not set a user' do
+                  let(:droplet_execution_metadata) { '{"entrypoint":["/image-entrypoint.sh"]}' }
+                  let(:droplet) do
+                    DropletModel.make(:docker, {
+                                        state: DropletModel::STAGED_STATE,
+                                        docker_receipt_image: 'the-image',
+                                        execution_metadata: droplet_execution_metadata
+                                      })
+                  end
+
+                  it 'creates a diego TaskActionBuilder' do
+                    expect do
+                      lifecycle_protocol.desired_lrp_builder(config, process)
+                    end.not_to raise_error
+                  end
+                end
+              end
+            end
+
+            context 'when root user IS NOT allowed' do
+              let(:app) { AppModel.make(:docker, { droplet: }) }
+
+              before do
+                TestConfig.override(allow_docker_root_user: false, additional_allowed_process_users: %w[root 0])
+              end
+
+              context 'and the process does not set a user' do
+                let(:process) { ProcessModel.make(:docker, { app: }) }
+
+                context 'and the droplet docker execution metadata sets the root user' do
+                  let(:droplet_execution_metadata) { '{"entrypoint":["/image-entrypoint.sh"],"user":"root"}' }
+                  let(:droplet) do
+                    DropletModel.make(:docker, {
+                                        state: DropletModel::STAGED_STATE,
+                                        docker_receipt_image: 'the-image',
+                                        execution_metadata: droplet_execution_metadata
+                                      })
+                  end
+
+                  it 'raises an error' do
+                    expect do
+                      lifecycle_protocol.desired_lrp_builder(config, process)
+                    end.to raise_error(::CloudController::Errors::ApiError, /Attempting to run process as root user, which is not permitted/)
+                  end
+                end
+
+                context 'and the droplet docker execution metadata sets the 0 user' do
+                  let(:droplet_execution_metadata) { '{"entrypoint":["/image-entrypoint.sh"],"user":"0"}' }
+                  let(:droplet) do
+                    DropletModel.make(:docker, {
+                                        state: DropletModel::STAGED_STATE,
+                                        docker_receipt_image: 'the-image',
+                                        execution_metadata: droplet_execution_metadata
+                                      })
+                  end
+
+                  it 'raises an error' do
+                    expect do
+                      lifecycle_protocol.desired_lrp_builder(config, process)
+                    end.to raise_error(::CloudController::Errors::ApiError, /Attempting to run process as root user, which is not permitted/)
+                  end
+                end
+
+                context 'and the droplet docker execution metadata does not set a user' do
+                  let(:droplet_execution_metadata) { '{"entrypoint":["/image-entrypoint.sh"]}' }
+                  let(:droplet) do
+                    DropletModel.make(:docker, {
+                                        state: DropletModel::STAGED_STATE,
+                                        docker_receipt_image: 'the-image',
+                                        execution_metadata: droplet_execution_metadata
+                                      })
+                  end
+
+                  it 'raises an error' do
+                    expect do
+                      lifecycle_protocol.desired_lrp_builder(config, process)
+                    end.to raise_error(::CloudController::Errors::ApiError, /Attempting to run process as root user, which is not permitted/)
+                  end
+                end
+              end
+
+              context 'and the process sets the root user' do
+                let(:process) { ProcessModel.make(:docker, { app: app, user: 'root' }) }
+
+                it 'raises an error' do
+                  expect do
+                    lifecycle_protocol.desired_lrp_builder(config, process)
+                  end.to raise_error(::CloudController::Errors::ApiError, /Attempting to run process as root user, which is not permitted/)
+                end
+              end
+
+              context 'and the process sets the 0 user' do
+                let(:process) { ProcessModel.make(:docker, { app: app, user: 0 }) }
+
+                it 'raises an error' do
+                  expect do
+                    lifecycle_protocol.desired_lrp_builder(config, process)
+                  end.to raise_error(::CloudController::Errors::ApiError, /Attempting to run process as root user, which is not permitted/)
+                end
+              end
+            end
           end
 
           describe '#task_action_builder' do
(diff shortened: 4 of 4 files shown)


acosta11:
Oh looks like I had pivotal-cf membership visibility set to private. Is there a way to rerun the CLA check to see if I need any other updates?

app/models/runtime/process_model.rb
```diff
@@ -576,9 +576,7 @@ def permitted_users
     end
 
     def docker_run_action_user
-      return AppModel::DEFAULT_CONTAINER_USER unless docker?
```
acosta11, line 579: I _think_ this ends up being a duplicate `docker?` check if docker_run_action_user only gets called after the model has checked that the process belongs to a docker lifecycle app.

Gerg:
@acosta11 you will also need to open a PR to add this new property to https://github.com/cloudfoundry/capi-release

app/models/runtime/droplet_model.rb
```diff
@@ -138,7 +138,7 @@ def docker_user
         end
       end
 
-      container_user.presence || AppModel::DEFAULT_DOCKER_CONTAINER_USER
+      container_user.presence || (Config.config.get(:allow_process_root_user) ? AppModel::DEFAULT_DOCKER_CONTAINER_USER : AppModel::DEFAULT_CONTAINER_USER)
```
Gerg: This is semantically a bit confusing, since it means in some cases the `docker_user` default is not the `DEFAULT_DOCKER_USER`, but is instead the `DEFAULT_CONTAINER_USER`. Also we have to duplicate the same ternary in both droplet and process model.

Maybe instead (roughly):
1. Encapsulate the ternary in a method (on AppModel?) `default_docker_user`
1. `DEFAULT_DOCKER_CONTAINER_USER` goes away (since it's only used in one place)

app/models/runtime/constraints/process_user_policy.rb
```diff
@@ -9,6 +9,9 @@ def initialize(process, allowed_users)
 
   def validate
     return if @process.user.blank?
+
+    return if @process.docker? && @process.user.downcase == 'root' && VCAP::CloudController::Config.config.get(:allow_process_root_user)
```
Gerg: 1. Is there a test for this `downcase`?
2. That said, do we actually need this `downcase`? Unix users are case-sensitive, so `root` != `RoOt`, right?

acosta11: That's fair, I sort of just pasta'd this from the existing test cases as I'm not familiar with the diego/garden stuff. Can update unless Tim recalls a specific reason to have included it.

app/models/runtime/droplet_model.rb
acosta11: Indeed, I could see trying to move this up higher in the hierarchy in general to avoid duplication. I'll give it a try and see how it looks.

acosta11:
> I haven't validated this myself yet, but I don't think this PR covers the case where the `root` user is coming from the Dockerfile. In that case, the user will be returned from staging as `root` and stored in the Droplet's `docker_exec_metadata`.

Indeed it's not exercised through that complete flow. I can look at adding a test to explicitly configure the user to root in the docker metadata. I think the setup that I copied was only setting image information so would have to update the call to the factory.

app/models/runtime/constraints/process_user_policy.rb
Gerg: Maybe because of Windows? I think Windows is not case-sensitive.

tcdowney: Yes, Windows is not case-sensitive and I was thinking with a Windows Mindset™ when I wrote that validation. It's possible it isn't useful/correct for Linux and can be revisited.

Gerg: We could be case-insensitive in CF, in the name of platform compatibility. I doubt anyone actually cares about making Docker images that run as `rOoT`.

acosta11: Went ahead and removed in 7344654760e795c474d2af8141dc3fa46595bd89 for consistency with existing behavior.

acosta11:
Wrapped up end to end testing, added some error handling from an issue detected as part of that process, and rebased on latest main with the task handler and ccng config updates. Should be good for review now.

See updated PR description for nsync edge case that I also validated in the e2e env. If anything, taking one last look at our handling of the `USER UID:GUID` directive instead of just `USER: root`.

acosta11:
Rebased on main and renamed property to `allow_docker_root_user` instead of `allow_process_root_user` since this applies to both Processes and Tasks. Bump @Gerg @tcdowney if y'all had any other thoughts.

app/models/runtime/constraints/process_user_policy.rb
```diff
@@ -9,6 +9,9 @@ def initialize(process, allowed_users)
 
   def validate
     return if @process.user.blank?
+
+    return if @process.docker? && @process.user == 'root' && VCAP::CloudController::Config.config.get(:allow_docker_root_user)
```
tcdowney: I think we should keep this separate from the concept of allowed_users. I think we should only allow setting process.user = 'root' if the operator adds `'root'` to the allowed user list.

app/models/runtime/droplet_model.rb
```diff
@@ -138,7 +138,7 @@ def docker_user
         end
       end
 
-      container_user.presence || AppModel::DEFAULT_DOCKER_CONTAINER_USER
+      container_user.presence || (Config.config.get(:allow_docker_root_user) ? AppModel::DOCKER_CONTAINER_ROOT_USER : AppModel::DEFAULT_CONTAINER_USER)
```
tcdowney: It took me a second to mentall parse this -- can we write it without the ternary? Same with the code below in the process model?

app/models/runtime/constraints/process_user_policy.rb
Gerg: I agree:
1. Setting process/task user via the API should be limited by `allowed_users`, including for Docker apps
2. Users extracted from Docker images should NOT be limited by `allowed_users`

Docker apps may have a bit of an awkward interaction with process/task user configuration. App devs can effectively specify the user via the Dockerfile anyway (assuming they control building/publishing the image).

One interesting usecase could be that the app dev is trying to push a docker app that runs as the Docker default, `root`, but doesn't actually require running as root. The operator has enabled "secure docker apps", so `root` is blocked. The app dev could use the process/task user to override `root` to a different user†, so they can run their otherwise-insecure docker app.

† assuming the image has another user available 🙂

lib/cloud_controller/diego/docker/lifecycle_protocol.rb
```diff
@@ -22,10 +22,18 @@ def staging_action_builder(config, staging_details)
           end
 
           def task_action_builder(config, task)
+            if task.run_action_user == 'root' && !Config.config.get(:allow_docker_root_user)
+              raise ::CloudController::Errors::ApiError.new_from_details('RunnerError', "'root' user not permitted.")
```
Gerg: This error message may be confusing/unclear to users, since they may not know their app is trying to run as root.

It'd be more luxurious to have something _like_ `attempting to run task as 'root' user, which is not permitted.`

spec/unit/lib/cloud_controller/diego/docker/lifecycle_protocol_spec.rb
```diff
@@ -118,6 +118,34 @@ module Docker
                 end
               end
             end
+
+            context 'when root user not allowed' do
```
Gerg: We probably want a test to ensure that `root` user is allowed when `allow_docker_root_user` is `true`.

spec/unit/lib/cloud_controller/diego/docker/lifecycle_protocol_spec.rb
```diff
@@ -118,6 +118,34 @@ module Docker
...
+              before do
+                TestConfig.override(allow_docker_root_user: false)
+              end
+
+              context 'and the process does not set a user' do
```
Gerg, line 202: I _think_ we also want to block the case where:
1. `allowed_users` includes `root`
2. process/task user is set to `root`
3. `allow_docker_root_user` is set to `false`

This case that may never happen in the real world, but I think that's the correct behavior.

spec/unit/lib/cloud_controller/diego/docker/lifecycle_protocol_spec.rb
```diff
@@ -118,6 +118,34 @@ module Docker
...
+                let(:app) { AppModel.make(:docker, { droplet: }) }
+                let(:process) { ProcessModel.make(:docker, { app: }) }
+
+                context 'and the droplet docker execution metadata sets the root user' do
+                  let(:droplet_execution_metadata) { '{"entrypoint":["/image-entrypoint.sh"],"user":"root"}' }
```
Gerg, line 206: We had discussed offline the case where the docker user is set to `0` (linux user id for the `root` user). Do we need to cover that case here also?

lib/cloud_controller/diego/desire_app_handler.rb
Gerg, line 1: Do we need equivalent logic for surfacing the error when running tasks?

lib/cloud_controller/diego/desire_app_handler.rb
```diff
@@ -12,6 +12,8 @@ def create_or_update_app(process, client)
               if e.name == 'RunnerError' && e.message['the requested resource already exists']
                 existing_lrp = client.get_app(process)
                 client.update_app(process, existing_lrp)
+              elsif e.name == 'RunnerError' && e.message["'root' user not permitted"]
```
Gerg: I know this is continuing an existing pattern, but this is pretty brittle to error message changes.

What about something like:
1. client/lifecycle raises a custom error type
2. handler catches that error type and re-raises API error

spec/unit/models/runtime/droplet_model_spec.rb
```diff
@@ -359,41 +359,91 @@ module VCAP::CloudController
...
+              end
+            end
+
+            context 'when the droplet execution metadata has invalid json' do
+              let(:droplet_execution_metadata) { '{' }
 
-          it 'defaults the user to root' do
-            expect(droplet_model.docker_user).to eq('root')
+              it 'defaults the user to root' do
+                expect(droplet_model.docker_user).to eq('vcap')
```
Gerg: This assertion doesn't match the test description. Same for others in the file.

spec/unit/models/runtime/task_model_spec.rb
```diff
@@ -214,41 +214,26 @@ module VCAP::CloudController
...
           end
-        end
-
-        context 'when the droplet execution metadata has invalid json' do
-          let(:droplet_execution_metadata) { '{' }
 
-          it 'defaults the user to root' do
-            expect(task.run_action_user).to eq('root')
+          it 'returns the default "vcap" user' do
+            expect(task.run_action_user).to eq('vcap')
```
Gerg: Why is the default changing here? Is this because of the `allow_docker_root_user` config used in tests? 

I wouldn't expect this to change, since the default should be the existing behavior (`root` is allowed, and `root` is the default user).

app/models/runtime/constraints/process_user_policy.rb
acosta11: Sounds good. Assuming the app is just going through the docker app cf push flow that doesn't set a user anyway, the short-circuiting in the lower level docker execution metadata based check may be sufficient to support this behavior. Will see how it runs. 👍

lib/cloud_controller/diego/desire_app_handler.rb
acosta11: Good question, I'm not familiar with task execution error cases and only found this particular layer with some guidance from Tim. Is it a problem for a new execution of a task to fail after the cutover to blocking the root user? I assume that a previously running task isn't an availability concern the same way an already deployed running app would be.

spec/unit/models/runtime/droplet_model_spec.rb
acosta11: Good catch, I think I messed up some merge conflict resolution on the rebase against main. Will clean these up. 👍

lib/cloud_controller/diego/desire_app_handler.rb
acosta11: Poked Tim and he pointed me to https://github.com/cloudfoundry/cloud_controller_ng/blob/main/lib/cloud_controller/diego/tasks_sync.rb . Sounds like Tasks are run at most once, so we should only need to handle new Tasks, which necessarily go through the model layer validation.

acosta11:
Also added dev note to PR description about potentially unexpected behavior of the Process/Task policy not applying to the users set in Dockerfile specifically. Running CATs on env now to see if anything weird pops up.

acosta11:
CATs appears to be working consistently for the standard app and docker suites. I'm mostly seeing what looks like flakes with logging tests.

```bash
$ cat cats-config.json
{
  "api": "<api>",
  "apps_domain": "<apps>",
  "admin_user": "admin",
  "admin_password": "<cred>",
  "skip_ssl_validation": true,
  "use_http": true,
  "include_apps": true,
  "include_capi_experimental":true,
  "include_detect": false,
  "include_security_groups": false,
  "include_services": false,
  "include_v3": true,
  "include_tasks": true,
  "include_backend_compatibility": false,
  "include_capi_no_bridge": false,
  "include_container_networking": false,
  "include_credhub" : false,
  "include_docker": true,
  "include_internet_dependent": false,
  "include_isolation_segments": false,
  "include_persistent_app": false,
  "include_private_docker_registry": false,
  "include_privileged_container_support": false,
  "include_route_services": false,
  "include_routing": false,
  "include_routing_isolation_segments": false,
  "include_service_discovery": false,
  "include_service_instance_sharing": false,
  "include_ssh": false,
  "include_sso": false,
  "include_zipkin": false,
  "java_buildpack_name": "java_buildpack_offline"
}

CONFIG=cats-config.json ./bin/test --procs=4
...
Summarizing 2 Failures:
  [FAIL] [app_syslog_tcp] Syslog Drain over TCP Syslog drains [It] forwards app messages to registered syslog drains via mtls
  /workspace/cf-acceptance-tests/app_syslog_tcp/syslog_drain.go:152
  [FAIL] [app_syslog_tcp] Syslog Drain over TCP Syslog drains [It] forwards app messages to registered syslog drains
  /workspace/cf-acceptance-tests/app_syslog_tcp/syslog_drain.go:113

Ran 80 of 275 Specs in 1513.412 seconds
FAIL! -- 78 Passed | 2 Failed | 2 Pending | 193 Skipped


Ginkgo ran 1 suite in 25m15.173901372s
```

Seems like I just need what looks like some flakey errors in the remaining mysql unit test here to pass.

lib/cloud_controller/diego/desire_app_handler.rb
```diff
@@ -12,6 +12,8 @@ def create_or_update_app(process, client)
               if e.name == 'RunnerError' && e.message['the requested resource already exists']
                 existing_lrp = client.get_app(process)
                 client.update_app(process, existing_lrp)
+              elsif e.name == 'UnprocessableEntity'
+                raise
```
Gerg, line 16: I wonder if we should just be re-raising all errors here. I'm checking with @sethboyles out-of-band about https://github.com/cloudfoundry/cloud_controller_ng/commit/133486e37c8a66883f8cad0da6fcaeefad0feda3.

spec/unit/lib/cloud_controller/diego/desire_app_handler_spec.rb
```diff
@@ -51,6 +51,18 @@ module Diego
             expect(client).to have_received(:get_app).exactly(2).times
           end
         end
+
+        context 'when root user is not permitted' do
```
Gerg, line 55: Nit: This context is more specific than the implemented logic; any `UnprocessableEntity` will be re-raised.

PR #4521, opened 2025-08-18
Move service binding uniquness validation into ccng

johha:
In the future, app developers will be able to create multiple service bindings for the same service instance and app.

This change is a preparation step:
- It modifies the service binding validation logic to no longer rely on database unique constraints.
- Instead, uniqueness is enforced within ccng.

A new config parameter, `max_service_credential_bindings_per_app_service_instance`, will eventually allow operators to set the maximum number of bindings. For now, this value is hard-coded to 1.

⚠️ **Important**:
This change must be deployed before the subsequent release/PR that removes the DB unique constraints. Otherwise, users might create multiple service bindings even though it is not yet allowed.

Further details are described in [RFC-0040](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0040-service-binding-rotation.md)

### DIFF
--- a/app/actions/service_credential_binding_app_create.rb
+++ b/app/actions/service_credential_binding_app_create.rb
@@ -30,7 +30,10 @@ module VCAP::CloudController
 
         ServiceBinding.new.tap do |new_binding|
           ServiceBinding.db.transaction do
-            validate_app_guid_name_uniqueness!(app.guid, message.name, service_instance.guid) # if max_bindings_per_app_service_instance == 1
+            # Lock the service instance to prevent multiple bindings being created when not allowed
+            service_instance.lock! if max_bindings_per_app_service_instance == 1
+
+            validate_app_guid_name_uniqueness!(app.guid, message.name, service_instance.guid)
 
             num_valid_bindings = 0
             existing_bindings = ServiceBinding.where(service_instance:, app:)
--- a/spec/unit/actions/service_credential_binding_app_create_spec.rb
+++ b/spec/unit/actions/service_credential_binding_app_create_spec.rb
@@ -197,11 +197,6 @@ module VCAP::CloudController
                 action.precursor(service_instance, app:, message:)
               end.not_to raise_error
 
-              # Mock the validation for the second request to simulate the race condition and trigger a
-              # unique constraint violation on service_instance_guid + app_guid
-              allow_any_instance_of(ServiceBinding).to receive(:validate).and_return(true)
-              allow(ServiceBinding).to receive(:first).with(service_instance:, app:).and_return(nil)
-
               # Second request, should fail with correct error
               expect do
                 action.precursor(service_instance, app: app, message: message2)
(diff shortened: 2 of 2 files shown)


app/actions/service_credential_binding_app_create.rb
```diff
@@ -19,37 +19,44 @@ def initialize(user_audit_info, audit_hash, manifest_triggered: false)
...
           credentials: {}
         }
 
-        ServiceBinding.new.tap do |b|
+        ServiceBinding.new.tap do |new_binding|
           ServiceBinding.db.transaction do
-            if binding
-              binding.destroy
-              VCAP::Services::ServiceBrokers::V2::OrphanMitigator.new.cleanup_failed_bind(binding)
+            validate_app_guid_name_uniqueness!(app.guid, message.name, service_instance.guid) # if max_bindings_per_app_service_instance == 1
```
kathap: Seems to be a leftover the comment here

app/actions/service_credential_binding_app_create.rb
```diff
@@ -66,11 +73,25 @@ def validate_service_instance!(app, service_instance, volume_mount_services_enab
...
+
+      def validate_app_guid_name_uniqueness!(target_app_guid, desired_binding_name, target_service_instance_guid)
+        return if desired_binding_name.nil?
+
+        dataset = ServiceBinding.where(app_guid: target_app_guid, name: desired_binding_name)
 
-        already_bound! if binding.create_succeeded? || binding.create_in_progress?
-        incomplete_deletion! if binding.delete_failed? || binding.delete_in_progress?
+        name_uniqueness_violation!(desired_binding_name) if max_bindings_per_app_service_instance == 1 && dataset.first
+        name_uniqueness_violation!(desired_binding_name) if dataset.exclude(service_instance_guid: target_service_instance_guid).first
```
kathap, line 97: Can we distinguish here between name uniqueness violation for same service instance and name uniqueness violation among several service instances? Maybe also include the service instance guid in the error message for case 2.

app/actions/service_credential_binding_app_create.rb
```diff
@@ -19,37 +19,44 @@ def initialize(user_audit_info, audit_hash, manifest_triggered: false)
...
+
+              validate_binding!(binding, desired_binding_name: message.name)
+              num_valid_bindings += 1
             end
-            b.save_with_attributes_and_new_operation(
-              binding_details,
+
+            validate_number_of_bindings!(num_valid_bindings)
+
+            new_binding.save_with_attributes_and_new_operation(
```
kathap, line 55: Is it proofed that there can not be 2 same inserts in parallel that succeed without the unique constraint in the db?

app/actions/service_credential_binding_app_create.rb
johha: It's easier to improve the error message after the unique constraint is removed otherwise it cannot be tested in unit tests right now. I'll add it to the follow up.

app/actions/service_credential_binding_app_create.rb
johha: Fixed with 2a1f574b1f20612f397e79cd141262935b078a23

kathap (approved):

PR #4522, opened 2025-08-18
Ignore ResourceExists errors from bbs_apps_client

Gerg:
- This is a re-implementation of https://github.com/cloudfoundry/cloud_controller_ng/commit/133486e37c8a66883f8cad0da6fcaeefad0feda3
- This reverts commits:
    - https://github.com/cloudfoundry/cloud_controller_ng/commit/e321fa8f014d57296abbec4572cb5525f4d96601
    - https://github.com/cloudfoundry/cloud_controller_ng/commit/fc96bb6e245e2b0e1e7ec62ed23c139bf1697452
    - https://github.com/cloudfoundry/cloud_controller_ng/commit/133486e37c8a66883f8cad0da6fcaeefad0feda3  
- This error can be raised by BBS when there is a race condition between
  the CC syncer and CC API when creating an LRP for a process
- Since we want to desire a process anyway, it already existing (with
  the correct version) is likely not an error case. Thus, we can behave
  idempotently.
- Improvements from previous implementation:
    - No longer ignores other types of APIErrors in
      desire_app_handler.rb
    - No longer coupled to the exact error message coming from BBS

### DIFF
--- a/lib/cloud_controller/diego/bbs_apps_client.rb
+++ b/lib/cloud_controller/diego/bbs_apps_client.rb
@@ -13,7 +13,7 @@ module VCAP::CloudController
           lrp = AppRecipeBuilder.new(config: @config, process: process).build_app_lrp
 
           response = @client.desire_lrp(lrp)
-          logger.info('desire.app.response', process_guid: lrp.process_guid, error: response.error)
+          logger.info('desire.app.response', process_guid: lrp.process_guid, error: response.error&.to_s)
 
           runner_invalid_request!(response.error.message) if response.error&.type == ::Diego::Bbs::ErrorTypes::InvalidRequest
           return response if [::Diego::Bbs::ErrorTypes::ResourceConflict, ::Diego::Bbs::ErrorTypes::ResourceExists].include?(response.error&.type)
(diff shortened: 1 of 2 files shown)


lib/cloud_controller/diego/desire_app_handler.rb
```diff
@@ -6,14 +6,7 @@ def create_or_update_app(process, client)
           if (existing_lrp = client.get_app(process))
             client.update_app(process, existing_lrp)
           else
-            begin
-              client.desire_app(process)
-            rescue CloudController::Errors::ApiError => e # catch race condition if Diego Process Sync creates an LRP in the meantime
-              if e.name == 'RunnerError' && e.message['the requested resource already exists']
-                existing_lrp = client.get_app(process)
-                client.update_app(process, existing_lrp)
```
Gerg, line 14: We do lose out on this update. I think the syncer will get to it anyway (or already got to it, since this is a race condition). I'll think about it some more.

Gerg: Yea, looking at `client.update_app`, it's getting the LRP based on the CC process's `guid` and `version`. If the `version`s match between the LRP and the CC process, then no update is needed (that's the purpose of `version`).

Thus, I don't think this additional update is needed.

PR #4523, opened 2025-08-19
feat: Platform Engineer can configure CF feature flags via bosh deplo…

hsinn0:
…yment

- Added the code to override default feature flags upon initialization of new deployment.
- TNZ-48301

  The objective: Platform engineer can override feature flags in the capi release bosh deployment manifest.

  Manifest example:
```
  - consumes: {}
    name: cloud_controller_ng
    properties:
      <snip>
      cc:
      <snip>
        feature_flag_overrides:
          diego_docker: true
          route_sharing: true
```

  See the above.
  
  https://github.com/cloudfoundry/capi-release/pull/567

### DIFF
--- a/spec/unit/models/runtime/feature_flag_spec.rb
+++ b/spec/unit/models/runtime/feature_flag_spec.rb
@@ -237,10 +237,6 @@ module VCAP::CloudController
           FeatureFlag.override_default_flags({ key => config_value })
         end
 
-        after do
-          FeatureFlag.find(name: key.to_s)&.destroy
-        end
-
         context 'and the value is not changed by admin' do
           it 'returns the config-set value' do
             expect(FeatureFlag.enabled?(key)).to be config_value
@@ -259,10 +255,6 @@ module VCAP::CloudController
             admin_override
           end
 
-          after do
-            admin_override.destroy
-          end
-
           it 'returns the admin-set value' do
             expect(FeatureFlag.enabled?(key)).to be admin_value
           end
@@ -274,10 +266,6 @@ module VCAP::CloudController
 
         before do
           FeatureFlag.make(name: key.to_s, enabled: admin_value)
-        end
-
-        after do
-          FeatureFlag.find(name: key.to_s)&.destroy
         end
 
         it 'overwrites the existing admin-set value' do
@@ -327,12 +315,6 @@ module VCAP::CloudController
           expect do
             FeatureFlag.override_default_flags({ diego_docker: !default_diego_docker_value, user_org_creation: !default_user_org_creation_value })
           end.not_to raise_error
-        end
-
-        after do
-          %i[diego_docker user_org_creation].each do |flag_key|
-            FeatureFlag.find(name: flag_key.to_s)&.destroy
-          end
         end
 
         it 'updates values' do


spec/unit/models/runtime/feature_flag_spec.rb
```diff
@@ -225,5 +225,126 @@ module VCAP::CloudController
...
+
+        it 'overwrites the existing admin-set value' do
+          expect(FeatureFlag.enabled?(key)).to be admin_value
+          FeatureFlag.override_default_flags({ key => !admin_value })
+          expect(FeatureFlag.enabled?(key)).to be !admin_value
+        end
+      end
+    end
+
+    describe '.update_default_flags' do
```
Gerg: Nit: This is the incorrect method name.

spec/unit/models/runtime/feature_flag_spec.rb
```diff
@@ -225,5 +225,126 @@ module VCAP::CloudController
         end
       end
     end
+
+    describe 'default flag override in config' do
```
Gerg, line 229: What do you think about nesting the `default flag override in config` under the `.override_default_flags` describe block (or vice-versa)? They appear to be testing the same method.

spec/unit/models/runtime/feature_flag_spec.rb
hsinn0: Fix in the new commit.

spec/unit/models/runtime/feature_flag_spec.rb
```diff
@@ -225,5 +225,126 @@ module VCAP::CloudController
...
+        end
+      end
+    end
+
+    describe '.update_default_flags' do
+      context 'with invalid flags' do
+        it 'raises an error for the one and only invalid name' do
+          expect do
+            FeatureFlag.override_default_flags({ an_invalid_name: true })
+          end.to raise_error(/Invalid feature flag name\(s\)/)
```
Gerg: Could we make stronger test assertions here (and for similar tests)? The error message is deterministic, right?

spec/unit/models/runtime/feature_flag_spec.rb
```diff
@@ -225,5 +225,126 @@ module VCAP::CloudController
...
+
+      context 'when there was no previously set conflicting value' do
+        let(:config_value) { !default_value }
+
+        before do
+          FeatureFlag.override_default_flags({ key => config_value })
+        end
+
+        after do
+          FeatureFlag.find(name: key.to_s)&.destroy
```
Gerg: Is this cleanup required? The tests should be running with isolated databases: https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/support/database_isolation.rb

Same for other tests.

app/models/runtime/feature_flag.rb
```diff
@@ -86,6 +86,23 @@ def self.admin_read_only?
       VCAP::CloudController::SecurityContext.admin_read_only?
     end
 
+    def self.override_default_flags(feature_flag_overrides)
+      raise "Invalid feature flag name(s): #{feature_flag_overrides}" unless feature_flag_overrides.keys.to_set <= DEFAULT_FLAGS.keys.to_set
```
Gerg: I think we can provide a more helpful error message here.  Maybe something like:
```rb
p "Invalid feature flag name(s): #{excess_keys.to_a}" if (excess_keys = {foo: false, bar: true, app_bits_upload: true, task_creation: false}.keys.to_set - FeatureFlag::DEFAULT_FLAGS.keys.to_set).any?
# "Invalid feature flag name(s): [:foo, :bar]"

p "Invalid feature flag name(s): #{excess_keys.to_a}" if (excess_keys = {app_bits_upload: true, task_creation: false}.keys.to_set - FeatureFlag::DEFAULT_FLAGS.keys.to_set).any?
# nil
```

app/models/runtime/feature_flag.rb
```diff
@@ -86,6 +86,23 @@ def self.admin_read_only?
...
+      raise "Invalid feature flag value(s): #{feature_flag_overrides}" unless feature_flag_overrides.values.all? { |item| item.is_a?(TrueClass) || item.is_a?(FalseClass) }
```
Gerg: Similar to above. Maybe an error message like: 

```rb
p "Invalid feature flag value(s): #{invalid_values}" if (invalid_values = {app_bits_upload: 'potato', task_creation: false, route_creation: 42 }.select { |k,v| !v.in? [true,false] }).any?
# "Invalid feature flag value(s): {:app_bits_upload=>\"potato\", :route_creation=>42}"
```

Gerg (changes requested):
Overall, this PR looks really good. 

I left a few comments with code style and error message suggestions. Happy to chat, if you want to go into more detail on any of them.

spec/unit/models/runtime/feature_flag_spec.rb
hsinn0: The intention of the former is to test the flag override behavior while the later is to test the method itself (with valid and invalid parameters). So I think that separating them makes sense.

spec/unit/models/runtime/feature_flag_spec.rb
hsinn0: The cleanup was necessary at least per my observation. I suppose that it is because the examples are under the same description or context?

spec/unit/models/runtime/feature_flag_spec.rb
hsinn0: Changed as suggested in the new commit.

app/models/runtime/feature_flag.rb
hsinn0: Changed as suggested in the new commit.

app/models/runtime/feature_flag.rb
hsinn0: Changed as suggested in the new commit.

spec/unit/models/runtime/feature_flag_spec.rb
Gerg: I tried removing them, and the tests passed locally for me. I opened a draft PR here to double check: https://github.com/cloudfoundry/cloud_controller_ng/pull/4536

hsinn0: If I remember correctly, without the clean-up, the result was inconsistent when I ran the whole spec file, depending on the randomized the test case execution order.

Gerg: I've been running 
```sh
while true; do bundle exec rspec spec/unit/models/runtime/feature_flag_spec.rb; done
```
for a while, and it has been consistently passing.

I still haven't seen evidence that these are needed. Let me know if you have reproduction steps for the issues you were seeing. Otherwise, I think we should remove them.

hsinn0: I cannot reproduce the inconsistent test result issue without the the clean-up code anymore. I will remove those.

hsinn0: Pushed new commit to remove the test clean-up code.

PR #4524, opened 2025-08-20
Better Error Response

serdarozerr:
In the event that a service-related job (such as creating/updating/deleting a service instance, or creating/deleting a service binding) fails when the service broker responds with 'failed' during the last operation state poll, then a dedicated exception is raised (LastOperationFailedState) which is not explicitly handled by a higher level which resulting in it being converted into an "UnknownError" by the PollableJobWrapper. With this PR we converted these error types to APIError with title of  "CF-UnableToPerform"


kathap (approved):

kathap (dismissed):
Just a hint, the tests e.g. for service_binding_create are passing with and without your changes, maybe they should be adapted?

PR #4525, opened 2025-08-20
Loosen file name validation for kubernetes service bindings

tomkennedy513:

The ServiceBinding spec has been updated to explicitly mention that underscores are valid in credential keys (https://github.com/servicebinding/spec/blob/main/README.md#workload-projection). Updating the implementation to match the change in teh spec

Many service bindings contain underscores in the credential keys. Previously these would return an error even though the spec only stated that credential keys should match the previous regex. The spec was updated to better clarify what should be allowed in credential keys

### DIFF
--- a/lib/cloud_controller/diego/service_binding_files_builder.rb
+++ b/lib/cloud_controller/diego/service_binding_files_builder.rb
@@ -37,7 +37,7 @@ module VCAP::CloudController
         @service_bindings.select(&:create_succeeded?).each do |service_binding|
           sb_hash = ServiceBindingPresenter.new(service_binding, include_instance: true).to_hash
           name = sb_hash[:name]
-          raise IncompatibleBindings.new("Invalid binding name: '#{name}'. Name must match #{binding_naming_convention.inspect}") unless valid_name?(name)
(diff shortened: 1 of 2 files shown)


philippthun:
@tomkennedy513 @sethboyles @Gerg @stephanme - As underscores were not valid before, we are transforming several `VCAP_SERVICES` attributes (e.g. binding_name -> binding-name): https://github.com/cloudfoundry/cloud_controller_ng/blob/a90419c5dcafce9cb209f124e789d255087c5c30/lib/cloud_controller/diego/service_binding_files_builder.rb#L110C11-L112

It would be nice to remove this transformation, although it's an incompatible change. But this feature is marked as experimental, so the question is: shall we dare?

tomkennedy513:
> @tomkennedy513 : we discussed changing the mapping of the `VCAP_SERVICES` attributes (currently: underscore to dash) in the [ARI WG meeting 26.8.25](https://docs.google.com/document/d/1wP21mVwaoZw45eU2YoDJg6KbvYJuD5ATrrhSY6_yHmM/edit?tab=t.0) and we think it would be good to use the original attribute names. Since the app feature `service-binding-k8s` is still documented as experimental, we should dare the change (now or never...).
> 
> A PR would be very appreciated.

@stephanme Here is the pr https://github.com/cloudfoundry/cloud_controller_ng/pull/4542

PR #4581, opened 2025-09-26
Read storage-cli config JSON from capi (AzureRM)

kathap:
Stop building storage-cli JSON inside CC; instead consume files rendered by capi job templates. StorageCliClient now selects the JSON file by `resource_type`.

Adopt storage_cli_client to read json configs rendered by capi. Detect correct json file by resource_type.
Switch responsibility of JSON generation to capi-release so CC only reads well-formed per-resource configs. This removes duplication of storage cli config file creation.
https://github.com/cloudfoundry/capi-release/pull/580

### DIFF
--- a/lib/cloud_controller/blobstore/storage_cli/azure_storage_cli_client.rb
+++ b/lib/cloud_controller/blobstore/storage_cli/azure_storage_cli_client.rb
@@ -5,7 +5,7 @@ module CloudController
         ENV['AZURE_STORAGE_CLI_PATH'] || '/var/vcap/packages/azure-storage-cli/bin/azure-storage-cli'
       end
 
-      CloudController::Blobstore::StorageCliClient.register('azure', AzureStorageCliClient)
+      CloudController::Blobstore::StorageCliClient.register('AzureRM', AzureStorageCliClient)
     end
   end
 end
--- a/spec/unit/lib/cloud_controller/blobstore/storage_cli/azure_storage_cli_client_spec.rb
+++ b/spec/unit/lib/cloud_controller/blobstore/storage_cli/azure_storage_cli_client_spec.rb
@@ -10,7 +10,7 @@ module CloudController
     RSpec.describe AzureStorageCliClient do
       let!(:tmp_cfg) do
         f = Tempfile.new(['storage_cli_config', '.json'])
-        f.write({ provider: 'azure',
+        f.write({ provider: 'AzureRM',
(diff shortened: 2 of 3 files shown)


lib/cloud_controller/blobstore/storage_cli/storage_cli_client.rb
```diff
@@ -16,29 +16,108 @@ class << self
...
+            case resource_type.to_s
+            when 'droplets', 'buildpack_cache' then :storage_cli_config_file_droplets
+            when 'buildpacks'                  then :storage_cli_config_file_buildpacks
+            when 'packages'                    then :storage_cli_config_file_packages
+            when 'resource_pool'               then :storage_cli_config_file_resource_pool
+            else
+              raise BlobstoreError.new("Unknown resource_type: #{resource_type}")
+            end
+
+          path = VCAP::CloudController::Config.config.get(key)
```
jochenehret: What if `config.get(key)` returns nil? Maybe include `key` in the error message?

lib/cloud_controller/blobstore/storage_cli/storage_cli_client.rb
```diff
@@ -16,29 +16,108 @@ class << self
...
+            json = Oj.load(File.read(path))
+          rescue StandardError => e
+            raise BlobstoreError.new("Failed to parse storage-cli config JSON at #{path}: #{e.message}")
+          end
+
+          validate_required_keys!(json, path)
+          json
+        end
+
+        def config_path_for!(resource_type)
```
jochenehret: Why an exclamation mark in the method name? It doesn't modify the passed object. Same for the other methods in this class.

lib/cloud_controller/blobstore/storage_cli/storage_cli_client.rb
kathap: If resource_type does not match any of the specified cases line 50-54, the else branch will raise a BlobstoreError before key is assigned, so execution will not reach line 59 with a nil value for key.

lib/cloud_controller/blobstore/storage_cli/storage_cli_client.rb
kathap: In cloud controller validation/check functions that raise an error end with an exclamation mark vs validation/check function that raise not an error do not end with an exclamation mark..

jochenehret (approved):

PR #4595, opened 2025-10-07
Some Rubocop fixes

moleske:

fix the easy rubocop issues

less output when running rubocop

### DIFF
--- a/app/access/organization_access.rb
+++ b/app/access/organization_access.rb
@@ -88,23 +88,23 @@ module VCAP::CloudController
     end
 
     def validate_remove_billing_manager_by_guid!(org)
-      return if org.billing_managers_dataset.many?
+      return if org.billing_managers_dataset.count > 1
 
       raise CloudController::Errors::ApiError.new_from_details('LastBillingManagerInOrg')
     end
 
     def validate_remove_manager_by_guid!(org)
-      return if org.managers_dataset.many?
+      return if org.managers_dataset.count > 1
 
       raise CloudController::Errors::ApiError.new_from_details('LastManagerInOrg')
     end
 
     def validate_remove_user_by_guid!(org, user_guid)
-      raise CloudController::Errors::ApiError.new_from_details('LastManagerInOrg') if org.managers_dataset.one? && org.managers.first.guid == user_guid
+      raise CloudController::Errors::ApiError.new_from_details('LastManagerInOrg') if org.managers_dataset.count == 1 && org.managers.first.guid == user_guid
 
-      raise CloudController::Errors::ApiError.new_from_details('LastBillingManagerInOrg') if org.billing_managers_dataset.one? && org.billing_managers.first.guid == user_guid
+      raise CloudController::Errors::ApiError.new_from_details('LastBillingManagerInOrg') if org.billing_managers_dataset.count == 1 && org.billing_managers.first.guid == user_guid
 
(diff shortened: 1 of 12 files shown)


lib/cloud_controller/seeds.rb
```diff
@@ -124,7 +124,7 @@ def find_routing_guid(domain)
       end
 
       def create_seed_security_groups(config)
-        return unless config.get(:security_group_definitions) && SecurityGroup.count == 0
```
philippthun, line 127: Rubocop gave a bad advice here. On a model/dataset we should use `empty?` as this issues an optimized SQL query. The `none?` function is implemented in `Enumerable`, thus all entries are fetched to check if there is any. `count` and `empty?` are `Sequel` functions instead:
```
.count == 0  =>  SELECT count(*) AS "count" FROM "security_groups" LIMIT 1

.none?       =>  SELECT * FROM "security_groups"

.empty?      =>  SELECT 1 AS\"one" FROM "security_groups" LIMIT 1
```

The other *good* function besides `empty?` is `any?` as we use the [any_not_empty](https://sequel.jeremyevans.net/rdoc-plugins/files/lib/sequel/extensions/any_not_empty_rb.html) extension. All other `Enumerable` functions should be used with care on a model/dataset.

app/jobs/runtime/failed_jobs_cleanup.rb
```diff
@@ -26,7 +26,7 @@ def perform
           orphaned_delayed_jobs = Delayed::Job.
                                   where(Sequel.lit("run_at < CURRENT_TIMESTAMP - INTERVAL '?' DAY", force_delete_after))
 
-          unless orphaned_delayed_jobs.count.zero?
+          unless orphaned_delayed_jobs.none?
```
philippthun: Can be changed to `if orphaned_delayed_jobs.any?`.

spec/unit/repositories/app_usage_event_repository_spec.rb
```diff
@@ -542,7 +542,7 @@ module Repositories
 
             started_app_count = ProcessModel.where(state: 'STARTED').count
 
-            expect(AppUsageEvent.count > 1).to be true
```
philippthun, line 545: `many?` also issues a `SELECT * FROM ...` - so here `count` seems to be the better option.

app/access/organization_access.rb
```diff
@@ -88,23 +88,23 @@ def validate!(org, params)
     end
 
     def validate_remove_billing_manager_by_guid!(org)
-      return if org.billing_managers_dataset.count > 1
```
philippthun, line 91: `many?` and `one?` also issue a `SELECT * FROM ...` - so here count seems to be the better option.

moleske:
> Actually I like using `many?` and `one?` and `none?` as they are more verbose end expressive. I'm wondering why Sequel does not override them... Maybe we should do this? Then the code would look cleaner and we would not have to care about accidentally using the wrong function...

Would you prefer I remove the commit for this `Style/CollectionQuerying` rubocop from this pr then?  Then we can spend some time figuring out why Sequel doesn't override and decide if we want to override.  Since the other rubocops seem fine might be worth doing that.  Otherwise I can `# rubocop:disable Style/CollectionQuerying` on the lines you've marked in the review for count.

Also I definitely didn't think about what queries would actually be so thanks for calling that out

PR #4609, opened 2025-10-15
Tighten the semantics of POST /v3/service_plans/:guid/visibility

### LINKED ISSUE #4606: Restrict POST /v3/service_plans/:guid/visibility to only support appending orgs (and not changing visibility type)
### Summary
The current implementation of [POST /v3/service_plans/:guid/visibility](https://v3-apidocs.cloudfoundry.org/version/3.203.0/index.html#apply-a-service-plan-visibility) is somewhat inconsistent with REST semantics and allows operations beyond simply appending organizations. Specifically, it can also change the visibility type (e.g., to _public_ or _admin_), which overlaps with the semantics of [PATCH /v3/service_plans/:guid/visibility](https://v3-apidocs.cloudfoundry.org/version/3.203.0/index.html#update-a-service-plan-visibility).

### Problem
From a REST perspective:
- **POST** should be used to add new resources or associations (in this case, adding orgs to an existing visibility list).
- **PATCH** should be used to modify existing state (e.g., changing the visibility type to _public_ or _admin_).

Currently, **POST** behaves more like a **PATCH**, as it allows changing the visibility type, not just appending orgs. This blurs the intended semantics and increases the risk of accidental type changes when clients are only trying to add organizations.

### Proposed Change
I propose tightening the semantics of `POST /v3/service_plans/:guid/visibility`:
- **Allow POST only when type = organization.**
- **Ensure the target plan already has visibility type organization.** If the plan’s visibility type is not organization, the request should fail with an appropriate error (e.g., `422 Unprocessable Entity`).

This ensures that:
- **POST** is used solely for appending organizations to organization-scoped plans.
- **PATCH** remains the correct method for changing the visibility type.

### Compatibility / Migration
This change could break existing clients that currently use **POST** to modify visibility type. However, those clients can simply switch to using **PATCH**, which already supports the same behavior. This aligns both endpoints with clearer REST semantics going forward.

philippthun:
- Allow `POST` only when `type = organization`. Otherwise fail with `422 Unprocessable Entity`.
- Ensure the target plan already has **visibility type** `organization`. If the plan’s visibility type is not `organization`, the request fails with `422 Unprocessable Entity`.

Fixes #4606 

### DIFF
--- a/app/actions/v3/service_plan_visibility_update.rb
+++ b/app/actions/v3/service_plan_visibility_update.rb
@@ -8,16 +8,21 @@ module VCAP::CloudController
       end
 
       def update(service_plan, message, append_organizations: false)
-        type = message.type
+        requested_type = message.type
         requested_org_guids = message.organizations&.pluck(:guid) || []
 
-        unprocessable!("cannot update plans with visibility type 'space'") if space?(service_plan)
+        unprocessable!("cannot update plans with visibility type 'space'") if space?(service_plan.visibility_type)
+
+        if append_organizations # i.e. POST request
+          unprocessable!("type must be 'organization'") unless org?(requested_type)
+          unprocessable!("can only append organizations to plans with visibility type 'organization'") unless org?(service_plan.visibility_type)
+        end
 
         service_plan.db.transaction do
           service_plan.lock!
-          service_plan.public = public?(type)
+          service_plan.public = public?(requested_type)
           service_plan.save
-          if org?(type)
+          if org?(requested_type)
             update_service_plan_visibilities(service_plan, requested_org_guids, append_organizations)
           else
             service_plan.remove_all_service_plan_visibilities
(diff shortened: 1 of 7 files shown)


app/actions/v3/service_plan_visibility_update.rb
```diff
@@ -11,7 +11,12 @@ def update(service_plan, message, append_organizations: false)
         type = message.type
```
johha: maybe there is a better name for this?

app/actions/v3/service_plan_visibility_update.rb
```diff
@@ -69,10 +74,14 @@ def unprocessable!(message)
...
         service_plan.visibility_type == VCAP::CloudController::ServicePlanVisibilityTypes::SPACE
       end
 
+      def visibility_type_org?(service_plan)
+        service_plan.visibility_type == VCAP::CloudController::ServicePlanVisibilityTypes::ORGANIZATION
+      end
+
       def org?(type)
         type == VCAP::CloudController::ServicePlanVisibilityTypes::ORGANIZATION
       end
```
johha: Can we combine this into a single method and use the same pattern for `visibility_type_space`?

app/actions/v3/service_plan_visibility_update.rb
philippthun: Renamed it to `requested_type` (there is also `requested_org_guids`).

app/actions/v3/service_plan_visibility_update.rb
philippthun: Done. Now there are only two methods, `space?(type)` and `org?(type)`.

PR #4611, opened 2025-10-16
Make many-to-many additions idempotent on unique constraint violation

johha:
Adds an `ignored_unique_constraint_violation_errors` option to the many_to_many association in the `VcapRelations` Sequel plugin. When this option is provided with a list of index name patterns, any `Sequel::UniqueConstraintViolation` error that occurs during an `add_` operation on an association and matches one of the patterns will be caught. The operation is wrapped in a transaction with a savepoint, which is rolled back upon catching the specific error, effectively making the addition idempotent. This prevents race conditions where concurrent requests attempt to add the same association. This issue can be observed when a user makes two POST requests at the same time to create a new resource instance.

This new option is applied to the `staging_spaces relationship` in the `SecurityGroup` model to handle potential concurrent updates. Further models will updated in separate commits.

### DIFF
--- a/lib/sequel_plugins/vcap_relations.rb
+++ b/lib/sequel_plugins/vcap_relations.rb
@@ -65,10 +65,17 @@ module Sequel::Plugins::VcapRelations
         # sequel is not capable of merging adds to a many_to_many association
         # like it is for a one_to_many and nds up throwing a db exception,
         # so lets squash the add
-        if other.is_a?(Integer)
-          super(other) unless send(ids_attr).include? other
-        else
-          super(other) unless send(name).include? other
+        db.transaction(savepoint: true) do
+          if other.is_a?(Integer)
+            super(other) unless send(ids_attr).include? other
+          else
+            super(other) unless send(name).include? other
+          end
+        rescue Sequel::UniqueConstraintViolation => e
+          # ignore the error and rollback the inner transaction
+          raise Sequel::Rollback if opts[:ignored_unique_constraint_violation_errors]&.any? { |pattern| e.message.include?(pattern) }
+
+          raise e
         end
       end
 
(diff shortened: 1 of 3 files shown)


spec/unit/lib/sequel_plugins/vcap_relations_spec.rb
```diff
@@ -297,6 +297,63 @@ def define_model(name)
...
+          db_connection[:dogs_names].db.transaction do
+            expect { @d1.add_name(@n2) }.not_to raise_error
+            expect(db_connection[:dogs_names].count).to eq(2)
+            expect(@d1.names).to include(@n2)
+          end
+        end
+      end
+
+      context 'when the join table does not have a unique constraint' do
+        # This test proofs that without a unique constraint or combined primary key duplicate entries can be created
```
jochenehret: "proves"

lib/sequel_plugins/vcap_relations.rb
```diff
@@ -65,10 +65,17 @@ def many_to_many(name, opts={})
         # sequel is not capable of merging adds to a many_to_many association
         # like it is for a one_to_many and nds up throwing a db exception,
         # so lets squash the add
-        if other.is_a?(Integer)
-          super(other) unless send(ids_attr).include? other
-        else
-          super(other) unless send(name).include? other
+        db.transaction(savepoint: true) do
```
jochenehret, line 68: Does this have side-effects on other add_ methods? Performance impacts?

johha: This will impact every insert for a many to many relationship. However I don't see any alternative how we could make parallel inserts idempotent.

jochenehret (approved):

PR #4618, opened 2025-10-24
Cloud Foundry do not parse SB response and do not include custom metadata

atanas-attodorov-wq:
* **A short explanation of the proposed change**:

Tanzu Data service tiles (like Postgres) bundle a specific engine version (e.g., Postgres 16.6), but this version is not visible on the ServiceInstance object in Cloud Foundry. This makes it hard for users to know which version they are consuming.

The Open Service Broker API (OSBI) already has a ServiceInstanceMetadata field to store this type of information, Even if the on-demand-service-broker (ODB) propagate the metadata it is not consumed during parsing in CAPI.

This change makes the service version visible in Cloud Foundry by using the existing OSBI metadata field. This is a three-part process:

Service Adapter: The service-adapter (e.g., the Postgres tile's adapter) will be updated to add the engine version as a key-value pair in the tags section of the BOSH manifest it generates (e.g., postgres_version: "16.6").

On-Demand Service Broker (ODB): The ODB will be modified to read this new version tag from the manifest. The ODB will then take this value and populate the Metadata field in its ProvisionedServiceSpec and UpdateServiceSpec responses to Cloud Foundry.

Cloud Controller  will parse that metadata and merge with labels that comes from the Service Broker. The labels which comes from Service Broker will be shown only if the SI is managed.

1. **postgres-tile**  https://github.gwd.broadcom.net/TNZ/postgres-tile/pull/394
2. **on-demand-service-broke**r https://github.com/pivotal-cf/on-demand-service-broker/pull/439

* [Yes ] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [Yes] I have viewed, signed, and submitted the Contributor License Agreement

* [ Yes] I have made this pull request to the `main` branch

* [Yes ] I have run all the unit tests using `bundle exec rake`

* [ No] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)

### DIFF
--- a/app/presenters/v3/service_instance_presenter.rb
+++ b/app/presenters/v3/service_instance_presenter.rb
@@ -50,7 +50,7 @@ module VCAP::CloudController
               }
             },
             metadata: {
-              labels: hashified_labels(service_instance.labels),
+              labels: merged_labels,
               annotations: hashified_annotations(service_instance.annotations)
             },
             links: {
@@ -77,7 +77,7 @@ module VCAP::CloudController
         end
 
         def hash_additions_managed
-          {
+          hash = {
             type: 'managed',
             maintenance_info: maintenance_info,
             upgrade_available: upgrade_available,
@@ -101,6 +101,13 @@ module VCAP::CloudController
               }
             }
           }
+          
+          # Add broker metadata if present
+          if service_instance.broker_metadata.present?
+            hash[:broker_metadata] = service_instance.broker_metadata
+          end
+          
+          hash
         end
 
         def hash_additions_user_provided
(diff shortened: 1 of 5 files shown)


app/presenters/v3/service_instance_presenter.rb
```diff
@@ -145,6 +152,21 @@ def last_operation
...
+          
+          # Get broker metadata labels if they exist (only for managed service instances)
+          broker_labels = {}
+          if service_instance.is_a?(ManagedServiceInstance) && service_instance.broker_metadata.present?
+            broker_metadata_labels = service_instance.broker_metadata.dig('labels')
+            broker_labels = broker_metadata_labels if broker_metadata_labels.is_a?(Hash)
+          end
+          
+          # Merge broker labels into CF labels (broker labels take precedence for conflicts)
+          cf_labels.merge(broker_labels)
```
Benjamintf1, line 168: I'm not sure merging the broker labels in with the user defined labels is the correct solution here. 

While, I think, without passing the broker attributes to the user it remains terribly useful, i think conflating the osabi labels with user labels because they are specified with the same word might be the wrong choice.

atanas-attodorov-wq: I agree . Can you have a look at this discussion?
https://github.com/cloudfoundry/cloud_controller_ng/discussions/4633
In the project i am working ,this will not be needed anymore. I am closing this PR. If someone say it is feasible , i can re-open the PR and contribute.

PR #4632, opened 2025-10-30
Add unique constraint for sidecar process types

johha:
Adds a database-level unique constraint to the `sidecar_process_types` and `revision_sidecar_process_types` tables on the `(sidecar_guid, type)` and `(revision_sidecar_guid, type)` columns, respectively.

This prevents duplicate process type entries for a given sidecar, which could previously occur under concurrent requests.

The migration handles de-duplicating any existing records before applying the new unique index. Additionally, the corresponding models have been updated to gracefully handle the `UniqueConstraintViolation` and raise a user-friendly validation error. The previous application-level validates_unique check, which was susceptible to race conditions, has been replaced.

### DIFF
--- a/app/models/runtime/revision_sidecar_process_type_model.rb
+++ b/app/models/runtime/revision_sidecar_process_type_model.rb
@@ -6,11 +6,19 @@ module VCAP::CloudController
                 key: :revision_sidecar_guid,
                 without_guid_generation: true
 
+    def around_save
+      yield
+    rescue Sequel::UniqueConstraintViolation => e
+      raise e unless e.message.include?('revision_sidecar_process_types_revision_sidecar_guid_type_index')
+
+      errors.add(%i[revision_sidecar_guid type], "Sidecar is already associated with process type #{type}")
+      raise Sequel::ValidationFailed.new(self)
+    end
+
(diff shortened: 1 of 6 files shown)


app/models/runtime/revision_sidecar_process_type_model.rb
```diff
@@ -6,11 +6,19 @@ class RevisionSidecarProcessTypeModel < Sequel::Model(:revision_sidecar_process_
                 key: :revision_sidecar_guid,
                 without_guid_generation: true
 
+    def around_save
+      yield
+    rescue Sequel::UniqueConstraintViolation => e
+      raise e unless e.message.include?('revision_sidecar_process_types_revision_sidecar_guid_type_index')
+
+      errors.add(%i[revision_sidecar_guid type], "Sidecar is already associated with process type #{type}")
```
jochenehret, line 14: Would it make sense to print different error messages for "revision" sidecars? (I guess not...)

johha: It's the same error message as before (see deleted line 13). As the user might be unaware of revisions I'd rather keep it out

PR #4650, opened 2025-11-07
Add idempotent droplet creation

kathap:
Add idempotent droplet creation:
- if there exists already a droplet, take that one and ensure buildpack_lifecycle_data is attached (but don’t overwrite)
- if not create a new droplet

Add idempotent droplet creation. This will add one simple SQL query to check for already existing droplet:
```SELECT * FROM "droplets" WHERE "build_guid" = 'abcd-1234' LIMIT 1;```
Fix hanging staging on cc-uploader kill caused by 2 created droplets with same build_guid. The Problem was if cc-uploader would run into uplaod timeout or cc-iuploader would be killed hard, then diego sees its upload request failed, diego sends a new upload reuqest to a healthy VM and with that a new droplet was created for the same build, which led to staging saving error ```"diego.staging.buildpack.saving-staging-result-failed"```, ```"error":"undefined method `buildpacks' for nil:NilClass","build"``` because completion handler tried to save the result with the first droplet guid but that droplet had no longer buildpack_lifecycle_data assigned, those were now assigned to the 2. droplet. Build and droplets remained in state STAGING forever and staging finally failed because of timeout.

### DIFF
--- a/app/actions/droplet_create.rb
+++ b/app/actions/droplet_create.rb
@@ -50,6 +50,7 @@ module VCAP::CloudController
     end
 
     def find_or_create_buildpack_droplet(build)
+      droplet = nil
       DropletModel.db.transaction do
         existing = DropletModel.where(build_guid: build.guid).first
         if existing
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/actions/droplet_create.rb [hunk 1]
```diff
@@ -49,22 +49,33 @@ def create_docker_droplet(build)
...
     end
 
-    def create_buildpack_droplet(build)
-      droplet = droplet_from_build(build)
-
+    def find_or_create_buildpack_droplet(build)
       DropletModel.db.transaction do
+        existing = DropletModel.where(build_guid: build.guid).first
+        if existing
+          if build.cnb_lifecycle?
```
jochenehret, line 58: Should we add some logging here (for the existing==true case)?

jochenehret (approved):

jochenehret (approved):

PR #4653, opened 2025-11-10
Update to Ruby 3.2.9

jochenehret:
Update Ruby to 3.2.9. Automatic update via Dependabot is not supported (can only update Gem dependencies).

Fixes a number of CVEs.


moleske (approved):
Note - the auto bump job won't update this file so we'll need to manually bump on our own.  Not really a problem as this is mostly for local development, but I think the github actions use this ruby-version file when getting ruby set up

jochenehret:
Yes, the "bump-ruby-package" job will only update capi-release. We are looking for a solution to automatically update cloud_controller_ng.

PR #4910, opened 2026-03-05
RFC0055 Identity-Aware Routing

rkoster:
## Summary

Implements Part 2 (CF Identity & Authorization) of the [RFC for Identity-Aware Routing for GoRouter](https://github.com/cloudfoundry/community/pull/1438). This adds a new Route Policies API that allows developers to control which Cloud Foundry apps can access routes on identity-aware domains.

## What This Enables

Developers can now secure routes on identity-aware domains with platform-enforced access control:

```bash
# Create a route on an identity-aware domain
cf map-route my-backend apps.identity --hostname backend

# Allow specific apps to access it
cf add-route-policy apps.identity --source-app frontend-app --hostname backend
cf add-route-policy apps.identity --source-app monitoring-app --hostname backend

# List route policies
cf route-policies apps.identity --hostname backend
```

GoRouter enforces these rules automatically - no code changes needed in the backend app.

### 1. Route Policies API (RFC-Compliant)

New `/v3/route_policies` endpoints for managing route-level access control:

| Method | Path | Description |
|--------|------|-------------|
| `GET` | `/v3/route_policies` | List route policies with filtering |
| `GET` | `/v3/route_policies/:guid` | Get a single route policy |
| `POST` | `/v3/route_policies` | Create a route policy |
| `PATCH` | `/v3/route_policies/:guid` | Update metadata (labels/annotations only) |
| `DELETE` | `/v3/route_policies/:guid` | Delete a route policy |

**Query Parameters:**
- `guids`, `route_guids`, `space_guids` - Filter by resource GUIDs
- `sources` - Filter by exact source strings
- `source_guids` - Text-match against GUIDs in sources (for stale policy detection)
- `include` - Sideload related resources: `route`, `source`

**Example Request:**
```json
POST /v3/route_policies
{
  "source": "cf:app:d76446a1-f429-4444-8797-be2f78b75b08",
  "metadata": {
    "labels": { "team": "payments" },
    "annotations": { "description": "Allow frontend to call payments API" }
  },
  "relationships": {
    "route": { "data": { "guid": "route-guid" } }
  }
}
```

**Response includes:**
- Source string (e.g., `cf:app:<guid>`, `cf:space:<guid>`, `cf:org:<guid>`, `cf:any`)
- Read-only relationships extracted from source (app, space, organization)
- Metadata (labels and annotations)
- Created/updated timestamps

### 2. Domain-Level Enforcement Configuration

Domains can now be configured with enforcement settings at creation time (immutable):

```bash
cf create-shared-domain apps.identity --enforce-route-policies --scope org
```

**New Domain Fields:**
- `enforce_route_policies` (boolean) - When true, GoRouter enforces route policies for routes on this domain
- `route_policies_scope` (string: `any`, `org`, `space`) - Operator-level boundary for allowed callers

These fields are set via `POST /v3/domains` and are **not** present on `PATCH /v3/domains` (immutable by design).

### 3. Validation Rules

Route policy creation enforces all RFC-specified validations:

- ✅ Route's domain must have `enforce_route_policies: true`
- ✅ Route's domain cannot be internal (internal routes bypass GoRouter)
- ✅ `cf:any` cannot be combined with other sources on the same route
- ✅ Duplicate sources per route are rejected
- ✅ Source GUIDs are not validated at creation time (intentional - allows cross-org sharing)

### 4. Metadata Support

Route policies support labels and annotations for auditability:

```json
{
  "metadata": {
    "labels": { "team": "frontend", "env": "prod" },
    "annotations": { 
      "contact": "frontend-team@example.com",
      "slack": "#frontend-support"
    }
  }
}
```

Implemented via:
- `RoutePolicyLabelModel` and `RoutePolicyAnnotationModel`
- Database tables: `route_policy_labels`, `route_policy_annotations`
- Migrations applied

### 5. Read-Only Relationships

Route policies expose read-only relationships extracted from the source:

| Source | Populated Relationship | Others |
|----------|----------------------|--------|
| `cf:app:<guid>` | `app: { data: { guid } }` | `space`, `organization` → `null` |
| `cf:space:<guid>` | `space: { data: { guid } }` | `app`, `organization` → `null` |
| `cf:org:<guid>` | `organization: { data: { guid } }` | `app`, `space` → `null` |
| `cf:any` | All `null` | — |

When the referenced resource no longer exists, the relationship data is `null` but the source string is preserved (for metadata like contact info).

### 6. Diego Sync Integration

Route policies are automatically translated into GoRouter-compatible route options during Diego sync:

```ruby
# Cloud Controller converts route policies to:
route.options = {
  "route_policy_scope": "org",  # From domain.route_policies_scope
  "route_policy_sources": "cf:app:guid1,cf:space:guid2"  # Comma-joined sources
}
```

Process timestamps are updated on route policy create/delete to trigger `ProcessesSync`.

## Database Migrations

Two new migrations:

1. **`20260421074454_add_enforce_route_policies_to_domains.rb`** - Domain enforcement fields
2. **`20260421074455_create_route_policies.rb`** - Route policies table plus labels/annotations metadata tables

## Files Changed

**Controllers:**
- `app/controllers/v3/route_policies_controller.rb` - Route Policies API implementation

**Models:**
- `app/models/runtime/route_policy.rb` - Core model with Diego sync callbacks
- `app/models/runtime/route_policy_label_model.rb` - Labels metadata
- `app/models/runtime/route_policy_annotation_model.rb` - Annotations metadata
- `app/models/runtime/domain.rb` - Add enforcement fields
- `app/models/runtime/route.rb` - Add route_policies relationship
- `app/models.rb` - Explicit requires for new models (Rails autoloading disabled)

**Messages:**
- `app/messages/route_policy_create_message.rb` - Create validation
- `app/messages/route_policy_update_message.rb` - Update validation (metadata only)
- `app/messages/route_policies_list_message.rb` - List filtering
- `app/messages/domain_create_message.rb` - Add enforcement fields

**Presenters:**
- `app/presenters/v3/route_policy_presenter.rb` - API response format
- `app/presenters/v3/domain_presenter.rb` - Add enforcement fields

**Decorators:**
- `app/decorators/include_route_policy_source_decorator.rb` - Sideload source resources
- `app/decorators/include_route_policy_route_decorator.rb` - Sideload routes

**Actions:**
- `app/actions/domain_create.rb` - Set enforcement fields (immutable after creation)

**Diego Integration:**
- `lib/cloud_controller/diego/protocol/routing_info.rb` - Generate route options

**Tests:**
- `spec/request/route_policies_spec.rb` - Full API integration tests
- `spec/unit/models/runtime/route_policy_spec.rb` - Model unit tests
- `spec/unit/messages/route_policy*_spec.rb` - Message validation tests

## Related PRs

- **RFC**: https://github.com/cloudfoundry/community/pull/1438 (merged)
- **routing-release** (GoRouter): https://github.com/cloudfoundry/routing-release/pull/535 (merged)
- **cf-deployment** (ops files): https://github.com/cloudfoundry/cf-deployment/pull/1353
- **CLI** (`cf add-route-policy`, `cf route-policies`, etc.): TBD

## Testing

All RFC requirements have test coverage:

- ✅ Route Policies CRUD operations
- ✅ Query parameter filtering (guids, route_guids, space_guids, sources, source_guids)
- ✅ Include parameter (route, source)
- ✅ Metadata (labels/annotations)
- ✅ Read-only relationships
- ✅ Validation rules (enforce_route_policies, internal domains, cf:any exclusivity, duplicates)
- ✅ Permissions (Space Developer required)
- ✅ Diego sync on create/delete

Run tests:
```bash
bundle exec rspec spec/request/route_policies_spec.rb
bundle exec rspec spec/unit/models/runtime/route_policy_spec.rb
```

## Deployment Notes

1. Run migrations: `rake db:migrate`
2. Ensure routing-release is deployed with mTLS support
3. No feature flags required - gated by domain configuration (`enforce_route_policies`)

## Breaking Changes

None - this is additive functionality. Existing routes are unaffected.


rkoster:
Now that the RFC: https://github.com/cloudfoundry/community/pull/1438 has been merged let's start the review process.

cppforlife:
> The operator sets a scope boundary at domain creation time that limits which callers can reach routes on the domain.

i was chewing on domain scope configuration -- it is a bit confusing. i think we are also missing perhaps the most important one -- the thing between `any` and `org`, that really represents i want at least one route-policy present on the route (more than one is fine) and they all eval to true. (as far as i understand, `any` just means that client certs match up with what's configured at bosh layer)

with that in mind, i think we should tweak it to allow following values:
- `only-ca`
- `policy` (this would be default option)
- `same-org-with-policy` (imho we should not have this)
- `same-space-with-policy` (imho we should not have this)

with small stylistic tweak:
```
cf create-shared-domain internal-corp.com --enforce-route-policies --default-enforcement only-ca/policy/same-org...
```

i still question the need to have same-* enforcement as it will create challenge for app operators to understand who is actually allowed to access their routes. (because there is effectively a hidden route policy that gets applied)

which then leads me to another question of whether we should not have any of this configuration be driven through CF CLI since you have to configure client verification certs via bosh anyway.

cppforlife:
> The community convention for identity-aware routing is `*.apps.identity`

the reason why `apps.internal` makes sense is `internal` tld is reserved for private use. identity tld is not though, and so would propose we change to something like `verified-apps.internal` (not too opinionated about the name, just tld)

beyhan:
Hi @cppforlife,

I was wandering here about the supported use case described in https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0055-identity-aware-routing-for-gorouter.md?plain=1#L33. I'm not sure about external exposed routes and whether having `internal` in the domain will be confusing. At least at SAP we handle domain's with `internal` wording in it accessible only SAP internally.

rkoster:
> > The community convention for identity-aware routing is `*.apps.identity`
> 
> the reason why `apps.internal` makes sense is `internal` tld is reserved for private use. identity tld is not though, and so would propose we change to something like `verified-apps.internal` (not too opinionated about the name, just tld)

The shift toward the `*.apps.identity` pattern (away from the `*.apps.mtls.internal` pattern that was part of the initial RFC) was an intentional refinement to address the specific concerns raised by @ameowlia [here](https://github.com/cloudfoundfly/community/pull/1438#discussion-3018024807). Where she emphasized that using `.internal` is so closely associated with C2C infrastructure that it creates significant confusion and potential for misconfiguration. By adopting `*.apps.identity`, the aim was to establish a distinct naming convention that clearly differentiates this identity aware routing from standard C2C traffic.

rkoster:
> > The operator sets a scope boundary at domain creation time that limits which callers can reach routes on the domain.
> 
> i was chewing on domain scope configuration -- it is a bit confusing. i think we are also missing perhaps the most important one -- the thing between `any` and `org`, that really represents i want at least one route-policy present on the route (more than one is fine) and they all eval to true. (as far as i understand, `any` just means that client certs match up with what's configured at bosh layer)
> 
> with that in mind, i think we should tweak it to allow following values:
> 
> * `only-ca`
> * `policy` (this would be default option)
> * `same-org-with-policy` (imho we should not have this)
> * `same-space-with-policy` (imho we should not have this)
> 
> with small stylistic tweak:
> 
> ```
> cf create-shared-domain internal-corp.com --enforce-route-policies --default-enforcement only-ca/policy/same-org...
> ```
> 
> i still question the need to have same-* enforcement as it will create challenge for app operators to understand who is actually allowed to access their routes. (because there is effectively a hidden route policy that gets applied)
> 
> which then leads me to another question of whether we should not have any of this configuration be driven through CF CLI since you have to configure client verification certs via bosh anyway.

I understand the concern about potential confusion, but I believe the functionality you're suggesting "requiring at least one route policy to be present" is already covered by the RFC.

When `enforce_route_policies` is set to true, the RFC introduces a default-deny model: 'all requests to routes without route policies are denied by default.' This effectively implements the 'must have at least one route-policy' requirement you're describing without needing a new scope. You can see this defined in the [Layered Authorization section](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0055-identity-aware-routing-for-gorouter.md#layered-authorization):
> When a domain has enforce_route_policies: true, all requests to routes without route policies are denied by default.

The any, org, and space scopes are intended to define the operator-level boundary (the maximum allowable access), whereas the Route Policies themselves handle the developer-level granularity. This separation allows the operator to set the 'ceiling' while the developer defines the specific 'allowed' sources within that ceiling."

ameowlia:
> The shift toward the `*.apps.identity` pattern (away from the `*.apps.mtls.internal` pattern that was part of the initial RFC) was an intentional refinement to address the specific concerns raised by @ameowlia [here](https://github.com/cloudfoundfly/community/pull/1438#discussion-3018024807). Where she emphasized that using `.internal` is so closely associated with C2C infrastructure that it creates significant confusion and potential for misconfiguration. By adopting `*.apps.identity`, the aim was to establish a distinct naming convention that clearly differentiates this identity aware routing from standard C2C traffic.

In addition to the fact that "internal" is closely related with c2c and internal bosh dns stuff, I also think it is confusing using "internal" as the default domain for something that is not internal. They are exposed via gorouter. They could be accessed via other foundations.

cppforlife:
re: identity tld -- i dont have an opinion which tld to use tbh (internal or not). imho, we are standing on the shaky ground recommending to use a tld that's not officially assigned a purpose (this is similar to us picking *.apps.com). also note that we use .internal in other places as well, such as bosh-dns aliasing: eg sql-db.service.cf.internal. 

side note: i did not have as strong concerns about *.bosh at the time but that probably was a mistake (though bosh is much less likely to become an official tld unlike identity).

> I was wandering here about the supported use case described in https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0055-identity-aware-routing-for-gorouter.md?plain=1#L33

@beyhan this would apply for identity as well, since it's pretty uncommon to decide to "take ownership" of a tld not under your control.

cppforlife:
> When enforce_route_policies is set to true, the RFC introduces a default-deny model: 'all requests to routes without route policies are denied by default.' This effectively implements the 'must have at least one route-policy' requirement

sounds good, looks like i misunderstood what's proposed.

> i still question the need to have same-* enforcement as it will create challenge for app operators to understand who is actually allowed to access their routes. (because there is effectively a hidden route policy that gets applied)

this i believe is still a problem -- how would users of routes know that admin configured domain to be only usable within a space. tbh im still not understanding the idea of "scoping user to a space or org" at the domain level -- under what conditions is this useful? it appears that it would just encourage users to use different routes to communicate between spaces instead of using proposed more secure mechanism. it also btw is asymmetric from network policies which do not try to limit their use at the domain level.

beyhan:
> https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0055-identity-aware-routing-for-gorouter.md?plain=1#L33

I also don't have strong opinion which tld we should go with but we have listed some good arguments which we should consider.

cppforlife:
do we agree that there are two separate classes of domains?:

a) default domain that we would use in CF if customer doesn't "configure the system", this primarily makes sense to only be used internally within the single CF system since DNS outside wouldn't resolve to gorouters (DNS for this one is fully managed by us internally i would have to assume)

b) customer provided domain which would be configured like our usual wildcard/non-wildcard domains. -- i believe that what's specified in RFC as use cases (link above) would fit into this category rather than the first one.

beyhan:
> do we agree that there are two separate classes of domains?:

This is also my understanding.

> a) default domain that we would use in CF if customer doesn't "configure the system", this primarily makes sense to only be used internally within the single CF system since DNS outside wouldn't resolve to gorouters (DNS for this one is fully managed by us internally i would have to assume)

We should be able to define any number of domains belonging to this class.

>b) customer provided domain which would be configured like our usual wildcard/non-wildcard domains. -- i believe that what's specified in RFC as use cases (link above) would fit into this category rather than the first one.

My understanding here is that we don't have any requirements on the domain itself. Or do we?

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+  def index
+    message = RoutePoliciesListMessage.from_params(query_params)
+    invalid_param!(message.errors.full_messages) unless message.valid?
+
+    dataset = build_dataset(message)
+
+    decorators = []
+    decorators << IncludeRoutePolicySourceDecorator if IncludeRoutePolicySourceDecorator.match?(message.include)
+    decorators << IncludeRoutePolicyRouteDecorator if IncludeRoutePolicyRouteDecorator.match?(message.include)
+
```
jcvrabo, line 22: it includes source and route includes, but it's missing the app space and organization

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    )
+  end
+
+  def show
+    route_policy = VCAP::CloudController::RoutePolicy.find(guid: hashed_params[:guid])
+    resource_not_found!(:route_policy) unless route_policy
+
+    route = route_policy.route
+    resource_not_found!(:route_policy) unless route && permission_queryer.can_read_from_space?(route.space.id, route.space.organization_id)
+
```
jcvrabo, line 41: the show has no includes, I would argue that it should also have the same includes as the listing.... source, route, org, space and app

jcvrabo (commented):
Heads-up regarding my expertise of ruby and the cloud controller code, which is little to none, but I think I have raised some valid points

cppforlife:
> We should be able to define any number of domains belonging to this class.

there would only be one that is pre-loaded by the system by default. i dont think this "class" has any other domains (same as apps.internal).

> My understanding here is that we don't have any requirements on the domain itself. Or do we?

these domains are no different from any other private/shared domains CC holds, especially if they are getting accessed from the outside. (apps outside and inside the diego need to be able to resolve the domain, and especially for cross-regional communication you need to ensure that these domains are not overlapping and are resolvable globally)

beyhan:
>there would only be one that is pre-loaded by the system by default. i dont think this "class" has any other domains (same as apps.internal).

You are right. Actually, there can be just one TLD for instance-identity-certificate-based mTLS. With our initial version, where org/app/space configuration was also possible on the GoRouter side, that wouldn't have worked. However, with the current approach, where policies are attached to routes, we don't need more than one such domain. I was still thinking based on the old approach. Also, combining multiple foundation will work because we can configure multiple CAs for the instance-identity-certificate-based mTLS domain.

>these domains are no different from any other private/shared domains CC holds, especially if they are getting accessed from the outside. (apps outside and inside the diego need to be able to resolve the domain, and especially for cross-regional communication you need to ensure that these domains are not overlapping and are resolvable globally)

Yes, this is what I meant, and we don't have any additional requirements here.

app/access/route_policy_access.rb
```diff
@@ -0,0 +1,66 @@
+module VCAP::CloudController
+  class RoutePolicyAccess < BaseAccess
```
philippthun: `BaseAccess` subclasses are the v2 permission layer used by `RestController` - not needed since there is no v2 endpoint for this feature.

app/models/runtime/route_policy.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    def after_destroy
+      super
+      touch_associated_processes
+    end
+
+    private
+
+    def touch_associated_processes
```
philippthun: I think this should be changed to `ProcessRouteHandler.new(process).notify_backend_of_route_update` wrapped in `db.after_commit`. Running processes should be updated, but the `updated_at` column of processes should not be changed.

app/models/runtime/route_policy.rb
```diff
@@ -0,0 +1,42 @@
+module VCAP::CloudController
+  class RoutePolicy < Sequel::Model(:route_policies)
+    many_to_one :route,
+                class: 'VCAP::CloudController::Route',
+                key: :route_id,
+                primary_key: :id,
+                without_guid_generation: true
```
philippthun: I think `without_guid_generation: true` is not needed as in this case an int column is used.

db/migrations/20260421074455_create_route_policies.rb
```diff
@@ -0,0 +1,58 @@
+Sequel.migration do
+  up do
+    unless table_exists?(:route_policies)
+      create_table :route_policies do
```
philippthun, line 4: Any reason for not using `VCAP::Migration.common(self, :route_policies)`? This migration leads to a different table layout (e.g. `DateTime` instead of `Timestamp`, not using `Sequel::CURRENT_TIMESTAMP` as default).

db/migrations/20260421074455_create_route_policies.rb
```diff
@@ -0,0 +1,58 @@
...
+        DateTime :updated_at, null: false
+
+        index :guid, unique: true, name: :route_policies_guid_index
+        index %i[route_id source], unique: true, name: :route_policies_route_id_source_index
+        foreign_key [:route_id], :routes, on_delete: :cascade, name: :fk_route_policies_route_id
+      end
+    end
+
+    unless table_exists?(:route_policy_labels)
+      create_table :route_policy_labels do
```
philippthun, line 16: Also for the labels and annotations tables, the `VCAP::Migration.common` helper should be used to have a consistent base.

devbox.d/mysql80/my.cnf
```diff
@@ -0,0 +1,6 @@
+# MySQL configuration file
```
philippthun: Please exclude from the commit.

devbox.json
```diff
@@ -0,0 +1,69 @@
+{
```
philippthun: Please exclude from the commit.

devbox.lock
```diff
@@ -0,0 +1,783 @@
+{
```
philippthun: Please exclude from the commit.

lib/cloud_controller/diego/protocol/routing_info.rb
```diff
@@ -37,17 +37,34 @@ def http_info(process_eager)
...
+
+          add_mtls_options(info, r) if r.domain.enforce_route_policies
+
+          info
+        end
+
+        def add_mtls_options(info, route)
+          # Inject mTLS policy options for enforce_route_policies domains.
+          # These are GoRouter-internal keys and are filtered from the /v3/routes API.
+          mtls_options = info['options']&.dup || {}
```
philippthun: I would find this approach cleaner:
```
mtls_options = {}
...
info['options'] = (info['options'] || {}).merge(mtls_options)
```

docs/v3/source/includes/resources/route_policies/_list.md.erb
```diff
@@ -0,0 +1,152 @@
...
+| **source_guids**         | _list of strings_                                | Comma-delimited list of GUIDs to text-match against source strings (useful for finding stale policies when resources are deleted)
+| **page**                 | _integer_                                        | Page to display; valid values are integers >= 1
+| **per_page**             | _integer_                                        | Number of results per page; valid values are 1 through 5000
+| **order_by**             | _string_                                         | Value to sort by. Defaults to ascending; prepend with `-` to sort descending. Valid values are `created_at`, `updated_at`
+| **label_selector**       | _string_                                         | A query string containing a list of [label selector](#labels-and-selectors) requirements
+| **include**              | _string_                                         | Optionally include related resources in the response; valid values are `route`, `app`, `space`, `organization`, and `source` (which includes app, space, or organization based on the source type)
```
philippthun: This is somehow confusing: `space_guids` filters by the route's space, but `include=space` includes the source space. Wouldn't `route` and `source` be enough as includes?

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+  end
+
+  def create
+    message = RoutePolicyCreateMessage.new(hashed_params[:body])
+    unprocessable!(message.errors.full_messages) unless message.valid?
+
+    route = find_and_authorize_route(message.route_guid)
+    validate_route_domain(route)
+
+    route_policy = VCAP::CloudController::RoutePolicy.db.transaction do
```
philippthun: This block should be extracted into a `RoutePolicyCreate` action.

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+      # Lock existing route policies for this route to prevent concurrent inserts
+      # from violating cf:any exclusivity or uniqueness constraints
+      VCAP::CloudController::RoutePolicy.where(route_id: route.id).for_update.all
+
+      validate_source_exclusivity(route, message.source)
+
+      policy = VCAP::CloudController::RoutePolicy.new(
+        guid: SecureRandom.uuid,
```
philippthun: `guid`, `created_at` and `updated_at` should not be set - this should be handled by the model (guid auto-generation, defaults).

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    unprocessable!("Cannot create route policies for route '#{route.guid}': the route's domain does not have enforce_route_policies enabled.")
+  end
+
+  def validate_source_exclusivity(route, source)
+    existing_sources = route.route_policies.map(&:source)
+
+    # Enforce cf:any exclusivity: if route already has a cf:any policy, reject new policies;
+    # if new policy is cf:any, reject if route already has any policies.
+    unprocessable!("Cannot add 'cf:any' source when other route policies already exist for this route.") if source == 'cf:any' && existing_sources.any?
+    unprocessable!("Cannot add source '#{source}': route already has a 'cf:any' policy.") if existing_sources.include?('cf:any') && source != 'cf:any'
```
philippthun: I think the `&& source != 'cf:any'` is not needed.

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    end
+    return if route.domain.enforce_route_policies
+
+    unprocessable!("Cannot create route policies for route '#{route.guid}': the route's domain does not have enforce_route_policies enabled.")
+  end
+
+  def validate_source_exclusivity(route, source)
+    existing_sources = route.route_policies.map(&:source)
+
+    # Enforce cf:any exclusivity: if route already has a cf:any policy, reject new policies;
```
philippthun: The order of the comment does not match the code.

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    message = RoutePolicyCreateMessage.new(hashed_params[:body])
+    unprocessable!(message.errors.full_messages) unless message.valid?
+
+    route = find_and_authorize_route(message.route_guid)
+    validate_route_domain(route)
+
+    route_policy = VCAP::CloudController::RoutePolicy.db.transaction do
+      # Lock existing route policies for this route to prevent concurrent inserts
+      # from violating cf:any exclusivity or uniqueness constraints
+      VCAP::CloudController::RoutePolicy.where(route_id: route.id).for_update.all
```
philippthun: As this returns the locked policies, it would be efficient to use them in the `validate_source_exclusivity` method, e.g.:
```
locked_policies = VCAP::CloudController::RoutePolicy.where(route_id: route.id).for_update.all
validate_source_exclusivity(locked_policies, message.source)
```
And in `validate_source_exclusivity`:
```
existing_sources = locked_policies.map(&:source)
```

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+  end
+
+  def update
+    route_policy = VCAP::CloudController::RoutePolicy.find(guid: hashed_params[:guid])
+    resource_not_found!(:route_policy) unless route_policy
+
+    route = route_policy.route
+    resource_not_found!(:route_policy) unless route && permission_queryer.can_read_from_space?(route.space.id, route.space.organization_id)
+    unauthorized! unless permission_queryer.can_write_to_active_space?(route.space.id)
+    suspended! unless permission_queryer.is_space_active?(route.space.id)
```
philippthun: Maybe `update` and `delete` could reuse a slightly changed `find_and_authorize_route` method. These three lines are identical.

philippthun (commented):
I'm not done yet with review, but wanted to share my comments anyway.

One general question: What do you think about having two columns in the `route_policy` table for the `source`: `source_type` and `source_guid`? Source type and guid are both things mentioned in the docs and for `source_guid` there even is a filter. This would make database queries faster as indexing these columns would actually allow easier access, i.e. no need for `LIKE` queries.

ameowlia:
I was able to create a route that is both internal AND and mtls domain.

```
go run . create-shared-domain apps.identity.meow --enforce-route-policies --scope any --internal
Creating shared domain apps.identity.meow as admin...
OK

TIP: Domain 'apps.identity.meow' is a shared identity-aware domain with route policy enforcement enabled. Routes on this domain require route policies to allow traffic.
```

Is this intentional? I think we should limit it to only allowing one or the other. Same with the router group flag.

docs/v3/source/includes/resources/route_policies/_list.md.erb
rkoster: Yeah the more I think about this, to more confusing it gets. So let's only use route and source. 

@jcvrabo do you agree?

app/controllers/v3/route_policies_controller.rb
rkoster: Added include support to the show endpoint in `89a2ac9`. It now supports the same `include` parameters as the list endpoint (`source`, `route`). Note: per philippthun's feedback in thread below, we're limiting includes to `route` and `source` only (dropping `space`, `org`, `app`) to avoid confusion with the space_guids filter semantics.

app/controllers/v3/route_policies_controller.rb
rkoster: Fixed in `238cc0e` — removed manual assignment; the model now handles guid auto-generation and timestamp defaults.

app/controllers/v3/route_policies_controller.rb
rkoster: Fixed in `238cc0e` — extracted a shared `find_and_authorize_route` helper reused across show, update, and destroy actions.

lib/cloud_controller/diego/protocol/routing_info.rb
rkoster: Refactored in the approach you suggested — building a fresh hash and merging at the end. Also renamed `add_mtls_options` → `add_route_policy_options` and the local variable to `route_policy_options` to align with the current route policy terminology (the `mtls_` prefix was from an earlier RFC iteration).

docs/v3/source/includes/resources/route_policies/_list.md.erb
rkoster: Agreed — limited includes to `route` and `source` only in `b62a643`. The `source` include already resolves the underlying app/space/org from the source string, making individual `app`/`space`/`organization` keywords redundant and confusing (since `space_guids` filters by the route's space, not the source space).

rkoster:
@ameowlia Thanks for raising this. Added validation in `8ef6200` to prevent combining `enforce_route_policies` with either `internal: true` or `router_group`. Internal domains bypass GoRouter (C2C networking), and TCP router groups do not support mTLS policy enforcement — so neither can enforce route policies.

app/actions/route_policy_create.rb
```diff
@@ -0,0 +1,34 @@
+module VCAP::CloudController
+  class RoutePolicyCreate
+    class Error < StandardError
+    end
+
+    def create(route:, message:)
+      RoutePolicy.db.transaction do
+        # Lock existing route policies for this route to prevent concurrent inserts
+        # from violating cf:any exclusivity or uniqueness constraints
+        locked_policies = RoutePolicy.where(route_id: route.id).for_update.all
```
philippthun: `for_update.all` only locks existing rows. Two concurrent creates on the same `route_id` will both read `[]` and both validate
`cf:any`. The `[route_id, source]` unique index doesn't catch it because the sources differ.

Lock the parent route instead:
```
Route.where(id: route.id).for_update.first
```

app/decorators/include_route_policy_source_decorator.rb
```diff
@@ -0,0 +1,74 @@
...
+          space_guids << resource_guid
+        when 'org'
+          org_guids << resource_guid
+        end
+      end
+
+      # Fetch and present resources
+      hash[:included][:apps] = fetch_and_present_apps(app_guids.uniq)
+      hash[:included][:spaces] = fetch_and_present_spaces(space_guids.uniq)
+      hash[:included][:organizations] = fetch_and_present_organizations(org_guids.uniq)
```
philippthun: No permission filter on the `where(guid: ...)` lookups. A space developer can write a policy on their own route with an app GUID in another org and gets the full `AppPresenter` payload back. Same for spaces and orgs.

Intersect each GUID list with `permission_queryer.readable_*_guids` before the fetch.

app/actions/route_policy_create.rb
```diff
@@ -0,0 +1,34 @@
+module VCAP::CloudController
+  class RoutePolicyCreate
+    class Error < StandardError
+    end
+
+    def create(route:, message:)
```
philippthun, line 6: The domain's `route_policies_scope` is documented as an enforcement boundary, but nothing here checks it. A domain with `route_policies_scope: 'space'` will happily accept `cf:org:<guid>` or `cf:any`. Maybe add a `validate_scope!` method called before the transaction.

db/migrations/20260421074455_create_route_policies.rb
```diff
@@ -0,0 +1,46 @@
+Sequel.migration do
+  up do
+    unless table_exists?(:route_policies)
+      create_table :route_policies do
+        VCAP::Migration.common(self, :route_policies)
+        String :source, size: 255, null: false
```
philippthun: I think it's better to create two columns instead of one: `source_type` and `source_guid` Source type and guid are both things mentioned in the docs and for `source_guid` there even is a filter. This would make database queries faster as indexing these columns would actually allow easier access, i.e. no need for `LIKE` queries.

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+    unprocessable!("Cannot create route policies for route '#{route.guid}': the route's domain does not have enforce_route_policies enabled.")
+  end
+
+  def build_dataset(message)
+    dataset = VCAP::CloudController::RoutePolicy.dataset
+
+    if permission_queryer.can_read_globally?
+      readable_route_ids = VCAP::CloudController::Route.select(:id)
+    else
+      readable_space_ids = permission_queryer.readable_space_scoped_spaces_query.select(:id)
```
philippthun: Switch to `readable_spaces_query` as `readable_space_scoped_spaces_query` excludes org managers, so they cannot list policies (but they can show them as other methods use `can_read_from_space?` which includes org managers). Ideally there should be request specs for all roles (there is a reusable pattern for that).

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+    resource_not_found!(:route_policy) unless route && permission_queryer.can_read_from_space?(route.space.id, route.space.organization_id)
+
+    decorators = []
+    decorators << IncludeRoutePolicySourceDecorator if IncludeRoutePolicySourceDecorator.match?(message.include)
+    decorators << IncludeRoutePolicyRouteDecorator if IncludeRoutePolicyRouteDecorator.match?(message.include)
+
+    render status: :ok, json: Presenters::V3::RoutePolicyPresenter.new(route_policy, decorators: decorators)
+  end
+
+  def create
```
philippthun, line 49: There are no audit events for create/update/destroy actions. Should those be added?

app/messages/route_policies_list_message.rb
```diff
@@ -0,0 +1,24 @@
+require 'messages/list_message'
+
+module VCAP::CloudController
+  class RoutePoliciesListMessage < ListMessage
```
philippthun: The docs mention that `label_selector` is supported. But this class does not inherit from `MetadataListMessage`... Would also need to be added to the `build_dataset` method in `route_policies_controller.rb`.

docs/v3/source/includes/resources/route_policies/_get.md.erb
```diff
@@ -0,0 +1,46 @@
...
+#### Permitted roles
+
+Role | Notes
+----- | ---
+Admin |
+Admin Read-Only |
+Global Auditor |
+Org Manager |
+Org Auditor |
+Org Billing Manager |
```
philippthun: Org auditor and billing manager have no "space read permissions", so they won't see anything.

docs/v3/source/includes/resources/route_policies/_list.md.erb
```diff
@@ -0,0 +1,152 @@
...
+#### Permitted roles
+
+Role | Notes
+----- | ---
+Admin |
+Admin Read-Only |
+Global Auditor |
+Org Manager |
+Org Auditor |
+Org Billing Manager |
```
philippthun: Org auditor and billing manager have no "space read permissions", so they will get an empty list.

app/actions/route_policy_create.rb
rkoster: Fixed in `a9053ea` — replaced `RoutePolicy.where(route_id: route.id).for_update.all` with `Route.where(id: route.id).for_update.first` followed by a plain `.all` read of policies.

`SELECT ... FOR UPDATE` on an empty table acquires no row locks, so two concurrent creates on the same route both see `[]` and both pass validation. Locking the parent route row (which always exists) serializes them at the database level regardless of how many policies currently exist.

Added `spec/unit/actions/route_policy_create_spec.rb` to document the fix.

app/decorators/include_route_policy_source_decorator.rb
rkoster: Fixed in `610f214` — added permission filtering to all three fetch methods, following the `IncludeSpaceDecorator` pattern.

For each resource type, the query is now gated on the current user's readable set:
- Apps: `.where(space_guid: permission_queryer.readable_space_guids_query)`
- Spaces: `.where(guid: permission_queryer.readable_space_guids_query)`
- Orgs: `.where(guid: permission_queryer.readable_org_guids_query)`

Global readers (admin, admin_read_only, global_auditor) skip the filter. A `Permissions` instance is built from `SecurityContext.current_user` inside `decorate`, consistent with how other decorators handle this.

Added `spec/unit/decorators/include_route_policy_source_decorator_spec.rb` with permission coverage.

app/actions/route_policy_create.rb
rkoster: Added `validate_scope!` called before the transaction.

`cf:org:<guid>` is rejected when `route_policies_scope` is `'space'` — an org-level source is too broad for a domain whose boundary is the space. `scope: 'org'` and `scope: 'any'` pass all source types through.

`cf:any` is intentionally allowed at every scope. Its meaning narrows with the domain boundary — on a `scope: 'space'` domain GoRouter interprets it as "any authenticated app within the same space", not globally. Since the contextual scoping happens at request time in GoRouter rather than at policy-creation time in CC, rejecting it here would be incorrect.

app/controllers/v3/route_policies_controller.rb
rkoster: Switched `readable_space_scoped_spaces_query` to `readable_spaces_query` in `build_dataset`. The former uses `SPACE_ROLES` which excludes `ORG_MANAGER`; the latter uses `ROLES_FOR_SPACE_READING` which includes it — consistent with how `show` already gates access via `can_read_from_space?`.

Added `it_behaves_like 'permissions for list endpoint'` and `'permissions for single object endpoint'` covering all roles in `spec/request/route_policies_spec.rb`. The `org_manager` case on the list endpoint was the failing RED test; the one-line swap made it green. `org_auditor`, `org_billing_manager`, and `no_role` get 200/empty and 404 respectively.

db/migrations/20260421074455_create_route_policies.rb
rkoster: Split `source` into `source_type` (`"app"` | `"space"` | `"org"` | `"any"`) and `source_guid` (UUID or `""` for `cf:any`) in e94dce58e. The DB unique index is now on `(route_id, source_type, source_guid)`, which turns the `source_guids` filter into a direct column lookup instead of a LIKE scan, and makes the `sources` filter a two-column exact match. The external API shape (composite `source` string) is unchanged — `RoutePolicy` exposes a virtual getter/setter that translates between the two representations.

docs/v3/source/includes/resources/route_policies/_get.md.erb
rkoster: Good catch — this was inconsistent with routes, which use `ROLES_FOR_ROUTE_READING` (includes org auditor). Fixed in 11b500736: route policies now use the same role set, so org auditors can read route policies. Org billing managers remain excluded (consistent with routes). Added a note for org billing manager in the permissions table.

docs/v3/source/includes/resources/route_policies/_list.md.erb
rkoster: Same fix as the show endpoint — 11b500736 switches route policy list to `ROLES_FOR_ROUTE_READING` so org auditors now see the policies they have access to. Org billing managers still get an empty list (consistent with routes); added a note for them in the permissions table.

app/controllers/v3/route_policies_controller.rb
rkoster: Added in 5cc32572a — `audit.route_policy.create`, `audit.route_policy.update`, and `audit.route_policy.delete` events via `RoutePolicyEventRepository`, consistent with the `audit.route.*` pattern.

app/messages/route_policies_list_message.rb
rkoster: Fixed in `df2b677` — `RoutePoliciesListMessage` now inherits from `MetadataListMessage`, which auto-registers `:label_selector` and wires `LabelSelectorRequirementValidator`. `build_dataset` in the controller now delegates to `LabelSelectorQueryGenerator.add_selector_queries` when a selector is requested.

rkoster:
@philippthun are there any more blockers on this PR? I'm asking because the routing-release changes have just been merged: https://github.com/cloudfoundry/routing-release/pull/535

app/messages/route_policies_list_message.rb
```diff
@@ -0,0 +1,24 @@
+require 'messages/metadata_list_message'
+
+module VCAP::CloudController
+  class RoutePoliciesListMessage < MetadataListMessage
+    register_allowed_keys %i[
+      guids
+      route_guids
+      space_guids
+      sources
```
philippthun, line 9: Each item from the `sources` array should be validated against the same regex as in the create message (helper method `source_format_valid`). Otherwise the current implementation of the controller will raise.

app/actions/route_policy_create.rb
```diff
@@ -0,0 +1,47 @@
...
+      validate_scope!(route.domain, message.source)
+
+      RoutePolicy.db.transaction do
+        # Lock the parent route row to serialize concurrent creates.
+        # SELECT ... FOR UPDATE on an empty policies table acquires no row locks,
+        # so two concurrent transactions can both read [] and both pass cf:any
+        # exclusivity validation. Locking the route row (which always exists)
+        # ensures they serialize regardless of how many policies currently exist.
+        Route.where(id: route.id).for_update.first
```
philippthun: There is still a small window for a race condition. If the route gets deleted right after it has been fetched by the controller, this will not fail but return `nil` and `RoutePolicy.create` won't succeed. You can simply add `or raise Error.new(<some meaningful message>)` to this line.

app/messages/route_policy_create_message.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def source_format_valid
+      return unless source.is_a?(String)
+      return if SOURCE_REGEX.match?(source)
+
+      errors.add(:source, "must be in format 'cf:app:<uuid>', 'cf:space:<uuid>', 'cf:org:<uuid>', or 'cf:any'")
+    end
+
+    def source_not_cf_any_with_others
+      # enforced at the controller level when checking existing policies on the route
```
philippthun: The comment is good, but there shouldn't be an empty method for this. So maybe move somewhere near to `validate :source_format_valid` and drop `source_not_cf_any_with_others`.

docs/v3/source/includes/resources/route_policies/_list.md.erb
```diff
@@ -0,0 +1,152 @@
...
+
+#### Query parameters
+
+| Name                     | Type                                             | Description
+| -----------              | --------                                         | -----------
+| **guids**                | _list of strings_                                | Comma-delimited list of route policy guids to filter by
+| **route_guids**          | _list of strings_                                | Comma-delimited list of route guids to filter by
+| **space_guids**          | _list of strings_                                | Comma-delimited list of space guids to filter by (filters by the route's space)
+| **sources**              | _list of strings_                                | Comma-delimited list of exact source strings to filter by (e.g., `cf:any`, `cf:app:guid`)
+| **source_guids**         | _list of strings_                                | Comma-delimited list of GUIDs to text-match against source strings (useful for finding stale policies when resources are deleted)
```
philippthun: `Comma-delimited list of source GUIDs to filter by exact match` - the text-match is gone now that source type and source guid have been split.

philippthun (commented):
Only a few little things - then it should be ready...

app/messages/route_policies_list_message.rb
rkoster: Moved `SOURCE_REGEX` to `RoutePolicy` (natural home for format constants shared across messages). Added `validates :sources, array: true, allow_nil: true` and a `sources_format_valid` validator to `RoutePoliciesListMessage` that checks each element against `RoutePolicy::SOURCE_REGEX`. Added tests covering nil, valid formats (all four types), fully invalid, and mixed valid/invalid arrays — all 35 specs pass. The decorator keeps its own local regex — structurally different (no `cf:any`, has capture groups for parsing).

app/actions/route_policy_create.rb
rkoster: Added `or raise Error.new("Route '#{route.guid}' not found.")` to the `for_update.first` call in 6d6706cd9. Added a test that stubs the route relation's `first` to return `nil` (using the same spy pattern as the existing concurrent-creates test) — 17 specs pass.

app/messages/route_policy_create_message.rb
rkoster: Dropped the empty `source_not_cf_any_with_others` method and the `validate` call for it. Moved the comment inline next to `validate :source_format_valid`.

docs/v3/source/includes/resources/route_policies/_list.md.erb
rkoster: Fixed in c28c42abc — updated the description to: "Comma-delimited list of GUIDs to filter by; matches the GUID portion of the source (e.g. the app, space, or org GUID)".

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,184 @@
...
+    resource_not_found!(:route_policy) unless route_policy
+
+    find_and_authorize_route_for_policy(route_policy)
+
+    route_policy_event_repository.record_route_policy_delete(
+      route_policy,
+      user_audit_info
+    )
+
+    route_policy.destroy
```
philippthun: First destroy, then record the audit event.

app/controllers/v3/route_policies_controller.rb
```diff
@@ -0,0 +1,184 @@
...
+    route = VCAP::CloudController::Route.find(guid: route_guid)
+    resource_not_found!(:route) unless route && permission_queryer.can_read_from_space?(route.space.id, route.space.organization_id)
+    unauthorized! unless permission_queryer.can_write_to_active_space?(route.space.id)
+    suspended! unless permission_queryer.is_space_active?(route.space.id)
+    route
+  end
+
+  def find_and_authorize_route_for_policy(route_policy)
+    route = route_policy.route
+    resource_not_found!(:route_policy) unless route && permission_queryer.can_read_from_space?(route.space.id, route.space.organization_id)
```
philippthun: Should also be `can_read_route_policy_from_space?`, or?

docs/v3/source/includes/resources/route_policies/_header.md
```diff
@@ -0,0 +1,11 @@
+## Route Policies
```
philippthun, line 1: We should mark this new feature as experimental.

philippthun (commented):
That's really the final review - I promise :smile:

db/migrations/20260421074455_create_route_policies.rb
```diff
@@ -0,0 +1,47 @@
...
+    unless table_exists?(:route_policy_labels)
+      create_table :route_policy_labels do
+        VCAP::Migration.common(self, :route_policy_labels)
+        String :resource_guid, null: false, size: 255
+        String :key_prefix, null: false, default: '', size: 253
+        String :key_name, null: false, size: 63
+        String :value, null: false, size: 63
+
+        index :resource_guid, name: :route_policy_labels_resource_guid_index
+        index %i[resource_guid key_prefix key_name], unique: true, name: :route_policy_labels_compound_index
```
philippthun: This unique indexes deviates from the convention introduced by `20240102150000_add_annotation_label_uniqueness`. That migration used `add_unique_constraint` with naming `<table>_unique` for every existing labels/annotations table.

Ideally switch to `add_unique_constraint` with the following names: `route_policy_labels_unique` / `route_policy_annotations_unique`

app/controllers/v3/route_policies_controller.rb
rkoster: Fixed in c8ee32eb — swapped the order so `route_policy.destroy` is called before `record_route_policy_delete`, ensuring no audit event is recorded if the destroy fails. Added a request spec to cover the failure case.

app/controllers/v3/route_policies_controller.rb
rkoster: Fixed in 66a48a6f — switched to `can_read_route_policy_from_space?` in `find_and_authorize_route_for_policy`, consistent with the `show` action.

docs/v3/source/includes/resources/route_policies/_header.md
rkoster: The API design was discussed at length during the RFC process (https://github.com/cloudfoundry/community/pull/1438) before being merged. Given that level of review, applying an experimental label feels inconsistent with the maturity of the design. Happy to discuss further if you feel strongly about it.

db/migrations/20260421074455_create_route_policies.rb
rkoster: Fixed in ab112db9 — moved the compound unique constraints out of the `create_table` blocks and into `alter_table` + `add_unique_constraint` calls with names `route_policy_labels_unique` and `route_policy_annotations_unique`, consistent with the convention from `20240102150000_add_annotation_label_uniqueness`.

docs/v3/source/includes/resources/route_policies/_header.md
philippthun: We marked all new features as experimental in the past. This is just a very cheap insurance against possible changes that may become necessary after the initial rollout. I'm not doubting the RFC's maturity, it's a pattern we used for other RFC-driven features as well (e.g. #4906). Removing this small addition in a few months is also very easy.

rkoster: Fair point — added the experimental marker in 6ae25754. Thanks for the context on the convention.

philippthun (approved):
Ready to get merged!

Note: I've updated the PR description to reflect the changed names, e.g. access rule → route policy.

PR #4981, opened 2026-03-30
Install buildpacks based on order on config

### LINKED ISSUE #3798: Cloud Controller should install new buildpacks in a more deterministic order
## Issue

Operators specify an array of buildpacks for Cloud Controller to install in [capi-release](https://github.com/cloudfoundry/capi-release/blob/458767f3c0b7049b9f732f14582d88f5100357f0/jobs/cloud_controller_ng/spec#L984). Cloud Controller [groups these by name](https://github.com/cloudfoundry/cloud_controller_ng/blob/02b5d20c3a83b7e8972dde0825cd7347e9b3eb5b/lib/cloud_controller/install_buildpacks.rb#L47) and then enqueues jobs to either update existing buildpacks or install new ones. These jobs are picked up by the local workers in the order in which they get to them. Since multiple workers are handling these jobs this can cause the buildpacks to be installed in non-deterministic orders.

## Context

Buildpack order matters since it affects [buildpack auto-detection](https://docs.cloudfoundry.org/buildpacks/understand-buildpacks.html#buildpack-detection). Many operators assume that the order they specify for their buildpacks in their CF deployment manifest will result in them using that order for auto-detection. The fact that the installation order/position for newly installed buildpacks is non-deterministic is surprising to operators.

## Steps to Reproduce

This is difficult to reproduce due to the randomness, but after multiple deploys installing a fresh set of buildpacks (not updating existing ones) you should expect to see the buildpacks installed in a different order each time.

## Expected result

I would expect the buildpacks to be installed in a consistent order.

## Current result

The buildpacks are _usually_ installed in the specified order (grouped by buildpack name in cases where multiple versions of buildpacks exist for different stacks). However occasionally the buildpacks will be installed in completely different orders which can cause issues with auto-detection.

Examples from two different installations:

```
➜ cf buildpacks | egrep "nginx|web_servers"
4          nginx_buildpack          cflinuxfs3    true      false    READY   nginx_buildpack-cached-cflinuxfs3-v1.2.14.zip
14         web_servers_cnb_beta     cflinuxfs4    true      false    READY   web-servers-cnb-buildpack-cflinuxfs4-v0.16.0.zip
22         nginx_buildpack          cflinuxfs4    true      false    READY   nginx_buildpack-cached-cflinuxfs4-v1.2.14.zip
```

```
➜ cf buildpacks | egrep "nginx|web_servers"
7          nginx_buildpack          cflinuxfs3   true      false    READY   nginx_buildpack-cached-cflinuxfs3-v1.2.14.zip
8          nginx_buildpack          cflinuxfs4   true      false    READY   nginx_buildpack-cached-cflinuxfs4-v1.2.14.zip
24         web_servers_cnb_beta     cflinuxfs4   true      false    READY   web-servers-cnb-buildpack-cflinuxfs4-v0.16.0.zip
```

## Possible Fix
Operators can reorder the buildpacks to their liking using `cf update-buildpack BUILDPACK_NAME --position <position>`. This updates the position column in the data base.  We could consider having our install buildpacks task specify a `position` based on the `install_buildpacks` order when installing new buildpacks instead of leaving it up to chance.

```
mysql> select name,position from buildpacks;
+-----------------------+----------+
| name                  | position |
+-----------------------+----------+
| staticfile_buildpack  |        1 |
| java_buildpack        |        2 |
| ruby_buildpack        |        3 |
| dotnet_core_buildpack |        4 |
| nodejs_buildpack      |        5 |
| go_buildpack          |        6 |
| python_buildpack      |        7 |
| nginx_buildpack       |        8 |
| php_buildpack         |        9 |
| r_buildpack           |       10 |
| binary_buildpack      |       11 |
+-----------------------+----------+
11 rows in set (0.04 sec)
```

tomkennedy513:
This change will install buildpacks in the same order as they are specified in the config without modifying the order of existing buildpacks. The buildpacks index in the config is passed to the install job that will shift the buildpack the final position.

This resolves #3798 

### DIFF
--- a/app/jobs/runtime/buildpack_installer_factory.rb
+++ b/app/jobs/runtime/buildpack_installer_factory.rb
@@ -56,7 +56,8 @@ module VCAP::CloudController
                                                                                                    name: buildpack_name,
                                                                                                    stack: buildpack_fields[:stack],
                                                                                                    file: buildpack_fields[:file],
-                                                                                                   options: buildpack_fields[:options]
+                                                                                                   options: buildpack_fields[:options],
+                                                                                                   config_index: buildpack_fields[:config_index]
                                                                                                  })
                             end
           end
--- a/app/jobs/runtime/create_buildpack_installer.rb
+++ b/app/jobs/runtime/create_buildpack_installer.rb
@@ -2,6 +2,13 @@ module VCAP::CloudController
   module Jobs
     module Runtime
       class CreateBuildpackInstaller < BuildpackInstaller
+        attr_accessor :config_index
+
+        def initialize(job_options)
+          super
+          @config_index = job_options[:config_index]
+        end
+
         def perform
           logger.info "Installing buildpack name `#{name}' with stack `#{stack_name}'"
 
@@ -11,6 +18,7 @@ module VCAP::CloudController
           buildpacks_lock.db.transaction do
             buildpacks_lock.lock!
             buildpack = Buildpack.create(name: name, stack: stack_name, lifecycle: options[:lifecycle])
+            buildpack.move_to(config_index + 1) if !config_index.nil? && !options.key?(:position)
           end
           begin
             buildpack_uploader.upload_buildpack(buildpack, file, File.basename(file))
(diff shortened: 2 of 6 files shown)


johha (approved):
LGTM

PR #5010, opened 2026-04-09
Fix route options serialization to preserve explicitly removed values

### LINKED ISSUE #551: Per-route load balancing algorithm is not reset to platform default when removed from a route
Current behavior
----------

When a per-route load balancing algorithm is configured (e.g., `hash` for hash-based routing) and later removed via `cf update-route -r loadbalancing`, gorouter's `setPoolLoadBalancingAlgorithm` ignores the empty value and the `EndpointPool` retains the previously set algorithm. This means the route continues to use the per-route algorithm (e.g., hash-based routing with its Maglev lookup table) instead of reverting to the platform default (e.g., `round-robin`).

The root cause is in `mbus/subscriber.go`: when the `loadbalancing` field is absent or empty in the NATS message, `makeEndpoint` leaves the endpoint's `LoadBalancingAlgorithm` as an empty string. In `route/pool.go`, `setPoolLoadBalancingAlgorithm` treats an empty string as "no change", so the pool keeps the old algorithm.

Desired behavior
----------

When the `loadbalancing` field is absent or empty in the NATS route registration message, gorouter should default the endpoint's load balancing algorithm to the platform default configured via `router.balancing_algorithm`. This ensures that removing a per-route algorithm causes the pool to revert to the platform default. Hash-based routing state (`HashLookupTable`, `HashRoutingProperties`) should be cleaned up when switching away from hash-based routing.

Affected Version
----------

develop

hoffmaen:
## Summary

When a route option (e.g., `loadbalancing`) is explicitly removed via the API, the value is set to `nil` internally. The `options=` setter previously called `.compact` on the options hash, which removed nil keys entirely from the serialized JSON. This made the resulting `"options":{}` indistinguishable from a route that never had options set, preventing downstream consumers (e.g., gorouter) from detecting the explicit removal and reverting to the platform default.

Fixes https://github.com/cloudfoundry/routing-release/issues/551
   
## Changes                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
                                                            
- Replace `.compact` with `.transform_values { |v| v.nil? ? '' : v }` in `Route#options_with_serialization=`, so that explicitly removed options are serialized as empty strings (e.g., `"options":{"loadbalancing":""}`) instead of being dropped                                                                                                                                                                                                                                     
- Update assertions in `route_update_spec.rb` and `manifest_route_update_spec.rb` to expect `{ 'loadbalancing' => '' }` instead of `{}`
- Add a new model-level test in `route_spec.rb` that explicitly verifies nil option values are serialized as empty strings                                                                                                                                                                                                                                                                                                                                                             
   
## Context                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
                                                            
This is the Cloud Controller counterpart to a gorouter issue (https://github.com/cloudfoundry/routing-release/issues/551) that adds support for reverting a per-route load balancing algorithm to the platform default. Without this CC change, the removal signal is lost before it reaches gorouter via NATS. 

### DIFF
--- a/app/models/runtime/route.rb
+++ b/app/models/runtime/route.rb
@@ -74,8 +74,9 @@ module VCAP::CloudController
       cleaned_opts = remove_hash_options_for_non_hash_loadbalancing(opts)
       rounded_opts = round_hash_balance_to_one_decimal(cleaned_opts)
       normalized_opts = normalize_hash_balance_to_string(rounded_opts)
-      # Remove nil values after all processing
-      normalized_opts = normalized_opts.compact if normalized_opts.is_a?(Hash)
+      # Convert nil values to empty strings to signal explicit removal to downstream consumers (e.g. gorouter),
+      # so they can distinguish "option was explicitly removed" from "option was never set".
+      normalized_opts = normalized_opts.transform_values { |v| v.nil? ? '' : v } if normalized_opts.is_a?(Hash)
       self.options_without_serialization = Oj.dump(normalized_opts)
     end
 
--- a/spec/unit/actions/manifest_route_update_spec.rb
+++ b/spec/unit/actions/manifest_route_update_spec.rb
@@ -650,8 +650,7 @@ module VCAP::CloudController
             ManifestRouteUpdate.update(app.guid, message, user_audit_info)
 
             route.reload
-            expect(route.options).to eq({})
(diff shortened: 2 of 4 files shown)


hoffmaen:
Cloud Controller change not required any more.

PR #5011, opened 2026-04-09
Recover stuck service operations after transient DB failures

serdarozerr:
## Problem
When CCDB becomes temporarily unreachable during a service broker polling cycle,
the polling job fails permanently even though the broker is still processing the
operation. This leaves the service resource in an inconsistent state:
- `last_operation.state = 'in progress'` (broker still working or already finished)
- `pollable_job.state = FAILED` (CC gave up)

## Solution
Adds a new periodic scheduled job `ServiceOperationsCreateInProgressCleanup` that
detects stuck `create` operations whose polling job has permanently failed and
resolves them by marking the operation as failed and triggering orphan mitigation
to deprovision any broker-side resource, giving clients a definitive final state.

**Detection chain:**
`service_instance/binding/key_operations.state = 'in progress'`
  AND `type = 'create'`
  AND `created_at` within the max async poll window
→ JOIN `service_instances/bindings/keys` (via foreign key)
→ JOIN `pollable_jobs` (via `resource_guid`) WHERE `state IN (POLLING, FAILED)`
  AND `operation = 'service_instance/bindings/keys.create'`
→ JOIN `delayed_jobs` (via `delayed_job_guid`) WHERE `failed_at IS NOT NULL`

**Recovery:**
Marks the stuck operation as `failed`, sets the pollable job to `FAILED`, and
calls `OrphanMitigator` to enqueue a broker-side DELETE for the potentially
orphaned resource. A row-level `FOR UPDATE SKIP LOCKED` guard prevents double
mitigation when multiple CC instances run concurrently.

**Scope:**
Covers `service_instance.create`, `service_bindings.create`, and
`service_keys.create`. Delete and update operations are intentionally excluded —
retrying a delete on the same resource GUID can cross-match with the old failed
pollable job, making safe recovery impossible without additional guards.

* Related links
https://github.com/cloudfoundry/capi-release/pull/653


app/jobs/runtime/delayed_jobs_recover.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        def default_maximum_duration_seconds
+          Config.config.get(:broker_client_max_async_poll_duration_minutes).minutes
+        end
+
+        def logger
+          @logger ||= Steno.logger('cc.background')
+        end
+
+        def batch_size
```
philippthun: This could be a constant: BATCH_SIZE.

app/jobs/runtime/delayed_jobs_recover.rb
```diff
@@ -0,0 +1,90 @@
+module VCAP::CloudController
+  module Jobs
+    module Runtime
+      class DelayedJobsRecover < VCAP::CloudController::Jobs::CCJob
+        def perform
+          logger.info('Recover halted delayed jobs')
```
philippthun: I would prefer: "Recovering stuck delayed jobs"

app/jobs/runtime/delayed_jobs_recover.rb
```diff
@@ -0,0 +1,90 @@
...
+          #   - delayed_jobs.failed_at IS NOT NULL: the delayed job permanently failed (exhausted max_attempts);
+          #     jobs still alive or locked have failed_at=NULL and must not be touched
+          cutoff_time = Time.now - default_maximum_duration_seconds
+          stuck = ServiceInstanceOperation.
+                  join(:service_instances, id: Sequel[:service_instance_operations][:service_instance_id]).
+                  join(:jobs, resource_guid: Sequel[:service_instances][:guid]).
+                  join(:delayed_jobs, guid: Sequel[:jobs][:delayed_job_guid]).
+                  where(Sequel[:service_instance_operations][:state] => 'in progress').
+                  where(Sequel[:service_instance_operations][:type] => 'create').
+                  where { Sequel[:service_instance_operations][:created_at] > cutoff_time }.
```
philippthun: You can use something similar to `FailedJobsCleanup` (only DB time is used):
```
where(Sequel.lit("#{Sequel[:service_instance_operations][:created_at]} > CURRENT_TIMESTAMP - INTERVAL '?' SECOND", default_maximum_duration_seconds.to_i))
```

app/jobs/runtime/delayed_jobs_recover.rb
```diff
@@ -0,0 +1,90 @@
...
+      class DelayedJobsRecover < VCAP::CloudController::Jobs::CCJob
+        def perform
+          logger.info('Recover halted delayed jobs')
+          recover
+        end
+
+        def max_attempts
+          1
+        end
+
```
philippthun: Please also add a `job_name_in_configuration` method.

app/jobs/runtime/delayed_jobs_recover.rb
```diff
@@ -0,0 +1,90 @@
...
+                   for_update.first
+            return unless pjob
+
+            # bring the pollable job into the clean polling state
+            pjob.update(cf_api_error: nil, state: PollableJobModel::POLLING_STATE)
+
+            # unwrap the serialized handler and re-enqueue via the reoccurring job's enqueue_next_job method
+            inner_job = Jobs::Enqueuer.unwrap_job(delayed.payload_object)
+            inner_job.send(:enqueue_next_job, pjob)
+          end
```
philippthun: I think it would be good to add an explicit log for every re-enqueued job.

app/jobs/runtime/delayed_jobs_recover.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+
+        def default_maximum_duration_seconds
+          Config.config.get(:broker_client_max_async_poll_duration_minutes).minutes
+        end
+
+        def logger
+          @logger ||= Steno.logger('cc.background')
```
philippthun: The logger should be more specific, e.g. `cc.background.delayed-jobs-recover`.

app/jobs/runtime/delayed_jobs_recover.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+
+        def reenqueue(pollable_guid, delayed)
+          # re-verify atomically that the pollable job still points to this dead delayed_job.
+          # if another process already re-enqueued a new job, pollable.delayed_job_guid was
+          # updated to the new delayed_job's guid, so where clause returns nil and we skip safely.
+          PollableJobModel.db.transaction do
+            pjob = PollableJobModel.where(guid: pollable_guid,
+                                          delayed_job_guid: delayed.guid).
+                   for_update.first
```
philippthun: Maybe add `skip_locked` - so if two processes try to lock the same row, the first one succeeds and the second one does nothing.

app/jobs/runtime/delayed_jobs_recover.rb
```diff
@@ -0,0 +1,90 @@
...
+          #     jobs still alive or locked have failed_at=NULL and must not be touched
+          cutoff_time = Time.now - default_maximum_duration_seconds
+          stuck = ServiceInstanceOperation.
+                  join(:service_instances, id: Sequel[:service_instance_operations][:service_instance_id]).
+                  join(:jobs, resource_guid: Sequel[:service_instances][:guid]).
+                  join(:delayed_jobs, guid: Sequel[:jobs][:delayed_job_guid]).
+                  where(Sequel[:service_instance_operations][:state] => 'in progress').
+                  where(Sequel[:service_instance_operations][:type] => 'create').
+                  where { Sequel[:service_instance_operations][:created_at] > cutoff_time }.
+                  where(Sequel[:jobs][:state] => [PollableJobModel::POLLING_STATE, PollableJobModel::FAILED_STATE]).
```
philippthun: After giving this some additional thoughts, I think that we should not take `failed` jobs into account here. This is a terminal state that we already exposed to the user - bringing a "dead job back to life" would violate user expectations.

What we might want to have in addition is triggering orphan mitigation for the combination pollable job is `failed` and service instance operation is `in progress`.

philippthun:
When focusing on failed jobs where the pollable job is still `POLLING`, this PR could be extended to all async operations:
```
service_instance.create, service_instance.update, service_instance.delete
service_binding.create, service_binding.delete
service_key.create, service_key.delete
service_route_binding.create, service_route_binding.delete
```

spec/migrations/20260505071445_add_jobs_operation_state_index_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        # Test down migration idempotency
+        expect { Sequel::Migrator.run(db, migrations_path, target: current_migration_index - 1, allow_missing_migration_files: true) }.not_to raise_error
+        expect(operation_state_partial_index_present).to be_falsey
+
+      elsif db.database_type == :mysql
+        # Test up migration
+        expect(db.indexes(:jobs)).not_to include(:jobs_operation_state_index)
+        expect { Sequel::Migrator.run(db, migrations_path, target: current_migration_index, allow_missing_migration_files: true) }.not_to raise_error
+        expect(db.indexes(:jobs)).to include(:jobs_operation_state_index)
```
johha: This should also work for postgres. Then you could get rid of the if.

service_operations_create_in_progress_cleanup
```diff
@@ -0,0 +1,194 @@
...
+        allow(Steno).to receive(:logger).and_return(fake_logger)
+        TestConfig.override(broker_client_max_async_poll_duration_minutes: max_poll_duration_minutes)
+      end
+
+      # Builds a fully stuck scenario that the job should pick up and re-enqueue by default.
+      # All filter conditions are satisfied: sio is in progress/create/within cutoff,
+      # pjob is FAILED with operation=service_instance.create, delayed_job has failed_at set.
+      # Override individual parameters to break a single filter and test exclusion.
+      def make_stuck_scenario(
+        sio_state: 'in progress',
```
johha: Very Minor: I prefer explicit names like `service_instance_state`.

app/jobs/runtime/delayed_jobs_recover.rb
```diff
@@ -0,0 +1,90 @@
+module VCAP::CloudController
+  module Jobs
+    module Runtime
+      class DelayedJobsRecover < VCAP::CloudController::Jobs::CCJob
```
johha: How about calling it `DelayedServiceJobsRecover`? Something to make clear that this is only for service related jobs.

service_operations_create_in_progress_cleanup
```diff
@@ -0,0 +1,194 @@
...
+      before do
+        allow(Steno).to receive(:logger).and_return(fake_logger)
+        TestConfig.override(broker_client_max_async_poll_duration_minutes: max_poll_duration_minutes)
+      end
+
+      # Builds a fully stuck scenario that the job should pick up and re-enqueue by default.
+      # All filter conditions are satisfied: sio is in progress/create/within cutoff,
+      # pjob is FAILED with operation=service_instance.create, delayed_job has failed_at set.
+      # Override individual parameters to break a single filter and test exclusion.
+      def make_stuck_scenario(
```
johha: Minor:  maybe `prepare_stuck_service_instance` fits better?

johha (changes requested):

app/jobs/runtime/service_operations_create_in_progress_cleanup.rb
```diff
@@ -0,0 +1,117 @@
+module VCAP::CloudController
+  module Jobs
+    module Runtime
+      class ServiceOperationsCreateInProgressCleanup < VCAP::CloudController::Jobs::CCJob
```
johha, line 4: Guess we need a new name to reflect also bindings and keys etc

app/jobs/runtime/service_operations_create_in_progress_cleanup.rb
```diff
@@ -0,0 +1,117 @@
...
+        BATCH_SIZE = 10
+
+        def perform
+          logger.info("Cleaning up service 'create' operations stuck in 'in progress'")
+          cleanup_operations(ServiceInstanceOperation, ServiceInstance, :service_instance_id, 'service_instance.create',      :cleanup_failed_provision)
```
johha, line 9: Maybe add the method names for better readability

app/jobs/runtime/service_operations_create_in_progress_cleanup.rb
```diff
@@ -0,0 +1,117 @@
...
+        def orphan_mitigator
+          @orphan_mitigator ||= VCAP::Services::ServiceBrokers::V2::OrphanMitigator.new
+        end
+
+        def default_maximum_duration_seconds
+          Config.config.get(:broker_client_max_async_poll_duration_minutes).minutes
+        end
+
+        def logger
+          @logger ||= Steno.logger('cc.background.service-operations-create-in-progress-cleanup')
```
johha, line 108: also adjust name here, maybe similar to class name?

spec/unit/jobs/runtime/service_operations_create_in_progress_cleanup.rb
```diff
@@ -0,0 +1,258 @@
...
+            expect(fake_mitigator).not_to have_received(:cleanup_failed_provision)
+          end
+        end
+
+        context 'when sio created_at is beyond the max polling window' do
+          it 'does not mitigate' do
+            scenario = prepare_stuck_service_instance(service_instance_created_at: Time.now - (max_poll_duration_minutes + 1).minutes)
+            job.perform
+            expect(scenario[:service_instance].last_operation.reload.state).to eq('in progress')
+            expect(fake_mitigator).not_to have_received(:cleanup_failed_provision)
```
johha: Minor: Could be simplified with a shared example for all `not_to have_received(:cleanup_failed_provision)` tests

johha (changes requested):

config/cloud_controller.yml
```diff
@@ -55,6 +55,9 @@ pollable_jobs:
 service_operations_initial_cleanup:
   frequency_in_seconds: 300
 
+service_operations_create_in_progress_cleanup:
+  frequency_in_seconds: 300
```
jochenehret: The broker_client_max_async_poll_duration default is 7 days. A job interval of 5 minutes seems to be very frequent. Would a longer frequency make sense (1 hour?)?

jochenehret (commented):
Looks good, just consider a lower job frequency.

app/jobs/runtime/service_operations_create_in_progress_cleanup.rb
serdarozerr: we have discussed and will stick to current name.

app/jobs/runtime/service_operations_create_in_progress_cleanup.rb
serdarozerr: we have discussed and will stick to current name.

app/jobs/runtime/service_operations_create_in_progress_cleanup.rb
serdarozerr: we have discussed and will stick to current name.

johha (approved):

jochenehret (approved):

PR #5045, opened 2026-04-20
Fix sporadic test failure in service broker deletion during creation

philippthun:
The test was flaky because both create and delete jobs could be enqueued with the same timestamp, making pickup order non-deterministic. When the delete job ran first, the create job failed with `ServiceBrokerGone`.

Fix by setting a higher priority number for the delete job so the create job always runs first.


johha (approved):
LGTM

PR #5065, opened 2026-04-28
Add lifecycle_type column to apps, droplets, and builds

philippthun:
Store `lifecycle_type` directly on the model instead of deriving it from associated lifecycle data records. This simplifies queries and makes the lifecycle type explicit.

- Add `lifecycle_type` column to apps, droplets, and builds tables
- Set `lifecycle_type` in create actions (`AppCreate`, `DropletCreate`)
- Add `before_create` hooks to builds/droplets to inherit from app
- Add model validation for `lifecycle_type` values
- Assign `lifecycle_data` back to model in lifecycle classes to ensure the in-memory association is set (avoids Sequel caching issues)
- Add lifecycle type validation to `DropletCopy`
- Update blueprints to set `lifecycle_type` explicitly
- Fix shared example `updated_at` filtering to use `allow_manual_update` and explicit update instead of `update_on_create: false`
- Replace inline timestamp workarounds with shared example usage in apps, droplets, and revisions request specs

### DIFF
--- a/app/models/runtime/build_model.rb
+++ b/app/models/runtime/build_model.rb
@@ -50,24 +50,40 @@ module VCAP::CloudController
     add_association_dependencies labels: :destroy
     add_association_dependencies annotations: :destroy
 
-    def lifecycle_type
-      return Lifecycles::BUILDPACK if buildpack_lifecycle_data
-      return Lifecycles::CNB if cnb_lifecycle_data
+    def validate
+      super
+      validates_includes Lifecycles::TYPES, :lifecycle_type
+    end
 
-      Lifecycles::DOCKER
+    def before_create
+      # Inherit lifecycle_type from associated app if not explicitly set
+      self[:lifecycle_type] = app&.lifecycle_type if self[:lifecycle_type].blank?
+
+      super
+    end
+
+    def lifecycle_type
+      return self[:lifecycle_type] if self[:lifecycle_type].present?
+
+      # Fallback for records written before the lifecycle_type column
+      # existed. Remove once existing rows are backfilled (see #5067).
+      return BuildpackLifecycleDataModel::LIFECYCLE_TYPE if buildpack_lifecycle_data
+      return CNBLifecycleDataModel::LIFECYCLE_TYPE if cnb_lifecycle_data
+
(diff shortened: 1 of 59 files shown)


app/models/runtime/build_model.rb
```diff
@@ -49,29 +45,44 @@ class BuildModel < Sequel::Model(:builds)
...
 
-    add_association_dependencies buildpack_lifecycle_data: :destroy, kpack_lifecycle_data: :destroy, cnb_lifecycle_data: :destroy
+    add_association_dependencies buildpack_lifecycle_data: :destroy, cnb_lifecycle_data: :destroy
 
     add_association_dependencies labels: :destroy
     add_association_dependencies annotations: :destroy
 
+    def validate
+      super
+      validates_includes %w[buildpack cnb docker], :lifecycle_type
```
johha: `%w[buildpack cnb docker]` should be a constant as it's used multiple times in various files

app/models/runtime/build_model.rb
```diff
@@ -49,29 +45,44 @@ class BuildModel < Sequel::Model(:builds)
...
+
+      super
+    end
+
     def lifecycle_type
-      return Lifecycles::BUILDPACK if buildpack_lifecycle_data
-      return Lifecycles::CNB if cnb_lifecycle_data
+      return self[:lifecycle_type] if self[:lifecycle_type].present?
+
+      # Fallback for records without lifecycle_type column value
```
johha: Maybe mention that this is to stay compatible with older schemas and will be removed at some point?

app/models/runtime/build_model.rb
philippthun: I've added a constant to [lib/cloud_controller/diego/lifecycles/lifecycles.rb](https://github.com/sap-contributions/cloud_controller_ng/blob/c25a9cfb74b04ee408d86de921fcaa50745b86d3/lib/cloud_controller/diego/lifecycles/lifecycles.rb#L6)

app/models/runtime/build_model.rb
philippthun: I've adjusted the comment in three places and linked to the ccng issue #5067.

johha (approved):

PR #5072, opened 2026-04-30
Add fog deprecation warning

johha:
Fog is deprecated and replaced by storage-cli, thus we should inform operators about the planned removal.
Further details in RFC: 
- https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0043-cc-blobstore-storage-cli.md
- https://github.com/cloudfoundry/community/issues/1274

### DIFF
--- a/lib/cloud_controller/blobstore/fog/fog_client.rb
+++ b/lib/cloud_controller/blobstore/fog/fog_client.rb
@@ -13,6 +13,9 @@ module CloudController
       attr_reader :root_dir
 
       DEFAULT_BATCH_SIZE = 1000
+      DEPRECATION_MESSAGE = '[DEPRECATION WARNING] The fog blobstore client is DEPRECATED and will be REMOVED. ' \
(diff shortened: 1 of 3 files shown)


lib/cloud_controller/blobstore/fog/fog_client.rb
```diff
@@ -30,6 +33,7 @@ def initialize(connection_config:,
         @max_size = max_size
         @aws_storage_options = aws_storage_options
         @gcp_storage_options = gcp_storage_options
+        logger.warn('blobstore.fog-deprecated', message: DEPRECATION_MESSAGE)
```
kathap, line 36: The warning logs on every FogClient instantiation. If the client is created frequently, this could spam logs.

johha: Operators should be aware of this change. If we only logs this during a restart people might only see this every couple of days and ignore it.

philippthun (approved):

PR #5079, opened 2026-05-04
Document case sensitivity of name fields in API docs

philippthun:
The API has inconsistent case sensitivity behavior across resources due to historical schema choices: some name columns use `citext` (case-insensitive) while others use plain `varchar` (case-sensitive). This was completely undocumented, causing confusion for API consumers.

Add (case-insensitive) or (case-sensitive) annotations to:
- Create endpoint name parameters, including uniqueness scope
- List endpoint name filter parameters
- Cross-resource name filters (e.g. `service_broker_names`)

Case-insensitive: apps, organizations, spaces, domains, isolation segments, organization quotas, service plans, service offerings, tasks

Case-sensitive: service instances, buildpacks, stacks, security groups, service brokers, space quotas, service credential bindings

Also remove the misspelled 'case_insenstive' option from three migrations (stacks, lockings, frameworks). The typo meant the option was silently ignored, so these columns were always case-sensitive. Removing the dead option makes the code reflect actual behavior.

### DIFF
--- a/db/migrations/20130131184954_new_initial_schema.rb
+++ b/db/migrations/20130131184954_new_initial_schema.rb
@@ -106,7 +106,7 @@ Sequel.migration do
     create_table :frameworks do
       VCAP::Migration.common(self)
 
-      String :name,        null: false, case_insenstive: true
+      String :name,        null: false
       String :description, null: false
       String :internal_info, null: false, size: 2048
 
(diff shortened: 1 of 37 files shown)


philippthun:
Backwards Compatibility Tests need this fix: #5085

johha (approved):

PR #5094, opened 2026-05-07
refactor: standardize permission layers

peterhaochen47:
- such that the codepaths involved become more aligned with the
following layer design/abstraction:
```
| Layer | Responsibility |
|-------|---------------|
| **Controller** | Asks `Permissions` (via `permission_queryer`) what the current user can see/do |
| **Permissions** | Translates "can user do X?" into role-set queries via `Membership` |
| **Membership** | Executes role-based database queries (which orgs/spaces does user belong to?) |
| **Models** | Data access, no RBAC logic |
```

### DIFF
--- a/lib/cloud_controller/permissions.rb
+++ b/lib/cloud_controller/permissions.rb
@@ -98,7 +98,7 @@ class VCAP::CloudController::Permissions
   end
 
   def is_org_manager?
-    VCAP::CloudController::OrganizationManager.where(user_id: @user.id).any?
+    membership.authorized_orgs_subquery(VCAP::CloudController::Membership::ORG_MANAGER).any?
   end
 
   def readable_org_guids
@@ -178,6 +178,10 @@ class VCAP::CloudController::Permissions
     can_read_globally? || membership.role_applies?(ROLES_FOR_SPACE_READING, space_id, org_id)
   end
 
+  def can_read_from_space_as_space_member?(space_id)
+    can_read_globally? || membership.role_applies?(SPACE_ROLES, space_id)
+  end
+
   def can_download_droplet?(space_id, org_id)
     can_read_globally? || membership.role_applies?(ROLES_FOR_DROPLET_DOWLOAD, space_id, org_id)
   end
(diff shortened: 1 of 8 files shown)


lib/cloud_controller/permissions.rb
```diff
@@ -98,7 +98,7 @@ def can_write_globally?
   end
 
   def is_org_manager?
-    VCAP::CloudController::OrganizationManager.where(user_id: @user.id).any?
+    membership.authorized_org_guids_subquery(VCAP::CloudController::Membership::ORG_MANAGER).any?
```
philippthun: Don't need the guid, so `authorized_orgs_subquery` would be simpler.

peterhaochen47: Addressed, also made an additional improvement, see message of new commit.

philippthun (commented):
Thanks for addressing the feedback. The second optimization is not needed as we are using the [any_not_empty](https://sequel.jeremyevans.net/rdoc-plugins/files/lib/sequel/extensions/any_not_empty_rb.html) extension: https://github.com/cloudfoundry/cloud_controller_ng/blob/b51602823dd52a0f2e423681753b77ccf89a4786/lib/cloud_controller/db.rb#L119

I don't care about using `!empty?` or `any?` in this file, but I think the commit message should be adapted as it might be misleading for others.

peterhaochen47:
> Thanks for addressing the feedback. The second optimization is not needed as we are using the [any_not_empty](https://sequel.jeremyevans.net/rdoc-plugins/files/lib/sequel/extensions/any_not_empty_rb.html) extension:
> 
> https://github.com/cloudfoundry/cloud_controller_ng/blob/b51602823dd52a0f2e423681753b77ccf89a4786/lib/cloud_controller/db.rb#L119
> 
> I don't care about using `!empty?` or `any?` in this file, but I think the commit message should be adapted as it might be misleading for others.

I see, updated, simply modified the first commit.

philippthun (approved):

PR #5098, opened 2026-05-12
Fix/speedup service brokers get endpoint

serdarozerr:
During get request to **service_brokers** endpoint, if service broker is associated with **space**, it fetch also space information, that leads to N+1 query problem. With this PR we add eager loading for :space association to load all the necessary space information with 1 query. 

### DIFF
--- a/app/fetchers/service_broker_list_fetcher.rb
+++ b/app/fetchers/service_broker_list_fetcher.rb
@@ -5,7 +5,7 @@ module VCAP::CloudController
   class ServiceBrokerListFetcher < BaseListFetcher
     class << self
       def fetch(message:, permitted_space_guids: nil, eager_loaded_associations: [])
-        dataset = ServiceBroker.dataset.eager(eager_loaded_associations).eager(:space)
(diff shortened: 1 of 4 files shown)


app/fetchers/service_broker_list_fetcher.rb
```diff
@@ -5,7 +5,7 @@ module VCAP::CloudController
   class ServiceBrokerListFetcher < BaseListFetcher
     class << self
       def fetch(message:, permitted_space_guids: nil, eager_loaded_associations: [])
-        dataset = ServiceBroker.dataset.eager(eager_loaded_associations)
+        dataset = ServiceBroker.dataset.eager(eager_loaded_associations).eager(:space)
```
philippthun: I think this should move into the `ServiceBrokerPresenter`, i.e.:
```
class << self
  # :labels and :annotations come from MetadataPresentationHelpers
  def associated_resources
    super + [:space]
  end
end
```

serdarozerr: fixed in adaf5e3

philippthun (approved):

PR #5121, opened 2026-05-23
Replace machinist with FactoryBot for test data setup

philippthun:
Per ADR 0015. machinist 1.0.6 has had no upstream activity since 2013 and machinist 2.0 is not a viable upgrade (no Sequel adapter, no Sham, non-persisting make). FactoryBot is actively maintained, has trivial Sequel integration, and provides equivalents for every pattern used by the existing blueprints.

Convert ~11k call sites from `Model.make` / `Model.make_unsaved` to `create(:foo)` / `build(:foo)`, replace ~130 blueprints with factory definitions under spec/support/factory_definitions/, and remove the machinist gem along with spec/support/fakes/blueprints.rb and spec/support/machinist_monkey_patch.rb.

Key technical decisions (see ADR 0015):
- Global `to_create { |i| i.save; i.refresh }` in spec/support/factories.rb. This matches machinist's Sequel adapter which both saved and refreshed; the refresh prevents stale-cache bugs whenever a factory's after-create hook mutates associations.
- Sham is preserved as a thin shim (spec/support/sham_shim.rb) that delegates `Sham.<name>` to `FactoryBot.generate(:sham_<name>)` sequences defined in spec/support/factories.rb. The shim mirrors the original `Sham.define` block 1:1 so existing call sites need no edits.
- Dynamic class -> factory-name conversion via `klass.name.demodulize.underscore.to_sym` is used in matchers and shared examples, so generic helpers continue to look up the right factory when given any model class.
- Named blueprints become traits: `Foo.blueprint(:bar)` turns into a trait `:bar` on the `:foo` factory; `Foo.make(:bar, x: 1)` becomes `create(:foo, :bar, x: 1)`.
- build replaces `make_unsaved` for the (rare) cases that wanted an unsaved instance.
- `:droplet_model` only auto-sets itself as the app's current droplet when no `app:` override is supplied (`set_as_current_droplet { app == :unset }`), matching machinist's blueprint where the default app block only ran in that case. Specs that previously relied on the auto-set side effect when passing `app:` explicitly call `app.update(droplet:)` just as the pre-migration versions did.
- `:revision_sidecar_process_type_model` builds its parent with the `:no_process_types` trait so `FactoryBot.lint` does not collide with the parent's after-create web row on the `(revision_sidecar_guid, type)` unique constraint.


Gemfile
```diff
@@ -63,7 +63,7 @@ end
 
 group :test do
   gem 'codeclimate-test-reporter', '>= 1.0.8', require: false
-  gem 'machinist', '~> 1.0.6'
+  gem 'factory_bot', '~> 6.5'
```
johha, line 61: Whats the reason for pinning the version?

spec/support/factories.rb
```diff
@@ -0,0 +1,80 @@
...
+FactoryBot.define do
+  to_create do |instance|
+    instance.save
+    instance.refresh
+  end
+
+  # Sequences must mirror BOTH the existing top-level sequences AND every name
+  # defined inside Sham.define in spec/support/fakes/blueprints.rb, because
+  # surviving blueprints reference them and the lint spec exercises both worlds
+  # during coexistence.
```
johha: Remove comment?

spec/support/factories.rb
```diff
@@ -0,0 +1,80 @@
...
+  sequence(:status)              { |_n| 'active' }
+  sequence(:token)               { |n| "token-#{n}" }
+  sequence(:type)                { |n| "type-#{n}" }
+  sequence(:unique_id)           { |n| "unique-id-#{n}" }
+  sequence(:url)                 { |n| "https://example.com/#{n}" }
+  sequence(:version)             { |n| "1.0.#{n}" }
+
+  # Sequences that mirror the original Sham.define block from
+  # spec/support/fakes/blueprints.rb byte-for-byte. The Sham compatibility
+  # shim in spec/support/sham_shim.rb delegates to these.
```
johha: Remove comment?

spec/support/factories.rb
```diff
@@ -0,0 +1,80 @@
...
+  sequence(:long_description)    { |n| "long-description-#{n}" }
+  sequence(:provider)            { |n| "provider-#{n}" }
+  sequence(:service_credentials) { |n| { "creds-key-#{n}" => "creds-val-#{n}" } }
+  sequence(:stack)               { |n| "stack-#{n}" }
+  sequence(:status)              { |_n| 'active' }
+  sequence(:token)               { |n| "token-#{n}" }
+  sequence(:type)                { |n| "type-#{n}" }
+  sequence(:unique_id)           { |n| "unique-id-#{n}" }
+  sequence(:url)                 { |n| "https://example.com/#{n}" }
+  sequence(:version)             { |n| "1.0.#{n}" }
```
johha: I'd prefer the old order from blueprints + the new stuff at the end. Makes the diff a bit nicer. Maybe also for the sham part and then highlight why/if values changed?

spec/support/factories.rb
```diff
@@ -0,0 +1,80 @@
...
+  # Sequences must mirror BOTH the existing top-level sequences AND every name
+  # defined inside Sham.define in spec/support/fakes/blueprints.rb, because
+  # surviving blueprints reference them and the lint spec exercises both worlds
+  # during coexistence.
+  sequence(:name)              { |n| "name-#{n}" }
+  sequence(:guid)              { |_n| SecureRandom.uuid.to_s }
+  sequence(:domain)            { |n| "domain-#{n}.example.com" }
+  sequence(:host)              { |n| "host-#{n}" }
+  sequence(:uaa_id)            { |n| "uaa-id-#{n}" }
+  sequence(:port)              { |n| n + 1024 }
```
johha: This used to be `n + 1000`. For this refactoring I'd not change that

spec/support/factories.rb
```diff
@@ -0,0 +1,80 @@
...
+  # spec/support/fakes/blueprints.rb byte-for-byte. The Sham compatibility
+  # shim in spec/support/sham_shim.rb delegates to these.
+  sequence(:sham_email)               { |n| "email-#{n}@somedomain.com" }
+  sequence(:sham_name)                { |n| "name-#{n}" }
+  sequence(:sham_label)               { |n| "label-#{n}" }
+  sequence(:sham_token)               { |n| "token-#{n}" }
+  sequence(:sham_auth_username)       { |n| "auth_username-#{n}" }
+  sequence(:sham_auth_password)       { |n| "auth_password-#{n}" }
+  sequence(:sham_provider)            { |n| "provider-#{n}" }
+  sequence(:sham_port)                { |n| n + 1024 }
```
johha: same as above

spec/support/factory_definitions/user.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :user, class: 'VCAP::CloudController::User' do
+    guid { Sham.uaa_id }
```
johha: rather `generate(:uaa_id)`?

Gemfile
philippthun: We also pin other test gems (rspec, rubocop). Dependabot will update it accordingly.

spec/support/factories.rb
philippthun: Done.

spec/support/factories.rb
philippthun: I kept a single line commit.

spec/support/factories.rb
philippthun: Reordered to follow the original Sham order; new sequences appended at the end.

spec/support/factories.rb
philippthun: Restored to `n + 1000`.

spec/support/factory_definitions/user.rb
philippthun: Switched to `generate(:uaa_id)`.

johha (approved):

PR #5137, opened 2026-06-01
speed up unit tests

johha:
Split migration specs into a dedicated parallel_rspec batch, cache
runtime logs across runs, and tune the Postgres/MySQL CI services for
non-durable test workloads. Trim hot-spot specs (RSA key reuse, smaller
factories, fake clocks) and switch to targeted per-test truncation.

### DIFF
--- a/spec/support/database_isolation.rb
+++ b/spec/support/database_isolation.rb
@@ -9,17 +9,30 @@ module DatabaseIsolation
   end
 
   # Sequel logger that records which tables an example wrote to, so we can truncate only those.
+  # Schema-qualified writes (e.g. INSERT INTO public.users) trigger a full truncate as a safe fallback.
   class WrittenTablesLogger
-    WRITE_REGEX = /\b(?:INSERT INTO|UPDATE|DELETE FROM|TRUNCATE TABLE|TRUNCATE)\s+[`"]?(\w+)/i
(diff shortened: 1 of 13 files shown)


spec/unit/lib/uaa/uaa_token_decoder_spec.rb
```diff
@@ -1,6 +1,9 @@
 require 'spec_helper'
 require 'cloud_controller/uaa/uaa_token_decoder'
 
+# Generate RSA-2048 keys once and reuse across examples; key randomness is irrelevant here.
+RSA_TEST_KEYS = Array.new(4) { OpenSSL::PKey::RSA.new(2048) }.freeze
```
philippthun: This constant should either go into a more specific namespace (i.e. declaration inside the `describe`) or be set in a `before(:all)`.

spec/support/database_isolation.rb
```diff
@@ -8,20 +8,47 @@ def self.choose(isolation, db)
     end
   end
 
+  # Sequel logger that records which tables an example wrote to, so we can truncate only those.
+  class WrittenTablesLogger
+    WRITE_REGEX = /\b(?:INSERT INTO|UPDATE|DELETE FROM|TRUNCATE TABLE|TRUNCATE)\s+[`"]?(\w+)/i
```
philippthun: This regex only works for non-qualified table names - which is what Sequel normally uses. If someone would issue an `INSERT INTO <schema>.<table>`, we would not clean this up correctly.

philippthun (approved):

PR #5149, opened 2026-06-05
Add admin_limit to cc.rate_limiter configuration

WeiQuan0605:

CF API rate limits currently exclude admin users entirely. This PR extends cc.rate_limiter with an admin_limit parameter (default -1 = no limit, preserving existing behavior) that when set to a positive value applies rate limiting to admin users using their own counter, independent of the general user limit.
  - Operators want to protect the CF API from abusive or runaway automation scripts running with admin credentials
  - Without this change, admin users can send unlimited requests and potentially overwhelm the API, while regular users are protected
  - With admin_limit: -1 (default), existing deployments are unaffected
Capi-release <https://github.com/cloudfoundry/capi-release/pull/655>

### DIFF
--- a/middleware/base_rate_limiter.rb
+++ b/middleware/base_rate_limiter.rb
@@ -141,10 +141,7 @@ module CloudFoundry
       end
 
       def exceeded_rate_limit(count, env)
-        limit = per_process_request_limit(env)
-        return false if limit == -1
-
-        count > limit
+        count > per_process_request_limit(env)
       end
 
       def estimate_remaining(env, new_count)
--- a/middleware/base_rate_limiter.rb
+++ b/middleware/base_rate_limiter.rb
@@ -141,7 +141,10 @@ module CloudFoundry
       end
 
       def exceeded_rate_limit(count, env)
-        count > per_process_request_limit(env)
+        limit = per_process_request_limit(env)
+        return false if limit == -1
+
+        count > limit
       end
 
       def estimate_remaining(env, new_count)
--- a/middleware/rate_limiter.rb
+++ b/middleware/rate_limiter.rb
@@ -19,7 +19,11 @@ module CloudFoundry
 
       def apply_rate_limiting?(env)
         request = ActionDispatch::Request.new(env)
-        !basic_auth?(env) && !internal_api?(request) && !root_api?(request) && per_process_request_limit(env) != -1
+        return false if basic_auth?(env)
+        return false if internal_api?(request) || root_api?(request)
+        return false if per_process_request_limit(env) < 0
+
+        true
       end
 
       def root_api?(request)
--- a/middleware/rate_limiter.rb
+++ b/middleware/rate_limiter.rb
@@ -10,6 +10,8 @@ module CloudFoundry
         @global_general_limit              = opts[:global_general_limit]
         @per_process_unauthenticated_limit = opts[:per_process_unauthenticated_limit]
         @global_unauthenticated_limit      = opts[:global_unauthenticated_limit]
+        @per_process_admin_limit           = opts[:per_process_admin_limit] || -1
+        @global_admin_limit                = opts[:global_admin_limit] || -1
         super(app, opts[:logger], EXPIRING_REQUEST_COUNTER, opts[:interval])
       end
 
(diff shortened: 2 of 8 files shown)


spec/unit/lib/cloud_controller/rack_app_builder_spec.rb
```diff
@@ -68,6 +68,8 @@ module VCAP::CloudController
               global_general_limit: 1230,
               per_process_unauthenticated_limit: 1,
               global_unauthenticated_limit: 10,
+              per_process_admin_limit: nil,
+              global_admin_limit: nil,
```
johha: Rather set this to `-1` or some other integer to represent the correct type

spec/request/rate_limit_spec.rb
```diff
@@ -73,4 +73,57 @@
       expect(parsed_response['errors'].first['detail']).to include('Rate Limit Exceeded: Unauthenticated requests from this IP address have exceeded the limit')
     end
   end
+
+  context 'as an admin' do
+    let(:admin_headers) { admin_headers_for(VCAP::CloudController::User.make) }
+
+    context 'when admin_limit is -1 (default, unlimited)' do
+      it 'is not rate limited' do
+        20.times do |n|
```
johha: It should be sufficient to do this get once as the rate limiter is skipped in that case and there will be no `X-RateLimit-Limit` header

serdarozerr (changes requested):
2 new parameters are introduced:   **per_process_admin_limit**,  **global_admin_limit** . The problem is during the calculation remaining rate limit, these new parameters are not used, so for the admin user it will still use the **per_process_request_limit** not the **per_process_admin_limit** .  Check this function to see how the reaming rate limit is calculated https://github.com/cloudfoundry/cloud_controller_ng/blob/main/middleware/base_rate_limiter.rb#L147

middleware/base_rate_limiter.rb
```diff
@@ -141,7 +141,10 @@ def per_process_request_limit(_env)
       end
 
       def exceeded_rate_limit(count, env)
-        count > per_process_request_limit(env)
+        limit = per_process_request_limit(env)
+        return false if limit == -1
```
johha: This would also be considered for other rate limit configurations like the general limit. Maybe this is an indicator that we should allow `-1` for all rate limit config parameters and disable it in  `RateLimiter::apply_rate_limiting`

johha (changes requested):

WeiQuan0605:
> 2 new parameters are introduced: **per_process_admin_limit**, **global_admin_limit** . The problem is during the calculation remaining rate limit, these new parameters are not used, so for the admin user it will still use the **per_process_request_limit** not the **per_process_admin_limit** . Check this function to see how the reaming rate limit is calculated https://github.com/cloudfoundry/cloud_controller_ng/blob/main/middleware/base_rate_limiter.rb#L147

Hi @serdarozerr, thanks for the review! The admin limits are actually correctly used in `estimate_remaining.` Both `per_process_request_limit(env)` and `global_request_limit(env`) already dispatch to the admin-specific values when the user is an admin. So `estimate_remaining` in base_rate_limiter.rb indirectly uses the admin limits through these methods

middleware/rate_limiter.rb
```diff
@@ -10,14 +10,16 @@ def initialize(app, opts)
...
+        @global_admin_limit                = opts[:global_admin_limit] || -1
         super(app, opts[:logger], EXPIRING_REQUEST_COUNTER, opts[:interval])
       end
 
       private
 
       def apply_rate_limiting?(env)
         request = ActionDispatch::Request.new(env)
-        !basic_auth?(env) && !internal_api?(request) && !root_api?(request) && !admin?
+        !basic_auth?(env) && !internal_api?(request) && !root_api?(request) && per_process_request_limit(env) != -1
```
johha: Two small things:
- Prefer `< 0` over `!= -1`. If an operator typos -2 `< 0` fails open as "unlimited" rather than 429-ing every request. 
- Worth splitting the boolean chain into guard clauses as each exemption stands on its own, easier to read + easier to add new conditions  later


```ruby
    return false if basic_auth?(env)
    return false if internal_api?(request) || root_api?(request)
    return false if per_process_request_limit(env) < 0

    true
```

johha (changes requested):

middleware/rate_limiter.rb
WeiQuan0605: This make sense

johha (approved):
LGTM

PR #5157, opened 2026-06-08
Fix v3/roles perf regression for regular users

johha:
## Summary

Fixes a ~2x perf regression on `GET /v3/roles` (as regular user) introduced by #5094.
Related performance tests results of capi release v1.236.0:
https://github.com/cloudfoundry/cf-performance-tests-pipeline/blob/main/results/rails/postgres/charts/roles-test-results/v1/roles-detailed-chart-with-most-recent-runs.png

Root cause: `Membership#authorized_spaces_subquery` returns `Space.where(id: ...).or(organization_id: ...)`, which Postgres compiles as a SubPlan re-evaluated once per branch of the `Role` UNION ALL — 8x `Seq Scan onspaces`.

## Perf measurements

Local repro: 5k orgs / 20k spaces / 50k users / ~880k role rows. Target user has mixed org + space role memberships. Benchmark calls `RolesController#readable_roles` directly via `bin/console` and runs `EXPLAIN ANALYZE`.

| State | Cost | Exec | Buffers |
|---|---|---|---|
| 1.236.0 (regression) | 133401 | 18.0 ms | 2369 |
| #5094 reverted (1.235.0) | 131804 | 4.6 ms | 532 |
| **This PR** | **131804** | **4.6 ms** | **532** |

Plan and buffer counts match the reverted state exactly.

## Fix

- `membership.rb`: rewrite `authorized_spaces_subquery` / `authorized_orgs_subquery`
    to push the OR inside a single `id IN (UNION)`. Same return contract; planner
    can now flatten. Benefits every caller of `readable_*_query`.
- `membership.rb`: add `authorized_space_ids_subquery` / `authorized_org_ids_subquery` + flat id-only UNIONs without the entity wrapper. Used by callers that filter
    by a raw `space_id` / `organization_id` FK column (avoids a redundant subplan over `spaces` / `organizations`).
- `permissions.rb`: expose `readable_space_ids_query` / `readable_org_ids_query`.
- `roles_controller.rb`: `readable_roles` uses the id-only queries.

### DIFF
--- a/lib/cloud_controller/membership.rb
+++ b/lib/cloud_controller/membership.rb
@@ -29,14 +29,7 @@ module VCAP::CloudController
     end
 
     def authorized_orgs_subquery(roles)
-      space_ids_query = member_space_ids(roles)
-      org_ids_from_space_roles = Space.where(id: space_ids_query).select(:organization_id) if space_ids_query
-
-      if org_ids_from_space_roles
-        Organization.where(id: member_org_ids(roles)).or(id: org_ids_from_space_roles)
-      else
-        Organization.where(id: member_org_ids(roles))
-      end
+      Organization.where(id: authorized_org_ids_subquery(roles))
     end
 
     def authorized_space_guids(roles)
@@ -48,11 +41,36 @@ module VCAP::CloudController
     end
 
     def authorized_spaces_subquery(roles)
-      org_ids = member_org_ids(roles)
-      if org_ids
-        Space.where(id: member_space_ids(roles)).or(organization_id: org_ids).select(:id, :guid)
+      Space.where(id: authorized_space_ids_subquery(roles)).select(:id, :guid)
+    end
+
+    # Like authorized_spaces_subquery but returns a flat id-only UNION (no Space.where wrapper).
+    # Use when the caller filters by a raw space_id FK column to avoid an extra subplan over `spaces`.
+    def authorized_space_ids_subquery(roles)
+      space_ids_query = member_space_ids(roles)
+      org_ids_query   = member_org_ids(roles)
+
+      space_ids_via_orgs = Space.where(organization_id: org_ids_query).select(:id) if org_ids_query
+
+      if space_ids_query && space_ids_via_orgs
+        space_ids_query.union(space_ids_via_orgs, from_self: false)
       else
-        Space.where(id: member_space_ids(roles)).select(:id, :guid)
+        space_ids_query || space_ids_via_orgs
+      end
+    end
+
+    # Like authorized_orgs_subquery but returns a flat id-only UNION (no Organization.where wrapper).
+    # Use when the caller filters by a raw organization_id FK column.
+    def authorized_org_ids_subquery(roles)
+      space_ids_query = member_space_ids(roles)
+      org_ids_query   = member_org_ids(roles)
+
+      org_ids_via_space_roles = Space.where(id: space_ids_query).select(:organization_id) if space_ids_query
+
+      if org_ids_query && org_ids_via_space_roles
+        org_ids_query.union(org_ids_via_space_roles, from_self: false)
+      else
+        org_ids_query || org_ids_via_space_roles
       end
     end
 
--- a/lib/cloud_controller/membership.rb
+++ b/lib/cloud_controller/membership.rb
@@ -32,6 +32,19 @@ module VCAP::CloudController
       Organization.where(id: authorized_org_ids_subquery(roles))
     end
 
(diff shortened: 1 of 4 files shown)


lib/cloud_controller/permissions.rb
```diff
@@ -168,6 +177,13 @@ def readable_spaces_query
     membership.authorized_spaces_subquery(ROLES_FOR_SPACE_READING)
   end
 
+  # Id-only variant of readable_spaces_query for callers filtering by a raw space_id FK column.
```
philippthun: I think this comment does not add any real value. The method name already explains what it returns.

lib/cloud_controller/permissions.rb
```diff
@@ -123,6 +123,15 @@ def readable_orgs_query
     end
   end
 
+  # Id-only variant of readable_orgs_query for callers filtering by a raw organization_id FK column.
```
philippthun: I think this comment does not add any real value. The method name already explains what it returns.

lib/cloud_controller/permissions.rb
```diff
@@ -123,6 +123,15 @@ def readable_orgs_query
...
+  def readable_org_ids_query
+    if can_read_globally?
+      VCAP::CloudController::Organization.dataset.select(:id)
```
philippthun: Although unrelated, could you also change `Organization.select` to `Organization.dataset.select` and `Space.select` to `Space.dataset.select` in this file - so that this is consistent.

lib/cloud_controller/membership.rb
```diff
@@ -48,11 +41,36 @@ def authorized_space_guids_subquery(roles)
...
 
     def authorized_spaces_subquery(roles)
-      org_ids = member_org_ids(roles)
-      if org_ids
-        Space.where(id: member_space_ids(roles)).or(organization_id: org_ids).select(:id, :guid)
+      Space.where(id: authorized_space_ids_subquery(roles)).select(:id, :guid)
+    end
+
+    # Like authorized_spaces_subquery but returns a flat id-only UNION (no Space.where wrapper).
+    # Use when the caller filters by a raw space_id FK column to avoid an extra subplan over `spaces`.
```
philippthun: I think this comment does not add any real value. The method name already explains what it returns.

lib/cloud_controller/membership.rb
```diff
@@ -48,11 +41,36 @@ def authorized_space_guids_subquery(roles)
...
+
+      if space_ids_query && space_ids_via_orgs
+        space_ids_query.union(space_ids_via_orgs, from_self: false)
+      else
+        space_ids_query || space_ids_via_orgs
+      end
+    end
+
+    # Like authorized_orgs_subquery but returns a flat id-only UNION (no Organization.where wrapper).
+    # Use when the caller filters by a raw organization_id FK column.
```
philippthun: I think this comment does not add any real value. The method name already explains what it returns.

lib/cloud_controller/membership.rb
```diff
@@ -48,11 +41,36 @@ def authorized_space_guids_subquery(roles)
...
+      Space.where(id: authorized_space_ids_subquery(roles)).select(:id, :guid)
+    end
+
+    # Like authorized_spaces_subquery but returns a flat id-only UNION (no Space.where wrapper).
+    # Use when the caller filters by a raw space_id FK column to avoid an extra subplan over `spaces`.
+    def authorized_space_ids_subquery(roles)
+      space_ids_query = member_space_ids(roles)
+      org_ids_query   = member_org_ids(roles)
+
+      space_ids_via_orgs = Space.where(organization_id: org_ids_query).select(:id) if org_ids_query
```
philippthun: Would `space_ids_via_org_roles` be a better name?

lib/cloud_controller/membership.rb
```diff
@@ -48,11 +41,36 @@ def authorized_space_guids_subquery(roles)
...
+      if space_ids_query && space_ids_via_orgs
+        space_ids_query.union(space_ids_via_orgs, from_self: false)
+      else
+        space_ids_query || space_ids_via_orgs
+      end
+    end
+
+    # Like authorized_orgs_subquery but returns a flat id-only UNION (no Organization.where wrapper).
+    # Use when the caller filters by a raw organization_id FK column.
+    def authorized_org_ids_subquery(roles)
```
philippthun: For consistent ordering this method should go after `def authorized_orgs_subquery(roles)`.

philippthun (commented):
Mostly cosmetic...

philippthun (approved):

PR #5196, opened 2026-06-16
Add lifecycle_type backfill clock job and rake task

philippthun:
A one-off clock job that backfills the `lifecycle_type` column on apps, droplets, and builds for rows that pre-date the column's introduction. Once all installations have run it long enough to drain those rows, this job will be removed.

The job processes up to 1000 (`BATCH_SIZE`) * 10 (`BATCHES_PER_RUN`) rows per table in short transactions, idempotent and bounded. It is defensive against missing columns: `schema(:table)` is checked on each invocation, so it would be safe to be shipped together with the migration that introduces the column.

Operators can also run `rake db:lifecycle_type_backfill[batch_size,batches_per_run]` manually; passing -1 for `batches_per_run` drains until no NULL rows remain.


johha (commented):
It might be nice if operators have the option run the backfill as rake task manually.

johha (approved):
LGTM

PR #5202, opened 2026-06-18
Fix readiness endpoint reporting OK when request backlog exceeds threads

svkrieger:
The all_workers_busy? check compared busy_threads == running, which is false when there is a request backlog: Puma inflates busy_threads to include queued requests (busy_threads = spawned - waiting + todo.size), so a fully saturated pool with queued work would have busy_threads > running and be incorrectly reported as not busy.

Fix by using pool_capacity == 0 instead, which is the correct signal that all threads are occupied regardless of queue depth.

### DIFF
--- a/lib/cloud_controller/api_metrics_webserver.rb
+++ b/lib/cloud_controller/api_metrics_webserver.rb
@@ -57,9 +57,23 @@ module VCAP
 
         unhealthy = determine_unhealthy_state(all_busy)
 
-        build_status_response(all_busy, unhealthy)
+        response = build_status_response(all_busy, unhealthy)
+
+        seconds_since_increase = @last_requests_count_increase_time ? (Time.now - @last_requests_count_increase_time).round(2) : nil
+        worker_stats = worker_statuses&.map do |w|
+          ls = w[:last_status] || {}
+          "pool_capacity=#{ls[:pool_capacity]} requests_count=#{ls[:requests_count]}"
+        end&.join(', ')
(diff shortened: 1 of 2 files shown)


lib/cloud_controller/api_metrics_webserver.rb
```diff
@@ -57,11 +57,25 @@ def status
...
 
-        build_status_response(all_busy, unhealthy)
+        response = build_status_response(all_busy, unhealthy)
+
+        seconds_since_increase = @last_requests_count_increase_time ? (Time.now - @last_requests_count_increase_time).round(2) : nil
+        worker_stats = worker_statuses&.map do |w|
+          ls = w[:last_status] || {}
+          "pool_capacity=#{ls[:pool_capacity]} requests_count=#{ls[:requests_count]}"
+        end&.join(', ')
+        log("[readiness] workers=[#{worker_stats}] all_busy=#{all_busy} secs_since_last_increase=#{seconds_since_increase.inspect} unhealthy=#{unhealthy} -> #{response[2].first}")
```
stephanme, line 67: How much log noise will this generate, i.e. how often will the status endpoint be called?
Maybe we should log at debug level here.

stephanme (approved):
LGTM, only one question regarding log level.

jochenehret (approved):

lib/cloud_controller/api_metrics_webserver.rb
svkrieger: We call the endpoint every 10 seconds. That will be 6 log lines per API VM per minute. I think it's reasonable for info logging.

PR #5181, opened 2026-06-12
Restrict suspending orgs to admins (#5173)

johha:
Mutating the `suspended` field is now admin-only on:
 - POST /v3/organizations
 - PATCH /v3/organizations/:guid
 - POST /v2/organizations (status: 'suspended')
 - PUT /v2/organizations/:guid (status: 'suspended')

Non-admins receive 403; sending the current value (no-op echo) remains allowed so GET-modify-PATCH clients are not broken. Fixes issue #5173

* [x] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [x] I have viewed, signed, and submitted the Contributor License Agreement

* [x] I have made this pull request to the `main` branch

* [x] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)

Issue #5173, opened 2026-06-10
OrgManager can suspend their own org but cannot un-suspend it


johha:
### Issue

`PATCH /v3/organizations/:guid` with body `{"suspended": true}` succeeds when the caller is an OrgManager (not a Cloud Controller admin). After the org transitions to `suspended: true`, the same OrgManager is then blocked from issuing the inverse PATCH `{"suspended": false}` because the suspension gate (`CF-OrgSuspended`, code 10017) fires on every write to a suspended org, including the un-suspend itself.

The result is a one-way trap: any OrgManager can suspend their own organization, after which only a Cloud Controller admin can recover it. This appears to invert the intended trust model of the `suspended` flag (a platform/operator control) and almost certainly creates an unintended self-DoS path for tenants.

### Context

Observed against `cloud_controller_ng` API version `3.221.0` on the v3 endpoints. The v2 path (`PUT /v2/organizations/:guid` with `{"status": "suspended"}`) was not retested for this specific authority bug, so this report is scoped to v3.

### Steps to Reproduce

Pre-conditions: a CC admin and a regular UAA user `t-orgmgr` who holds the `OrgManager` role on the org under test (call its GUID `$ORG`). The org starts active (`suspended: false`).

1. As **CC admin**, confirm initial state:
   ```
   cf curl /v3/organizations/$ORG | jq '{suspended}'
   # {"suspended": false}
   ```
2. Authenticate as **t-orgmgr** (OrgManager on the org, no admin scope):
   ```
   cf auth t-orgmgr <password>
   ```
3. As **t-orgmgr**, suspend the org via the v3 PATCH:
   ```
   cf curl -X PATCH /v3/organizations/$ORG -d '{"suspended": true}'
   ```
   → Returns **200 OK** with `"suspended": true` in the body. No error, no authorization rejection.
4. Confirm the state took effect:
   ```
   cf curl /v3/organizations/$ORG | jq '{suspended}'
   # {"suspended": true}
   ```
5. Still as **t-orgmgr**, attempt to un-suspend:
   ```
   cf curl -X PATCH /v3/organizations/$ORG -d '{"suspended": false}'
   ```
   → Returns **403** with:
   ```json
   {
     "errors": [
       {"detail": "The organization is suspended", "title": "CF-OrgSuspended", "code": 10017}
     ]
   }
   ```
6. Recovery now requires a Cloud Controller admin to issue the un-suspend PATCH.

[...]

Issue #3589, opened 2024-01-09
Make recursive delete operations more intuitive


svkrieger:
## Issue

Currently recursive delete operations like org, space, app delete, which implicitly delete service bindings and/or service instances, will fail if one of the service related deletions are handled asynchronously by the service broker. This is not optimal as users have to trigger the deletion of the parent resource again. Also currently users will get the following error message, which does not really reveal what is going on:
`An operation for the service binding between app myapp and service instance myinstance is in progress.`

## Context

[provide more detailed introduction]

## Steps to Reproduce

For example delete an app, which is bound to a service instance. The broker should answer unbinding requests with a 202.

## Current result

The following tables describe the resulting behaviour for different resources and responses by the broker. In general all the recursive deletions fail if a sub-resource gets deleted asynchronously. The behaviour is the same for service instances and service bindings.

### Delete service instance when service binding present

|API call|resulting request to broker|broker response|current behaviour|
|---|---|---|---|
|DELETE /v3/service_instance/ <service_instance_guid>|DELETE /v2/service_instances/ <si_guid>/service_bindings/<sb_guid>|202|Starts polling service binding last operation and sets service instance delete job and service instance last operation to failed with message "delete could not be completed: An operation for the service binding between app myapp and service instance myinstance is in progress."|
|||500|Both the service instance and service binding delete operations fail|
|||200|Binding will be gone immediately, then a delete service instance request will be sent to the broker and either its gone too or polling starts|

### Delete app when bound to service

|API call|resulting request to broker|broker response|current behaviour|
|---|---|---|---|
|DELETE /v3/apps/<app_guid>|DELETE /v2/service_instances/ <si_guid>/service_bindings/<sb_guid>|202|Starts polling service binding last operation and sets app delete job to failed with message "Job (3d1d051f-6c94-47ab-85e2-dec27e0db75a) failed: An operation for the service binding between app myapp and service instance myinstance is in progress."|
|||500|The app delete job fails and the service binding operations state is set to failed. Error message: "Job (17662f61-4111-4712-91da-d3ab11629ba7) failed: Service broker failed to delete service binding for instance myinstance: The service broker returned an invalid response. Status Code: 500 Internal Server Error, Body: {"state":"in progress"}"|
|||200|Service binding gets deleted, app gets deleted and app delete job set to COMPLETE|

### Delete space which contains a service binding

|API call|resulting request to broker|broker response|current behaviour|
|---|---|---|---|
[...]

PR #5206, opened 2026-06-19
Add suspended state for spaces and deleting state for spaces and orgs

johha:
### A short explanation of the proposed change
 
- Adds a suspended state for spaces (v3-only), mirroring the existing org-suspended feature. While the writes blocked by space-suspended are now well-defined, OrgManagers act as the "admins of an org" and can toggle space suspended; SpaceManagers and below cannot. Admins continue to bypass.

- Also introduces internal deleting state plumbing for both spaces and orgs (no controller path flips it yet,  it's prep for an upcoming async-delete flow - see #3589 ). Surfaces a new CF-ResourceBeingDeleted (10020) error so the rest of the system can react to it.

- Branch is split into 4 small commits to make review easier:
    1. Schema + model + permission helpers
    1. Mechanical rename of is_*_active? guards → require_writable_*! across v3 controllers
    1. v3 feature wiring (messages, actions, presenter, errors, docs, request specs)
    1. v2 access classes aligned with the new state (v2 just respects state — the suspended-space feature itself is v3-only)
  
 - Operators want to lock down a space (block apps, routes, services from being modified) without deleting it - same use case that already exists at the org level. Today operators have to suspend the whole org or remove all roles manually. The deleting state plumbing supports a separate upcoming change for async deletion.

## Links to any other associated PRs:
 - Builds on #5173 / #5181 (org-side admin-only suspended fix). This PR mirrors that authority model on spaces, with OrgManager added as the space-scope equivalent of "org admin".
 - Pre-existing API quirks on the suspended field surfaced during black-box testing are tracked in #TODO (out of scope for this PR).
 
  - [x] I have reviewed the contributing guide
  - [x] I have viewed, signed, and submitted the Contributor License Agreement
  - [x] I have made this pull request to the main branch
  - [x] I have run all the unit tests using bundle exec rake
  - [ ] I have run CF Acceptance Tests


app/access/process_model_access.rb
```diff
@@ -54,7 +54,7 @@ def index_with_token?(_)
 
     def create?(app, _params=nil)
       return true if admin_user?
-      return false if app.in_suspended_org?
+      return false if app.in_suspended_or_deleting_org? || app.space&.suspended_or_deleting?
```
philippthun: Helper method? `app.in_suspended_or_deleting_space?`

app/access/route_access.rb
```diff
@@ -76,7 +76,7 @@ def index_with_token?(_)
 
     def can_write_to_route(route, is_create=false)
       return true if context.queryer.can_write_globally?
-      return false if route.in_suspended_org?
+      return false if route.in_suspended_or_deleting_org? || route.space.suspended_or_deleting?
```
philippthun: Helper method? `route.in_suspended_or_deleting_space?`

Why not `route.space&`?

app/access/service_instance_access.rb
```diff
@@ -56,14 +56,14 @@ def create?(service_instance, _params=nil)
       return true if admin_user?
 
       FeatureFlag.raise_unless_enabled!(:service_instance_creation)
-      return false if service_instance.in_suspended_org?
+      return false if service_instance.in_suspended_or_deleting_org? || service_instance.space&.suspended_or_deleting?
```
philippthun: Helper method? `service_instance.in_suspended_or_deleting_space?`

app/controllers/v3/application_controller.rb
```diff
@@ -150,6 +158,33 @@ def permission_queryer
...
+    when :deleting
+      resource_being_deleted!('organization')
+    when :suspended
+      suspended!
+    end
+  end
+
+  def require_writable_space!(space)
+    org_state = permission_queryer.writable_org_state(space.organization_id)
+    return resource_being_deleted!('organization') if org_state == :deleting
```
philippthun: These `return` statements are not needed. Do we use this pattern elsewhere? And without `return` we can skip the empty lines (maybe keep one before `permission_queryer.writable_space_state`).

app/controllers/v3/application_controller.rb
```diff
@@ -150,6 +158,33 @@ def permission_queryer
...
+
+  def require_writable_org_id!(org_id)
+    case permission_queryer.writable_org_state(org_id)
+    when :active
+      nil
+    when :deleting
+      resource_being_deleted!('organization')
+    when :suspended
+      suspended!
+    end
```
philippthun: This would look more compact to me. Does Rubocop complain then?
```
org_state = permission_queryer.writable_org_state(org_id)
resource_being_deleted!('organization') if org_state == :deleting
suspended! if org_state == :suspended
```

app/controllers/v3/builds_controller.rb
```diff
@@ -40,8 +40,8 @@ def create
 
     package = PackageModel.where(guid: message.package_guid).
               eager(:app, :space, space: :organization, app: %i[buildpack_lifecycle_data cnb_lifecycle_data]).first
-    unprocessable_package! unless package &&
-                                  permission_queryer.can_manage_apps_in_active_space?(package.space.id) && permission_queryer.is_space_active?(package.space.id)
+    unprocessable_package! unless package && permission_queryer.can_manage_apps_in_active_space?(package.space.id)
+    require_writable_space!(package.space) if package
```
philippthun: `if package` is not needed, this would already be an `unprocessable_package!`.

app/controllers/v3/roles_controller.rb
```diff
@@ -71,10 +71,10 @@ def destroy
 
     if role.for_space?
       unauthorized! unless permission_queryer.can_update_active_space?(role.space_id, role.space.organization_id)
-      suspended! unless permission_queryer.is_space_active?(role.space_id)
+      require_writable_space!(role.space)
     else
       unauthorized! unless permission_queryer.can_write_to_active_org?(role.organization_id)
-      suspended! unless permission_queryer.is_org_active?(role.organization_id)
+      require_writable_org!(role.organization)
```
philippthun: `require_writable_org_id!(role.organization_id)` should save a SELECT.

app/controllers/v3/space_quotas_controller.rb
```diff
@@ -64,7 +64,7 @@ def update
 
     unauthorized! unless permission_queryer.can_write_globally? ||
                          (space_quota && permission_queryer.can_write_to_active_org?(space_quota.organization_id))
-    suspended! unless space_quota && permission_queryer.is_org_active?(space_quota.organization_id)
+    require_writable_org!(space_quota.organization) if space_quota
```
philippthun: `require_writable_org_id!(space_quota.organization_id)`

app/controllers/v3/space_quotas_controller.rb
```diff
@@ -87,7 +87,7 @@ def apply_to_spaces
...
```
philippthun: `require_writable_org_id!(space_quota.organization_id)`

app/controllers/v3/space_quotas_controller.rb
```diff
@@ -112,7 +112,7 @@ def remove_from_space
...
```
philippthun: `require_writable_org_id!(space_quota.organization_id)`

app/controllers/v3/space_quotas_controller.rb
```diff
@@ -134,7 +134,7 @@ def destroy
...
```
philippthun: `require_writable_org_id!(space_quota.organization_id)`

app/controllers/v3/spaces_controller.rb
```diff
@@ -70,11 +71,13 @@ def update
     space = fetch_space(hashed_params[:guid])
     space_not_found! unless space && permission_queryer.can_read_from_space?(space.id, space.organization_id)
     unauthorized! unless permission_queryer.can_update_active_space?(space.id, space.organization_id)
-    suspended! unless permission_queryer.is_space_active?(space.id)
+    require_writable_space!(space)
 
     message = VCAP::CloudController::SpaceUpdateMessage.new(hashed_params[:body])
     unprocessable!(message.errors.full_messages) unless message.valid?
 
+    unauthorized! if suspended_by_unauthorized?(message, space.organization)
```
philippthun: Pass `organization_id` instead and change implementation of `suspended_by_unauthorized?`.

app/controllers/v3/spaces_controller.rb
```diff
@@ -86,7 +89,7 @@ def destroy
     space = fetch_space(hashed_params[:guid])
     space_not_found! unless space && permission_queryer.can_read_from_space?(space.id, space.organization_id)
     unauthorized! unless permission_queryer.can_write_to_active_org?(space.organization_id)
-    suspended! unless permission_queryer.is_org_active?(space.organization_id)
+    require_writable_org!(space.organization)
```
philippthun: `require_writable_org_id!(space.organization_id)`

app/models/services/service_instance.rb
```diff
@@ -175,8 +175,8 @@ def credentials_with_serialization
     alias_method 'credentials_without_serialization', 'credentials'
     alias_method 'credentials', 'credentials_with_serialization'
 
-    def in_suspended_org?
-      space&.in_suspended_org?
+    def in_suspended_or_deleting_org?
+      space&.in_suspended_or_deleting_org?
```
philippthun: `|| false` as in app/models/runtime/process_model.rb.

Other models use `delegate` - why is there a difference?

lib/cloud_controller/permissions.rb
```diff
@@ -149,21 +153,50 @@ def can_write_to_active_org?(org_id)
...
+
+  def space_state(space_id)
+    space = VCAP::CloudController::Space.where(id: space_id).select(:status, :organization_id).first
+    return :active if space.nil?
+
+    org = org_state(space.organization_id)
+    return :deleting if org == :deleting
+    return :deleting if space.status == VCAP::CloudController::Space::DELETING
+    return :suspended if org == :suspended
+    return :suspended if space.status == VCAP::CloudController::Space::SUSPENDED
```
philippthun:
```
return :deleting if org == :deleting || space.status == VCAP::CloudController::Space::DELETING
return :suspended if org ... || space.status ...
```

lib/cloud_controller/permissions.rb
```diff
@@ -149,21 +153,50 @@ def can_write_to_active_org?(org_id)
...
+    space = VCAP::CloudController::Space.where(id: space_id).select(:status, :organization_id).first
+    return :active if space.nil?
+
+    # Org state is never bypassed by org managers — a suspended/deleting org locks out everyone except admins
+    org = org_state(space.organization_id)
+    return :deleting if org == :deleting
+    return :suspended if org == :suspended
+
+    # Org managers can manage spaces in their active orgs even if the space itself is suspended/deleting
+    return :active if membership.role_applies?(VCAP::CloudController::Membership::ORG_MANAGER, nil, space.organization_id)
```
philippthun: Can use `is_org_manager_for_org?`.

lib/cloud_controller/permissions.rb
```diff
@@ -149,21 +153,50 @@ def can_write_to_active_org?(org_id)
...
-  def is_space_active?(space_id)
-    return true if can_write_globally? # admins can modify suspended orgs
+  def writable_org_state(org_id)
+    return :active if can_write_globally? # admins can modify suspended/deleting orgs
+
+    org_state(org_id)
+  end
+
+  def space_state(space_id)
+    space = VCAP::CloudController::Space.where(id: space_id).select(:status, :organization_id).first
```
philippthun, line 170: Maybe add a `_space_state` helper that returns the state as symbol (similar to `org_state`). As the query and the status checks are needed several times, this would make the code simpler (more readable), I guess.

spec/unit/models/helpers/org_space_status_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        expect(resource.active?).to be true
+
+        resource.status = 'suspended'
+        expect(resource.active?).to be false
+
+        resource.status = 'deleting'
+        expect(resource.active?).to be false
+
+        resource.status = nil
+        expect(resource.active?).to be false
```
philippthun: The db column is non-nullable, so not sure if this test adds any value.

spec/unit/messages/space_update_message_spec.rb
```diff
@@ -88,6 +88,23 @@ module VCAP::CloudController
           expect(message).to be_valid
         end
       end
+
+      describe 'suspended' do
+        it 'validates that it is a boolean' do
+          body = { suspended: 1 }
```
philippthun: Reuse `body` as in spec/unit/messages/spaces/space_create_message_spec.rb, e.g. `body[:suspended] = 1`?

spec/unit/messages/space_update_message_spec.rb
```diff
@@ -88,6 +88,23 @@ module VCAP::CloudController
...
+          message = SpaceUpdateMessage.new(body)
+
+          expect(message).not_to be_valid
+          expect(message.errors.full_messages).to include('Suspended must be a boolean')
+        end
+
+        it 'is not required' do
+          body = {}
```
philippthun: Reuse `body`?

app/actions/space_update.rb
```diff
@@ -11,6 +11,10 @@ def update(space, message)
       space.db.transaction do
         space.lock!
         space.name = message.name if message.requested?(:name)
+        if message.requested?(:suspended)
+          error!("Space '#{space.name}' is being deleted and cannot be reactivated.") if space.deleting? && message.suspended == false
```
philippthun: There is no check for setting a DELETING space to SUSPENDED. I think this should also be forbidden.

app/actions/organization_update.rb
```diff
@@ -15,6 +15,7 @@ def update(org, message)
         AnnotationsUpdate.update(org, message.annotations, OrganizationAnnotationModel)
 
         if message.requested?(:suspended)
+          error!("Organization '#{org.name}' is being deleted and cannot be reactivated.") if org.deleting? && message.suspended == false
```
philippthun: There is no check for setting a DELETING org to SUSPENDED. I think this should also be forbidden.

app/controllers/v3/application_controller.rb
```diff
@@ -150,6 +158,33 @@ def permission_queryer
...
+  end
+
+  def require_writable_space!(space)
+    org_state = permission_queryer.writable_org_state(space.organization_id)
+    return resource_being_deleted!('organization') if org_state == :deleting
+
+    space_state = permission_queryer.writable_space_state(space.id)
```
philippthun: Pass `org_state` as optional parameter, so that `writable_space_state` does not fetch it again. This reduces the total number of DB statements issued by `require_writable_space!` from 3 to 2.

db/migrations/20260522120000_add_status_to_spaces.rb
```diff
@@ -0,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table :spaces do
+      add_column :status, String, null: false, default: 'active', size: 255 unless @db.schema(:spaces).map(&:first).include?(:status)
```
philippthun: Claude suggests to omit the `null: false` as this could be expensive on MySQL and the `org.status` field is also nullable at the moment.

At a later point in time we could (should) mark the column in both tables as `NOT NULL` and also set a size (255) for `org.status`.

lib/cloud_controller/permissions.rb
```diff
@@ -149,21 +153,50 @@ def can_write_to_active_org?(org_id)
     membership.role_applies?(ROLES_FOR_ORG_WRITING, nil, org_id)
   end
 
-  def is_org_active?(org_id)
-    return true if can_write_globally? # admins can modify suspended orgs
+  def org_state(org_id)
+    status = VCAP::CloudController::Organization.where(id: org_id).get(:status)
+    return :active if status.nil?
```
philippthun, line 158: Does `:active` for a non-existing org (and space) make sense? What about returning `nil` instead?

app/controllers/runtime/route_mappings_controller.rb
```diff
@@ -34,7 +34,7 @@ def create
       raise CloudController::Errors::ApiError.new_from_details('RouteNotFound', request_attrs['route_guid']) unless route
       raise CloudController::Errors::ApiError.new_from_details('AppNotFound', request_attrs['app_guid']) unless process
       raise CloudController::Errors::ApiError.new_from_details('NotAuthorized') unless permissions.can_write_to_active_space?(process.space.id)
-      raise CloudController::Errors::ApiError.new_from_details('OrgSuspended') unless permissions.is_space_active?(process.space.id)
+      raise CloudController::Errors::ApiError.new_from_details('OrgSuspended') unless permissions.writable_space_state(process.space.id) == :active
```
philippthun, line 37: This raises `OrgSuspended` also for a suspended space or a deleting org/space. Not sure if we want to fix this for the v2 controllers (app/controllers/runtime/route_mappings_controller.rb and app/controllers/services/service_bindings_controller.rb)?

johha: My idea was to make only minimal changes to v2. If users want to use the suspended spaces (+async deletion later) they need to use v3 anyways.

lib/cloud_controller/permissions.rb
johha: Before it also returned `:active`, so I'd prefer to keep it as is

philippthun (approved):

PR #5210, opened 2026-06-19
Keep and backfill Usage Event records of running apps, tasks, and service instances

### LINKED ISSUE #4182: Problems related to pruning old AppUsageEvent and ServiceUsageEvent records

joyvuu-dave:
This PR fixes #4182.

Consumers of usage events had no way to find the current state of running apps, tasks, and service instances.

With this change, usage event records for running apps, tasks, and service instances are kept from being pruned during the normal cleanup job. A backfill also seeds a baseline event (`WAS_RUNNING` / `TASK_WAS_RUNNING`) for resources that are still running but whose start event was already pruned, so consumers can reconstruct the current state.

### Why a new state value

The baseline uses a new state value, `WAS_RUNNING` (`TASK_WAS_RUNNING` for tasks), rather than a plain `STARTED`. A plain `STARTED` would look exactly like a scale event: scaling a running process already writes `STARTED` with `previous_state = STARTED`, and purge-and-reseed writes the same pair. A consumer could not tell a baseline from a real start, and the baseline's `created_at` would read as the start time, which is the exact complaint in #4182 about reseed. The cost of the new value is that a consumer that checks states strictly has to learn it before the upgrade. We think that is the right trade: the point of the baseline is that consumers can see it is one.

Seeded baseline events are never deleted, because consumers may already have read them. If a resource stops while the backfill is running and its baseline is left without a matching ending event, the backfill adds the missing ending event (`STOPPED` / `DELETED` / `TASK_STOPPED`) instead. Two consequences for consumers, both documented on the V3 usage event resources: an added ending event carries the time of the repair rather than the exact stop time, so the interval it closes can run slightly long; and consumers should close an interval on the first ending event they see and ignore any duplicates. Task stop events are also only emitted when the task has a start event or baseline on record, so consumers never see a stop they cannot pair with a start. When that skip happens for a task that was running, Cloud Controller logs a warning, because a running task always had a start event and something must have deleted it; the warning names `rake db:was_running_backfill` as the repair.

### A note for reviewers: this is the codebase's first advisory lock

Only one backfill may run at a time. If two ran at once, both could see that the same ending event is missing, and both would write it. Because the backfill never deletes rows, the duplicate would be permanent. So the backfill takes a database advisory lock (`pg_advisory_lock` on Postgres, `GET_LOCK` on MySQL): the seed migrations wait for it, and the rake task fails fast with a clear message.

Here is why none of the existing locks would work.

The `lockings` row lock (used for buildpacks) ends when its transaction commits. The backfill commits many times, once per batch, so a second run could start between two batches.

The bigint backfill uses `FOR UPDATE SKIP LOCKED`. That lock lets two runs share the work. We need the opposite: only one run at a time.

Locket is a separate service that needs certificates and a running connection. A database migration cannot depend on that.

A session-scoped advisory lock holds across many transactions and frees itself if its session dies. Rails' migration runner guards against concurrent migrations the same way; Sequel just does not ship it.

Since this codebase has not used an advisory lock before, we would welcome maintainer judgment here.

### DIFF
--- a/app/jobs/runtime/events_cleanup.rb
+++ b/app/jobs/runtime/events_cleanup.rb
@@ -9,7 +9,7 @@ module VCAP::CloudController
         end
 
         def perform
-          Database::OldRecordCleanup.new(Event, cutoff_age_in_days).delete
+          Database::OldRecordCleanup.new(Event, cutoff_age_in_days:).delete
         end
 
         def job_name_in_configuration
--- a/app/models/runtime/app_usage_event.rb
+++ b/app/models/runtime/app_usage_event.rb
@@ -9,6 +9,23 @@ module VCAP::CloudController
                       :buildpack_guid, :buildpack_name,
                       :package_state, :previous_package_state, :parent_app_guid,
                       :parent_app_name, :process_type, :task_name, :task_guid
+
+    def self.usage_lifecycles
+      [
+        {
+          beginning_states: [ProcessModel::STARTED, Repositories::AppUsageEventRepository::WAS_RUNNING_EVENT_STATE],
+          ending_state: ProcessModel::STOPPED,
+          guid_column: :app_guid
+        },
+        {
+          beginning_states: [Repositories::AppUsageEventRepository::TASK_STARTED_EVENT_STATE,
+                             Repositories::AppUsageEventRepository::TASK_WAS_RUNNING_EVENT_STATE],
+          ending_state: Repositories::AppUsageEventRepository::TASK_STOPPED_EVENT_STATE,
+          guid_column: :task_guid
+        }
+      ].freeze
+    end
+
     AppUsageEvent.dataset_module do
       def supports_window_functions?
         false
--- a/app/models/runtime/task_model.rb
+++ b/app/models/runtime/task_model.rb
@@ -155,13 +155,38 @@ module VCAP::CloudController
         Repositories::AppUsageEventRepository::TASK_WAS_RUNNING_EVENT_STATE
       ]
       started = app_usage_repo.find_by_task_and_state(task: self, state: start_evidence_states)
-      return if started.nil?
+      if started.nil?
+        warn_if_running_task_has_no_start_record
+        return
+      end
 
       create_stop_event
     end
 
     def create_stop_event
       Repositories::AppUsageEventRepository.new.create_from_task(self, 'TASK_STOPPED')
+    end
+
+    # The skip above is silent. That is fine for a task that never ran. But a
+    # task can only reach RUNNING through the state change that also writes
+    # its TASK_STARTED event, in the same transaction. So a running task with
+    # no start record means something deleted the record later: the
+    # destructive v2 purge, or the events cleanup job of an older version
+    # that did not yet keep the start events of running tasks. Someone should
+    # find out which, so log a warning. We warn only for RUNNING because only
+    # RUNNING is certain: a task canceled while still PENDING also passes
+    # through CANCELING, so a CANCELING task may never have run.
+    def warn_if_running_task_has_no_start_record
+      state_before_stop = column_changed?(:state) ? initial_value(:state) : state
+      return unless state_before_stop == RUNNING_STATE
+
+      logger.warn("Not writing a TASK_STOPPED usage event for task #{guid}: the task was running but has no " \
+                  'TASK_STARTED or TASK_WAS_RUNNING usage event on record. This should not happen. Run ' \
+                  "'rake db:was_running_backfill' and check whether usage events were deleted or purged.")
+    end
+
+    def logger
+      @logger ||= Steno.logger('cc.models.task')
     end
   end
 end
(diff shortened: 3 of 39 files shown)


philippthun:
Heads-up: #5121 merged, replacing machinist + Sham with FactoryBot. This PR will need a rebase and a fixup - your diff adds around 29 `.make` / Sham lines. Mechanical conversions:

- AppUsageEvent.make(args) → create(:app_usage_event, args)
- ServiceUsageEvent.make(args) → create(:service_usage_event, args)
- ProcessModelFactory.make(...) stays as-is (kept as a thin wrapper)
- Other ClassName.make(args) → create(:symbol, args) (e.g. AppModel.make → create(:app_model))
- Sham.foo → generate(:foo) only inside spec/support/factory_definitions/; elsewhere a compatibility shim keeps Sham.foo working

joyvuu-dave:
Done.

PeteLevineA:
I'm a little worried about the following: "Task stop events are also only emitted when the task has a start event or baseline on record, so consumers never see a stop they cannot pair with a start." I know the code originally had a `return if start_event.nil?` Thinking through this should we at least warn when a task is found stopped with no start event at least for debugging?

"After deploying this change, run rake db:was_running_backfill once so we repair any events that happened while old API servers were still serving traffic."
I see this is documented here: https://github.com/cloudfoundry/cloud_controller_ng/pull/5210/changes#diff-2539fbe604329b0f121ef317e716169f17d1d0c17c01575b8a2e2082598c45d9 should we also emit a warning for users who might not have seen the latest documentation?

joyvuu-dave:
@PeteLevineA I guess what I hear you saying is that maybe this is good opportunity to start logging out when we know we were about to create a STOP event but stopped ourself due to there not being a START event. 

I can see this being beneficial. I've just added code that does that now.

**Edit (2026-08-25):** I also added a note telling operators to run the backfill rake task once after the deploy. I've since taken it back out. Here's why.

It covered one narrow case: the old cleanup job runs during the rollout window and prunes a running app's start event before the new servers take over. That leaves one app with no baseline until it restarts. The odds of this happening are pretty low. I don't think asking operators to do this on every deploy is really the right call.

PR #5237, opened 2026-06-26
Fix LocalClient#delete_all_in_path not using partitioned key

WeiQuan0605:

`LocalClient#delete_all_in_path` was building the target path with the raw key instead of `partitioned_key(key)`, so it looked in the wrong directory and silently deleted nothing.

All other `LocalClient` methods (`exists?`, `cp_to_blobstore`, `file_path`, etc.) use `partitioned_key` consistently, which maps a key like `abcdef1234` to `ab/cd/abcdef1234`. `delete_all_in_path` was the only method missing this, making it a no-op in practice.

- `lib/cloud_controller/blobstore/local/local_client.rb`: use `partitioned_key(path)` in `delete_all_in_path`
- `spec/unit/lib/cloud_controller/blobstore/local/local_client_spec.rb`: rewrite the test using real blobstore keys so the partitioned path is actually exercised (the old test used raw directory names and did not catch the bug)

### DIFF
--- a/lib/cloud_controller/blobstore/local/local_client.rb
+++ b/lib/cloud_controller/blobstore/local/local_client.rb
@@ -97,7 +97,7 @@ module CloudController
       end
 
       def delete_all_in_path(path)
-        dir = File.join(@base_path, path)
+        dir = File.join(@base_path, File.dirname(partitioned_key(path)))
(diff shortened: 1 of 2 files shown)


johha (approved):
LGTM

PR #5243, opened 2026-06-30
Drain local workers by working off remaining jobs on SIGTERM

svkrieger:

This adds LocalWorkerDrainPlugin, which replaces the SIGTERM handler for local workers only. Instead of stopping, the worker sets exit_on_complete so it finishes all remaining jobs in the queue before exiting, avoiding dangling PROCESSING_UPLOAD/PENDING states across restarts.

The plugin is only registered for local workers, draining of other workers like cc-generic remain unchanged.

Local workers handle jobs that require local filesystem access (package, droplet, and buildpack uploads). When a local worker receives SIGTERM during drain, the default delayed_job handler calls stop(), which exits after the current job but leaves remaining queued jobs unprocessed until the worker restarts.

### DIFF
--- a/lib/delayed_job/local_worker_drain_plugin.rb
+++ b/lib/delayed_job/local_worker_drain_plugin.rb
@@ -1,0 +1,13 @@
+# Replaces the default SIGTERM handler for local workers to drain all remaining jobs before exiting.
+class LocalWorkerDrainPlugin < Delayed::Plugin
+  callbacks do |lifecycle|
+    lifecycle.before(:execute) do |worker|
+      trap('TERM') do
+        Thread.new { worker.say 'Draining: will exit after finishing remaining jobs' }
+        worker.class.exit_on_complete = true
+      end
+    end
+  end
+end
+
+Delayed::Worker.plugins << LocalWorkerDrainPlugin
--- a/lib/tasks/jobs.rake
+++ b/lib/tasks/jobs.rake
@@ -26,6 +26,7 @@ namespace :jobs do
   desc 'Start a delayed_job worker that works on jobs that require access to local resources.'
 
   task :local, [:name] => :environment do |_t, args|
+    require 'delayed_job/local_worker_drain_plugin'
     puts RUBY_DESCRIPTION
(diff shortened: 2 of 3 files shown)


lib/tasks/jobs.rake
```diff
@@ -1,4 +1,5 @@
 require 'delayed_job/quit_trap'
+require 'delayed_job/local_worker_drain_plugin'
```
johha: I would move this `require` into the local worker task. With this the plugin is independent of the queue name. The corresponding unit test could check if the plugin is loaded

lib/delayed_job/local_worker_drain_plugin.rb
```diff
@@ -0,0 +1,21 @@
+require 'jobs/queues'
+
+# Local workers process jobs that require access to local filesystem resources (e.g. buildpack, droplet,
+# and package uploads via nginx). The 'delayed_job' gem's default SIGTERM handler calls stop(), which
+# exits the worker after the current job finishes - leaving the rest of the queue unprocessed.
+# This plugin replaces the gem's SIGTERM handler for local workers so that all remaining jobs in the
+# queue are worked off before the worker exits, ensuring no jobs are left dangling when draining.
```
johha: Usually we try avoid those multi line AI comments

lib/delayed_job/local_worker_drain_plugin.rb
```diff
@@ -0,0 +1,21 @@
...
+class LocalWorkerDrainPlugin < Delayed::Plugin
+  callbacks do |lifecycle|
+    lifecycle.before(:execute) do |worker|
+      next unless worker.class.queues.length == 1 && VCAP::CloudController::Jobs::Queues.local?(worker.class.queues.first)
+
+      trap('TERM') do
+        Thread.new { worker.say 'Exiting...' }
```
johha: How about `Draining: will exit after finishing remaining jobs`?

johha (changes requested):

johha (approved):

johha (approved):

PR #5286, opened 2026-07-13
Fix lifecycle_type_backfill job never running and piling up

philippthun:
The job's queue was missing from the generic worker's allowlist, so it was enqueued but never executed.

Also make the scheduler skip enqueuing when an untouched job is already queued, and keep only the newest untouched job, removing the rest.

The generic worker's queue list was maintained by hand and had to be kept in sync with the clock scheduler. A missing entry meant a job was enqueued but never consumed. Build the list from the scheduler's job definitions so new periodic jobs are wired up automatically.


johha (approved):
GETM

PR #5314, opened 2026-07-27
Document VCAP_SERVICES to service binding files translation

### LINKED ISSUE #4036: Technical Specification: Translation of VCAP_SERVICES to Service Binding Files
This document describes how service binding details, exposed to application processes in the `VCAP_SERVICES` environment variable, are translated into a list (tree) of files known as _Service Binding Files_.

## Translation Process

1. The translation process begins with the system checking if the application process is enabled for file-based service bindings. If enabled, it builds a set of valid service binding files.

2. Each service binding's name is validated for uniqueness and compatibility. Duplicate or invalid binding names result in an `IncompatibleBindings` error. Binding names must match `[a-z0-9\-.]{1,253}` as specified in the Service Binding Specification for Kubernetes [1].

3. Multiple service binding files are generated for each service binding. The service binding's name acts as the directory name, while various properties of the service binding (such as credentials and attributes) act as file names.

4. The `credentials` attribute in `VCAP_SERVICES` is a JSON object. Each top-level key within this object will be matched to a file within the service binding's directory, and the corresponding value will be the file content. In case `credentials` contain a deeply nested structure, still only the top-level keys will be represented as files. Nested structures inside their value will be serialized as JSON objects. The same applies if the value is a list. (See example no. 1)

5. Existing credential keys can be overwritten by other attributes, that are `VCAP_SERVICES` attributes as defined in the Cloud Foundry Documentation [2] plus `type` and `provider`. The full list of reserved file names is: `binding-guid`, `binding-name`, `instance-guid`, `instance-name`, `name`, `label`, `tags`, `plan`, `syslog-drain-url`, `volume-mounts`, `type`, and `provider`. Overwriting an existing key does not result in an error. (See example no. 2)

6. All file names must match the same schema as the service binding's name. An invalid name results in an `IncompatibleBindings` error. The `VCAP_SERVICES` attributes are transformed to comply to this schema, i.e. underscores are replaced by hyphens, so that e.g. `VCAP_SERVICES` attribute `binding_guid` becomes file name `binding-guid`. (See example no. 3)

7. If a service binding value represents a list (like it is the case for tags and volume mounts), the file content will be a JSON representation of the array.

8. Empty lists or `null` values are omitted, i.e. no file will be created. (See example no. 3)

9. The total bytesize of the generated files is validated. If the total bytesize (of file paths and contents) exceeds the maximum allowed bytesize limit (default is 1,000,000 bytes), an `IncompatibleBindings` error is raised.

## Examples

```
Example No. 1
-------------

VCAP_SERVICES= {
  "foo": [
    {
      "name": "foo",
      "credentials": {
        "simple": "value",
        "deeply": {
          "nested": "value"
        },
        "list": ["v", "a", "l", "u", "e"]
      }
    }
  ]
}

Service Binding Files:
  foo/name: foo
  foo/simple: value
  foo/deeply: {"nested":"value"}
  foo/list: ["v","a","l","u","e"]


Example No. 2
-------------

VCAP_SERVICES= {
  "foo": [
    {
      "name": "foo",
      "credentials": {
        "name": "user",
        "secret": "password"
      }
    }
  ]
}

Service Binding Files:
  foo/name: foo
  foo/secret: password


Example No. 3
-------------

VCAP_SERVICES= {
  "foo": [
    {
      "name": "foo",
      "binding_guid": "45436ca8-0a7c-45e3-9439-ca1b44db7a2b",
      "syslog_drain_url": null,
      "volume_mounts": []
    }
  ]
}

Service Binding Files:
  foo/name: foo
  foo/binding-guid: 45436ca8-0a7c-45e3-9439-ca1b44db7a2b
```

[1] https://servicebinding.io/spec/core/1.1.0/#workload-projection
[2] https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-SERVICES

philippthun:
Add a v3 API docs section describing how the service-binding-k8s and file-based-vcap-services app features expose service bindings on the container file system, including the per-binding file layout, naming rules, reserved attributes, size limit, and examples.

### DIFF
--- a/docs/v3/source/includes/resources/app_features/_service_binding_files.md.erb
+++ b/docs/v3/source/includes/resources/app_features/_service_binding_files.md.erb
@@ -1,0 +1,105 @@
+### Service binding files
+
+By default, service binding details are exposed to an app through the `VCAP_SERVICES` environment variable. Two mutually exclusive app features instead expose them as files on the container's file system. Only one of the two features may be enabled for an app at a time.
+
+- **file-based-vcap-services** writes the full contents of `VCAP_SERVICES` verbatim to a single file named `vcap_services`.
+- **service-binding-k8s** translates each service binding into its own directory of files, as described below.
+
+#### Translation of VCAP_SERVICES to service binding files
+
+When **service-binding-k8s** is enabled, the service binding details in `VCAP_SERVICES` are converted into a tree of files, following the [servicebinding.io](https://servicebinding.io/spec/core/1.1.0/#workload-projection) workload-projection specification. The binding's name becomes a directory name, and each of the binding's properties becomes a file within that directory whose contents are the property's value.
+
+The translation follows these rules:
+
+- Each binding name must be unique and valid; duplicate or invalid names cause an error. A name must match `[a-z0-9\-.]{1,253}` (per the Kubernetes/servicebinding.io spec).
+- The binding's name is used as the directory name; its properties become the file names.
+- The `credentials` attribute is a JSON object: each top-level key becomes a file whose content is that key's value. Nested objects and lists are serialized as JSON.
+- Reserved attributes may overwrite credential keys of the same name without error. The reserved attributes are: `binding_guid`, `binding_name`, `instance_guid`, `instance_name`, `name`, `label`, `tags`, `plan`, `syslog_drain_url`, `volume_mounts`, `type`, and `provider`.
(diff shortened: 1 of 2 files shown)


jochenehret (approved):
Looks good! But from where exactly is this page reachable?

philippthun:
> ... But from where exactly is this page reachable?

It will be available in the navigation pane after "Supported app features":
image

PR #5330, opened 2026-07-31
Replace netaddr gem with stdlib IPAddr

johha:
The netaddr gem is effectively unmaintained (no releases despite open requests). Security group rule validation only needs parsing, family detection, and range-endpoint ordering, all of which IPAddr covers.

Behaviour is preserved:
- leading-zero IPv4 octets are normalised before parsing (IPAddr rejects them, netaddr accepted them);
- range endpoints must be plain addresses, so a prefix is rejected there.

Add a rule_validator spec covering the v2 validate_destination path, which previously had no dedicated coverage.


lib/cloud_controller/rule_validator.rb
```diff
@@ -122,26 +120,46 @@ def self.no_leading_zeros(destination)
...
+
+    # A range endpoint must be a plain address, so a prefix is not allowed here.
+    private_class_method def self.parse_address(val, family)
+      raise IPAddr::InvalidAddressError if val.include?('/')
+
+      parse_cidr(val, family)
+    end
+
+    private_class_method def self.parse_cidr(val, family)
+      ip = IPAddr.new(drop_leading_zeros(val))
```
kathap: Quick question: ::ffff:010.0.0.1 (IPv6-mapped IPv4 with zero-padded octets) was accepted by netaddr but will be rejected by IPAddr. Intentional?

johha: Confirmed, and intentional. CC passes the destination to Diego verbatim, and Diego validates with Go's net.ParseIP, which rejects ::ffff:010.0.0.1 too (Go refuses zero-padded octets since 1.17). So netaddr was accepting a rule that never actually enforces — IPAddr rejecting it up front just aligns CC with Diego. Padded IPv6 (2001:0db8::1) stays valid; it's specifically the ambiguous padded IPv4 octet that's now rejected. Added a spec to cover it.

kathap (approved):

PR #5332, opened 2026-07-31
Replace vmstat gem with /proc reads

johha:
The vmstat gem is a native C extension with no Windows support (its build fails there), and it surfaced in an outdated-dependency audit. On the Linux ccng runtime all three values it provided are available directly from /proc, so the gem can be dropped from lib/vcap/stats.rb.

Behaviour is preserved. vmstat's Linux backend reads /proc/meminfo, so:
- memory_free_bytes = (Inactive + MemFree) * 1024
- memory_used_bytes = (MemTotal - Inactive - MemFree) * 1024 (equals vmstat's active+wired by conservation - the buckets sum to MemTotal)
- cpu_load_average  = first field of /proc/loadavg

Verified against the live vmstat gem on a production CF API VM: 0.0% diff on both memory metrics, load average identical to rounding. Non-Linux dev machines (Mac/Windows) return 0, where these metrics have no consumer.

Add a stats spec covering all three methods on both the Linux and non-Linux paths, which previously had no direct coverage.

### DIFF
--- a/lib/vcap/stats.rb
+++ b/lib/vcap/stats.rb
@@ -1,5 +1,4 @@
 require 'vcap/pid_file'
-require 'vmstat'
 
 module VCAP
   class Stats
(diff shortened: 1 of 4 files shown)


lib/vcap/stats.rb
```diff
@@ -16,21 +15,35 @@ def process_memory_bytes_and_cpu
...
       end
 
       private
 
+      def linux?
+        RUBY_PLATFORM.match?(/linux/)
+      end
+
+      # Reads a kB-valued field from /proc/meminfo (e.g. "MemTotal:  16793990 kB").
+      def meminfo_kb(key)
```
kathap: This will trigger more reads (5) for one metrics update vs once in the vmstat version. Is that problematic?

kathap (approved):

PR #5360, opened 2026-08-12
Allow current=true filter on GET /v3/droplets to return only current droplets across all apps

WeiQuan0605:
 
  GET /v3/droplets now accepts current=true as a query parameter, returning only droplets that are set as the current droplet of an app. Previously this filter was only available on the app-nested endpoint GET /v3/apps/:guid/droplets?current=true.

  Clients that need to display stack information for multiple apps (e.g. to warn about deprecated stacks) previously had to make one request per app to fetch the current droplet. With this change, they can fetch current droplets for all apps in a single request by combining with existing filters:
 GET /v3/droplets?current=true&app_guids=guid1,guid2,...

  This reduces N+1 API requests to 2 requests total.

### DIFF
--- a/app/messages/droplets_list_message.rb
+++ b/app/messages/droplets_list_message.rb
@@ -18,7 +18,8 @@ module VCAP::CloudController
     validates :states, array: true, allow_nil: true
     validates :space_guids, array: true, allow_nil: true
     validates :organization_guids, array: true, allow_nil: true
-    validates :current, inclusion: { in: ['true'], message: 'only accepts the value \'true\'' }, allow_nil: true
+    validates :current, inclusion: { in: ['true'], message: 'only accepts the value \'true\'' }, allow_nil: true,
+                        unless: -> { package_guid.present? }
     validate :app_nested_request, if: -> { app_guid.present? }
     validate :package_nested_request, if: -> { package_guid.present? }
 
--- a/app/messages/droplets_list_message.rb
+++ b/app/messages/droplets_list_message.rb
@@ -20,6 +20,7 @@ module VCAP::CloudController
     validates :organization_guids, array: true, allow_nil: true
     validates :current, inclusion: { in: ['true'], message: 'only accepts the value \'true\'' }, allow_nil: true
     validate :app_nested_request, if: -> { app_guid.present? }
+    validate :package_nested_request, if: -> { package_guid.present? }
 
     def to_param_hash
       super(exclude: %i[app_guid package_guid])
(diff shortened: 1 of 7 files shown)


app/messages/droplets_list_message.rb
```diff
@@ -32,12 +31,6 @@ def self.from_params(params)
 
     private
 
-    def not_app_nested_request
```
kathap, line 35: Should we convert that method into a package-specific validator rather than removing it? 
```not_app_nested_request``` gated all routes where app_guid is absent, that's both the global ```GET /v3/droplets``` route and the package-nested ```GET /v3/packages/:guid/droplets``` route. Removing it enables ```current=true``` on the global route as intended, but also silently enables it on the package route, which flips that endpoint from returning 422 Unknown query parameter to 200 with a join executed. That's outside the PR's stated scope and untested. A ```package_nested_request validator``` (firing on package_guid.present?) would still reject current there while allowing it on the global route.

File: spec/unit/messages/droplets_list_message_spec.rb [hunk 2]
```diff
@@ -125,10 +125,15 @@ module VCAP::CloudController
 
         context 'when the query is not nested under an app' do
```
kathap, line 126: Can we add a test asserting current is still rejected on the package-nested route (package_guid present, app_guid absent)?

spec/unit/fetchers/droplet_list_fetcher_spec.rb
```diff
@@ -125,6 +125,48 @@ module VCAP::CloudController
...
+
+          let(:filters) { { current: 'true', app_guids: [app1.guid] } }
+
+          it 'returns only the current droplet for the specified app' do
+            results = fetcher.fetch_all(message).all
+            expect(results).to contain_exactly(staged_droplet_for_app1)
+          end
+        end
+      end
     end
```
kathap, line 198: ```current=true``` combined with organization_guids or space_guids isn't covered, those add a second join to apps (apps_orgs/apps_spaces alongside the new apps_current). Could we add fetcher tests for those two combinations?

kathap (changes requested):

spec/unit/messages/droplets_list_message_spec.rb
WeiQuan0605: Added

spec/unit/fetchers/droplet_list_fetcher_spec.rb
WeiQuan0605: Added

spec/request/droplets_spec.rb
```diff
@@ -659,6 +659,7 @@
           space_guids
           app_guids
           organization_guids
+          current
```
kathap, line 662: This moves ```current``` into the valid-params list, but there's no request-level test that actually calls ```GET /v3/droplets?current=true``` and asserts the response returns only current droplets. The filtering is covered at the fetcher unit level, but a request spec for the happy path would cover the full controller→message→fetcher→presenter path. Non-blocking, but worth adding.

kathap (dismissed):

kathap (dismissed):

johha (approved):
LGTM

PR #5361, opened 2026-08-12
WIP: Add concurrent request rate limiter and optimize middleware stack

serdarozerr:
## Summary

**Concurrent request rate limiter:** Introduced `ConcurrencyRateLimiter` middleware that limits the number of concurrent in-flight requests per user (or IP for unauthenticated requests). Supports separate `logging_limit` (warn only) and `blocking_limit` (return 429) for smooth rollout. Disabled by default via config. Backed by Redis (with TTL-based safety cleanup) or in-memory store for single-process deployments. Admins are rate limited the same as regular users.

**Middleware stack optimization:** Split `SecurityContextSetter` into two concerns — token decoding only (`configure_token_only`) happens early in the stack, and user DB lookup (`configure_user`) is deferred to a new `UserContextSetter` middleware placed after rate limiters. This means rate limiting never triggers a DB lookup for blocked requests. `RequestMetrics` is also moved after rate limiters so that blocked 429 responses are excluded from request metrics, keeping metrics focused on requests that actually reached the application.

## Changes

- `middleware/concurrency_rate_limiter.rb` — new middleware with `ConcurrencyRateLimiter`, `ConcurrencyLimiter`, `ConcurrentRedisStore`, `ConcurrentInMemoryStore`
- `middleware/user_context_setter.rb` — new middleware for deferred user DB lookup
- `middleware/security_context_setter.rb` — now calls `configure_token_only` instead of full `configure`
- `lib/cloud_controller/security/security_context_configurer.rb` — added `configure_token_only` and `configure_user` methods
- `lib/cloud_controller/rack_app_builder.rb` — wired new middlewares, moved `RequestMetrics` after rate limiters
- `lib/cloud_controller/config_schemas/api_schema.rb` — added `concurrency_rate_limiter` config schema
- `Retry-After` header added on 429 responses from the concurrent rate limiter.

* Related PRs
   https://github.com/cloudfoundry/capi-release/pull/679

### DIFF
--- a/config/cloud_controller.yml
+++ b/config/cloud_controller.yml
@@ -275,6 +275,14 @@ rate_limiter_v2_api:
   global_admin_limit: 20000
   reset_interval_in_minutes: 60
 
+concurrency_rate_limiter:
+  enabled: false
+  blocking_limit: 10
+  logging_limit: 10
+  redis_connection_pool_size: 40
+  redis_counter_ttl_seconds: 60
+
+
 temporary_enable_v2: true
 
 max_concurrent_service_broker_requests: 0
--- a/lib/cloud_controller/rack_app_builder.rb
+++ b/lib/cloud_controller/rack_app_builder.rb
@@ -7,13 +7,16 @@ require 'security_context_setter'
 require 'rate_limiter'
 require 'service_broker_rate_limiter'
 require 'rate_limiter_v2_api'
+require 'concurrency_rate_limiter'
 require 'new_relic_custom_attributes'
 require 'zipkin'
 require 'block_v3_only_roles'
 require 'below_min_cli_warning'
+require 'user_context_setter'
 
 module VCAP::CloudController
   class RackAppBuilder
+    # rubocop:disable Metrics/MethodLength, Metrics/BlockLength
     def build(config, request_metrics, request_logs)
       token_decoder = VCAP::CloudController::UaaTokenDecoder.new(config.get(:uaa))
       configurer = VCAP::CloudController::Security::SecurityContextConfigurer.new(token_decoder)
@@ -21,7 +24,6 @@ module VCAP::CloudController
       logger = access_log(config)
 
       Rack::Builder.new do
-        use CloudFoundry::Middleware::RequestMetrics, request_metrics
         use CloudFoundry::Middleware::Cors, config.get(:allowed_cors_domains)
         use CloudFoundry::Middleware::VcapRequestContextSetter
         use CloudFoundry::Middleware::BelowMinCliWarning if config.get(:warn_if_below_min_cli_version)
@@ -29,6 +31,17 @@ module VCAP::CloudController
         use CloudFoundry::Middleware::SecurityContextSetter, configurer
         use CloudFoundry::Middleware::Zipkin
         use CloudFoundry::Middleware::RequestLogs, request_logs
+
+        if config.get(:concurrency_rate_limiter, :enabled)
+          use CloudFoundry::Middleware::ConcurrencyRateLimiter, {
+            logger: Steno.logger('cc.concurrency_rate_limiter'),
+            blocking_limit: config.get(:concurrency_rate_limiter, :blocking_limit),
+            logging_limit: config.get(:concurrency_rate_limiter, :logging_limit),
+            redis_connection_pool_size: config.get(:concurrency_rate_limiter, :redis_connection_pool_size),
+            redis_counter_ttl_seconds: config.get(:concurrency_rate_limiter, :redis_counter_ttl_seconds)
+          }
+        end
+
         if config.get(:rate_limiter, :enabled)
           use CloudFoundry::Middleware::RateLimiter, {
             logger: Steno.logger('cc.rate_limiter'),
@@ -59,6 +72,9 @@ module VCAP::CloudController
           }
         end
 
+        use CloudFoundry::Middleware::RequestMetrics, request_metrics
+        use CloudFoundry::Middleware::UserContextSetter, configurer
+
         use CloudFoundry::Middleware::CefLogs, Logger.new(config.get(:security_event_logging, :file)), config.get(:local_route) if config.get(:security_event_logging, :enabled)
         use Rack::CommonLogger, logger if logger
 
@@ -76,6 +92,7 @@ module VCAP::CloudController
         end
       end
     end
+    # rubocop:enable Metrics/MethodLength, Metrics/BlockLength
 
     private
 
--- a/lib/cloud_controller/rack_app_builder.rb
+++ b/lib/cloud_controller/rack_app_builder.rb
@@ -30,7 +30,6 @@ module VCAP::CloudController
         use CloudFoundry::Middleware::NewRelicCustomAttributes if config.get(:newrelic_enabled)
         use CloudFoundry::Middleware::SecurityContextSetter, configurer
         use CloudFoundry::Middleware::Zipkin
-        use CloudFoundry::Middleware::RequestLogs, request_logs
 
         if config.get(:concurrency_rate_limiter, :enabled)
           use CloudFoundry::Middleware::ConcurrencyRateLimiter, {
@@ -74,6 +73,7 @@ module VCAP::CloudController
 
         use CloudFoundry::Middleware::RequestMetrics, request_metrics
         use CloudFoundry::Middleware::UserContextSetter, configurer
+        use CloudFoundry::Middleware::RequestLogs, request_logs
 
         use CloudFoundry::Middleware::CefLogs, Logger.new(config.get(:security_event_logging, :file)), config.get(:local_route) if config.get(:security_event_logging, :enabled)
         use Rack::CommonLogger, logger if logger
--- a/lib/cloud_controller/security/security_context_configurer.rb
+++ b/lib/cloud_controller/security/security_context_configurer.rb
@@ -6,14 +6,31 @@ module VCAP::CloudController
       end
 
       def configure(header_token)
+        configure_token_only(header_token)
+        return unless VCAP::CloudController::SecurityContext.valid_token?
+
+        configure_user
+      rescue VCAP::CloudController::UaaTokenDecoder::BadToken
+        VCAP::CloudController::SecurityContext.set(nil, :invalid_token, header_token)
+      end
+
+      def configure_token_only(header_token)
         VCAP::CloudController::SecurityContext.clear
         decoded_token = decode_token(header_token)
+        VCAP::CloudController::SecurityContext.set_token_only(decoded_token, header_token)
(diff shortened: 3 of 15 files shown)


lib/cloud_controller/security/security_context_configurer.rb
```diff
@@ -6,14 +6,31 @@ def initialize(token_decoder)
...
         VCAP::CloudController::SecurityContext.clear
         decoded_token = decode_token(header_token)
+        VCAP::CloudController::SecurityContext.set_token_only(decoded_token, header_token)
+      rescue VCAP::CloudController::UaaTokenDecoder::BadToken
+        VCAP::CloudController::SecurityContext.set(nil, :invalid_token, header_token)
+      end
+
+      def configure_user
+        decoded_token = VCAP::CloudController::SecurityContext.token
+        return unless decoded_token && decoded_token != :invalid_token
```
philippthun, line 27: You can also use the helper method:
```
return unless VCAP::CloudController::SecurityContext.valid_token?
```

lib/cloud_controller/security_context.rb
```diff
@@ -12,6 +12,12 @@ def self.set(user, token=nil, auth_token=nil)
       Thread.current[:vcap_auth_token] = auth_token
     end
 
+    def self.set_token_only(token, auth_token=nil)
```
philippthun, line 15: I think this method is not necessary. From the name one could not tell if it sets the user to nil or not. I would prefer to use `.set(nil, decoded_token, header_token)` instead as it is more explicit.

lib/cloud_controller/security/security_context_configurer.rb
```diff
@@ -6,14 +6,31 @@ def initialize(token_decoder)
       end
 
       def configure(header_token)
+        configure_token_only(header_token)
+        return unless VCAP::CloudController::SecurityContext.valid_token?
+
+        configure_user
+      rescue VCAP::CloudController::UaaTokenDecoder::BadToken
+        VCAP::CloudController::SecurityContext.set(nil, :invalid_token, header_token)
```
philippthun, line 14: I think this method can simply be:
```
configure_token_only(header_token)
configure_user
```

spec/unit/middleware/user_context_setter_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        it 'returns the app response' do
+          status, headers, body = middleware.call(env)
+          expect(status).to eq(200)
+          expect(headers).to eq({})
+          expect(body).to eq('a body')
+        end
+
+        it 'calls configure_user before the app' do
+          expect(security_context_configurer).to receive(:configure_user).ordered
+          expect(app).to receive(:call).ordered.and_return([200, {}, 'a body'])
```
philippthun, line 32: Minor: the return is not necessary for this test.

spec/unit/lib/cloud_controller/security/security_context_configurer_spec.rb
```diff
@@ -197,6 +197,81 @@ module Security
...
+
+          it 'sets invalid token without raising' do
+            expect { configurer.configure_token_only(auth_token) }.not_to raise_error
+            expect(SecurityContext.token).to eq(:invalid_token)
+          end
+        end
+      end
+
+      describe '#configure_user' do
+        let(:token_information) { { 'user_id' => 'user-id-1' } }
```
philippthun, line 235: The token_information can be inlined in this context.

spec/unit/lib/cloud_controller/security/security_context_configurer_spec.rb
```diff
@@ -197,6 +197,81 @@ module Security
...
+
+      describe '#configure_token_only' do
+        let(:auth_token) { 'auth-token' }
+        let(:token_information) { { 'user_id' => 'user-id-1' } }
+
+        before do
+          allow(token_decoder).to receive(:decode_token).with(auth_token).and_return(token_information)
+        end
+
+        it 'sets the token without looking up the user in the DB' do
```
philippthun, line 209: To really check for DB calls there is a nice test helper method:
```
expect { configurer.configure_token_only(auth_token) }.to have_queried_db_times(/select .* from .users./i, 0)
```

middleware/concurrency_rate_limiter.rb
```diff
@@ -0,0 +1,228 @@
...
+        if apply_rate_limiting?(env)
+          user_guid = get_user_id(env)
+          incremented = @concurrency_limiter.try_increment?(user_guid)
+          return too_many_requests!(env) unless incremented
+        end
+
+        status, headers, body = @app.call(env)
+        [status, headers, body]
+      ensure
+        @concurrency_limiter.decrement(user_guid) if incremented
```
philippthun, line 176: This should happen inside `try_increment?`.

middleware/concurrency_rate_limiter.rb
```diff
@@ -0,0 +1,228 @@
...
+        incremented = false
+
+        if apply_rate_limiting?(env)
+          user_guid = get_user_id(env)
+          incremented = @concurrency_limiter.try_increment?(user_guid)
+          return too_many_requests!(env) unless incremented
+        end
+
+        status, headers, body = @app.call(env)
+        [status, headers, body]
```
philippthun, line 174: Can be simplified to:
```
@app.call(env)
```

middleware/concurrency_rate_limiter.rb
```diff
@@ -0,0 +1,228 @@
...
+        @store = if redis_socket.nil?
+                   ConcurrentInMemoryStore.new
+                 else
+                   ConcurrentRedisStore.new_socket(redis_socket, connection_pool_size: @redis_connection_pool_size, counter_ttl_seconds: @redis_counter_ttl_seconds)
+                 end
+      end
+    end
+
+    class ConcurrencyRateLimiter
+      include CloudFoundry::Middleware::ClientIp
```
philippthun, line 149: We should re-organize stuff that is used in different places. Here is a proposal:

- new mixin CloudFoundry::Middleware::UserId
  - includes CloudFoundry::Middleware::ClientIp
  - public method: get_user_id (from BaseRateLimiter)
  - private method: user_token? (from BaseRateLimiter)

- move method basic_auth? from BaseRateLimiter to CloudFoundry::BasicAuth::BasicAuthAuthenticator
  - extract private helpers if feasible

- new mixin CloudFoundry::Middleware::InternalOrRootApi
  - public method: internal_api? (from RateLimiter)
  - public method: root_api? (from RateLimiter)

- new mixin CloudFoundry::Middleware::TooManyRequests
  - public method: too_many_requests! (from BaseRateLimiter, new param error_name)
  - private method: rate_limit_error (from RateLimiter)
  - also use in ServiceBrokerRateLimiter

middleware/concurrency_rate_limiter.rb
```diff
@@ -0,0 +1,228 @@
...
+        rand(1..5).to_i
+      end
+
+      def error_name
+        'ConcurrentRequestLimitExceeded'
+      end
+
+      def error_name_ip_based
+        'IPBasedConcurrentRequestLimitExceeded'
+      end
```
philippthun, line 128: I think these methods belong into the ConcurrencyRateLimiter class.

middleware/concurrency_rate_limiter.rb
```diff
@@ -0,0 +1,228 @@
...
+          return 0 unless @data.key?(key)
+
+          @data[key] -= 1
+          @data.delete(key) if @data[key] <= 0
+          @data[key] || 0
+        end
+      end
+    end
+
+    class ConcurrencyLimiter
```
philippthun, line 63: I think you should move your enhancements into the already existing ConcurrentRequestCounter and it's inner RedisStore class (and move them out of service_broker_rate_limiter.rb).

When you add the ttl handling there (which the ServiceBrokerRateLimiter should also get) as well as the logging feature (for ServiceBrokerRateLimiter blocking and logging limits both should equal the existing max_concurrent_requests param), the same counter class can be used for both rate limiters.

config/cloud_controller.yml
```diff
@@ -275,12 +275,15 @@ rate_limiter_v2_api:
   global_admin_limit: 20000
   reset_interval_in_minutes: 60
 
-temporary_enable_v2: true
+concurrency_rate_limiter:
+  enabled: false
+  blocking_limit: 10
+  logging_limit: 10
```
philippthun, line 281: Should be 5 as in the capi spec. Just for symmetry...

middleware/concurrency_rate_limiter.rb
```diff
@@ -0,0 +1,228 @@
...
+      def user_token?(env)
+        !!env['cf.user_guid']
+      end
+
+      def too_many_requests!(env)
+        headers = {}
+        headers['Retry-After'] = @concurrency_limiter.suggested_retry_after.to_s
+        headers['Content-Type'] = 'text/plain; charset=utf-8'
+        message = rate_limit_error(env).to_json
+        headers['Content-Length'] = message.length.to_s
```
philippthun, line 194: Claude says this should be `bytesize` not `length`. Sounds reasonable.

lib/cloud_controller/rack_app_builder.rb
```diff
@@ -7,28 +7,40 @@
...
+require 'concurrency_rate_limiter'
 require 'new_relic_custom_attributes'
 require 'zipkin'
 require 'block_v3_only_roles'
 require 'below_min_cli_warning'
+require 'user_context_setter'
 
 module VCAP::CloudController
   class RackAppBuilder
+    # rubocop:disable Metrics/MethodLength, Metrics/BlockLength
```
philippthun, line 19: Maybe you could refactor this in order to reduce the method length...

serdarozerr: Refactoring is possible but would require removing the builder block and introducing extra methods, which might hurts readability. So i thought better to keep current form since the full middleware stack is visible in order in one place.

lib/cloud_controller/security/security_context_configurer.rb
serdarozerr: resolved in [f708371](https://github.com/cloudfoundry/cloud_controller_ng/pull/5361/commits/f70837133b7f51fb98e95b8d8f1a129b010f923a)

spec/unit/lib/cloud_controller/security/security_context_configurer_spec.rb
serdarozerr: resolved in commit [f708371](https://github.com/cloudfoundry/cloud_controller_ng/pull/5361/commits/f70837133b7f51fb98e95b8d8f1a129b010f923a)

config/cloud_controller.yml
serdarozerr: set 10 in capi.

PR #5399, opened 2026-08-25
Add ccdb.ssl_mode to control DB TLS without a CA

philippthun:
When no `ccdb.ca_cert` is configured, the DB connection options factories previously emitted no SSL options, leaving the behavior to the client library default.

Add an optional `ccdb.ssl_mode` property, honored only when no CA is configured, so operators can pin an explicit mode:
- MySQL: disabled, required
- Postgres: disable, allow, prefer, require

Unknown values raise a clear error. A configured CA still takes the verifying path unchanged.

The property is added to all db-connection config schemas (api, worker, clock, deployment_updater, migrate, rotate_database_key, route_syncer, blobstore_benchmarks), because the rendered config is validated under
several contexts.

Also remove dead code from the MySQL factory: the `options[:sslmode]` assignments on the CA path were never read by mysql2, which only honors `options[:ssl_mode]`. Server-certificate verification on the CA path is
driven by `sslca` and `sslverify`, so removing the ignored `sslmode` keys does not change behavior. The new `ssl_mode` branch correctly emits `options[:ssl_mode]`. Postgres continues to use `sslmode`, which its
Sequel adapter does read.

### DIFF
--- a/lib/cloud_controller/db_connection/mysql_options_factory.rb
+++ b/lib/cloud_controller/db_connection/mysql_options_factory.rb
@@ -1,6 +1,8 @@
 module VCAP::CloudController
   module DbConnection
     class MysqlOptionsFactory
+      SUPPORTED_SSL_MODES = %i[disabled required].freeze
+
       def self.build(opts)
         options = {
           charset: 'utf8'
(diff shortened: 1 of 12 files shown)


lib/cloud_controller/db_connection/mysql_options_factory.rb
```diff
@@ -1,6 +1,8 @@
 module VCAP::CloudController
   module DbConnection
     class MysqlOptionsFactory
+      SUPPORTED_SSL_MODES = %i[disabled required].freeze
```
kathap, line 4: Is it on purpose that these are 'required'. 'disabled' and the ones for postgres are 'disable' and 'require'?

philippthun: Yes, these are the modes the different database connectors understand. For Postgres there are also more possible values.

kathap (approved):

PR #5401, opened 2026-08-27
Remove lifecycle_type_backfill job

WeiQuan0605:
Related to #5067

* A short explanation of the proposed change:
    Removes the one-off `lifecycle_type_backfill` clock job, its configuration,
    and related scheduler entries. This job was introduced in Deploy 1 to
    backfill NULL lifecycle_type values and is no longer needed.

* An explanation of the use cases your change solves
    Cleans up the temporary backfill job once all installations have had
    sufficient time to run it.

* [X] I have reviewed the [contributing guide](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/CONTRIBUTING.md)

* [X] I have viewed, signed, and submitted the Contributor License Agreement

* [X] I have made this pull request to the `main` branch

* [X] I have run all the unit tests using `bundle exec rake`

* [ ] I have run [CF Acceptance Tests](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/spec/README.md#cf-acceptance-tests-cats)

Issue #5067, opened 2026-04-28
Proposal: Add lifecycle_type column to apps, droplets, and builds


philippthun:
## Problem

Lifecycle type (buildpack, docker, cnb) is not stored directly on apps, droplets, or builds. It is derived at runtime by checking for the existence of associated lifecycle data records:

```ruby
# Current: checks multiple associations
def lifecycle_type
  return 'buildpack' if buildpack_lifecycle_data
  return 'cnb' if cnb_lifecycle_data
  'docker'
end
```

This causes:
1. **Expensive list queries** - `GET /v3/apps?lifecycle_type=docker` requires excluding apps from *both* `buildpack_lifecycle_data` and `cnb_lifecycle_data` tables via subqueries
2. **Unnecessary eager loading** - Presenters load lifecycle data associations just to determine the type string, even when the full lifecycle data is not needed
3. **Implicit semantics** - "docker" is the absence of other types, not an explicit value

## Proposed Solution

Add a `lifecycle_type` VARCHAR column (with index) to `apps`, `droplets`, and `builds` tables.

## Deployment Strategy (Two Phases, Zero Downtime)

**Deploy 1** (PR: #5065):
- Add nullable columns with concurrent index
- Populate for new records via `AppCreate`, `V2::AppCreate`, and `before_create` hooks on droplets/builds
- Fallback to association lookup for existing records
- Model validation ensures only valid lifecycle types (buildpack, cnb, docker)

**Deploy 2**:
- Backfill existing records based on lifecycle data tables
- Make columns non-nullable
- Remove fallback logic
- Optimize `AppListFetcher`: subqueries → simple `WHERE` clause
- Optimize presenters: skip eager-loading lifecycle data for type determination

## Impact After Deploy 2

| Area | Before | After |
|------|--------|-------|
| `AppListFetcher` lifecycle_type filter | 3 subqueries on lifecycle data tables | 1 simple `WHERE lifecycle_type = ?` |
| Presenters (App, Build, Droplet, Process) | Eager-load lifecycle data associations to determine type | Read column value directly |
| Lifecycle type checks in controllers | Association lookups | O(1) column read |
| Docker type semantics | Implicit (absence of other types) | Explicit column value |

## Endpoints That Benefit

**Critical:** `GET /v3/apps` with `lifecycle_type` filter (subquery elimination)

**High:** All list and detail endpoints for apps, droplets, builds, and processes (presenter eager-load avoidance)

[...]

PR #5402, opened 2026-08-27
Enforce lifecycle_type on apps, droplets, and builds

WeiQuan0605:
Related to #5067

> **Note**: This PR depends on #5401 being merged first. CI failures are expected until that happens.

    Backfills NULL lifecycle_type values via a migration, enforces NOT NULL
    constraint, removes fallback logic from models, optimizes AppListFetcher
    (subqueries → simple WHERE clause), and optimizes ProcessPresenter eager
    loading.

    Allows lifecycle_type to be determined directly from the column without
    joining lifecycle data tables, reducing N+1 queries and simplifying code.
    Enables efficient filtering via GET /v3/apps?lifecycle_type=...

### DIFF
--- a/db/migrations/20260825120100_make_lifecycle_type_non_nullable.rb
+++ b/db/migrations/20260825120100_make_lifecycle_type_non_nullable.rb
@@ -1,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table(:apps)     { set_column_not_null :lifecycle_type }
+    alter_table(:droplets) { set_column_not_null :lifecycle_type }
+    alter_table(:builds)   { set_column_not_null :lifecycle_type }
+  end
+
+  down do
+    alter_table(:apps)     { set_column_allow_null :lifecycle_type }
+    alter_table(:droplets) { set_column_allow_null :lifecycle_type }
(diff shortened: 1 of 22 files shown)


db/migrations/20260825120100_make_lifecycle_type_non_nullable.rb
```diff
@@ -0,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table(:apps)     { set_column_not_null :lifecycle_type }
```
svkrieger: Have you tested this migration on a table which hold the expected amount of datasets? This might result in ACCESS EXCLUSIVE locks on the tables and take a while to scan, leading to service degradation. We should do this in a zero-downtime fashion like done in other migrations i.e. lib/database/bigint_migration.rb

WeiQuan0605: Thanks for the feedback! Updated the migration to follow the ADD CONSTRAINT ... NOT VALID + VALIDATE CONSTRAINT pattern. Tested locally with db:migrate and db:rollback.

PR #5448, opened 2026-09-15
Redact credentials from buildpack URLs in telemetry logs

WeiQuan0605:
  Custom buildpack URLs containing credentials (e.g. https://user:secret@github.com/org/bp) were written as plain text to telemetry logs. This PR applies UrlSecretObfuscator to buildpack URLs before telemetry emission, so credentials appear as ***: *** in logs.
  When a space developer configures an app with a credential-bearing custom buildpack URL and triggers a restage or build, those credentials were leaking into the telemetry log file on the server. Anyone with access to that log file (operators, log aggregation systems) could read them.


kathap (changes requested):
in controllers/internal/staging_completion_controller.rb is there missing to add .obfuscated_buildpacks ? line 81

app/models/helpers/lifecycle_data_model_mixin.rb
```diff
@@ -0,0 +1,7 @@
+require 'cloud_controller/url_secret_obfuscator'
```
kathap, line 1: The module is defined at the top level, but all other helpers in app/models/helpers/ wrap their content in module VCAP::CloudController. It works right now via Ruby's constant lookup fallback, but wrapping it consistently would be safer and less surprising.

kathap:
Should LifecycleDataModelMixin also be included in app/models.rb?

kathap:
Maybe a test in ```spec/unit/controllers/internal/staging_completion_controller_spec.rb``` after existing test 'emits a telemetry event for the completed build' would be good, e.g. 'redacts credentials from custom buildpack URLs in telemetry'. It could reuse the same build, url, staging_response, and Timecop.freeze pattern, and check both what the URL looks like after redaction and that the raw secret is gone.

WeiQuan0605:
> Maybe a test in `spec/unit/controllers/internal/staging_completion_controller_spec.rb` after existing test 'emits a telemetry event for the completed build' would be good, e.g. 'redacts credentials from custom buildpack URLs in telemetry'. It could reuse the same build, url, staging_response, and Timecop.freeze pattern, and check both what the URL looks like after redaction and that the raw secret is gone.

Good point. I will add this.

kathap:
LGTM

kathap (approved):

PR #5452, opened 2026-09-16
Update internal API docs from consumer audit

svkrieger:
Add readiness_changed and rescheduling endpoints; fix consumer/auth fields for staging, log_access, asg_latest_update, tasks/completed, metrics, and buildpacks download; add auth-mechanism legend; flag dead-consumer endpoints as removal candidates.


docs/internal/README.md
```diff
@@ -6,14 +6,16 @@ CC's internal API is a patchwork of endpoints used by internal components.
...
+- **OAuth** — Served on the public TLS listener; the caller presents a UAA bearer token and CC validates its scope in-app. Note that being under an `/internal/*` path does not by itself imply mTLS — OAuth endpoints (`log_access`, `ssh_access`) travel the public listener.
+- **Staging basic auth** — HTTP Basic auth with the staging user/password, enforced in-app by a Sinatra `before '/staging/*'` filter. These `/staging/*` download endpoints only work with a **local blobstore** (`blobstore_type: local` or `local-temp-storage`, backed by the `LocalClient` filesystem storage on the CC VM). URLs CC hands to Diego embed the credentials. In practice this is primarily a **development** configuration: production deploys use `webdav` or `storage-cli`, where Diego fetches a direct blobstore URL instead and these endpoints are dormant (nginx returns 403 for `/staging/*` when no local blobstore is configured).
```
stephanme: [RFC-0043](https://github.com/cloudfoundry/community/issues/1274) deprecated the `webdav` provider -> production deployments use `storage-cli` 

storage-cli supports webdav as well.

docs/internal/README.md
```diff
@@ -29,39 +31,32 @@ We do not recommend using internal API endpoints for anything other than their i
 **Auth Mechanism:** OAuth
 
 ### GET /v2/buildpacks/:guid/download
-**Description:** Download a buildpack file
+**Description:** Download a buildpack file. Note: this is a `/v2` public-namespace path (served on the public listener), not an `/internal/*` endpoint — its `download` action is exempt from user OAuth and gated by staging basic auth instead.
```
stephanme, line 50: This also means: when CF API v2 is disabled, local blobstore is broken.

Future: I suggest to move this endpoint to either `/staging/v2/buildpacks/:guid/download` or `/internal/v2/buildpacks/:guid/download`

docs/internal/README.md
```diff
@@ -29,39 +31,32 @@ We do not recommend using internal API endpoints for anything other than their i
...
 
-**Intended Consumer:** ??? (Probably intended for Diego staging task, but it looks like they use direct download from the blobstore now)
+**Intended Consumer:** Diego staging task (local blobstore; remote blobstores get a direct signed URL and bypass this endpoint)
 
-**Auth Mechanism:** mTLS
+**Auth Mechanism:** Staging basic auth
 
 ### GET /staging/packages/:guid
```
stephanme, line 58: Future: I would move this endpoint to `/internal/staging/packages/:guid` or `/internal//v2/staging/packages/:guid`.
This makes clear that this is an internal API that can be changed/dropped at any point in time.

stephanme (commented):
Maybe we add per endpoint a "Routing" info (internal=port 9023, external=port 9022) so that routing is not implicit from Auth.

In the future, I would move most internal endpoints to internal routing. The access check endpoints should not require external routing but still need OAuth.
The only exceptions should be the download endpoints for local blobstore (needed for staging container). For production (no local blobstore), all internal endpoints should use internal routing.

My comments about changing the url of internal endpoints is for discussion only. This PR documents the current setup.

stephanme (approved):

docs/internal/README.md
svkrieger: Yes, even though the scenario is very unlikely as local blobstore is a dev tool and we don't usually have a working Diego instance running when working locally anyways. But it makes sense to move it in the future as someone could potentially use a local blobstore in a full deployment.

